PR #49, opened 2022-06-14
fix: added node v. 16 to pipeline buildspec

wolanlu:
fix: 
- added node v. 16 to pipeline buildspec
- skipping manual approval for pipelines that have only one step

### Feature or Bugfix
- Bugfix

### Detail
- buildspec does not specify node version which default to node v. 12 which is already not supported by cdk. This ends with pipeline not building
- it a minor but when specifying pipeline stages user can choose to have only one stage. This one stage pipeline should not have a manual approval step in the end regardless the name of the stage.

### Relates


dlpzx:
thank you for your contribution :)

Issue #45, opened 2022-06-13
Error while deploying locally on MacBook Pro M1 


pruthvirajksuresh:
Hello team ,

I am getting the below error while deploying the data all on macbookpro m1 

container Name :-  "aws-dataall_cdkproxy)"

Error Message:-
"qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory"'

Any solution would be helpful. 

Thank you , 
Raj

louishourcade:
Hello,
Thanks for raising this issue. I personally don't work with a Mac M1 but I know some colleagues that do, and managed to solve the issue with data.all local deployment. I'm forwarding them the question internally to see if they can help you.
Best

pruthvirajksuresh:
Thank you @louishourcade

PR #67, opened 2022-06-27
Bug fix: M1 Mac docker error

eugeneteoh:
### Feature or Bugfix
- Bugfix

### Detail
- M1 Macs does not properly build the `cdkproxy` container. To fix this, the container should be built on `linux/amd64` platform.

### Relates
- #45

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -22,6 +22,8 @@ services:
       - $HOME/.aws/credentials:/root/.aws/credentials:ro
     restart:
       on-failure:60
+    platform:
+      linux/amd64
 
   graphql:
     build:


eugeneteoh:
Github Actions currently do not have M1 VMs. But it should be added when available.
Reference: https://github.com/actions/virtual-environments/issues/2187

eugeneteoh:
Can someone else with an M1 test this?
CC: @dlpzx @louishourcade @pruthvirajksuresh

dlpzx:
I have an Intel one, let me forward the issue inside the community.

eric-leblouch-al:
The platform needs also to be added for graphql and frontend since they also are built.

eugeneteoh:
> The platform needs also to be added for graphql and frontend since they also are built.

From my side, `cdk-proxy` is the only one causing trouble. I would not recommend moving the other containers to amd64 as it significantly slows down the performance (for M1).

eric-leblouch-al:
I did a `docker-compose build` successfully using this PR on a M1.

PR #82, opened 2022-07-11
Feat/sagemakerdomain

JahedZ:
### Feature or Bugfix
- Feature

### Detail
- In this PR, we provide the code to deploy a Sagemaker Domain with basic settings upon creation of a data.all workspace
- We also extend the Sagemaker permissions available to users, while ensuring they can only access their team's resources (using a resource tagging strategy)
- This takes as an input your default VPC, and assumes it has the appropriate networking configurations. To learn about the network requirements for Sagemaker Studio, read this documentation https://docs.aws.amazon.com/sagemaker/latest/dg/studio-notebooks-and-internet-access.html. For advanced networking customizations, stacks/environment.py should be modified according to your ecosystem's needs.

### Relates
- This was requested in issue 48 
- https://github.com/awslabs/aws-dataall/issues/48

### DIFF
--- a/backend/dataall/cdkproxy/stacks/environment.py
+++ b/backend/dataall/cdkproxy/stacks/environment.py
@@ -209,14 +209,14 @@ class EnvironmentSetup(Stack):
                 ),
             )
 
-try:
-      default_vpc = ec2.Vpc.from_lookup(self, 'VPCStudio', is_default=True)
-except Exception as e:
-       logger.error(f"Default VPC not found, Exception: {e}")
-            vpc_id = default_vpc.vpc_id
-            subnet_ids = [private_subnet.subnet_id for private_subnet in default_vpc.private_subnets]
-            subnet_ids += [public_subnet.subnet_id for public_subnet in default_vpc.public_subnets]
-            subnet_ids += [isolated_subnet.subnet_id for isolated_subnet in default_vpc.isolated_subnets]
+            try:
+                default_vpc = ec2.Vpc.from_lookup(self, 'VPCStudio', is_default=True)
+                vpc_id = default_vpc.vpc_id
+                subnet_ids = [private_subnet.subnet_id for private_subnet in default_vpc.private_subnets]
+                subnet_ids += [public_subnet.subnet_id for public_subnet in default_vpc.public_subnets]
+                subnet_ids += [isolated_subnet.subnet_id for isolated_subnet in default_vpc.isolated_subnets]
+            except Exception as e:
+                logger.error(f"Default VPC not found, Exception: {e}. If you don't own a default VPC, modify the networking configuration, or disable ML Studio upon environment creation.")
 
             sagemaker_domain = sagemaker.CfnDomain(
                 self,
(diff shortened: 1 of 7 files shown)


backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -162,19 +159,91 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
+                                iam.Role.from_role_arn(self, 'DomainRole', role_arn=sagemaker_domain_role.role_arn),
+                                iam.Role.from_role_arn(self, 'EnvironmentDefaultRole', role_arn=self.environment_admins_group.environmentIAMRoleArn),
+                            ] + [iam.Role.from_role_arn(self, f'{group.groupUri}Role', role_arn=group.environmentIAMRoleArn) for group in self.environment_groups],
+                            actions=['kms:*'],
+                        )
+                    ],
+                ),
+            )
+
+            default_vpc = ec2.Vpc.from_lookup(self, 'VPCStudio', is_default=True)
```
dosiennik, line 212: What is going to happen in case of no default VPC in the AWS account?

dlpzx: the PR assumes the VPC as input:
_"This takes as an input your default VPC, and assumes it has the appropriate networking configurations. To learn about the network requirements for Sagemaker Studio, read this documentation https://docs.aws.amazon.com/sagemaker/latest/dg/studio-notebooks-and-internet-access.html. For advanced networking customizations, stacks/environment.py should be modified according to your ecosystem's needs."_

dosiennik: yeah but here in line 212  is see a lookup for the default vpc but there will be cases where the account will not have the default vpc at all so the question is how it is going to work? Shouldn't it look more for the vpc when the data.all is being deployed (vpcId passed as an input or the one which was created as a part of deploying the data.all) or am I missing something?

dosiennik: ah ok I see, so it simply assumes the environment which we link in data.all is having a default vpc

dlpzx: By looking at the code, you can also choose to create a SageMaker Studio domain before creating the environment and avoid data.all to create it for you

dlpzx: @JahedZ  Here a suggestion:

```suggestion
try:
      default_vpc = ec2.Vpc.from_lookup(self, 'VPCStudio', is_default=True)
except Exception as e:
       logger.error(f"Default VPC not found, Exception: {e}")
```
```

**Review by dlpzx — CHANGES_REQUESTED** [review]

File: backend/dataall/cdkproxy/stacks/environment.py (hunk 1 above)
Comment (JahedZ) [reply backend/dataall/cdkproxy/stacks/environment.py:212]: Indeed, this is thought as the "minimal network setup" to have a viable Studio environment - provided that you don't already own a Studio domain. Thanks for the suggestion, it's a very good addition

Comment (JahedZ) [reply backend/dataall/cdkproxy/stacks/environment.py]: But it would be interested to have some kind of "Bring your own VPC" Setup, upon workspace creation.
Similar to what we do with Sagemaker notebook instances

File: backend/dataall/cdkproxy/stacks/environment.py [hunk 2]
```diff
@@ -162,19 +159,91 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
+            vpc_id = default_vpc.vpc_id
+            subnet_ids = [private_subnet.subnet_id for private_subnet in default_vpc.private_subnets]
+            subnet_ids += [public_subnet.subnet_id for public_subnet in default_vpc.public_subnets]
```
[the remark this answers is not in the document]
JahedZ: Thanks, added a corresponding commit

degoldner:
Any update on this issue?

PR #107, opened 2022-08-05
Bugfix: Change nginx config file to fix 404 issue

LEUNGUU:
### Feature or Bugfix

- Bugfix

### Detail
-  For the vpc-facing mode, frontend will be deployed in Farget Service. We found that the current nginx configuration file would cause nginx to return 404 error while we refresh the page.The PR change the nginx configuration file and set it as the default configuration for nginx service.

### Relates
- No related issue. Found it while deployment.


louishourcade:
Thanks for your contribution. Tested and it works for for the "VPC-facing" deployment

PR #113, opened 2022-08-23
Add ram:AssociateResourceShare to pivot role template

degoldner:
### Feature or Bugfix
- Bugfix

### Detail
During the testing of sharing glue tables across multiple environments in different AWS accounts we ran into the following error: 

Could not grant principal xxxxxxxxx permissions ['DESCRIBE', 'SELECT'] to table dataall_demodataset_6_28ibmjc5.dataall_demodataset_6_28ibmjc5 due to: An error occurred (AccessDeniedException) when calling the GrantPermissions operation: User: arn:aws:sts::xxxxxxxxx:assumed-role/dataallPivotRole/dataallPivotRole is not authorized to perform: ram:AssociateResourceShare on resource: arn:aws:ram:eu-central-1:xxxxxxxxx:resource-share/0c58101d-ca4b-48ee-8cd8-5b9a8bd9a68d (Service: AWSRAM; Status Code: 403; Error Code: AccessDeniedException; Request ID: acb6b394-24ef-4c74-bfbe-8fbcd71dca21; Proxy: null)_

Upon investigation we identified that the issue occurs during the execution of the ECS task share_manager.approve_share(). The code is using the data.all pivot role to execute api calls against Lakeformation and RAM to share the glue table with the other environment.

Adding the IAM action to the data.all pivot role solved the error message and the table is shared successfully between the environments.

### DIFF
--- a/deploy/pivot_role/pivotRole.yaml
+++ b/deploy/pivot_role/pivotRole.yaml
@@ -362,6 +362,14 @@ Resources:
                   - LakeFormation*
           - Effect: Allow
             Action:
+              - 'ram:AssociateResourceShare'
+            Resource: !Sub 'arn:aws:ram:*:${AWS::AccountId}:resource-share/*'
+            Condition:
+              'ForAllValues:StringLike':
+                'ram:ResourceShareName':
+                  - LakeFormation*
+          - Effect: Allow
+            Action:
               - 'ram:DeleteResourceShare'
             Resource: !Sub 'arn:aws:ram:*:${AWS::AccountId}:resource-share/*'
             Condition:
@@ -455,7 +463,7 @@ Resources:
               - 'cloudformation:DescribeStackResource'
             Resource:
               - !Sub 'arn:aws:cloudformation:*:${AWS::AccountId}:stack/${EnvironmentResourcePrefix}*/*'
-              - !Sub 'arn:aws:cloudformation:*:${AWS::AccountId}:stack/CDKToolkit/*'
+              - !Sub 'arn:aws:cloudformation:*:${AWS::AccountId}:stack/CDKToolkitDataAll/*'
       ManagedPolicyName: !Sub ${EnvironmentResourcePrefix}-pivotrole-policy-1
       Roles:
         - !Ref PivotRoleName


deploy/pivot_role/pivotRole.yaml
```diff
@@ -455,7 +463,7 @@ Resources:
               - 'cloudformation:DescribeStackResource'
             Resource:
               - !Sub 'arn:aws:cloudformation:*:${AWS::AccountId}:stack/${EnvironmentResourcePrefix}*/*'
-              - !Sub 'arn:aws:cloudformation:*:${AWS::AccountId}:stack/CDKToolkit/*'
+              - !Sub 'arn:aws:cloudformation:*:${AWS::AccountId}:stack/CDKToolkitDataAll/*'
```
dlpzx, line 466: Can you also include
- !Sub 'arn:aws:cloudformation:*:${AWS::AccountId}:stack/CDKToolkit/*' ?

degoldner: Yes sure, this line was not intended to be changed upstream as we use a separate workspace naming only in our fork.

dlpzx (approved):
lgtm, thank you

degoldner:
> lgtm, thank you

@dlpzx, no worries, another question I have is how this change will be applied to already deployed workloads. The creation of the pivot role is done once during bootstrapping, so it will not be updated automatically with this commit as I understand.

dlpzx:
You are correct, the pivot role for existing environments would need to be updated. Another option would be to automate the bootstraping steps in a CICD pipeline. In a current deployment we have translated the CloudFormation pivot role template to a CDK stack and we deploy and run the bootstraping commands from the tooling account (with a different CICD pipeline)

degoldner:
Another option would be to move the pivot role to the environment CF stack where then it is managed by data.all application as part of the environment and it can be updated as we update the CF stack for the environment via the UI.

dlpzx:
I think it is possible, however I see one main drawback. By requiring an Admin to manually create the pivotrole  in the data account, you are validating that "someone" with the proper permissions is explicitly establishing a trust with the data.all account (besides the --trust relationship in the CDK bootstrap of the data account).

In other words, an Admin in the data account is saying "I not only trust the cdk to create stacks from the data.all account, I explicitly trust data.all to create stacks and perform SDK calls in this account".

I think this explicit trust is important. There are other ways to automate the deployment/update of the pivotRole. For example, including it in a "baseinfra" CICD pipeline controlled by the admins of data.all

degoldner:
@dlpzx I do not understand how the explicitly creating trust on the pivot role adds value after explicitly creating trust for the cdk to basically create any resources in the target account. 

When you say that we could add a CICD pipeline around the pivotRole then this pipeline would basically deploy the pivotRole via cdk, similar as the data.all application would do?

chamcca:
Setting up a trust for the CDK to deploy into the account from the data.all account does implicitly grant trust for any application/stack from the primary data.all account to be pushed to the target account. But, given the size, complexity, and data access that data.all enables in the target account, we feel it is best that an administrator of the target account explicitly establish a trust for data.all by creating the pivot role. 

Requiring an explicit trust for our application is safer for users than relying on an implicit trust for any application established by the CDK.

PR #128, opened 2022-09-13
Change the OAuth2 grant type to code to benefit from refresh token, r…

dosiennik:
In order to provide our customers with the best user experience the expiration of the ID & Access tokens every hour shouldn't force the users to log in again. Additionally the user is even not redirected to the login page automatically but gets stuck with the following error displayed by the application:
image
As a result the user needs to refresh the page to be redirected to the login page.

To reproduce:
- log in to the data.all
- do something else in the meantime and get back  to the UI after more than one hour (once the token expires) and click on the Catalog menu on the left (as an example)
- the error will be displayed, refresh the page to be redirected to the login page

The fix I propose:
- changes the OAuth2 grant type to 'CODE', as a result the refresh token is also returned by the Cognito
- under the hood data.all uses Amplify.js which is luckily able to make use of the refresh token automatically, so once the id token & access token expire it will send a request to Cognito to get new ones
- make a small change in the front-end code to redirect to the login page automatically

### DIFF
--- a/frontend/src/contexts/AmplifyContext.js
+++ b/frontend/src/contexts/AmplifyContext.js
@@ -19,7 +19,7 @@ Auth.configure({
     domain: process.env.REACT_APP_COGNITO_DOMAIN,
     redirectSignIn: process.env.REACT_APP_COGNITO_REDIRECT_SIGNIN,
     redirectSignOut: process.env.REACT_APP_COGNITO_REDIRECT_SIGNOUT,
-    responseType: 'token'
+    responseType: 'code'
   }
 });
 
@@ -129,6 +129,7 @@ export const AuthProvider = (props) => {
     <AuthContext.Provider
       value={{
         ...state,
+        dispatch,
         platform: 'Amplify',
         login,
         logout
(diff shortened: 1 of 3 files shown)


frontend/docker/dev/Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN . ~/.nvm/nvm.sh && npm install -g npm yarn
 
 COPY package.json yarn.lock ./
 
-RUN . ~/.nvm/nvm.sh && yarn install
+RUN . ~/.nvm/nvm.sh && yarn install --network-timeout 1000000
```
LEUNGUU, line 23: I think this is not a common situation. Maybe we could add the timeout parameter when we need instead of adding it here for all the users.

dosiennik: yeah I can remove it, for me I couldn't build it without this param

dosiennik: @LEUNGUU removed that change, thx!

dlpzx (approved):
lgtm

PR #137, opened 2022-09-22
fix count of required permissions during init

wolanlu:
fix count of required permissions: using map instead of list to eliminate duplicates and provide interface for defining descriptions

### Feature or Bugfix

- Bugfix

### Detail
Every time graphql lambda is initialised it is performing permission initialisation which is unnecessary. The root cause of this error is incorrect length calculation done agains list of all permissions which has duplicates. A map was introduced to count only unique values and clean up the code creating descriptions for permissions.

### Relates


dlpzx (approved):
lgtm, thank you. I also find it annoying the repeated definition in every start of the lambda!

PR #149, opened 2022-10-03
Bugfix/query organization environments fix

wolanlu:
### Feature or Bugfix

- Bugfix

### Detail
When entering any filter criteria for environments on organisation view the result in empty. Expected to filter environments with name containing entered term. 
Added integration test to check this and fixed query (the was wrong table used for filter condition) 

### Relates
N/A

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
     -   repo: https://github.com/psf/black
-        rev: 21.6b0
+        rev: 22.8.0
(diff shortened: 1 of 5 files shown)


wolanlu:
I've bumped the version of black because the old one could not work with current version of click which made pre-commit hook not working. Also, since there were many formatting changes in files I touched (I am not sure why, maybe a global reformatting is needed) I moved these to separate commit

dlpzx (approved):

PR #191, opened 2022-11-04
Changed function to an function that can be found in older versions o…

rb201:
### Feature or Bugfix
- Bugfix
- https://github.com/awslabs/aws-dataall/issues/188
- replaced the function with another function that can be found in older python versions
### Detail
- `removesuffix` function is not found in older versions of python


dlpzx (approved):
lgtm

PR #192, opened 2022-11-04
189 acm certificate

noah-paige:
Bug Fix
- For `internet_facing=False` deployments allows creation and import of ACM Certificate for ALB to use for the application to circumvent stuck in Pending Validation state (originally caused due to validating an ACM public certificate using a domain record in a Route53 private hosted zone)

Relate to Issue - https://github.com/awslabs/aws-dataall/issues/189


dlpzx (approved):
lgtm

PR #200, opened 2022-11-08
Fix removed_folders none object for no Access point

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
- if we are sharing tables only, there are no pre-existing access points and the get_removed_folders method returns a NoneType object that is then treated as a list, therefore failing.

By explicitly returning an empty list for this case we avoid the error.
It has been tested in branch v1m2m0

### Relates
- #199 

### DIFF
--- a/backend/dataall/tasks/data_sharing/common/s3_approve_share.py
+++ b/backend/dataall/tasks/data_sharing/common/s3_approve_share.py
@@ -172,3 +172,5 @@ class S3ShareApproval(S3ShareManager):
                     ) for prefix in removed_prefixes
                 ]
                 return removed_folders
+            return []
+        return []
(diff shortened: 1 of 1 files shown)


backend/dataall/tasks/data_sharing/common/s3_approve_share.py
```diff
@@ -172,3 +172,5 @@ def get_removed_prefixes(
                     ) for prefix in removed_prefixes
                 ]
                 return removed_folders
+            return []
```
LEUNGUU, line 175: Maybe we can remove this one and just keep the last return.  Because if no condition matches, then it will always go to the last return

LEUNGUU (commented):
Please check my comments above. Thanks.

LEUNGUU (approved):

PR #224, opened 2022-11-23
Remove comma which creates a tuple and duplicate SSM parameters.

mgwidmann:
### Detail
- Extra comma in assignment causes array to be wrapped in a 2-element tuple and then JSII cannot convert it. See https://github.com/awslabs/aws-dataall/issues/216
- SSM parameters are duplicated if `custom_domain` is used. First they are created in the backend stack, then later when the front end is deployed it recreates the SSM parameters with the same name. This only happens when `custom_domain` and `internet_facing` are both being used.

### Relates
- https://github.com/awslabs/aws-dataall/issues/216


dlpzx (approved):
lgtm

Issue #217, opened 2022-11-21
dbmigration is failing


haljeshi:
Hello,

I have deployed data.all several times for different organizations. I have faced [this issue](https://github.com/awslabs/aws-dataall/issues/101) before and I have solved it by increasing the Concurrently running builds and it solved the issue in my first deployment.

In other deployments, everything went smoothly. But in this particular deployment, I'm facing [this issue](https://github.com/awslabs/aws-dataall/issues/133) and I'm unable to solve it. I have increased the build timeout as suggested by the OP but it didn't solve the issue. The concurrently running builds has been increased from 60 to 180 and yet the issue is still there.

Can you please assist me in solving the issue?

The build logs:

```
[Container] 2022/11/21 07:27:45 going inside waitForAgent
[Container] 2022/11/21 07:27:45 Waiting for agent ping
[Container] 2022/11/21 07:27:46 Waiting for DOWNLOAD_SOURCE
[Container] 2022/11/21 07:27:47 Phase is DOWNLOAD_SOURCE
[Container] 2022/11/21 07:27:47 finished waitForAgent
[Container] 2022/11/21 07:27:47 CODEBUILD_SRC_DIR=/codebuild/output/src944698246/src
[Container] 2022/11/21 07:27:47 YAML location is /codebuild/readonly/buildspec.yml
[Container] 2022/11/21 07:27:47 Setting HTTP client timeout to higher timeout for S3 source
[Container] 2022/11/21 07:27:47 Processing environment variables
[Container] 2022/11/21 07:27:47 No runtime version selected in buildspec.
[Container] 2022/11/21 07:27:49 Moving to directory /codebuild/output/src944698246/src
[Container] 2022/11/21 07:27:49 Configuring ssm agent with target id: codebuild:c67d65a9-13a6-4235-9e6f-83a7d129f8aa
[Container] 2022/11/21 07:27:49 Successfully updated ssm agent configuration
[Container] 2022/11/21 07:27:49 Registering with agent
[Container] 2022/11/21 07:27:49 Phases found in YAML: 1
[Container] 2022/11/21 07:27:49  BUILD: 10 commands
[Container] 2022/11/21 07:27:49 Phase complete: DOWNLOAD_SOURCE State: SUCCEEDED
[Container] 2022/11/21 07:27:49 Phase context status code:  Message: 
[Container] 2022/11/21 07:27:49 Entering execCommands
[Container] 2022/11/21 07:27:49 Entering phase INSTALL
[...]

PR #228, opened 2022-11-24
Added uploadToS3 stage when quality gate = False

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
For the setup of quality_gate set to false we create the UploadToS3 stage
image

### Relates
- #217 

### DIFF
--- a/deploy/stacks/pipeline.py
+++ b/deploy/stacks/pipeline.py
@@ -463,6 +463,38 @@ class PipelineStack(Stack):
                     security_groups=[self.codebuild_sg],
                 ),
             )
+        else:
+            it_project_role = iam.Role(
+                self,
+                id=f'ItCobdeBuildRole{self.git_branch}',
+                role_name=f'{self.resource_prefix}-{self.git_branch}-integration-tests-role',
+                assumed_by=iam.CompositePrincipal(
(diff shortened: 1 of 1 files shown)


deploy/stacks/pipeline.py
```diff
@@ -463,6 +463,38 @@ def set_quality_gate_stage(self):
...
+                role_name=f'{self.resource_prefix}-{self.git_branch}-integration-tests-role',
+                assumed_by=iam.CompositePrincipal(
+                    iam.ServicePrincipal('codebuild.amazonaws.com'),
+                    iam.AccountPrincipal(self.account),
+                ),
+            )
+            for policy in self.codebuild_policy:
+                it_project_role.add_to_policy(policy)
+
+            gate_quality_wave = self.pipeline.add_wave('QualityGate')
```
noah-paige, line 479: Can we change the id to `UploadToS3` instead of `QualityGate` to avoid confusion when quality gate set to false - otherwise looks good

noah-paige (approved):

deploy/stacks/pipeline.py
dlpzx: You are right. Added :)

Issue #249, opened 2022-12-12
Safety check fails for SQLalchemy<2.0.0b1. Non-affected versions are still in beta


dlpzx:
### Describe the bug

The following vulnerability has been found in our version of SQLalchemy (found [in pyup](https://pyup.io/vulnerabilities/PVE-2022-51668/51668/) and more detailed [in SQLalchemy GitHub repo](https://github.com/sqlalchemy/sqlalchemy/pull/8563). 

Safety checks running in the QualityGate stage of the CICD pipeline fail and prevent any upgrades or development to data.all.

The only non-affected versions of SQLalchemy are still in preview (versions 2.0.0bX) which introduces the risk of using non-prod versions, plus release 2.X is a major release which also involves additional checks that need to be performed to ensure backwards compatibility.

Here are the vulnerability's details: 

"Currently, careless uses of str(engine.URL()) in logs and prints can lead to leaking clear text passwords to the open. "

**We are not using this method in any place in our code, which means that this vulnerability does not affect our implementation**

### How to Reproduce

Execute the data.all CICD pipeline by releasing a change to the CodePipeline pipeline, the Quality stage should fail with the following message:

![image](https://user-images.githubusercontent.com/71252798/207057559-abb2379e-4d57-42be-b375-577653e74547.png)

![image](https://user-images.githubusercontent.com/71252798/207057660-01c4c2ab-03d2-40a9-89a6-fa15be072ff4.png)

### OS

Mac

### Python version

3.8

### AWS data.all version

v1.2.3

PR #250, opened 2022-12-12
Added --ignore for SQLAlchemy vulnerability on all non-beta versions

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
- Added --ignore for the vulnerability https://pyup.io/vulnerabilities/PVE-2022-51668/51668/ It is explained in detail in #249 
- Once there is a production version of SQLAlchemy that we can upgrade to we will upgrade the library and remove this --ignore flag
### Relates
#249 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ check-security: upgrade-pip install-backend install-cdkproxy
 	pip install bandit
 	pip install safety
 	bandit -lll -r backend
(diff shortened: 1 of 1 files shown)


NickCorbett (approved):
As noted in the issue, this check is removed as we have manually reviewed the code and it does not affect the project

Issue #242, opened 2022-12-07
Unauthorized data preview within the same Dataset


aaronchong888:
### Describe the bug

Currently the data classification is only at the Dataset level (Unclassified/ Official/ Secret). When a Dataset contains multiple tables, when a user request access for only 1 table within the Dataset, once it is granted, the user will be able to view ALL the other tables Preview Data.

### How to Reproduce

```
*P.S. Please do not attach files as it's considered a security risk. Add code snippets directly in the message body as much as possible.*
```
1. Create a dataset with Classification set to "Official"
2. Create 2 different tables within the dataset
3. User request access for only 1 table in the dataset
4. Grant access to the user
5. User will be able to preview data for both tables

### Expected behavior

Users should only be able to preview data if they are granted access to the corresponding tables (unless it is an Unclassified dataset).

### Screenshots

![Screenshot 2022-12-02 at 2 53 48 PM](https://user-images.githubusercontent.com/8871864/206139104-204b462c-56ba-4a69-b825-c795941e3da3.png)

![Screenshot 2022-12-02 at 2 51 30 PM](https://user-images.githubusercontent.com/8871864/206139138-c2521b1c-20a1-4e59-826e-636d136c25fb.png)

### OS

AWS Deployment

### Python version

3.8

### AWS data.all version

1.2.2

noah-paige:
Hi @aaronchong888 - thanks for calling this out! I was able to replicate this bug and am working on pushing a fix in this PR https://github.com/awslabs/aws-dataall/pull/265

PR #265, opened 2022-12-22
242 unauthorized data preview

noah-paige:
### Feature or Bugfix
- Bugfix

### Detail
- Only be able to preview data if they are granted access to the corresponding tables (unless it is an Unclassified dataset).

### Relates
- #242 


backend/dataall/db/api/share_object.py
```diff
@@ -258,13 +258,23 @@ def approve_share_object(
...
         share.status = ShareObjectStatus.Approved.value
 
-        ResourcePolicy.attach_resource_policy(
-            session=session,
-            group=share.principalId,
-            permissions=permissions.DATASET_READ,
-            resource_uri=dataset.datasetUri,
-            resource_type=models.Dataset.__name__,
-        )
+        # GET TABLES SHARED AND APPROVE SHARE FOR EACH TABLE
```
dlpzx, line 261: I will need to make updates to this piece with the new design of the sharing state machine, but with this implementation I think it is fine

backend/migrations/versions/d05f9a5b215e_backfill_dataset_table_permissions.py
```diff
@@ -0,0 +1,131 @@
...
+        print(f'Failed to backfill dataset table permissions for owners/stewards due to: {e}')
+
+    try:
+        bind = op.get_bind()
+        session = orm.Session(bind=bind)
+        print('Back-filling dataset table permissions for shared principals...')
+        share_table_items: [ShareObjectItem] = session.query(ShareObjectItem).filter(
+            (
+                and_(
+                    ShareObjectItem.status == ShareObjectStatus.Share_Succeeded.value,
```
dlpzx, line 111: I like that the addition of permissions happens after the share has succeeded, in the implementation we grant when a table is approved, but I think it is a bit difficult to avoid doing it at a different stage.

dlpzx (approved):

dlpzx (approved):
Sorry for this

PR #266, opened 2022-12-30
Update the region in which parameters are looked for

louishourcade:
### Feature or Bugfix
- Bugfix

### Detail
In the current version, if `internet_facing==True`, then `cognito_urls_configs.py` and `frontend_config.py` look for parameters in `us-east-1`.
But if `custom_domain` is configured, then the parameters `frontend/custom_domain_name` and `userguide/custom_domain_name` are created with the backend stack (thus in the region of the backend).
With this code changes, `cognito_urls_configs.py` and `frontend_config.py` look for parameters in `us-east-1` only if `internet_facing ==True` and `custom_domain==False`. If this is not the case, it uses parameter in the backend region.

### DIFF
--- a/deploy/configs/cognito_urls_config.py
+++ b/deploy/configs/cognito_urls_config.py
@@ -24,11 +24,9 @@ def setup_cognito(
         'Parameter'
     ]['Value']
 
-    if internet_facing == 'True':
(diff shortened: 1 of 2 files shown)


dlpzx (approved):
lgtm, thank you Louis

Issue #255, opened 2022-12-15
Removing share item (folder) does not work


degoldner:
### Describe the bug

After having successfully shared folders cross account the removal of certain folders from the share object is not working as expected. The permission is not revoked immediately and also after subsequent submission the permission does not get revoked correctly, though it is not shown in the data.all UI anymore.

### How to Reproduce

**WHEN**
1. Share a dataset with two folders with another environment (cross account)
2. Approve share
3. Owner of dataset removes one of the folders, share object turns into DRAFT mode
4. Requester submits share request again with only the one remaining folder
5. Owner approves share request again

**THEN**
After owner removes folder from share object the permission is NOT revoked and the owner has no possibility to revoke the whole share object any more.

After approving share object with only one remaining folder the revoking of removed of folders fails and thus no permission to the removed folder is revoked. The requester still has access to the folder.

We identified that the issue resides in the removal of folders in the S3ShareApproval.py class.
When removing a share item from a share object it is deleted from the share object in the database.
**Later during approval of the share object we look for the removed folders as a share item and get a None exception:**

`
Sharing task failed due to: 'NoneType' object has no attribute 'status'`

With the following code being the root cause:
![image](https://user-images.githubusercontent.com/107395339/207883138-1685e173-9127-48d6-a5fd-a1177333228d.png)

### Expected behavior

When the owner of a dataset edits items in the share object he/she should trigger an immediate action which is reflected also in the AWS permission layer or get better feedback about the status of the share item. Removing a share item currently is not equal to actually revoking permissions.

### OS

AWS deployment

### Python version

3.7

### AWS data.all version

v1.2.0

dlpzx:
Hi @degoldner I just tested it and you are right, the sharing stays. I did not get the error `Sharing task failed due to: 'NoneType' object has no attribute 'status'` but I was able to list the objects in a folder previously shared that I removed from the share request.

degoldner:
@dlpzx , the error happens only when the share request is submitted and approved again after a successfully shared folder has been removed. Then only the share manager task will try to revoke the access for the removed folders as part of the anew approval for the share object.

degoldner:
That said I think our issue here is not how we handle the removal of folders for a share request which has been approved anew but rather we have to think how we handle the removal of share items in the first place and the state associated to them. 

I could imagine to only directly remove/delete share items in certain states (DRAFT e.g.) and introduce new logic for revoking permissions on item level for share items which are in status Share_Succeeded.

[...]

PR #267, opened 2023-01-04
Introduce sharing State Machines

dlpzx:
### Feature or Bugfix
- Feature
- Refactoring

### Detail
- embed sharing actions into State machines
- Redesign the sharing workflow to keep track of revoke of access
- Adjust and simplify sharing managers and tasks
- Redesign clean-up process
- Added info in share statistics in inbox and outbox:

image

(Check out the conversation in the below issue for more details)

### Relates
- #255 


degoldner:
Hi @dlpzx, I haven't looked in detail into the implementation but definitely appreciate the work you put into this improvement!

One thing that caught my attention is that we are lacking tests for the implementation of the sharing logic in the BE, for example the modification of the share item status and the required interactions for sharing S3 or LF.

Our team already implemented a couple of tests, specifically for the s3_share_manager.py and we would be happy to contribute those.

Additionally with the change of the state machine I believe that we should check the tests in test_share.py which test the share logic of the graphql BE from an API level and extend them to the new behavior.

LEUNGUU (approved):
Thanks for adding me as a reviewer. 
I like the state machine which gives us a really clear way to track the share item status, especially for us, developer.

dlpzx:
As @degoldner has pointed out (thank you), we are lacking tests. Please don't merge until we have added those. In particular we will need:
1. tests for SM testing the statuses, actions and transitions
2. tests for the sharing tasks (share managers) --> @degoldner I would love to use your contributions for the S3 share manager.

degoldner:
@dlpzx, I do not have permissions to commit to this branch directly. Could an an administrator please give me permissions to commit to this feature branch? Alternatively I can create a fork of this branch for the unit tests. What do you prefer?

dlpzx:
Hi @degoldner, The second would be fine (for the first option we would need to open a permission process that takes longer)

dlpzx:
Besides testing, we need to review all resolvers that involve shares (for example, dataset deletion checking approved shares) @noah-paige just FYI I will continue working on this PR

degoldner:
@dlpzx, I opened the following PR where I added the tests for the **s3_share_manager.py**. I will finalize the tests as some changes with the consumption role affected the earlier version we worked on and then can **add the tests** for the **s3_process_share.py**. 

https://github.com/awslabs/aws-dataall/pull/271

As you can see I added a separate file for the tests of the test_s3_share_manager.py as the current test_share_manager.py is to generic for all the different classes we want to test.

degoldner:
@dlpzx, we are picking up the test for the share object api (test_share.py), the s3_process_share.py and the lakeformation managers and processors (same account).

We will collect the changes here:
https://github.com/awslabs/aws-dataall/pull/275

frontend/src/views/Shares/ShareView.js
```diff
@@ -157,6 +161,28 @@ function ShareViewHeader(props) {
...
+        },
+        variant: 'success'
+      });
+      await fetchItems();
+      await fetchItem();
+    } else {
+      dispatch({ type: SET_ERROR, error: response.errors[0].message });
+    }
+    setRevoking(false);
+  };
```
luxatwoodmark, line 185: Regarding Lines 197-212: seems like the "shares" breadcrumb is there twice by mistake

backend/migrations/versions/509997f0a51e_sharing_state_machines_v1_4_0.py
```diff
@@ -0,0 +1,139 @@
...
+        session = orm.Session(bind=bind)
+        print('Updating share Objects...')
+        (
+            session.query(ShareObject)
+            .filter(
+                ShareObject.status == 'Completed'
+            )
+            .update(
+                {
+                    ShareObject.status == 'PendingApproval'
```
degoldner, line 117: In the update statement it should be a mapping:
                {
                    ShareObject.status**:**'PendingApproval'
                }

dlpzx: You are right, thank you. I don't know how this passed my tests. thank you

frontend/src/views/Shares/ShareView.js
dlpzx: Yeeep, thank you again

dlpzx:
@degoldner I have added tests for the sharing api calls and for the LakeFormation share manager. Critics are more than welcome :)

frontend/src/views/Worksheets/WorksheetView.js
```diff
@@ -32,7 +32,7 @@ import listEnvironmentGroups from '../../api/Environment/listEnvironmentGroups';
 import { SET_ERROR } from '../../store/errorReducer';
 import listDatasetsOwnedByEnvGroup from '../../api/Environment/listDatasetsOwnedByEnvGroup';
 import listDatasetTables from '../../api/Dataset/listDatasetTables';
-import getSharedDatasetTables from '../../api/Dataset/getSharedDatasetTables';
+import getSharedDatasetTables from '../../api/DatasetTable/getSharedDatasetTables';
 import listDatasetTableColumns from '../../api/DatasetTable/listDatasetTableColumns';
 import searchEnvironmentDataItems from '../../api/Environment/listDatasetsPublishedInEnvironment';
```
noah-paige, line 37: Currently, if I share N Tables in 1 dataset with a Consumer Group, the glue database name shows up N times if the consumer is querying the data using Worksheets feature.

I believe we use this `searchEnvironmentDataItems` api to query and set shared databases --> may just need to add a new query that pulls queries the same with a `.distinct()` filter on database name.

noah-paige (commented):
Left 2 minor comments - let me know your thoughts @dlpzx and otherwise looks good to me!

backend/dataall/tasks/data_sharing/share_processors/s3_process_share.py
```diff
@@ -0,0 +1,186 @@
...
+            dataset: models.Dataset,
+            share: models.ShareObject,
+            revoke_folders: [models.DatasetStorageLocation],
+            source_environment: models.Environment,
+            target_environment: models.Environment,
+            source_env_group: models.EnvironmentGroup,
+            env_group: models.EnvironmentGroup
+    ) -> bool:
+        """
+        1) (one time only) manage_bucket_policy - grants permission in the bucket policy
```
degoldner, line 112: The docs here for the process_revoked_shares do not match, I would suggest the following:
`        """
        1) Remove revoked folder from S3 access point policy
        2) Update_share_item_status

        Returns
        -------
        True if all revoke_folders were revoked successfully
        """`

backend/dataall/tasks/data_sharing/share_managers/s3_share_manager.py
```diff
@@ -286,61 +303,77 @@ def delete_access_point_policy(self):
...
+            target_environment: models.Environment,
+    ):
         logger.info(
             'Deleting target role IAM policy...'
         )
+        access_point_name = S3ShareManager.build_access_point_name(share)
         existing_policy = IAM.get_role_policy(
-            self.target_account_id,
-            self.target_requester_IAMRoleName,
+            dataset.AwsAccountId,
```
degoldner, line 334: Here you we want to delete the policy on the requester account, therefore we should use the target_environments AwsAccountId as input for the IAM.get_role_policy() fct.


`        existing_policy = IAM.get_role_policy(
            target_environment.AwsAccountId,
            share.principalIAMRoleName,
            "targetDatasetAccessControlPolicy",
        )`

degoldner:
Regarding the SM we have some improvements we would like to suggest from our side:

This is the current setup:
image

1. First of all it would be great in the diagram for the share item SM to differentiate between share item actions and share object actions better to make it easier to understand
2. Second we suggest to introduce a new state called "REVOKED" for the share object which is the result of the "REVOKE_ALL" share object action as opposed to the "REJECTED" state which is the result of the "REJECT" action. The difference between them is that rejecting a share object means to reject NEW PENDING changes but to not act upon already granted access. On the other hand the revoke_all action will revoke all permissions.
3. When the approver revokes a single item the share request goes into DRAFT again and has to be submitted again before revoking can happen. We think the approver should be able to revoke access to single items directly.

dlpzx:
Hi @degoldner, I will review the points above (as always thank you for your work)

Regarding the SM, is the current figure representing more or less what you would want?
image


From the points you mentioned:
- separation between item and object actions. This is possible until some extend, because they affect each other. The principle that I followed is that a shareObjects are treated as "requests" which means no information about the success or failure of the sharing is in their workflow, because we do not share requests, we share items. Similarly, the shareItem is not submitted, it is pending approval. I know that it can still be confusing so any proposition that you have drafted we will consider it.
- revoked state: agree, I see the point and value of it. I will add it to the SM, it should not be too difficult :) In the figure I just drafted where it will be, but it might happen that some of the transitions (arrows from/to revoked) change a bit.
- revokeItem for approvers without requester submission. I also see the value of this feature but it is a little bit more difficult to integrate it in the current SM without disturbing the sharing workflow. I will work on it today and get back to you with a design. In the above figure it is clear how to revoke one item from the item SM, but from the share object SM I need to have a look into how to trigger the process task to batch multiple revoke items together instead of having one task per item

tests/api/test_share.py
```diff
@@ -503,432 +965,408 @@ def test_add_item(
...
-        username=user3.userName,
-        shareUri=share_object.shareUri,
-        groups=[group3.name],
-    )
-    assert response.data.approveShareObject.status == 'Approved'
-    assert response.data.approveShareObject.userRoleForShareObject == 'Approvers'
+    # Then shared item changes to status Revoke_Approved
+    sharedItem = response.data.revokeAllShareObject.get('items').nodes[0]
+    status = sharedItem['status']
+    status == dataall.api.constants.ShareItemStatus.Revoke_Approved.value
```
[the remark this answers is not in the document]
dlpzx: Yeeep, thank you

frontend/src/views/Worksheets/WorksheetView.js
dlpzx: I am going to fix this in a separate issue as it is a generic bug. Also from what I see in worksheets we are not using `paginated_shared_with_environment_group_datasets` and we are querying all environment databases.

dlpzx:
Hi @degoldner, I have included changes to implement this SM:

When an approver revokes 1 single item, the item goes into revoke approved directly. Instead of triggering an ECS task for each revoked item (let's say the approver revokes 3 items), we wait 5 minutes until we trigger the ECS task and revoke the single revoked items at once. What do you think?

image

backend/dataall/db/api/dataset.py
```diff
@@ -491,12 +500,12 @@ def list_dataset_shares(session, dataset_uri) -> [models.ShareObject]:
...
 
     @staticmethod
-    def list_dataset_approved_shares(session, dataset_uri) -> [models.ShareObject]:
+    def list_dataset_shares_with_existing_shared_items(session, dataset_uri) -> [models.ShareObject]:
         query = session.query(models.ShareObject).filter(
             and_(
                 models.ShareObject.datasetUri == dataset_uri,
                 models.ShareObject.deleted.is_(None),
-                models.ShareObject.status == 'Approved',
+                models.ShareObject.existingSharedItems.is_(True),
```
degoldner, line 508: This call does not work in my opinion as existingSharedItems is a query expression which need to be defined, see https://docs.sqlalchemy.org/en/14/orm/mapped_sql_expr.html#query-time-sql-expressions-as-mapped-attributes

Instead we could join here with the ShareObjectItem table to verify if shared items exist.

dlpzx:
Hi @degoldner , @luxatwoodmark you probably noticed that I have been doing major changes the last 2 days. After discussing the design internally we realized that it could be simplified to fulfill your request of "allowing direct revoking of items by approvers". I was also not 100% convinced by the mix of "approving" and "revoking" in the sharing tasks and the general user experience, so this was a good opportunity to re-iterate on our initial design. 

image

The big portion of the implementation is done, but I still need to review it from beginning to end, I am sure we can find many bugs right now. Still to do:
- review share manager, sharing resolvers and state machines (maybe simplify some of the names)
- review share or revoke failure workflow
- fix tests when needed
- review associated apis: requests from me/ to me. Environment, Dataset, Redshift cluster apis
- review associated frontend views: statistics, color schemas for statuses...

But as an spoiler, here are a few snapshots of the implementation:

image

In this case there is only one folder, but I have tested it with multiple items :) It also removes the need of the revokeAll api, because we can just select all items in this window.
image

What are your thoughts?

backend/dataall/tasks/data_sharing/share_processors/s3_process_share.py
dlpzx: On it :)

dlpzx:
I included all changes except for this point:
- I think in the share object SM the transition form COMPLETED to SUBMITTED and form COMPLETED to APPROVED should be possible too, for example when we have items in Draft state and call the RevokeItems action then we will end up in COMPLETED state but maybe then directly want we want to submit the items from there which are still in status PendingApproval, similar for approval

I need to think on how to "store" the previous state maybe it is not from the COMPLETED, but from the REVOKE_IN_PROGRESS that we have to add arrows. Let me give it some thought. The rest it is already pushed :)

tests/api/test_share.py
```diff
@@ -17,139 +15,226 @@ def cpltz(l):
...
-    return response.data.createShareObject
-
 
-def test_init(tables1, tables2):
-    assert True
+    assert 'Unauthorized' in response.errors[0].message
 
 
-def test_request_access_unauthorized(client, dataset1, env2, group2, group3):
+@pytest.mark.dependency()
```
degoldner, line 230: Regarding these test for the share api we refactored our forked version again into a setup where each test is independent of one another and instead the required state is constructed in each test with a cleanup after each tests which ensures that each test can run independent from another.

For this we first moved all the graphql queries/mutations into separate functions, so the calls can be reused in multiple test cases:

`def add_share_item(shareUri, itemUri, itemType, client, userName, groupUri):
    query = """
    mutation AddSharedItem($shareUri:String!,$input:AddSharedItemInput){
        addSharedItem(shareUri:$shareUri,input:$input){
            shareUri
            shareItemUri
            itemUri
            status
            action
        }
    }
    """

    response = client.query(
        query,
        username=userName,
        groups=[groupUri],
        shareUri=shareUri,
        input={
            "itemUri": itemUri,
            "itemType": itemType,
        },
    )

    print("Response from addSharedItem: ", response)

    return response`


We added a cleanup part to the share fixture:
`@pytest.fixture(scope="function")
def share1(env2, user, group, user2, group2, dataset1, share, client, db):
    share1 = share(
        environmentUri=env2.environmentUri, 
        userName=user2.userName, 
        groupUri=group2.groupUri, 
        datasetUri=dataset1.datasetUri
    )

    yield share1

    # Cleanup share
    # First try to direct delete
    delete_share_object_response = delete_share_object(shareUri=share1.shareUri, client=client, userName=user2.userName, groupUri=group2.groupUri)

    if delete_share_object_response.data.deleteShareObject == True:
        return

    # Then delete via revoke if still items present
    revoke_all(share1.shareUri, client, user2.userName, group2.groupUri)

    succesfull_processing_of_share_object(db, share1.shareUri)

    delete_share_object(
        shareUri=share1.shareUri,
        client=client, 
        userName=user2.userName, 
        groupUri=group2.groupUri
    )
`

tests/api/test_share.py
```diff
@@ -330,139 +509,393 @@ def test_list_dataset_share_objects(
...
-        },
-    )
-    assert 'UnauthorizedOperation' in response.errors[0].message
+    # Then shared item was added to share object in status PendingApproval
+    assert response.data.addSharedItem.shareUri == share1.shareUri
+    assert response.data.addSharedItem.status == \
+        dataall.api.constants.ShareItemStatus.PendingApproval.name
+
+
+@pytest.mark.dependency(depends=["test_add_share_item"])
```
degoldner, line 680: Regarding these test for the share api we refactored our forked version again into a setup where each test is independent of one another and instead the required state is constructed in each test with a cleanup after each tests which ensures that each test can run independent from another.

For this we first moved all the graphql queries/mutations into separate functions, so the calls can be reused in multiple test cases:

    def add_share_item(shareUri, itemUri, itemType, client, userName, groupUri):
    query = """
    mutation AddSharedItem($shareUri:String!,$input:AddSharedItemInput){
        addSharedItem(shareUri:$shareUri,input:$input){
            shareUri
            shareItemUri
            itemUri
            status
            action
        }
    }
    """

    response = client.query(
        query,
        username=userName,
        groups=[groupUri],
        shareUri=shareUri,
        input={
            "itemUri": itemUri,
            "itemType": itemType,
        },
    )

    print("Response from addSharedItem: ", response)

    return response


We added a cleanup part to the share fixture:

     @pytest.fixture(scope="function")
     def share1(env2, user, group, user2, group2, dataset1, share, client, db):
     share1 = share(
         environmentUri=env2.environmentUri, 
         userName=user2.userName, 
         groupUri=group2.groupUri, 
         datasetUri=dataset1.datasetUri
     )

    yield share1

    # Cleanup share
    # First try to direct delete
    delete_share_object_response = delete_share_object(shareUri=share1.shareUri, client=client, 
    userName=user2.userName, groupUri=group2.groupUri)

    if delete_share_object_response.data.deleteShareObject == True:
        return

    # Then delete via revoke if still items present
    revoke_all(share1.shareUri, client, user2.userName, group2.groupUri)

    succesfull_processing_of_share_object(db, share1.shareUri)

    delete_share_object(
        shareUri=share1.shareUri,
        client=client, 
        userName=user2.userName, 
        groupUri=group2.groupUri
    )

Finally the test cases are setting up there own required test state:

     def test_submit_share_request_with_share_item(share1, table1, client, user2, group2):
     # Given
     # Existing share object in status Draft (-> fixture share1)
     # and share item added
     add_share_item(share1.shareUri, table1.tableUri, dataall.api.constants.ShareableType.Table.name, client, 
     user2.userName, group2.groupUri)

     get_share_object_response = get_share_object(
        share1.shareUri,
        client,
        user2.userName,
        group2.groupUri,
    )

    assert get_share_object_response.data.getShareObject.status == 
    dataall.api.constants.ShareObjectStatus.Draft.name

    shareItem = get_share_object_response.data.getShareObject.get("items").nodes[0]
    assert shareItem.status == dataall.api.constants.ShareItemStatus.PendingApproval.value
    assert get_share_object_response.data.getShareObject.get("items").count == 1

    # When
    # Submit share object
    submit_share_object_response = submit_share_object(
        share1.shareUri,
        client,
        user2.userName,
        group2.groupUri,
    )
    # Then ....

dlpzx: Hi @degoldner thank you for the suggestion! Would you mind open a PR to this branch for this? Then I will have a look at the changes. I will continue with the remaining implementation tasks today/tomorrow, let's see if we can have this done by this week or early next week

dlpzx:
@noah-paige @degoldner Except for the refractoring of the tests (which now work with the previous setup) the rest is ready for review.

dlpzx:
@noah-paige @degoldner 
Fixed issues:
- failure and cleanup logic
Found issues:
1. folder sharing failure when folder is shared with a second group
2. pagination issue in revoke items window for more than 10 items
3. refractoring tests (optional)

dlpzx:
After fixing:
- folder sharing failure when folder is shared with a second group --> it was due to the naming convention needed for access points that did not allow "_"
- pagination issue in revoke items window for more than 10 items --> filtering defaults

Today I will work on:
- improving UI for revoke modal so that it refreshes the table of items
- improving testing of share API and make the tests independent form each other

dlpzx:
@degoldner I refractored the sharing api tests. Thank you for pointing it out, I like this last version better :)

noah-paige (approved):
I have gone through testing all the new and existing sharing functionality for folders and tables - all seems to be working well and really enjoy the new design and sharing process leveraging the Share SMs

I opened 2 more general issues that I found that are not specific to sharing ( #288 and #287 ) approving this PR to be merged to v1m4m0 branch

Issue #283, opened 2023-01-27
Add visibility for shared tables and folder consumption


dlpzx:
**Is your idea related to a problem? Please describe.**
It is difficult to know how to consume data once I have requested access to tables and folders

**Describe the solution you'd like**
I would like a copy/paste command to the access points of the shared folders and the clear information of the shared database that I can use to query shared tables.

noah-paige:
good to close this out @dlpzx ?

PR #289, opened 2023-02-01
Visibility for shared tables and folder consumption

dlpzx:
### Feature or Bugfix
- Feature

### Detail
- Added "shared" db name and access point names in the UI
- Added "shared" db name and access point names as part of the shareObject graphQL schema
The result is that users can view relevant information to consume data that has been shared with them directly from the shares view. They can copy to their clipboards a command to list the objects inside a shared folders using the AWS CLI or they can copy an SQL statement to use in Athena and query shared tables.

image

### Relates
- #283 


noah-paige (approved):
looks good!

PR #259, opened 2022-12-16
219 pipeline filter

noah-paige:
### Feature or Bugfix
- Feature

### Detail
- Add filter to pipeline list view
- Filter by region (of CICD Env), devStrategy, and tags 

Screen Shot 2022-12-16 at 3 24 32 PM

### Relates
- [#219 ](https://github.com/awslabs/aws-dataall/issues/219)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #290, opened 2023-02-02
V1.4.0 Features

dlpzx:
### Feature or Bugfix
- v.1.4.0 Features

### Detail
- #267 - Sharing State Machines
- #289 - Visibility of shared access points and databases
- #259  - Pipeline filter to search them


backend/dataall/db/api/share_object.py
```diff
@@ -309,28 +604,61 @@ def reject_share_object(
...
+            Item_SM.update_state(session, share.shareUri, new_state)
+
+        Share_SM.update_state(session, share, new_share_state)
+
+        ResourcePolicy.delete_resource_policy(
+            session=session,
+            group=share.groupUri,
+            resource_uri=dataset.datasetUri,
+        )
+        api.Notification.notify_share_object_approval(session, username, dataset, share)
```
noah-paige, line 625: Think this should be calling `notify_share_object_rejection` rather than `notify_share_object_approval`

dlpzx: You are right, let me review the notifications. I have also noticed that in general the notifications can be improved. They do not show the principal of the share just the dataset that has been approved. I will open another issue for this.

dlpzx: https://github.com/awslabs/aws-dataall/issues/291 Here is the issue for notifications

backend/migrations/versions/509997f0a51e_sharing_state_machines_v1_4_0.py
```diff
@@ -0,0 +1,163 @@
+"""sharing_state_machines_v1_4_0
+
```
chamcca, line 2: is this file supposed to be included?

dlpzx: Yes, why?

backend/dataall/db/api/dataset.py
```diff
@@ -15,14 +15,22 @@
...
-from ..models.Enums import Language, ConfidentialityClassification
+from ..models.Enums import Language, ConfidentialityClassification, ShareObjectStatus, ShareItemStatus
 from ...utils.naming_convention import (
     NamingConventionService,
     NamingConventionPattern,
 )
 
 logger = logging.getLogger(__name__)
 
+SHARE_ITEM_SHARED_STATES = [
```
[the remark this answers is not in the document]
dlpzx: Good idea. I have added it as part of the Share items state machine class in `backend/dataall/db/api/share_object.py`. I had to resolve some circular dependencies, but it looks good now.

backend/dataall/tasks/data_sharing/share_managers/lf_share_manager.py
```diff
@@ -0,0 +1,551 @@
...
+            f'Revoking source table access '
+            f'on {self.source_environment.AwsAccountId}/{self.dataset.GlueDatabaseName}/{table.GlueTableName} '
+            f'for principal {self.env_group.environmentIAMRoleArn}'
+        )
+        LakeFormation.revoke_source_table_access(
+            target_accountid=self.target_environment.AwsAccountId,
+            region=self.target_environment.region,
+            source_database=self.dataset.GlueDatabaseName,
+            source_table=table.GlueTableName,
+            target_principal=self.env_group.environmentIAMRoleArn,
```
noah-paige, line 361: Should we be using the share principal IAM Role here as the `target_principal` rather than EnvGroup IAM Role

dlpzx: Yes, during the testing I also ran into this issue when revoking to principals that are not an environment admin group (e.g. when testing consumption roles). O started working on a fix yesterday

dlpzx:
In the validation another intermittent error appeared: for folder sharing if it was the first folder shared we create an access point and then we attach a policy. In some cases the access point does not have enough time to be created before we attach a policy. Then, this first folder goes to "share_failed", when retrying a second time it works as expected. But to avoid this issue I have added a 5 seconds waiting time after we call the createAccessPoint API call.

noah-paige (approved):
changes look good to me - approving!

PR #292, opened 2023-02-03
Support OpenSearch Serverless

kukushking:
### Feature or Bugfix
- Feature

### Detail
- Add OpenSearch Serverless stack & corresponding feature flag
- Update client request signature ("aoss"/"es") and create service SSM parameter
- Update domain / collection references
- Add ECS task role to OpenSearch Serverless principals
- Update CDK library to latest (2.61.1)
- Update Pipeline CodeBuild image to `AMAZON_LINUX_2_4` with latest node version to use latest CDK and remove unnecessary packages
- Pass VPC endpoints security group to OpenSearch Serverless stack
- Update template_cdk.json
- Add monitoring


deploy/stacks/monitoring.py
```diff
@@ -131,11 +137,19 @@ def create_cw_alarms(
         self.set_aurora_alarms(
             f'{resource_prefix}-{envname}-aurora-alarm', database, self.cw_alarm_action
         )
-        self.set_es_alarms(
-            f'{resource_prefix}-{envname}-opensearch-alarm',
-            openseach_domain,
-            self.cw_alarm_action,
-        )
+        if openseach_domain:
```
dlpzx, line 140: What about when there is no opensearch_domain? no alarm?

deploy/stacks/monitoring.py
```diff
@@ -108,6 +112,8 @@ def create_cw_alarms(
         lambdas,
         database,
         openseach_domain,
```
dlpzx, line 114: this is a typo right? opensearch

kukushking: Yes, thanks for the catch!

deploy/stacks/monitoring.py
kukushking: If there is no `opensearch_domain`, there should be `opensearch_serverless_collection_id` that has separate monitoring, including alarms on availability of the collection (collections are domains in serverless). The reason I separated the two is because serverless provides a different set of metrics. 

I can either put serverless monitoring into else clause here or raise an error if none of these two parameters are provided - whichever works best for you?

template_cdk.json
```diff
@@ -31,7 +31,8 @@
         "prod_sizing": "boolean_SET_INFRA_SIZING_TO_PROD_VALUES_IF_TRUE|DEFAULT=true",
         "enable_cw_rum":  "boolean_SET_CLOUDWATCH_RUM_APP_MONITOR|DEFAULT=false",
         "enable_cw_canaries": "boolean_SET_CLOUDWATCH_CANARIES_FOR_FRONTEND_TESTING|DEFAULT=false",
-        "shared_dashboards_sessions": "string_TYPE_SESSION_SHARED_DASHBOARDS|(reader, anonymous) DEFAULT=anonymous"
+        "shared_dashboards_sessions": "string_TYPE_SESSION_SHARED_DASHBOARDS|(reader, anonymous) DEFAULT=anonymous",
+        "enable_opensearch_serverless": "boolean_USE_OPENSEARCH_SERVERLESS|DEFAULT=false"
```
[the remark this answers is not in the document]
kukushking: We did discuss it with @dlpzx and decided to keep it backwards-compatible and rather opt-in using a feature flag. Besides, at the time of the PR publication AOSS wasn't even GA yet. Happy to discuss this further though.

deploy/stacks/backend_stack.py
```diff
@@ -251,21 +253,25 @@ def __init__(
...
+            "ecs_security_groups": ecs_stack.ecs_security_groups,
+            "ecs_task_role": ecs_stack.ecs_task_role,
+            "prod_sizing": prod_sizing,
             **kwargs,
-        )
+        }
+        if enable_opensearch_serverless:
+            opensearch_stack = OpenSearchServerlessStack(self, f'OpenSearchServerless', **opensearch_args)
+        else:
+            opensearch_stack = OpenSearchStack(self, f'OpenSearch', **opensearch_args)
```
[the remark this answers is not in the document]
kukushking: Thanks!

deploy/stacks/opensearch.py
```diff
@@ -20,8 +21,10 @@ def __init__(
         envname='dev',
         resource_prefix='dataall',
         vpc: ec2.Vpc = None,
+        vpc_endpoints_sg: ec2.SecurityGroup = None,
         lambdas=None,
         ecs_security_groups: [ec2.SecurityGroup] = None,
+        ecs_task_role: Optional[iam.Role] = None,
```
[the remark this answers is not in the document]
kukushking: Not in server-full stack, no.

deploy/stacks/monitoring.py
```diff
@@ -424,6 +438,45 @@ def set_es_alarms(self, alarm_name, domain_name, cw_alarm_action):
             cw_alarm_action,
         )
 
+    def set_aoss_alarms(self, alarm_name, collection_id, collection_name, cw_alarm_action):
+        self._set_aoss_alarm(
+            collection_id,
+            collection_name,
+            f'{alarm_name}-collection-ActiveCollection',
+            'ActiveCollection',
+            1,
```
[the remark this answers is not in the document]
kukushking: Agreed, thanks.

template_cdk.json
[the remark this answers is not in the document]
nikpodsh: Makes sense. Let's keep it as it is. Thanks for the clarification!

deploy/stacks/backend_stack.py
```diff
@@ -251,21 +253,25 @@ def __init__(
...
             ],
-            ecs_security_groups=ecs_stack.ecs_security_groups,
-            prod_sizing=prod_sizing,
+            "ecs_security_groups": ecs_stack.ecs_security_groups,
+            "ecs_task_role": ecs_stack.ecs_task_role,
+            "prod_sizing": prod_sizing,
             **kwargs,
-        )
+        }
+        if enable_opensearch_serverless:
```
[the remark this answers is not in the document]
kukushking: I've implemented your suggestion with the only exception which is `create_opensearch_serverless_stack` is still called by the backend stack based on the passed feature flag. We do follow the same approach for other feature flags as well. The main reason is that I want backend stack to be self-contained to be able to synth/deploy it w/o having to rely on the stage having to call either of the opensearch methods.  Let me know if that works!

deploy/stacks/opensearch.py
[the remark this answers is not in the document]
kukushking: Removed

deploy/stacks/monitoring.py
kukushking: Updated

deploy/stacks/backend_stack.py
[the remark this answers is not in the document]
nikpodsh: Sure, looks good! Thank you for addressing it.

deploy/stacks/pipeline.py
```diff
@@ -207,11 +207,9 @@ def __init__(
                 'Synth',
                 input=source,
                 build_environment=codebuild.BuildEnvironment(
-                    build_image=codebuild.LinuxBuildImage.AMAZON_LINUX_2_3,
+                    build_image=codebuild.LinuxBuildImage.STANDARD_6_0,
                 ),
                 commands=[
-                    'n 16.15.1',
-                    'yum -y install shadow-utils wget && yum -y install openssl-devel bzip2-devel libffi-devel postgresql-devel',
```
nikpodsh, line 214: Looking into https://github.com/aws/aws-codebuild-docker-images/blob/master/ubuntu/standard/6.0/Dockerfile I couldn't find `shadow-utils` (should be `libvshadow-utils` in Ubuntu) and something to related to `postgre` (since we are deleting `postgresql-devel`). Are they shipped with `STANDARD_6_0`?
I am not sure if we use these tools at all at the moment , but wanted to double check with you that we removed them intentionally and there is no need for something like `apt-get install -y libvshadow-utils` 🙂

deploy/stacks/backend_stack.py
```diff
@@ -309,6 +318,23 @@ def __init__(
...
+            cw_alarm_action=self.monitoring_stack.cw_alarm_action,
+        )
+
+    def create_opensearch_serverless_stack(self):
+        aoss_stack = OpenSearchServerlessStack(self, 'OpenSearchServerless', **self.opensearch_args)
+        self.monitoring_stack.set_aoss_alarms(
+            alarm_name=f'{self.resource_prefix}-{self.envname}-opensearch-serverless-alarm',
+            collection_id=aoss_stack.collection_id,
+            collection_name=aoss_stack.collection_name,
+            cw_alarm_action=self.monitoring_stack.cw_alarm_action,
```
nikpodsh, line 339: Super minor: looks like `set_aoss_alarms` is called only from this place and we don't need `set_aoss_alarms` to have a parameter `cw_alarm_action` since we can use it directly from `monitoring_stack`

deploy/stacks/pipeline.py
kukushking: Tbh I wasn't sure if those are needed as well. This specific CodeBuild is only responsible for running `cdk synth` so they seemed unnecessary. It's worth to double check with @dlpzx though.

On a related note, since I've updated this one, might as well upgrade other CBs to `STANDARD_6_0` but I will wait for the confirmation first.

deploy/stacks/backend_stack.py
kukushking: Thanks, resolved

nikpodsh (approved):
Since [this](https://github.com/awslabs/aws-dataall/pull/292#discussion_r1102877558) seems to work fine, I think we can go without it:)
Approved.  Great job! 👍

kukushking:
Thanks for the review @nikpodsh!

deploy/stacks/pipeline.py
dlpzx: The reason behind installing node directly in CodeBuild is that node16 is needed for CDK2 and it was not available at runtime in `AMAZON_LINUX_2_3` images. With `AMAZON_LINUX_2_4` node16 is installed at runtime so we can remove those lines. At first sight I would prefer to use `AMAZON_LINUX_2_4` instead of Ubuntu to avoid changes that could cause errors (plus documentation changes), but maybe I am missing something.

I am also testing the libraries that we are installing ('yum -y install shadow-utils wget && yum -y install openssl-devel bzip2-devel libffi-devel postgresql-devel') caused by the networking changes imposed in #307.

dlpzx (commented):
Before merging we can include what we discussed yesterday around existing customers that want to take advantage of this feature.

How do we migrate the information from OpenSearch non-serverless to serverless. At least we need to provide some info

deploy/stacks/pipeline.py
kukushking: Thanks @dlpzx, makes sense. Let me give `AMAZON_LINUX_2_4` a try. I'll also keep the libraries to reduce blast radius.

kukushking: Upgraded all CBs to `AMAZON_LINUX_2_4` & performed a test deployment - all good.

kukushking: @dlpzx all of the packages are already in 4.0 image so we're good removing them. Removed & ran through the deployment pipeline to verify 👍

dlpzx:
Thank you for cleaning up the unnecessary packages from the CICD pipeline! The testing in v1m5m0 branch is ready, so you can go ahead and merge your changes. Great contribution @kukushking 🚀

Issue #287, opened 2023-01-31
Worksheet is showing Database Name N duplicate times for N number of tables shared within the DB


noah-paige:
### Describe the bug

If I have N # of tables shared to a consumer environment that all exists in the same data all dataset (i.e. Glue DB) - when querying that data using data all Worksheets the `Database` drop down shows the same database name N # of times (one for each table shared)

### How to Reproduce

```
*P.S. Please do not attach files as it's considered a security risk. Add code snippets directly in the message body as much as possible.*
```

1. Create a Data All Dataset and add 2 or more Tables to that dataset
2. Share 2 or more tables from that 1 dataset to another env/group
3. Once the share is approved - create a Worksheet for the group the tables are shared too
4. Select the consumer Environment Name and Group --> Check the drop-down for `Database` to see the duplicate names

### Expected behavior

Have 1 DB name for 1 or more tables shared to the consumer env/group 

### Screenshots

![Screen Shot 2023-01-31 at 4 22 14 PM](https://user-images.githubusercontent.com/69586985/215886233-8395d279-6ac3-4046-afe9-16bf550e423d.png)

### OS

n/a

### Python version

n/a

### AWS data.all version

v1.3.1

PR #304, opened 2023-02-09
Make query distinct on dataaset for worksheet view

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Fix bug in worksheet

### Relates
- #287 

### DIFF
--- a/frontend/src/views/Worksheets/WorksheetView.js
+++ b/frontend/src/views/Worksheets/WorksheetView.js
@@ -161,7 +161,8 @@ const WorksheetView = () => {
             page: 1,
             pageSize: 10000,
             term: '',
-            uniqueDatasets: true
+            uniqueDatasets: true,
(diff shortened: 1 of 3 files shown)


frontend/src/views/Worksheets/WorksheetView.js
```diff
@@ -161,7 +161,7 @@ const WorksheetView = () => {
             page: 1,
             pageSize: 10000,
             term: '',
-            itemTypes: 'DatasetTable'
```
dlpzx, line 164: If we remove this line, what happens with folders? Aren't we going to have response nodes for the folders? I think it would only affect the case in which we have a dataset of folders and no tables

noah-paige: Good catch! Let me push a fix here

dlpzx (approved):
thank you!

Issue #321, opened 2023-02-20
Check if dataset storage location was already registered when importing datasets


dlpzx:
**Is your idea related to a problem? Please describe.**
[FORWARDED FROM CUSTOMER]
We create data pipelines and run crawlers to create tables, we use Lake Formation to grant access to ETL resources and crawlers to the tables that we are creating. The S3 location is registered in Lake Formation. 

Then we import the dataset in data.all, but the creation of the stack fails because it tries to register a data location that is already registered.

**Describe the solution you'd like**
I would like data.all to register the location only when it is needed. That means, first it should check if the S3 location is already registered and when it is not, then it should register it.

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

dlpzx:
I reviewed the [documentation](https://docs.aws.amazon.com/lake-formation/latest/dg/registration-role.html) to ensure this change would not affect sharing with data.all. 

From data.all we register the S3 location with the pivotRole. Lake Formation assumes the pivotRole to access the data in that S3 location. If the S3 location was previously registered with another IAM role the only difference is that under-the-hood Lake Formation uses a different role to access the data, but LF permissions work independently from the underlying IAM role that is assumed. 

In short, I confirm that we can use a pre-existing registration of the S3 location.

PR #324, opened 2023-02-21
Check existing Lake Formation data lake location and make dataset resources names unique

dlpzx:
### Feature or Bugfix
- Feature

### Detail
- When a dataset is imported, the dataset stack checks whether the specified S3 location has been registered with Lake Formation. If it has been already registered data.all does nothing, if it was not registered it registers it as a Lake Formation data location.
- For the Glue profiling job, the data quality job and the crawler a suffix with the datasetUri has been added to the name to track them better and to avoid issues in the future if we decide to implement "multiple databases mapping to one single Bucket" features.

### Relates
- #321 

### DIFF
--- a/backend/dataall/aws/handlers/lakeformation.py
+++ b/backend/dataall/aws/handlers/lakeformation.py
@@ -23,14 +23,15 @@ class LakeFormation:
 
             response = lf_client.describe_resource(ResourceArn=resource_arn)
 
-            log.debug(f'LF data location already registered: {response}')
+            log.info(f'LF data location already registered: {response}')
 
             return response['ResourceInfo']
 
         except ClientError as e:
-            log.error(
+            log.info(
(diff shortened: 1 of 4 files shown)


dlpzx:
Hi @noah-paige, I have tested it locally:
- try to import a dataset from an already registered S3 Bucket and check that it skipped the LF registration. Glue resources created successfully.
- try to create a new dataset and checked that it registered it in LF. Glue resources created successfully.
- try to update an existing dataset and checked Glue resources created successfully and LF registration unaffected.

noah-paige (approved):
looks good to me!

dlpzx:
I tested the functionality in a real deployment:
- try to import a dataset from an already registered S3 Bucket and check that it skipped the LF registration. Glue resources created successfully.
- try to create a new dataset and checked that it registered it in LF. Glue resources created successfully.
- try to delete the imported dataset and check that the LF location was still registered.

Issue #307, opened 2023-02-14
Restrict networking configuration in Tooling account


dlpzx:
**Is your idea related to a problem? Please describe.**
[FORWARDED REQUEST FROM CUSTOMER] Currently the CICD resources in the tooling account are hosted in a VPC whose NACL rules include an **Allow ALL traffic** inbound and outbound rule. In our organization this rule is removed automatically by our security teams and we cannot deploy data.all. This is because data.all needs to install packages from the internet and without this rule it cannot install node16 and npm and pip packages.

**Describe the solution you'd like**
Since data.all deploys VPC endpoints, I would like to use those directly and avoid access to the internet. Moreover, data.all deploys CodeArtifact repositories that we can use to install the necessary packages. Finally, node16 is now available at runtime for Linux:v4 images, so we can upgrade the CodeBuild images to use node16 and avoid installing it ourselves.

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

dlpzx:
Update on the investigations regarding this issue. There are a couple of enhancements identified:
1. Update CodeBuild Linux images to v4 (if possible with CDK) and use node on runtime instead of installing node in CodeBuild
2. Remove unnecessary installation of yum packages that are already installed in Linux v4
3. Ensure that we use CodeArtifact to install npm and pip packages instead of installing directly from the internet
4. (Maybe) Explore alternatives to safety when to run vulnerability scanning. For example we could use AWS Inspector

dlpzx:
Solved in #409

PR #337, opened 2023-03-01
Configurable restricted VPC for tooling account

dlpzx:
### Feature or Bugfix
- Feature

### Detail
- New cdk.json configuration parameter `tooling_vpc_restricted_nacl`
- If set to `true` we create a custom NACL for the tooling data.all created VPC with the following inbound rules. Outbound allows all traffic.

image

- We have enabled DNS private names for CodeArtifact endpoints, to correctly resolve them inside the VPC and keep traffic within the VPC.
- Modified some CodeBuild steps to install pip and npm packages from CodeArtifact instead of from the internet.
- In #323, CodeBuild Linux images were updated, eliminating the need to install yum packages and node.

### Relates
- #307 

### DIFF
--- a/deploy/stacks/vpc.py
+++ b/deploy/stacks/vpc.py
@@ -130,9 +130,9 @@ class VpcStack(pyNestedClass):
             nacl = ec2.NetworkAcl(
                 self, "RestrictedNACL",
                 vpc=self.vpc,
-                network_acl_name=f'{resource_prefix}-{envname}-restrictedNACL'
+                network_acl_name=f'{resource_prefix}-{envname}-restrictedNACL',
+                subnet_selection=ec2.SubnetSelection(subnets=self.vpc.private_subnets + self.vpc.public_subnets),
             )
-            #TODO: COMPLETE NACL RULES
             nacl.add_entry(
                 "entryOutbound",
                 cidr=ec2.AclCidr.any_ipv4(),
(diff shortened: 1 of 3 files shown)


dlpzx:
Hi @kukushking, I tested the above configuration in a fresh deployment and it completed successfully :)

deploy/stacks/vpc.py
```diff
@@ -125,6 +126,55 @@ def create_new_vpc(self, cidr, envname, resource_prefix):
...
+                "entryInboundCustomTCP",
+                cidr=ec2.AclCidr.any_ipv4(),
+                traffic=ec2.AclTraffic.tcp_port_range(start_port=1024, end_port=65535),
+                rule_number=102,
+                direction=ec2.TrafficDirection.INGRESS,
+                rule_action=ec2.Action.ALLOW
+            )
+            nacl.add_entry(
+                "entryInboundAllInVPC",
+                cidr=ec2.AclCidr.ipv4("10.0.0.0/16"),
```
kukushking, line 170: How about referring to`self.vpc.cidr` here just in case the user changes the cidr from the default?

dlpzx: Good idea! I just added it

dlpzx: Tested 👍

kukushking (approved):
Looks great!

PR #339, opened 2023-03-01
fix: Dashboard Quicksight button leads to an inaccessibility page

wolanlu:
### Feature or Bugfix

- Bugfix

### Detail
Dashboard Quicksight button leads to an inaccessibility page on every first hit

**WHEN** user hits button “View in Quicksight” on any dashboard in the platform

Expected Result:
**THEN** the user gets directed to the requested dashboard in a new full size browser tab

Actual Result:
**THEN** the user gets directed to a page that displays an error message: “Embedding failed because of invalid URL or authorization code. Both of these must be valid and the authorization code must not be expired for embedding to work.”

NOTE: When the page gets refreshed via the browser button, it will show the dashboard. But without any hint, users will not know or try this.

### Relates


dlpzx (approved):
lgtm, well spotted!

PR #350, opened 2023-03-03
fix pivot role logs actions privileges

dosiennik:
### Feature or Bugfix
- Bugfix

### Detail
For for setting up privileges for logs related actions. Currently due to the incorrect settings logs for Custom Resources and Glue are not visible.


dlpzx (approved):
lgtm. I checked and yes you are right lambda logs are not named aws-lambda

Issue #251, opened 2022-12-12
PivotRole stack as CDK stack


dlpzx:
**Is your idea related to a problem? Please describe.**
I'm always frustrated when new permissions are needed in the PivotRole and I have to update all my CloudFormation stacks (one in each environment account)

**Describe the solution you'd like**
I would like to implement some sort of automation outside of data.all to update this IAM role in a more effective way. For that I would like to have the CDK version of the PivotRole available as part of the repository as a reference

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

PR #355, opened 2023-03-08
feat: include iam pivot role as part of the environment cdk stack

dlpzx:
### Feature or Bugfix
- Feature

#### ⚠️ ⚠️ ⚠️ IMPORTANT ⚠️ ⚠️ ⚠️ 
IF YOU ARE MIGRATING FROM MANUAL TO CDK-CREATED ROLE, WHILE THE PIPELINE IS UPDATING THE BACKEND CODE, UPDATING ENVIRONMENT/DATASET STACKS AND UPDATING THE FRONTEND CODE YOU WILL EXPERIENCE DOWN TIMES. PLAN THE MIGRATION (AROUND 1-2H) IN A TIME WINDOW THAT DOES NOT IMPACT YOUR USERS.

#### Configuration of manual or cdk-created pivotRole
With this PR customers will be able to use data.all and link environments using a manually created pivotRole or a cdk-pivotRole deployed as part of the environment stack. What to use is configurable in the `cdk.json` for each infra account. For example, customers can have manually created pivotRoles in "dev" and cdk-created ones in "prod". The configuration affects all environments of the deployment. In this example, environments linked to "prod" will always deploy a cdk-PivotRole, it is not possible to have a mix of manual and cdk roles in the same infra deployment.

- Added configuration parameter in `cdk.json` to enable or disable pivotRole creation as part of environment in each development environment: 
 `"enable_pivot_role_auto_create": "boolean_ENABLE_PIVOT_ROLE_AUTO_CREATE_IN_ENVIRONMENT|DEFAULT=false",`

- Creation of SSM Parameter that stores this flag in each of the development environments. (Screenshot outdated, not the parameter is called `enablePivotRoleAutoCreate` to match the cdk.json parameter name
image

- Modification of the "dataall-pivotrole-name" Secret in central account. For pivotRole created manually it stays in "dataallPivotRole" for cdk-created ones "dataallPivotRole-cdk" to avoid issues with conflicting IAM roles.

- Added CDK nested stack for PivotRole IAM role and add it to the Environment stack. It is deployed depending on the value of the created SSM parameter. Permissions that reference the pivotRole are adapted to use the pivotRole-cdk when it is enabled.

- Conditional pre-requisites box in CreateEnvironment UI view depending on the value of the SSM Parameter. The value of the SSM parameter is written to the `.env` file of the React application and is used in the `frontend/src/views/Environments/EnvironmentCreateForm.js`.
(With `enable_pivot_role_auto_create: false`)
image

(With `enable_pivot_role_auto_create: true`)
image

#### Refactor AWS account checks
In the resolvers and in the environment stack deployment originally we performed some checks in the account with boto3 calls assuming that the pivotRole was created. In this PR some of this checks are now effectuated with the `cdk look up role` and other tests have been moved to other places in the code. In addition, more tests are added to consider all scenarios (e.g. check that pivotRole exists in manual-pivot role deployments).

- Modified check of CDKToolkit CFN stack to avoid using pivotRole
image

- Added check of pivotRole creation (for manual created pivotRole) with CDK look-up role.
image

- Added check_environment methods to the update_environment API call.

- Previous check of existing SageMaker Studio domain when MLStudio is enabled is now executed with cdk look up role
- Previous check of Quicksight subscription when Dashboards is enabled now happens after environment creation, when a Dataset is created or when a Dashboard-Quicksight session is started. 
- Previous creation of data.all Quicksight default group when Dashboards is enabled now happens after environment creation, when a Dataset is created or when a Dashboard-Quicksight session is started.

#### Migrating from manual to cdk-created pivotRole
This PR takes into consideration the scenario in which customers already have a deployment of data.all with environments linked using a manually created pivotRole and want to leverage cdk-created pivotRoles. 

To do that, customers just need to add the `enable_pivot_role_auto_create` parameter in their `cdk.json` configuration and set it to `true`. Once the CICD pipeline has completed: new linked environments will contain the nested cdk-pivotRole stack (no actions needed) and existing environments can be updated by:
a) manually, by clicking on "update stack" in the environment>stack tab
b) automatically, wait for the `stack-updater` ECS task that runs daily overnight
c) automatically, set the added `update_dataall_stacks_in_cicd_pipeline` parameter to `true` in the `cdk.json` config file. The `stack-updater` ECS task will be triggered from the CICD pipeline

- Added configuration parameter in `cdk.json` to add a CodeBuild stage in CICD pipeline to trigger update of environments and datasets stacks:
`"enable_update_dataall_stacks_in_cicd_pipeline": "boolean_ENABLE_UPDATE_DATAALL_STACKS_IN_CICD_PIPELINE|DEFAULT=false"`
This parameter can be set back to `false` whenever customers do not foresee many changes in the dataset and environment stacks and back to true when we want to force the update of stacks. This can be useful to ensure changes in the stacks are applied immediately instead of waiting for the stack-updater to run overnight.

image

- Modified `stacks-updater` ECS task to run environment updates first, wait until completion and then update dataset stacks. The reason of keeping this order is that the datasets of an environment use Lambda functions created in the environment stack as Custom resources (glue-db-handler). The original Lambda function needs to be updated to use the new pivotRole before being used in the Dataset stack. Plus, Quicksight group creation which happens in the dataset stack uses the pivotRole because the cdk-look-up role has no Quicksight permissions.

- Replaced the CFN `AWS::LakeFormation::Resource` by a Custom Resource and added the necessary 'lakeformation:De/registerResource` permissions to Pivot Role. Using the Cfn resource directly caused internal Failure errors. To have more control over the update we now execute it as a custom resource (in the same way as we do with the glue-db-handler). 

image

#### ⚠️ ⚠️ ⚠️ Migration risks ⚠️ ⚠️ ⚠️ 

Removing the AWS::LakeFormation::Resource resource and replacing it with a Custom Resource within the same/single update of the dataset stack causes a race condition since:
1) deleting AWS::LakeFormation::Resource triggers deletion of the data location resource from Lake Formation.
2) adding a custom resource or actually replacing AWS::LakeFormation::Resource with a custom resource creates or updates the data location resource.

Then the output of this update of the dataset stack depends on the order in which 1 & 2 occurs, in the worst case 2 happens before 1, which results in deletion of the Lake Formation data location resource. This is explained because both AWS::LakeFormation::Resource and Custom Resource point to the same DataLakeLocation which can be registered just once.

This error will be solved in a second update of stacks, in which the 'on_update' event of the datalakelocation handler checks if the location was registered in Lake Formation and creates the corresponding data location resource if it is missing. A second update can be triggered manually, by going to the UI, selecting the dataset and clicking on update stack in the stack tab. If it is not triggered manually, an scheduled task runs this update daily.

#### Documentation
- Userguide changes in environment section
- GitHub pages changes (in a different PR)

#### Additional enhancements
- For the dataset custom resources, now the provider creation has been moved to the environment stack avoiding [Lambda policy size limitations](https://repost.aws/knowledge-center/lambda-resource-based-policy-size-error), the SSM parameters for the Lambda arn and name are left for reference if needed.
- Modified Default Lake Formation settings custom resource in environment stack: added validation of IAM roles when adding LakeFormation data lake admins, if the IAM roles do not exist, data.all removes them from data lake admins to avoid errors. It now removes pivotRole (manual or cdk) from LakeFormation data lake admins on environment stack deletion.
- Make team IAM roles policies unique by adding environmentUri to their policy names
- bugfix: [Fix local imports of shareItemSM for Docker and quicksight properties](https://github.com/awslabs/aws-dataall/commit/527f5619b6f35324e85d961b180d9114afc4e6ff)
- Isolated SageMaker Studio domain stack in a different nested stack in the Environment stack for readability

### Relates
- #251 


backend/dataall/aws/handlers/cloudformation.py
```diff
@@ -16,16 +16,16 @@ def __init__(self):
...
-    def check_existing_cdk_toolkit_stack(AwsAccountId, region):
-        cfn = CloudFormation.client(AwsAccountId=AwsAccountId, region=region)
+    def check_existing_cdk_toolkit_stack(AwsAccountId, region, role=None):
         try:
+            cfn = CloudFormation.client(AwsAccountId=AwsAccountId, region=region, role=role)
             response = cfn.describe_stacks(StackName='CDKToolkit')
-        except cfn.exceptions.ClientError as e:
+        except ClientError as e:
             print(e)
             raise Exception('CDKToolkitNotFound')
```
dosiennik, line 30: although you didn't touch this place in the code if I take a look at it closer there is a dead code:

` stacks = response['Stacks']
        if not len(stacks):
            raise Exception('CDKToolkitNotFound')`

This code will never be executed since if the stack with name 'CDKToolkit' doesn't exist describe_stacks always throws an error, it will not return an empty stacks list.

Have a look at the note in here: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/cloudformation/describe-stacks.html

'If the stack doesn’t exist, an ValidationError is returned.'

backend/dataall/aws/handlers/parameter_store.py
```diff
@@ -16,10 +16,12 @@ def __init__(self):
         pass
 
     @staticmethod
-    def client(AwsAccountId=None, region=None):
+    def client(AwsAccountId=None, region=None, role=None):
         if AwsAccountId:
-            session = SessionHelper.remote_session(AwsAccountId)
+            print(f"SSM Parameter remote session with role:{role if role else 'PivotRole'}")
```
dosiennik, line 21: can we use logger for it?

backend/dataall/aws/handlers/sts.py
```diff
@@ -173,9 +155,18 @@ def get_delegation_role_arn(cls, accountid):
...
+
+    @classmethod
+    def get_cdk_look_up_role_arn(cls, accountid, region):
+        """Returns the name that will be assumed to perform IAM actions on a given AWS accountid using CDK Toolkit role
+        Args:
+            accountid(string) : aws account id
+        Returns:
+                string : arn of the CDKToolkit role on the target aws account id
+        """
+        print(f"Getting CDK look up role: arn:aws:iam::{accountid}:role/cdk-hnb659fds-lookup-role-{accountid}-{region}")
```
dosiennik, line 168: can we use logger for it?

backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
```diff
@@ -25,6 +25,9 @@ def on_event(event, context):
 
 
 def on_create(event):
+    """Creates if it does not exist Glue database for the data.all Dataset
```
dosiennik, line 28: just a note here for the future, using custom resource for glue db is not needed anymore, there is a construct for it: https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-glue-alpha.Database.html

backend/dataall/cdkproxy/assets/lakeformationdefaultsettings/index.py
```diff
@@ -8,6 +8,21 @@ def clean_props(**props):
     return data
 
 
+def validate_principals(principals):
+    iam_client = boto3.client('iam')
+    validated_principals = []
+    for principal in principals:
+        if ":role/" in principal:
+            print(f'Principal {principal} is an IAM role, validating....')
```
dosiennik, line 16: I have noticed print in multiple places. I think we need to replace it with the logger

backend/dataall/cdkproxy/assets/lakeformationdefaultsettings/index.py
```diff
@@ -66,4 +85,43 @@ def on_update(event):
 
 
 def on_delete(event):
-    pass
+    """"Removes the PivotRole from the existing Data Lake Administrators
+    Before adding any principal, it validates it exists if it is an IAM role
+    """
+    AWS_ACCOUNT = os.environ.get('AWS_ACCOUNT')
```
dosiennik, line 91: I think we can also move the code for reading aws account and region env variables in each method to do it outside of the  handler to do it just once and not repeat ourself (DRY)

tests/api/test_datapipelines.py
```diff
@@ -22,6 +25,9 @@ def env2(env, org1, user, group, tenant, module_mocker):
     module_mocker.patch(
         'dataall.api.Objects.Environment.resolvers.check_environment', return_value=True
     )
+    module_mocker.patch(
+        'dataall.api.Objects.Environment.resolvers.get_pivot_role_as_part_of_environment', return_value=False
```
dosiennik, line 29: I see we repeat it in so many places, maybe we can mock just somewhere in one place like a fixture with autoUse set to true?

frontend/src/api/Environment/getPivotRoleAsPartOfEnvironment.js
```diff
@@ -0,0 +1,11 @@
+import { gql } from 'apollo-boost';
+
+const getPivotRoleAsPartOfEnvironment = () => ({
```
dosiennik, line 3: do we need to expose a dedicated query for it? maybe we should think about something more generic like query for returning the config?
Another option we can use 'REACT_APP_' variables for React as we do for other stuff:
REACT_APP_GRAPHQL_API={graphql_api_url}
REACT_APP_SEARCH_API={search_api_url}
REACT_APP_COGNITO_USER_POOL_ID={user_pool_id}
REACT_APP_COGNITO_APP_CLIENT_ID={app_client}
REACT_APP_COGNITO_DOMAIN={domain}
REACT_APP_COGNITO_REDIRECT_SIGNIN=https://{signin_singout_link}
REACT_APP_COGNITO_REDIRECT_SIGNOUT=https://{signin_singout_link}
REACT_APP_COGNITO_HOSTED_UI_ENABLED={'true' if cognito_hosted_ui_enabled == 'True' else 'false'}
REACT_APP_COGNITO_AUTHENTICATION_PROVIDER_NAME={cognito_authentication_provider_name}
REACT_APP_USERGUIDE_LINK=https://{user_guide_link}

template_cdk.json
```diff
@@ -32,7 +32,10 @@
         "enable_cw_rum":  "boolean_SET_CLOUDWATCH_RUM_APP_MONITOR|DEFAULT=false",
         "enable_cw_canaries": "boolean_SET_CLOUDWATCH_CANARIES_FOR_FRONTEND_TESTING|DEFAULT=false",
         "shared_dashboards_sessions": "string_TYPE_SESSION_SHARED_DASHBOARDS|(reader, anonymous) DEFAULT=anonymous",
-        "enable_opensearch_serverless": "boolean_USE_OPENSEARCH_SERVERLESS|DEFAULT=false"
+        "enable_quicksight_monitoring": "boolean_ENABLE_CONNECTION_QUICKSIGHT_RDS|DEFAULT=false",
+        "enable_opensearch_serverless": "boolean_USE_OPENSEARCH_SERVERLESS|DEFAULT=false",
+        "pivot_role_as_part_of_environment": "boolean_CREATE_PIVOT_ROLE_AS_PART_OF_ENVIRONMENT_STACK|DEFAULT=false",
```
dosiennik, line 37: naming is hard :) it may be a personal opinion but I am not a big fan of this name, since it doesn't say anything if the pivot role is created automatically or manually, it just mentions that the pivot role is part of the environment stack

Did you consider other names for this parameter? For some params I see that we use a convention of 'enable_something'. Maybe we should think about a consistency here:
- "enable_pivot_role_auto_create"
or something else

backend/dataall/tasks/stacks_updater.py
```diff
@@ -18,32 +19,39 @@
...
+def update_stack(session, envname, target_uri, wait=False):
     stack: models.Stack = db.api.Stack.get_stack_by_target_uri(
         session, target_uri=target_uri
     )
     cluster_name = Parameter().get_parameter(env=envname, path='ecs/cluster/name')
-    if not Ecs.is_task_running(cluster_name, f'awsworker-{stack.stackUri}'):
-        stack.EcsTaskArn = Ecs.run_cdkproxy_task(stack.stackUri)
+    if not Ecs.is_task_running(cluster_name=cluster_name, started_by=f'awsworker-{stack.stackUri}'):
+        stack.EcsTaskArn = Ecs.run_cdkproxy_task(stack_uri=stack.stackUri)
+        if wait:
```
dosiennik, line 45: looks like a very custom solution/approach, in general as we already discussed we need to refactor the way how we  handle starting all of the ecs tasks, to queue them for instance and have a better control over the tasks run in parallel to have a better control of ENI/IPs consumed etc

backend/dataall/tasks/data_sharing/share_managers/lf_share_manager.py
```diff
@@ -60,11 +60,14 @@ def get_share_principals(self) -> [str]:
...
         if self.target_environment.dashboardsEnabled:
-            q_group = Quicksight.get_quicksight_group_arn(
-                self.target_environment.AwsAccountId
+            group = Quicksight.create_quicksight_group(
+                self.target_environment.AwsAccountId, Quicksight.DEFAULT_GROUP_NAME
             )
-            if q_group:
-                principals.append(q_group)
+            if group and group.get('Group', {}).get('Arn'):
+                q_group = group.get('Group', {}).get('Arn')
```
dosiennik, line 67: group and q_group make the code less readable, can we name it group_arn in here?

backend/dataall/api/Objects/Environment/queries.py
```diff
@@ -11,6 +11,12 @@
     test_scope='Environment',
 )
 
+getCreatePivotRole = gql.QueryField(
```
dosiennik, line 14: I commented it somewhere below, I feel like we don't need to expose an additional query for it

backend/dataall/api/Objects/Environment/resolvers.py
```diff
@@ -30,16 +31,38 @@ def get_trust_account(context: Context, source, **kwargs):
     return current_account
 
 
-def check_environment(context: Context, source, input=None):
+def get_pivot_role_as_part_of_environment(context: Context, source, **kwargs):
+    ssm_param = ParameterStoreManager.get_parameter_value(region=os.getenv('AWS_REGION', 'eu-west-1'), parameter_path=f"/dataall/{os.getenv('envname', 'local')}/pivotRole/createdAsPartOfEnvironmentStack")
```
dosiennik, line 35: I see in multiple places in the code returning a default region as Ireland? do we really need it?

backend/dataall/api/Objects/Environment/resolvers.py
```diff
@@ -30,16 +31,38 @@ def get_trust_account(context: Context, source, **kwargs):
...
-    if input.get('dashboardsEnabled'):
-        existing_quicksight = Quicksight.check_quicksight_enterprise_subscription(AwsAccountId=account)
+    pivot_role_arn = SessionHelper.get_delegation_role_arn(accountid=account_id)
+    cdk_look_up_role_arn = SessionHelper.get_cdk_look_up_role_arn(
+        accountid=account_id, region=region
+    )
+    cdk_role_name = CloudFormation.check_existing_cdk_toolkit_stack(
+        AwsAccountId=account_id, region=region, role=cdk_look_up_role_arn
+    )
+    if pivot_role_as_part_of_environment == False:
```
dosiennik, line 58: can be personal, not a big fan comparing to == False, can we just do `if not `?

backend/dataall/api/Objects/Environment/resolvers.py
```diff
@@ -30,16 +31,38 @@ def get_trust_account(context: Context, source, **kwargs):
...
     if ENVNAME == 'pytest':
         return 'CdkRoleName'
-    account = input.get('AwsAccountId')
-    region = input.get('region')
-    cdk_role_name = CloudFormation.check_existing_cdk_toolkit_stack(AwsAccountId=account, region=region)
 
-    if input.get('dashboardsEnabled'):
-        existing_quicksight = Quicksight.check_quicksight_enterprise_subscription(AwsAccountId=account)
+    pivot_role_arn = SessionHelper.get_delegation_role_arn(accountid=account_id)
+    cdk_look_up_role_arn = SessionHelper.get_cdk_look_up_role_arn(
```
dosiennik, line 52: this is a technical detail how we check the existence of the cdk_toolkit_stack, can be encapsulated/moved to the CloudFormation.check_existing_cdk_toolkit_stack method

backend/dataall/aws/handlers/quicksight.py
```diff
@@ -11,10 +11,11 @@
 logger = logging.getLogger('QuicksightHandler')
 logger.setLevel(logging.DEBUG)
 
-DEFAULT_GROUP_NAME = 'dataall'
-
 
 class Quicksight:
+
+    _DEFAULT_GROUP_NAME = 'dataall'
```
dosiennik, line 17: for the future we can make the default group name configurable

backend/dataall/cdkproxy/stacks/dataset.py
```diff
@@ -389,53 +309,94 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
                 iam.ArnPrincipal(
-                    f'arn:aws:iam::{dataset.AwsAccountId}:role/{pivot_role_name}'
+                    f'arn:aws:iam::{dataset.AwsAccountId}:role/{self.pivot_role_name}'
                 ),
             ),
         )
         dataset_admin_policy.attach_to_role(dataset_admin_role)
 
-        glue_db_handler_arn = ssm.StringParameter.from_string_parameter_name(
+        # Datalake location custom resource: registers the S3 location in LakeFormation
```
dosiennik, line 318: I would be happy if we remove it to keep the code clean, we can always find it in the history once the bug is fixed and we need to revert it back

backend/dataall/cdkproxy/stacks/sagemakerstudio.py
```diff
@@ -1,21 +1,128 @@
...
+    def check_sagemaker_studio(self, environment: models.Environment):
+        logger.info('check sagemaker studio domain creation')
+        try:
+            cdk_look_up_role_arn = SessionHelper.get_cdk_look_up_role_arn(
+                accountid=environment.AwsAccountId, region=environment.region
+            )
+            dataall_created_domain = ParameterStoreManager.client(
+                AwsAccountId=environment.AwsAccountId, region=environment.region, role=cdk_look_up_role_arn
+            ).get_parameter(Name=f'/dataall/{environment.environmentUri}/sagemaker/sagemakerstudio/domain_id')
+            return None
```
dosiennik, line 45: question: why do we explicitly return None here?

backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -502,93 +521,47 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
+        if self._environment.mlStudiosEnabled:
+            # Create dependency group - Sagemaker depends on group IAM roles
+            sagemaker_dependency_group = DependencyGroup()
+            sagemaker_dependency_group.add(default_role)
+            for group_role in group_roles:
+                sagemaker_dependency_group.add(group_role)
+            vpc_id = None
+            subnet_ids = []
             try:
+                print("looking for default VPC")
```
dosiennik, line 538: logger

backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -378,28 +370,55 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
-            properties={
-                'DataLakeAdmins': [
-                    f'arn:aws:iam::{self._environment.AwsAccountId}:role/{self.pivot_role_name}',
-                ]
+            "DatalakeLocationCustomResourceHandler",
+            function_name=f'{self._environment.resourcePrefix}-datalakelocation-handler-{self._environment.environmentUri}',
+            role=self.pivot_role,
+            handler="index.on_event",
+            code=_lambda.Code.from_asset(entry_point),
+            memory_size=1664,
```
dosiennik, line 390: just a general question, how memory and timeout settings were selected for all this custom resource lambda functions?

backend/dataall/cdkproxy/stacks/pivot_role.py
```diff
@@ -0,0 +1,822 @@
...
+                    actions=['cloudwatch:PutMetricData', 'cloudwatch:GetMetricData', 'cloudwatch:GetMetricStatistics'],
+                    resources=['*'],
+                ),
+                # Logs
+                iam.PolicyStatement(
+                    sid='Logs',
+                    effect=iam.Effect.ALLOW,
+                    actions=['logs:CreateLogGroup', 'logs:CreateLogStream', 'logs:PutLogEvents'],
+                    resources=[
+                        f'arn:aws:logs:*:{self.account}:/aws-glue/*',
```
dosiennik, line 171: should be:
`    f"arn:aws:logs:*:{self.account}:log-group:/aws-glue/*",
                        f"arn:aws:logs:*:{self.account}:log-group:/aws/lambda/*",
                        f"arn:aws:logs:*:{self.account}:log-group:/{env_resource_prefix}*",
                    ],`

backend/dataall/cdkproxy/stacks/pivot_role.py
```diff
@@ -0,0 +1,822 @@
...
+                    effect=iam.Effect.ALLOW,
+                    actions=['ec2:CreateTags', 'ec2:DeleteTags'],
+                    resources=[f'arn:aws:ec2:*:{self.account}:network-interface/*'],
+                ),
+                iam.PolicyStatement(
+                    sid='DeleteENI',
+                    effect=iam.Effect.ALLOW,
+                    actions=['ec2:DeleteNetworkInterface'],
+                    resources=['*'],
+                    conditions={'StringEquals': {'aws:ResourceTag/dataall': 'True'}},
```
dosiennik, line 388: this is incorrect, if we attach pivot role to the lambda function or the crawler ENIs created by the lambda or glue service doesn't contain this dataall tag, as a result ENIs are not released which causes abnormal consumption of IP addresses in the VPC

backend/dataall/cdkproxy/stacks/dataset.py
```diff
@@ -389,53 +309,94 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
+        # Get the lambda arn from SSM, this Lambda is created as part of the environment stack
+        # It replaces the following (just because it causes Cfn issues when handling upgrades of pivotRole)
+
+        # storage_location = CfnResource(
```
dosiennik, line 322: I think I have already mentioned it, what may happen during the migration since I have encountered it by myself.
Removing this AWS::LakeFormation::Resource resource and adding a Custom Resource within the same process of updating the dataset stack causes a race condition since:
1. deleting AWS::LakeFormation::Resource from the cdk code triggers deletion of the data location resource from the lake ormation
2. adding a custom resource or actually replacing AWS::LakeFormation::Resource with a custom resource creates or updates the data location resource

Then the output of an update of the dataset stack depends on the order in which 1 & 2 occurs, in the worst case which happened to me (2 executed at first, 1  as second step):
2. replacing AWS::LakeFormation::Resource with a custom resource updated the datalake location with a new iam role
1. although updated correctly, removal of AWS::LakeFormation::Resource triggers deletion of the datalake location
As a result of this update, the datalake location is not anymore registered in the lake formation.

It is due to both AWS::LakeFormation::Resource and Custom Resource pointing to the same DataLakeLocation which can be registered just once.

I just want to raise it as a risk for the migration from the manual to cdk managed approach.
To mitigate it as a first step of the migration we need to update all the dataset stacks setting the AWS::LakeFormation::Resource deletion policy to RETAIN. Then regardless of the order we get rid of the risk of deleting the datalake location.

backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
dlpzx: I think the issue has never been that the database construct did not exist, I think the main limitation is that the database does not come with the Lake Formation permissions. But I am seeing that this CFN resource exists now: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lakeformation/CfnPermissions.html. So a note for the future, we can think about replacing the custom resource by real resources!

dosiennik: yep exactly, I think we can do something like this for setting up the privileges (as an example):
````
database_permissions = lf.CfnPermissions(
                self,
                f"GlueDatabasePermissions-{uuid.uuid4()}",
                data_lake_principal=lf.CfnPermissions.DataLakePrincipalProperty(data_lake_principal_identifier=role_arn),
                resource=lf.CfnPermissions.ResourceProperty(
                    database_resource=lf.CfnPermissions.DatabaseResourceProperty(catalog_id=catalog_id, name=glue_database.database_name),
                ),
                permissions=["CREATE_TABLE", "ALTER", "DROP"],
                permissions_with_grant_option=["CREATE_TABLE", "ALTER", "DROP"],
            )
            tables_permissions = lf.CfnPermissions(
                self,
                f"GlueDatabaseTablePermissions-{uuid.uuid4()}",
                data_lake_principal=lf.CfnPermissions.DataLakePrincipalProperty(data_lake_principal_identifier=role_arn),
                resource=lf.CfnPermissions.ResourceProperty(
                    table_resource=lf.CfnPermissions.TableResourceProperty(
                        catalog_id=catalog_id, database_name=glue_database.database_name, table_wildcard=lf.CfnPermissions.TableWildcardProperty()
                    ),
                ),
                permissions=["SELECT", "ALTER", "DESCRIBE"],
                permissions_with_grant_option=["SELECT", "ALTER", "DESCRIBE"],
            )
            database_permissions.node.add_dependency(glue_database)
            tables_permissions.node.add_dependency(glue_database)
````

Comment (dlpzx) [reply backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py:28]: We are going to forget, so I opened an[ issue](https://github.com/awslabs/aws-dataall/issues/381). Maybe we can wait for the storage location CFN error to be fixed to replace both custom resources.

File: backend/dataall/tasks/stacks_updater.py [hunk 24]
```diff
@@ -18,32 +19,39 @@
...
+            retries = 1
+            while Ecs.is_task_running(cluster_name=cluster_name, started_by=f'awsworker-{stack.stackUri}'):
+                log.info(f"Update for {stack.name}//{stack.stackUri} is not complete, waiting for 30 seconds...")
+                time.sleep(30)
```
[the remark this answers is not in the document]
dlpzx: Let's first move it to constants. I think that what we discussed around putting an SQS queue in front of ECS-cdkproxy is the way to go instead of this custom queueing

backend/dataall/tasks/stacks_updater.py
dlpzx: By the way I faced very strange problems when running all environments in parallel. I shared with you a testing document with this behavior. That is why I left them running in sequence

backend/dataall/tasks/stacks_updater.py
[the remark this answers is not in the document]
dosiennik: Fine to me!

backend/dataall/tasks/stacks_updater.py
dosiennik: I see, I will take a look

backend/dataall/aws/handlers/quicksight.py
dosiennik: I see, great idea!

backend/dataall/cdkproxy/stacks/environment.py
dlpzx: To my knowledge I think there was no standard way of defining them. Do you have any ideas?

backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
dlpzx: Another motivation to avoid custom resources: I have done some "limitations" testing on data.all and I have reached a limitation on the number of datasets that can be created in an environment. The limitation comes from the custom resources Lambda policy size (check [this](https://repost.aws/knowledge-center/lambda-resource-based-policy-size-error)). I want to do some last minute changes to the custom resources providers in the dataset stack as part of this PR (needed for a customer).

backend/dataall/cdkproxy/stacks/environment.py
dosiennik: I think this code does very simple stuff like calling the API only. So memory and timeout settings seems to be a little bit to high:

There is a way at least I have used to find the best settings: https://github.com/alexcasalboni/aws-lambda-power-tuning

backend/dataall/cdkproxy/stacks/dataset.py
dosiennik: Weird that I didn't happen for your tests. It depends on the order so you may be more lucky than I am. 

I just changed the deletion policy to RETAIN (separate commit), deployed the changes to the AWS, then executed stacks-updater task using aws-cli. Then in the yet another next commit I proceeded with removal of the original DataLakeLocation from the cdk code and replacing it with the Custom Resource, but it wasn't very convenient as you may imagine :D

' I am also thinking, maybe we can check in the on_update custom resource to create if not created. That way if in the first update (on_create) it fails, in the second one it will succeed. ' - I was also thinking about it, then the update would check if the datalakelocation is still there, if not, would recreate it. That would be more convenient.

But to be sure that the state of the aws resources is as expected we would need to execute update stacks task two times during the migration from manual to automated approach for the pivot role. So we need to mention it in the migration steps.

backend/dataall/api/Objects/Dataset/resolvers.py
```diff
@@ -22,8 +23,18 @@
 log = logging.getLogger(__name__)
 
 
+def check_dataset_account(environment):
+    if environment.dashboardsEnabled:
+        Quicksight.check_quicksight_enterprise_subscription(AwsAccountId=environment.AwsAccountId)
+        Quicksight.create_quicksight_group(AwsAccountId=environment.AwsAccountId)
+    return True
```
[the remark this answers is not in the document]
dlpzx: I see, fixing that to make it more explicit

backend/dataall/api/Objects/Environment/resolvers.py
dosiennik: I am sorry, I think I wasn't clear enough, it is actually minor, I meant we can move lines 52-54 to the CloudFormation.check_existing_cdk_toolkit_stack method then the code here would be like:

```` cdk_role_name = CloudFormation.check_existing_cdk_toolkit_stack(
        AwsAccountId=account_id, region=region
    )`
    
    and implementation of the CloudFormation.check_existing_cdk_toolkit_stack will take care of obtaining the cdk role arn and making use of it as an internal detail how we check the existence of the cdk toolkit stack.

File: backend/dataall/cdkproxy/stacks/environment.py (hunk 20 above)
Comment (dosiennik) [reply backend/dataall/cdkproxy/stacks/environment.py:390]: https://github.com/awslabs/aws-dataall/issues/382 does it make sense? :)

Comment (dlpzx) [reply backend/dataall/cdkproxy/stacks/environment.py:390]: Thank you!!

File: backend/dataall/cdkproxy/stacks/dataset.py (hunk 23 above)
Comment (dlpzx) [reply backend/dataall/cdkproxy/stacks/dataset.py]: Worst case scenario the daily task would trigger the update, so it would be a mismatch of a couple of hours

File: backend/dataall/api/Objects/Environment/resolvers.py (hunk 15 above)
Comment (dlpzx) [reply backend/dataall/api/Objects/Environment/resolvers.py:51]: I see, yes, let's keep it simple we do not need to specify the role for this method

File: backend/dataall/api/Objects/Environment/resolvers.py (hunk 13 above)
Comment (dlpzx) [reply backend/dataall/api/Objects/Environment/resolvers.py:35]: I will have opened another issue for this. I think it has to do with the docker deployment of data.all. But it is good that you bring it up :)

File: frontend/src/api/Environment/getPivotRoleAsPartOfEnvironment.js (hunk 8 above)
Comment (dlpzx) [reply frontend/src/api/Environment/getPivotRoleAsPartOfEnvironment.js:3]: I am going to discuss it with Amr, to see if this is something we can do as part of the frontend refactoring. I foresee us having many of this kind of parameters in the future

File: backend/dataall/api/Objects/Environment/resolvers.py (hunk 13 above)
Comment (dosiennik) [reply backend/dataall/api/Objects/Environment/resolvers.py]: sure thing, thanks!

File: backend/dataall/cdkproxy/stacks/dataset.py (hunk 23 above)
Comment (dosiennik) [reply backend/dataall/cdkproxy/stacks/dataset.py]: yeah just to mention about this risk

File: backend/dataall/aws/handlers/cloudformation.py (hunk 1 above)
Comment (dlpzx) [reply backend/dataall/aws/handlers/cloudformation.py]: I removed the dead lines of code :) Also, I think that in any case it will fail even before the describe-stacks call because the cfn client is trying to assume the cdk-look-up role

File: backend/dataall/api/Objects/Environment/resolvers.py [hunk 26]
```diff
@@ -52,7 +75,13 @@ def create_environment(context: Context, source, input=None):
         )
 
     with context.engine.scoped_session() as session:
-        cdk_role_name = check_environment(context, source, input=input)
+        pivot_role_as_part_of_environment = get_pivot_role_as_part_of_environment(context, source)
```
kukushking, line 78: It seems you can move `get_pivot_role_as_part_of_environment` check inside `check_environment` to  have the function that performs all necessary checks and returns the correct role.

deploy/stacks/backend_stack.py
```diff
@@ -65,6 +66,8 @@ def __init__(
         vpc = self.vpc_stack.vpc
         vpc_endpoints_sg = self.vpc_stack.vpce_security_group
 
+        self.pivot_role_name = f"dataallPivotRole{'-cdk' if enable_pivot_role_auto_create else ''}"
```
kukushking, line 69: During a migration to new auto-create pivot roles, there will be a point in time when backend has updated and is configured to use the new pivot roles (with `-cdk` suffix), while environment updates are still pending. As I understand, the IAM policies will change allowing only new pivot roles to perform actions. Is it expected that environments won't be usable until they upgrade? 

Also, I'm not as familiar with the codebase but we need to be extra cautious here to avoid being locked out. Pivot role isn't itself used to perform the deployment right?

backend/dataall/cdkproxy/assets/lakeformationdefaultsettings/index.py
```diff
@@ -1,17 +1,39 @@
...
+    for principal in principals:
+        if ":role/" in principal:
+            log.info(f'Principal {principal} is an IAM role, validating....')
+            try:
+                iam_client.get_role(RoleName=principal.split("/")[-1])
+                log.info(f'Adding principal {principal} to validated principals')
+                validated_principals.append(principal)
+            except Exception as e:
+                log.exception(f'Failed to get role {principal} due to: {e}')
+                raise Exception(f'Failed to get role {principal} due to: {e}')
```
kukushking, line 32: Wasn't the intent here to return the list of valid principals, ignoring the errors? Otherwise `on_create`/`on_delete` will fail.

backend/dataall/aws/handlers/sagemaker_studio.py
```diff
@@ -7,17 +7,18 @@
...
     @staticmethod
-    def client(AwsAccountId, region):
-        session = SessionHelper.remote_session(AwsAccountId)
+    def client(AwsAccountId, region, role=None):
+        session = SessionHelper.remote_session(accountid=AwsAccountId, role=role)
         return session.client('sagemaker', region_name=region)
 
     @staticmethod
-    def get_sagemaker_studio_domain(AwsAccountId, region):
+    def get_sagemaker_studio_domain(AwsAccountId, region, role):
```
nikpodsh, line 15: We should add a role to `resolvers` of `SagemakerStudio`. Otherwise, we will crash when we try to create a profile

backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
```diff
@@ -1,18 +1,49 @@
...
     return data
 
 
+def lambda_policy_remove_duplicates(props):
+    log.info(f"Removing unnecessary Lambda policy statements from Lambda: {props['LambdaArn']}")
+    response = lambda_client.get_policy(FunctionName=props["LambdaArn"])
+    policy = json.loads(response.get("Policy"))
+    log.info(f"Lambda Policy statements:{policy.get('Statement')}")
+
+    for statement in policy.get("Statement")[:-1]:
```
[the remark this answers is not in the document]
dlpzx: It is removing all except the last one. What happened was:
1. We deploy the Lambda function as part of the environment CDK stack
2. We deploy a custom resource that invokes this Lambda in every dataset CDK stack
3. The custom resource creates an IAM role to invoke the Lambda
4. The custom resource adds an statement to the original Lambda policy to allow the created IAM role permission to invoke the function.
5. After around 60 datasets the Lambda policy size reaches its limit :( So we constrain the number of datasets that can be created in an environment.

What happens now:
1. We deploy the Lambda function as part of the environment CDK stack
2. We deploy a custom resource that invokes this Lambda in every dataset CDK stack
3. We **assign an IAM role** to invoke the Lambda - pivotRole
4. The custom resource adds an statement to the original Lambda policy to allow the pivotRole permission to invoke the function.
5. We remove all statements except the last one from the Lambda policy (if we don't with every dataset a new statement allowing the pivotRole is added to the policy)
6. I tested and now we can create hundreds of datasets without issue


I really tried looking for a way in which the custom resource does not add this invoke statement, but there is nothing in CDK/CFN that allows this option.

deploy/stacks/backend_stack.py
dlpzx: I added a "capital letter" warning in the description and we will make sure to add it in the release notes, but thank you you are right there is some downtime until everything is complete.

backend/dataall/aws/handlers/sts.py
```diff
@@ -173,9 +155,18 @@ def get_delegation_role_arn(cls, accountid):
...
+
+    @classmethod
+    def get_cdk_look_up_role_arn(cls, accountid, region):
+        """Returns the name that will be assumed to perform IAM actions on a given AWS accountid using CDK Toolkit role
+        Args:
+            accountid(string) : aws account id
+        Returns:
+                string : arn of the CDKToolkit role on the target aws account id
+        """
+        log.info(f"Getting CDK look up role: arn:aws:iam::{accountid}:role/cdk-hnb659fds-lookup-role-{accountid}-{region}")
```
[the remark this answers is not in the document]
dlpzx: Yes, it looks weird. It comes from the CDK [bootstrapping ](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html)command, it is a qualifier that is used to avoid naming conflicts. The default value is `hnb659fds`, so if customers run the command as we define it in the frontend they would get the default. If they have any other kind of bootstrapping standard we would run into issues. I think there is already an issue about custom bootstrapping processes.

backend/dataall/aws/handlers/sagemaker_studio.py
dlpzx: You are right, I fixed it by adding a default for the parameter and letting the pivotRole be the used role

backend/dataall/api/Objects/Environment/resolvers.py
dlpzx: Done! thank you

backend/dataall/cdkproxy/assets/lakeformationdefaultsettings/index.py
dlpzx: Yeeep, the exception should only be logged. Thank you

backend/dataall/aws/handlers/cloudformation.py
dosiennik: indeed since the role will simply not be there

backend/dataall/cdkproxy/assets/datalakelocationcustomresource/index.py
```diff
@@ -0,0 +1,102 @@
...
+    raise Exception(f"Invalid request type: {request_type}")
+
+
+def on_create(event):
+    """ Checks if the S3 location is already registered in Lake Formation.
+    If already registered it updated the roleArn for the location.
+    If not registered, it registers the location.
+    """
+    props = event["ResourceProperties"]
+    lambda_policy_remove_duplicates(props)
```
nikpodsh, line 32: Why do we have duplicates? Is there a way to avoid creation of duplicated policy for lambda not to delete this duplication after?

nikpodsh: Looks like we are doing it to replace the old manual pivot role. Is that correct?

backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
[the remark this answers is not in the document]
dosiennik: @dlpzx so if we would create a dedicated IAM role for the custom resource in the env stack and would add this iam role in advance to the lambda policy which handles the custom resource, anyway the cr.Provider will trigger adding the role to the lambda policy?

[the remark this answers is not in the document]
dlpzx: @dlpzx I think so. What I tested was to specify the role in the provider and check whether it tried to recreate the same statement and it did. I think @kukushking has given me a possible solution, create the Provider directly in the environment stack and store the service_token as SSM parameter. Then in the dataset stack we just create the CustomResource with the service token retrieved from SSM. I pushed this changes in my last commit, but I need to test

backend/dataall/cdkproxy/assets/datalakelocationcustomresource/index.py
dlpzx: Nope, check the conversation above

backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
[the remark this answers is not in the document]
dosiennik: @dlpzx I see, let me know if it works. Would be a nice learning to me as well :)

deploy/stacks/pipeline.py
```diff
@@ -611,6 +610,38 @@ def set_db_migration_stage(
             ),
         )
 
+    def set_stacks_updater_stage(
```
nikpodsh, line 613: I still have doubts if we really need a dedicated stage to update stacks? 🤔 
1) `stack_updates` runs every night on schedule, so env stacks should automatically be updated on the next day (Is there a customer that needs to have all stack updates ASAP without waiting until next day?)
2) Okay, we can assume that this feature can also be used for some other changes in future. In this case, why wouldn't we introduce a button to update all stacks. We already have a button to update one stack, so it should be easy to do for many stacks. Otherwise, just to use this feature, a customer has to commit a change with `enable_update_dataall_stacks_in_cicd_pipeline=true`, deploy it, wait for downtime while all environment and dataset stacks are updated, and not to forget to commit `enable_update_dataall_stacks_in_cicd_pipeline=false` on the next round (or get more 1-2 hours of downtime). 

Please, feel free to correct me since I might have misunderstood something :)

dlpzx: I think for prod deployments, the less downtime the better. In this case downtime is not just that the application does not work, users might try to perform actions that break the CloudFormation stacks that are not updated.
@dosiennik @kukushking I would also want your opinion.

dosiennik: if I understood the question correctly, I think adding this stage is not only for the purpose of migration, like enabling it and disabling once the job of updating the stacks is done, the step is added to the pipeline also for the future changes into the pivot role and applying the changes as a part of the deployment - not to wait until night but to see the deployed changes/hot-fixes-  if something is missing in the pivot role - as soon as possible

frontend/src/views/Environments/EnvironmentCreateForm.js
```diff
@@ -131,6 +131,8 @@ const EnvironmentCreateForm = (props) => {
     });
   };
   useEffect(() => {
+    console.log("ENABLE_PIVOT_ROLE_AUTO_CREATE")
```
nikpodsh, line 134: Looks like it was left after debugging. Let's remove it

backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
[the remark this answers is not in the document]
dlpzx: I also tested an existing environment handler. 
When creating a new dataset - The Lambda does not change, it keeps the previously added policies but it does not add any new ones.

When updating an existing dataset - it fails 😞 Apparently we cannot modify the service token
image

nikpodsh:
I think it'd be nice to have a test that compares a manual pivot role and the automated one. Just in case when we added a new changes in one of them, but forgot to replicated it into the other. What do you think?

backend/dataall/api/Objects/Dataset/resolvers.py
```diff
@@ -212,6 +228,9 @@ def get_dataset_stewards_group(context, source: models.Dataset, **kwargs):
 
 def update_dataset(context, source, datasetUri: str = None, input: dict = None):
     with context.engine.scoped_session() as session:
+        dataset = Dataset.get_dataset_by_uri(session, datasetUri)
+        environment = Environment.get_environment_by_uri(session, dataset.environmentUri)
+        check_dataset_account(environment=environment)
```
nikpodsh, line 233: Minor: looks like we can create `check_dataset_account(dataset_uri=uri)` and move  `get_environment_by_uri` into this method.

backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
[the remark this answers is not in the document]
dlpzx: We just needed to rename the CFN resources :) 
image

deploy/stacks/pipeline.py
kukushking: Couple of thoughts: for this specific pivot role change we do want to roll out the update as soon as possible to minimize downtime (and avoid "wait until midnight" type of instructions in the readme :)). Auto-update feature is one of the ways to achieve that (albeit with a bit of friction having to enable it) and it probably is a useful feature beyond this one change. 

I do like the idea of "update all" button though, as a way to force the update once. This will address the use case where users want to keep auto-update off (either for cost optimisation or something else). But all of this is probably out of scope of this PR, as long as there is an update path it's fine.

backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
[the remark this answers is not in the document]
kukushking: Awesome!

backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -364,42 +358,96 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
+            f'{self._environment.resourcePrefix}GlueDbCustomResourceProvider',
+            on_event_handler=gluedb_custom_resource
+        )
+
         ssm.StringParameter(
             self,
             'GlueCustomResourceFunctionArn',
             string_value=gluedb_custom_resource.function_arn,
-            parameter_name=f'/dataall/{self._environment.environmentUri}/cfn/custom-resources/lambda/arn',
+            parameter_name=f'/dataall/{self._environment.environmentUri}/cfn/custom-resources/gluehandler/lambda/arn',
```
kukushking, line 371: Very minor - are these SSM parameters (`GlueCustomResourceFunctionName`, `GlueCustomResourceFunctionArn`) needed? Are they used?

kukushking (approved):
Looks great!

backend/dataall/cdkproxy/stacks/pivot_role.py
```diff
@@ -0,0 +1,821 @@
...
+                    actions=['cloudwatch:PutMetricData', 'cloudwatch:GetMetricData', 'cloudwatch:GetMetricStatistics'],
+                    resources=['*'],
+                ),
+                # Logs
+                iam.PolicyStatement(
+                    sid='Logs',
+                    effect=iam.Effect.ALLOW,
+                    actions=['logs:CreateLogGroup', 'logs:CreateLogStream', 'logs:PutLogEvents'],
+                    resources=[
+                        f'arn:aws:logs:*:{self.account}:/aws-glue/*',
```
dosiennik, line 171: seems like for glue we forgot to change it, same in line 173

deploy/stacks/pipeline.py
nikpodsh: Yep, sounds good!

backend/dataall/cdkproxy/stacks/environment.py
dlpzx: I had the same thought, they are not needed in any piece of code. But I have used them in the past to troubleshoot and find the original Lambda of the custom resource. I have left them on purpose but I do not have a strong opinion about them

backend/dataall/api/Objects/Dataset/resolvers.py
dlpzx: In the update_dataset method yes, but in the create method it is not possible. When we create we are checking the environment account before the dataset object exists

backend/dataall/cdkproxy/stacks/pivot_role.py
dlpzx: @dosiennik, `aws-glue` is actually intended but maybe I am making a mistake.
From what I see the name of the log group of glue is:
image
That is why I have written it this way

dosiennik: but I think what is wrong here: loggroup part is missing in the arn

dosiennik: this is how the arn looks like:
arn:aws:logs:eu-central-1:423128813120:log-group:/aws-glue/crawlers:*
but in the config 'log-group' is missing if I see it well :)

backend/dataall/cdkproxy/assets/datalakelocationcustomresource/index.py
```diff
@@ -0,0 +1,81 @@
...
+    if not _is_resource_registered(props["ResourceArn"]):
+        register(props)
+    else:
+        update(props)
+
+
+def _is_resource_registered(resource_arn: str):
+    lf_resources = lf_client.list_resources(FilterConditionList=[{"Field": "RESOURCE_ARN", "ComparisonOperator": "EQ", "StringValueList": [resource_arn]}])
+    return len(lf_resources["ResourceInfoList"]) > 0
+
```
dlpzx, line 40: @dosiennik I found out that when we have more than 100 registered locations the pagination of this api call prevents us from seeing the registered resource. It tries to re-register it and it fails :( 
So, let's say we have around 500 registered locations and a particular dataset is returned in the 4th page when we list_resources without any filter condition. We are updating a dataset, the handler is triggered and the list_resources with filter returns nothing as it is trying to filter from the first page of results.

I am working on a fix for this, I think I am going to try to register the resource and check the exception

backend/dataall/cdkproxy/stacks/pivot_role.py
dlpzx: True! thank you!

backend/dataall/cdkproxy/assets/datalakelocationcustomresource/index.py
dosiennik: oh, that's interesting, thanks for the finding! we are actually applying a filter 
`lf_client.list_resources(FilterConditionList=[{"Field": "RESOURCE_ARN", "ComparisonOperator": "EQ", "StringValueList": [resource_arn]}])`

but maybe it works another way under the hood

dlpzx: Exactly, I think that filter is only filtering the results from the first page of results. I have tested it and it seems to be the case because NextToken is returned.

```
[DEBUG]	2023-04-03T08:05:38.808Z	1c09bb21-65be-47fd-9de2-29a77d273c02	Response body:
b'{
    "NextToken": "eyJsYXN0RXZhbHVhdGVkS2V5Ijp7IkhBU0hfS0VZIjp7InMiOiI4NTUyMjU2OTU4NzlfODU1MjI1Njk1ODc5In0sIlJBTkdFX0tFWSI6eyJzIjoiczM6Ly9kYXRhYWxsLWFwaWRhdGFzZXR5bGltaXQtMTkyLTlyYnE3bGpvIn19LCJleHBpcmF0aW9uIjp7InNlY29uZHMiOjE2ODA1OTU1MzgsIm5hbm9zIjo4MDUwMDAwMDB9fQ==",
    "ResourceInfoList": []
}'

```

dosiennik: whoops, so I need to take a look as well, for now I have no other idea than iterating page by page until the next token is returned from the response :/

dlpzx: That is what I just pushed :)
Take a look at the logs, this is for an environment with around 500 datasets. Until the fourth call it does not find the resource:
image

Issue #295, opened 2023-02-07
Modularization of data.all code


dlpzx:
**Is your feature request related to a problem? Please describe.**
- As a client I am always frustrated when I customize data.all and then I ran into conflicts whenever I try get a more upgraded version.
- Since developing a new feature involves changing many details across the repo, the learning curve can be very steep at the beginning.
- I also find it difficult to contribute back, since my developed features collide with some features or modify many different pieces of code in various files across the repo.
- Some customers want to use a subset of data.all modules but instead they need to deploy the whole platform and maintain the whole code even when using only the core functionalities.

**Describe the solution you'd like**
- Code should be refractored to group features/modules together. It should be easy to find all code related to a feature and it should be easy to replace it with another piece of code.
- Platform should be configurable and customers need to easily define the modules that build their platform and what is the source code for it.

**Describe alternatives you've considered**
Modularization can be done at multiple levels. 
- Architecture modularization: similar to what is already implemented. Each component is isolated (db, api, aws...)
- Code modularization: refractor code but keep the above infrastructure in single components for all features.
- Features as microservices: Each feature module having its own resources

**Additional context**

*P.S. Please Don't attach files. Add code snippets directly in the message body instead.*

dlpzx:
Completed with V2.0 release 🚀

PR #359, opened 2023-03-09
backend: Pluginarchitecture

nikpodsh:
This is a draft PR for showing purposes. There are still some minor issues that needs to be addressed.

### Feature or Bugfix
- Refactoring

### Detail
There are following changes under this PR:
1. Modularization + Refactoring of notebooks
There are new modules that will play a major role in the future refactoring:

   * Core = contains the code need for application to operate correctly
   * Common = common code for all modules
   * Modules = the plugin/feature that can be inserted into the system (at the moment only notebooks)

   A small instruction how to import module can be found in [backend/dataall/modules/__init__.py](https://github.com/awslabs/aws-dataall/compare/main...nikpodsh:aws-dataall:pluginarchitecture?expand=1#diff-fee040ec737598e874f2e707883daf2e6b00a71c550a06f60aa8bbebd6550205) + see `loader.py`
The other part that is related to modularization is the creation of environment parameters and resources. 
Environment parameter will replace all hardcoded parameters of  the environment configuration.
There is a new file - config.json that allows you to configure an application configuration.
All existing parameters will be migrated via db migration in AWS

2. Extracting permissions and request context (Optional for the modularization)

   Engine, user, and user groups had been passed as a parameter of context in the request. This had forced to pass a lot of parameters to other methods that weren't even needed. This information should be as a scope of the request session.
There is a new way to retrieve the information using `RequestContext.`
There is also a new way to use permission checks that require less parameters and make code cleaner. The old way was marked as deprecated 

3. Restructure of the code (Optional for the modularization)

   Since the modularization will touch all the places in the API code it can be a good change to set a new structure of the code. There are small re-organization in notebook module to address 

   * Allocating the resources before the validating parameters
   * Not clear responsibility of the classes
   * Mixed layers 
  
   There are new structure :
   - resolvers = validate and pass code to service layer
   - service layer = bisnesss logic
   - repositories = database logic (all queries should be placed here)
   - aws = contains a wrapper client upon boto3
   - cdk = all logic related to create stacks or code for ecs
   - tasks = code that will be executed in AWS lambda (short-living tasks)

   All names can be changed.

### Relates
[#295](https://github.com/awslabs/aws-dataall/issues/295)


deploy/stacks/pipeline.py
```diff
@@ -452,6 +452,7 @@ def set_quality_gate_stage(self):
                     commands=[
                         'mkdir -p source_build',
                         'mv backend ./source_build/',
+                        'mv config.json ./source_build/',
```
dlpzx, line 455: [Clarification] - In this step we are uploading the backend code to an S3 Bucket. This step in necessary for the migration stage that accesses the migration scripts on the bucket from the infra account.Now we also need to access the config.json so we need to add it to the bucket

deploy/stacks/container.py
```diff
@@ -512,3 +486,11 @@ def set_scheduled_task(
             # security_groups=[security_group],
         )
         return scheduled_task
+
```
dlpzx, line 489: [Clarification] - Method to simplify env creation. Added `config_location` to the environment of the ECS container

backend/requirements.txt
```diff
@@ -12,4 +12,5 @@ pyjwt==2.4.0
 PyYAML==6.0
 requests==2.27.1
 requests_aws4auth==1.1.1
-sqlalchemy==1.3.16
\ No newline at end of file
+sqlalchemy==1.3.24
```
dlpzx, line 15: [Clarification] - Bug on sqlalchemy and alembic versions. Unrelated to modularization

dlpzx:
Hi @nikpodsh, I have changed the target branch of the merge to a `modularization` branch.

backend/dataall/modules/notebooks/aws/client.py
```diff
@@ -0,0 +1,62 @@
...
+        """Stops the notebooks instance by sending a request to AWS"""
+        try:
+            self._client.stop_notebook_instance(NotebookInstanceName=self._instance_name)
+        except ClientError as e:
+            raise e
+
+
+def client(notebook: SagemakerNotebook) -> SagemakerClient:
+    """Factory method to retrieve the client to send request to AWS"""
+    return SagemakerClient(notebook)
```
dlpzx, line 62: I would rename the client to 'sagemaker_client' for cases that use more than one AWS service
Also, do you think that we could have a`aws` package as part of the `common` package? Some of the aws clients are used in several modules. For example the IAM handler

dlpzx: There is also the option of having the whole `aws` package as part of `common` and remove it from the modules, but as downside we will end up with "module code" outside of the module directory

backend/dataall/cdkproxy/stacks/__init__.py
dlpzx, line 1: No need to load it, we remove it because we load it with loader

backend/dataall/cdkproxy/stacks/dataset.py
```diff
@@ -545,6 +545,6 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
 
         Tags.of(self).add('Classification', dataset.confidentiality)
 
-        TagsUtil.add_tags(self)
+        TagsUtil.add_tags(stack=self, model=models.Dataset, target_type="dataset")
```
dlpzx, line 548: [Clarification] - belongs to a different PR

backend/dataall/cdkproxy/stacks/__init__.py
```diff
@@ -2,7 +2,6 @@
 from .environment import EnvironmentSetup
 from .pipeline import PipelineStack
 from .manager import stack, instanciate_stack, StackManager
-from .notebook import SagemakerNotebook
```
dlpzx, line 5: [Clarification ] No need to load it, we remove it because we load it with loader

backend/dataall/cdkproxy/stacks/pipeline.py
```diff
@@ -370,7 +370,7 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
             value=codepipeline_pipeline.pipeline_name,
         )
 
-        TagsUtil.add_tags(self)
+        TagsUtil.add_tags(stack=self, model=models.DataPipeline, target_type="pipeline")
```
dlpzx, line 373: [Clarification] - belongs to a different PR

backend/dataall/cdkproxy/stacks/policies/service_policy.py
```diff
@@ -106,27 +93,9 @@ def generate_policies(self) -> [aws_iam.ManagedPolicy]:
 
         services = ServicePolicy.__subclasses__()
 
-        if permissions.CREATE_REDSHIFT_CLUSTER not in self.permissions:
```
dlpzx, line 109: [clarification] - the check of the permissions needed is moved to the modules (I really like this). @nikpodsh, I am thinking that for some modules we will need permissions form different services. e.g. It is not the same the permissions needed for Notebook than for SM Studio (although both are SageMaker), right now we are taking the sum of both for both modules, if instead we defined the permissions not by service, but by module we would have less inter-dependency and we would be able to restrict permissions to only those needed for each module. (Of course it has the downside of duplicating permissions, but I think in this case security comes always first)

backend/dataall/cdkproxy/stacks/policies/_lambda.py
```diff
@@ -1,9 +1,13 @@
+from dataall.db import permissions
```
dlpzx, line 1: [Question] dataall.db permissions is still a file that remains common to all modules right? @nikpodsh How would this piece work for users? Do they have to all modify this file and add their permissions?

backend/dataall/cdkproxy/stacks/policies/cloudformation.py
```diff
@@ -3,7 +3,7 @@
 
 
 class Cloudformation(ServicePolicy):
```
dlpzx, line 5: [Question] - and for example this policy is not dependent from any module. In conclusion, we need to analyze the policies needed for each module, those that are generic and how they are assigned

backend/dataall/cdkproxy/stacks/redshift_cluster.py
```diff
@@ -184,6 +184,6 @@ def __init__(self, scope, id: str, target_uri: str = None, **kwargs) -> None:
             hosted_rotation=aws_secretsmanager.HostedRotation.redshift_single_user(),
         )
 
-        TagsUtil.add_tags(self)
+        TagsUtil.add_tags(stack=self, model=models.RedshiftCluster, target_type="redshift")
```
dlpzx, line 187: [Clarification] - belongs to another PR

backend/dataall/cdkproxy/stacks/sagemakerstudio.py
```diff
@@ -93,6 +93,6 @@ def __init__(self, scope, id: str, target_uri: str = None, **kwargs) -> None:
             .to_string()
         )
 
-        TagsUtil.add_tags(self)
+        TagsUtil.add_tags(stack=self, model=models.SagemakerStudioUserProfile, target_type="mlstudio")
```
dlpzx, line 96: [Clarification] - belongs to another PR

backend/dataall/modules/__init__.py
```diff
@@ -0,0 +1,39 @@
+"""
+Contains all submodules that can be plugged into the main functionality
+
+How to migrate to a new module:
+1) Create your python module
+2) Create the following submodules:
+    a) gql
+    b) tasks
+    3) cdk
```
dlpzx, line 9: [Question] - @nikpodsh IMPORTANT! something that we have not considered yet are other long-running tasks besides the cdk task. There are 7 ECS tasks, from which 2 run on-demand (cdkproxy and share-manager), the other 5 run on different schedules:
image

I am not sure about the utility of some of these scheduled tasks (e.g. subscriptions seems to be a legacy task). I can review which ones we really need. I can confirm that cdkproxy, share-manager, stack-updater, catalog-indexer and table-syncer should be included

backend/dataall/modules/notebooks/gql/types.py
dlpzx, line 1: [Clarification] -  This is a name change from the previous "schema". I agree that schema was misleading, it seemed that it defined the schema of the Notebooks only, I like types better, also because we are defining gql ObjectTypes

backend/local_cdkapi_server.py
dlpzx, line 1: [Clarification ] renaming the local.cdkapi.server files, naming was a little weird

tests/core/test_config.py
```diff
@@ -0,0 +1,7 @@
+from dataall.core.config import config
+
+
+def test_config():
+    """Checks that properties are read correctly"""
+    modules = config.get_property("modules")
+    assert "notebooks" in modules
```
dlpzx, line 7: [Question] - What happens when notebooks is not added to the modules? Is this test going to be particular for each implementation

tests/modules/notebooks/test_notebook_stack.py
```diff
@@ -0,0 +1,18 @@
+
+def test_notebook_stack(client, sgm_notebook, group):
```
dlpzx, line 2: Replaces test in`tests/api/test_stack.py`? @nikpodsh

tests/modules/notebooks/test_sagemaker_notebook.py
```diff
@@ -10,75 +20,35 @@ def org1(org, user, group, tenant):
...
     module_mocker.patch(
         'dataall.api.Objects.Environment.resolvers.check_environment', return_value=True
     )
-    env1 = env(org1, 'dev', user.userName, group.name, '111111111111', 'eu-west-1')
+    env1 = env(org1, 'dev', user.userName, group.name, '111111111111', 'eu-west-1',
+               parameters={"notebooksEnabled": "True"})
     yield env1
 
 
-@pytest.fixture(scope='module', autouse=True)
```
dlpzx, line 22: Moved to `tests/modules/notebooks/conftest.py` right?

tests/modules/notebooks/test_sagemaker_notebook.py
```diff
@@ -10,75 +20,35 @@ def org1(org, user, group, tenant):
 
 
 @pytest.fixture(scope='module')
-def env1(env, org1, user, group, tenant, module_mocker):
+def env1(env, org1, user, group, tenant, db, module_mocker):
```
dlpzx, line 23: Why do we need `db`? @nikpodsh

dlpzx (commented):
First look

backend/dataall/modules/notebooks/aws/client.py
nikpodsh: Tbh, I was thinking about naming it as `sagemaker_client`. The only think that stopped me is that we already have a convention to name  submodules like `resolvers`, `queries`, and etc. 
I think we should be consistent in the naming of all submodules. If we are going to rename `client` to `sagemaker_client` we should rename `mutations` to `sagemaker_mutations`.  If we don't have such convention we end up having some file with prefixes and others not.

I see a huge plus from your suggestion: it should help easier navigate across the project (when you want to open a specific file), but as a downside, we will have more changes in the PR 

How would you like to proceed on this one? 🙂 


About `aws` module. I don't see an advantage of having it in `common`. I think every module will have their own `client.py` (or `module_client.py` as we decide)  in `aws` submodule instead of having a bigger module of all clients in `common`. It will help us encapsulate the logic of wrapper upon boto3 for each specific module. 
However, if some modules will require to use the same code, we can extract this code to `common`. 
Let me know what you think 🙂

backend/dataall/cdkproxy/stacks/policies/service_policy.py
nikpodsh: Yes, the permission should be defined by the module in the most cases. The example can be found in `modules/notebooks/services/permissions.py. 
But here we have a special case: notebooks and ML studio share the permissions in a few places. It was the reason why I put a few permissions in common. 

The right solution would be to define more granular permissions for each module, but I had had to write a migration script for existing permissions and wanted to avoided this overhead to keep the PR smaller 🙂

backend/dataall/cdkproxy/stacks/policies/_lambda.py
nikpodsh: I think this module will migrate to core and will have only some basic permissions that needed for the core part to function well. Then every module will register their permissions into the core part (by adding them into permission containers). I can define a better API for the permissions if you think that would be helpful

backend/dataall/modules/__init__.py
nikpodsh: They don't need loading of modules as far as I see. They are either isolated or use small portion of the code that they load explicitly. I think most of them will live under core, but something like sharing can be moved to modules (under cdk or we can add another submodule just for them).

tests/core/test_config.py
nikpodsh: This test is needed to make sure that we don't miss any module. By default, we want to include all modules in `config.json`, so the test will have all modules. If a customer wants to change the config she/he should also change the test accordingly. It helps prevent issues, then a person deleted another module or made changes in the config file unintentionally

tests/modules/notebooks/test_notebook_stack.py
nikpodsh: Yes, I extracted the part related to notebooks from `tests/api/test_stack.py` to a dedicated test

tests/modules/notebooks/test_sagemaker_notebook.py
nikpodsh: Yeah! Since it's related only to notebooks I decided to move tests related to notebooks to `modules` folder

tests/modules/notebooks/test_sagemaker_notebook.py
nikpodsh: No, we don't need it. Thanks!

backend/dataall/cdkproxy/stacks/sagemakerstudio.py
nikpodsh: It's related to modularization :) 
Before that, we had the mapping inside `TagUtil`. I had to move it out since notebooks were part of this mapping and pass it as a parameter

backend/dataall/cdkproxy/stacks/policies/cloudformation.py
nikpodsh: If a service policy doesn't depend on the other module and should be included every single time then it should be part of the core.

deploy/stacks/container.py
nikpodsh: I'd say it's more about reducing the duplication of the code 😄

backend/dataall/modules/notebooks/aws/client.py
```diff
@@ -0,0 +1,62 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SagemakerClient:
+    """
+    A Sagemaker notebooks proxy client that is used to send requests to AWS
+    """
+
+    def __init__(self, notebook: SagemakerNotebook):
```
dosiennik, line 15: Did you think about organising it in a bit different way? 
Seems like we will be creating a separate SageMaker client per the SageMaker notebook instance individually but the clients are rather aws account id, region specific. So if there are requests for notebooks belonging to the same awsAccount, region not sure if it will be the most optimal solution to create a new client for each of these?

I see we are passing the full notebook objects to the constructor but actually only the accountId, region, instance_name are used.

I would rather think of creating a client per aws account id/region and use a flyweight pattern to provide notebook or instance name to the methods for getting a status etc.

backend/dataall/modules/notebooks/db/models.py
```diff
@@ -1,11 +1,14 @@
...
 from sqlalchemy import Column, String, Integer
-from sqlalchemy.orm import query_expression
 
-from .. import Base
-from .. import Resource, utils
+from dataall.db import Base
+from dataall.db import Resource, utils
 
 
 class SagemakerNotebook(Resource, Base):
```
dosiennik, line 9: Shouldn't db layer be part of each module? So if I want to introduce a new module for instance a SagemakerNotebook related one I would have to define a db model/migration script for it in a separate directory for a particular module to not mix all the stuff together in one db/models directory?

backend/dataall/modules/notebooks/db/repositories.py
```diff
@@ -0,0 +1,53 @@
...
+Provides the API to retrieve / update / delete notebooks
+"""
+from sqlalchemy import or_
+from sqlalchemy.orm import Query
+
+from dataall.db import paginate
+from dataall.modules.notebooks.db.models import SagemakerNotebook
+
+
+class NotebookRepository:
```
dosiennik, line 12: I would introduce some hierarchy to enforce naming convention for CRUD operations that needs to be implemented like for instance Spring Data module does.

backend/dataall/modules/__init__.py
```diff
@@ -0,0 +1,39 @@
+"""
+Contains all submodules that can be plugged into the main functionality
+
+How to migrate to a new module:
+1) Create your python module
+2) Create the following submodules:
+    a) gql
```
dosiennik, line 7: curious if we can be less technology specific in naming the submodules, just an idea:
- api
- tasks (are we going to divide this into short vs long running tasks ? )
- infra

config.json
```diff
@@ -0,0 +1,3 @@
+{
+    "modules": ["notebooks"]
```
dosiennik, line 2: may be a silly question since I am not very advance in python, is there an option to register a module similar way how we currently register stacks or worker handlers? by using an annotation?

backend/api_handler.py
```diff
@@ -12,7 +12,9 @@
 from dataall.api.Objects import bootstrap as bootstrap_schema, get_executable_schema
```
dosiennik, line 12: we may think a little bit about the folder structure since we have a back-end then inside we have a data.all, isn't the full project data.all specific? looks like api_handler.py, aws_handler.py, cdkproxymain.py by its folder structure are not data.all specific? Similar applies to docker/prod, migrations

backend/dataall/modules/notebooks/db/repositories.py
```diff
@@ -0,0 +1,53 @@
...
+    """DAO layer for notebooks"""
+    _DEFAULT_PAGE = 1
+    _DEFAULT_PAGE_SIZE = 10
+
+    def __init__(self, session):
+        self._session = session
```
dosiennik, line 18: not sure if session should be a part of the constructor, I think it should be rather obtainer on a request basic since it stores/tracks all the objects which are part of the session

backend/dataall/core/config.py
```diff
@@ -0,0 +1,44 @@
...
+            raise KeyError(f"Couldn't find a property {key} in the config")
+
+        return copy.deepcopy(self._config[key])
+
+    def set_property(self, key: str, value: Any) -> None:
+        """Sets a property into the config"""
+        self._config[key] = value
+
+    @staticmethod
+    def _read_config_file() -> Dict[str, Any]:
```
dosiennik, line 28: Could you elaborate a bit about benefits of storting the config in the json file?

I am personally not very convinced about this approach, I see like this file is added to the container. If at some point we switch to using 'zip' files for lambda functions we will be adding this file to the zip file as well?

I would suggest to consider using any AWS service for storing the config file which were designed for it, did you consider https://docs.aws.amazon.com/appconfig/latest/userguide/what-is-appconfig.html or it doesn't fit into the requirements in here?

backend/dataall/modules/notebooks/aws/client.py
nikpodsh: I don't see a big difference of two solutions from the technical perspective. But in your provided example, we will do something like this in every single line when we want to use:
```
client(notebook.AWSAccountId, notebook.region).start_instance(notebook.NotebookInstanceName)
```
That will make us to write more code.  But let's also assume what happens if we need more parameters in the future? For example, we want to pass VPC, subnets, and InstanceType. The number of parameters will keep growing.

About the caching: this code will be executed in lambda, so the creation of the local cache with future re-using is not possible. 

For me, the original version has a better meaning: it's a AWS client for a specific Sagemaker notebook. You just need to pass the instance and the client will encapsulate the logic how the request to AWS will be executed. 
 
Of course, it can be revisited later when we see how the other `clients` will be organize. 🙂

dosiennik: I think like we don't need to create a new client for every request/a need to execute a sagemaker api operation.
For lambda the best practices the client should be created before the handler is executed and if the same lambda container will be reused for the next call we can reuse the client (for same region,account id) otherwise we are increasing the request handling time by creating a new client every time. 

 But I guess it is not the most important for now so I agree that we can get back to it at some point.

backend/dataall/modules/notebooks/db/repositories.py
nikpodsh: I actually like the idea of creating some convension, but I think Spring Data is a different story here: it uses `bytebuddy` to create an implementation of a repository based upon your interface. If you don't provide the method name in specific way, the spring boot magic won't work (e.g. `findNotebooksByInstanceTypeAndRegion`). This **enforce** you to write a code in the specific way.

We can create a hierarchy trying to achive the same, e.g. a base class with a generic methods like `save/delete/find`. All repos will extend from a the base repo and can re-use.this method. Hovewer, I still have a concern that when you need a specific method nothing will stop  you from creating something like that `remove_something_by_key` instead of `delete_by_key`, `create_object_with_default` instead of `save_with_default`. So we can end up with a useless base class at the end that brings only complexity. 🤔 
But please let me know what you think.

P.S. I believe if we truly want to create conventions we should create a code style guide where we can describe rules how the code should be written

backend/dataall/modules/notebooks/aws/client.py
dosiennik: but we indeed here have a bit of a challenge since we don't know the account id / region before the request will be sent so let's revisit it later

backend/dataall/modules/__init__.py
nikpodsh: I was trying to follow the existing approach and named them after the original modules. But I am happy to change them to be more generic. 


`api` - is a great name, but we already have `dataall.api` and `dataall.db.api` packages, so if it doesn't bring confusion to anyone we can use this (as alternative we can use `schema`, but it's more specific) 
`infra` - this one can be confusing from my point of view, since we can do infrastructure changes in lambda and ecs tasks. I was thinking about `ecs` as it's where the code will run (but again platform-specific)
`tasks` - this can be tricky. I thought about two approaches: 
1) To name it tasks and have two submodules `short` and `long` that represent short-living and long-living tasks accordingly (or `lambda` and `ecs` tasks, but it will be specific). This might be more clear for understanding
2) To rename it to `handlers` (as it used to be) and put there only short-living tasks. Long-living tasks will be placed under `ecs/cdk/infra` module. This might make a module importing a little bit easier.

That's very good that you brought this up, now we can decide how to make it correctly!
Naming is hard, anyone is very welcome to add their opinion on how to name  and structure it :)

backend/api_handler.py
nikpodsh: I can only try to speculate why it was done this way. 

I think that we can move all handlers to be part of `dataall` package, but they should be treated very carefully since there are docker and docker-compose files that depend on them as a entrypoint. And we should do it in a dedicated PR, just in case. 
About docker and migrations, I don't think that we should move them at all. Since `dataall` is a python package we don't want to import a migration somehow from the code. Let better them stay remote)

backend/dataall/modules/common/sagemaker/permissions.py
```diff
@@ -0,0 +1,27 @@
+"""
```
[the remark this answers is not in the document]
nikpodsh: In an ideal world, permissions should be module specific like we already have notebook's permissions in `dataall.modules.notebooks.services.permissions.py`.
But we also have a set of permissions that are shared between notebooks and ML studio. That was the primary reason why they are in `common`. Plus under the hood, ML studio uses the same policy as notebooks

`common` exists to avoid the code duplication between modules. Though ML studio is not a module yet.

At the moment `common` has a flat structure, since we expect that the modules will be pretty isolated and there will be not much intersections. But if we see a lot of `common` code we can create a deeper structure by creating of submodules  under `common` . Each submodule will represent a specific code share like `sagemaker` where we put the policy and the permissions between notebooks and ML studio there, or `shares` to put code for duplication of different share mechanisms, and etc

backend/migrations/versions/5fc49baecea4_add_enviromental_parameters.py
```diff
@@ -0,0 +1,145 @@
...
+        4) Migration every resource allocated for the environment to the environment_resources
+    """
+    try:
+        bind = op.get_bind()
+        session = orm.Session(bind=bind)
+
+        print("Creating of environment_parameters table...")
+        op.create_table(
+            "environment_parameters",
+            Column("environmentUri", String, primary_key=True),
```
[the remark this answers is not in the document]
nikpodsh: Nice catch! Thanks!

config.json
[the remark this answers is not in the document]
nikpodsh: There is a requirement that customer should be able to turn on and turn off every module. Not all customers are using all features of data.all. Another case is when you want to replace some functionality with your own e.g. you don't want to use existing mechanism for sharing datasets. In this case, you just delete a line in the config and put your module name with your code.

By default, all refactored modules will be added to config.json, but then you can change it for your specific needs.

The customization can be go beyond that and instead of having just name of module we can have a configuration for every module:
```
{
    "modules" : [
           "notebooks" : {
                    "use_free_tier_only:": true,
                     "notebook_prefix": "my_company.com_notebook"
            },
           "datasets": {
                   "some_prop": false
           },
          etc
    ]
}
```

backend/dataall/modules/loader.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Since there are different infrastructure components that requires only part
+    of functionality to be loaded, there should be different loading modes
+
+    Keys represent loading mode while value a suffix im module loading.
+    For example, API will try to load a graphql functionality under a module directory
+    The values represent a submodule and should exist
+    """
+
+    API = "gql"
```
dlpzx, line 25: [Idea/Question] - Customers usually customize certain APIs from a module. How are we going to handle small-customizations of a module? How do we allow customers to do add configurations to the module? One way to solve this with the current design is by copy/pasting the module and create a new "NotebooksCustom" module. It is a solution, but it can result of many custom" forks" which is what we are trying to avoid. I think it is a little soon to be included in this PR, but Ifor the future I would like to have something in the config.json such as:
- modules:
    - notebooks:
       max_notebooks_per_user: 5
       automatic_stop_instance: true

backend/api_handler.py
```diff
@@ -42,7 +45,6 @@ def adapted(obj, info, **kwargs):
                 username=info.context['username'],
                 groups=info.context['groups'],
                 schema=info.context['schema'],
-                cdkproxyurl=info.context['cdkproxyurl'],
```
dlpzx, line 45: [Clarification] - Like the es (ElasticSearch/OpenSearch) engine parameter, cdkproxyurl is now part of the context

backend/api_handler.py
```diff
@@ -23,6 +25,7 @@
 for name in ['boto3', 's3transfer', 'botocore', 'boto']:
     logging.getLogger(name).setLevel(logging.ERROR)
 
+load_modules(modes=[ImportMode.API])
```
dlpzx, line 28: [Clarification] Here is where the magic occurs, we import the modules `api` mode submodules

backend/cdkproxymain.py
```diff
@@ -20,7 +21,7 @@
     f"Application started for envname= `{ENVNAME}` DH_DOCKER_VERSION:{os.environ.get('DH_DOCKER_VERSION')}"
 )
 
-
+load_modules(modes=[ImportMode.CDK])
```
dlpzx, line 24: [Clarification] - for the case of cdkproxy we import each of the stacks. Then in the next line "registered_stacks()" we are regisstering all the imported stacks

backend/dataall/api/Objects/DataPipeline/resolvers.py
```diff
@@ -52,7 +52,7 @@ def create_pipeline(context: Context, source, input=None):
                 payload={'account': pipeline.AwsAccountId, 'region': pipeline.region},
             )
 
-    stack_helper.deploy_stack(context, pipeline.DataPipelineUri)
+    stack_helper.deploy_stack(pipeline.DataPipelineUri)
```
dlpzx, line 55: [Calrification] We use RequestContext instead of passing context.

backend/dataall/api/Objects/DataPipeline/schema.py
```diff
@@ -1,6 +1,8 @@
 from ... import gql
 from .resolvers import *
 from ...constants import DataPipelineRole
+from dataall.api.Objects.Environment.resolvers import resolve_environment
```
dlpzx, line 4: [Clarification ] Nice addition - Instead of using a resolver in each object that basically did the same (retrieve environment), we use the resolve_environment generic method to avoid duplications

backend/dataall/api/Objects/Environment/schema.py
```diff
@@ -42,6 +42,14 @@
...
 
 
+EnvironmentParameter = gql.ObjectType(
+    name='EnvironmentParameter',
+    fields=[
+        gql.Field(name='key', type=gql.String),
+        gql.Field(name='value', type=gql.String),
+    ]
+)
+
```
dlpzx, line 52: [Clarification] - Environment Parameter is actually the link of the Environment with the modules

backend/dataall/api/Objects/Stack/stack_helper.py
dlpzx, line 1: The location of this file has always been a bit confusing for me. All gql objects have the same structure except for this file that is used across many objects, it is like a handler. What do you think?

backend/dataall/core/context.py
```diff
@@ -0,0 +1,42 @@
...
+
+from dataclasses import dataclass
+from typing import List
+
+from dataall.db.connection import Engine
+from threading import local
+import opensearchpy
+
+
+_request_storage = local()
```
dlpzx, line 18: [Clarification] - it creates a local instance for this thread. In this local instance we can store info about the context of this particular API call keeping it isolated from other API calls

backend/api_handler.py
```diff
@@ -135,21 +137,25 @@ def handler(event, context):
             print(f'Error managing groups due to: {e}')
             groups = []
 
+        set_context(RequestContext(ENGINE, username, groups, ES))
+
         app_context = {
```
dlpzx, line 142: [Question] - this app_context can be replaced later in graphql_sync by get_context right?

backend/dataall/api/context.py
dlpzx, line 1: [Clarification] Notice that for RequestContext we use dataclasses which simplifies init

backend/dataall/core/environment/models.py
```diff
@@ -0,0 +1,33 @@
...
+from dataall.db import Resource, Base
+
+
+class EnvironmentParameter(Base):
+    """Represent the parameter of the environment"""
+    __tablename__ = 'environment_parameters'
+    environmentUri = Column(String, ForeignKey("environment.environmentUri"), primary_key=True)
+    key = Column('paramKey', String, primary_key=True)
+    value = Column('paramValue', String, nullable=True)
+
```
dlpzx, line 13: I am not understanding the primary keys here. `key` is not going to be unique and environmentUri will be duplicated. Can we have a unique primary key=environmentUri-key?

backend/dataall/core/permission_checker.py
```diff
@@ -0,0 +1,126 @@
...
+    # returns a function to call and static decorator if applied
+    return fn, staticmethod if static_func else no_decorated
+
+
+def has_resource_permission(permission):
+    """
+    Decorator that check if a user has access to the resource.
+    The method or function decorated with this decorator must have a URI of accessing resource
+    Good rule of thumb: if there is a URI that accesses a specific resource,
+    hence it has URI - it must be decorated with this decorator
```
dlpzx, line 79: <3

backend/dataall/modules/notebooks/aws/client.py
kukushking: An alternative is to cache the clients per account/region e.g.:

```
from functools import cache

@cache
def _get_client(account, region):
    session = ...
    return session.client(...)
```

backend/dataall/db/api/environment.py
```diff
@@ -1285,61 +1299,45 @@ def paginated_environment_redshift_clusters(
         ).to_dict()
 
     @staticmethod
-    def list_environment_objects(session, environment_uri):
-        environment_objects = []
+    def count_environment_objects(session, environment_uri):
```
dlpzx, line 1302: [Clarification] - nice addition, this method is used to check that environment deletion happens after all child objects are deleted. We do not need to list them, just count them

backend/dataall/db/api/environment.py
```diff
@@ -1372,9 +1370,9 @@ def delete_environment(session, username, groups, uri, data=None, check_perm=Non
             'environment', Environment.get_environment_by_uri(session, uri)
         )
 
-        environment_objects = Environment.list_environment_objects(session, uri)
+        has_resources = Environment.count_environment_objects(session, uri)
 
-        if environment_objects:
+        if has_resources:
```
dlpzx, line 1375: @nikpodsh but I think this should reference the output of count_environments. Right now I think this condition will always be true, no?

backend/dataall/modules/common/sagemaker/permissions.py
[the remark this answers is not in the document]
kukushking: > But we also have a set of permissions that are shared between notebooks and ML studio
Does this mean if I import both ML studio and notebooks and grant notebooks permissions, I will automatically gain some permissions for ML studio resources? Shouldn't these be separated?

backend/dataall/modules/loader.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class ImportMode(Enum):
+    """Defines importing mode
+
+    Since there are different infrastructure components that requires only part
+    of functionality to be loaded, there should be different loading modes
+
+    Keys represent loading mode while value a suffix im module loading.
+    For example, API will try to load a graphql functionality under a module directory
```
kukushking, line 21: Can modules be self-contained and import all infra components they require to properly function themselves e.g. in the modules `__init__.py`? Or there is a use case where user might only want to import APIs from the module but not the tasks for example (given that if there's both in the module, the APIs probably launch tasks)?

backend/dataall/modules/notebooks/db/repositories.py
dlpzx: Is there a way to generate a boilerplate module? Like introducing the module name "notebooks" and have some sort of command line that generates a directory of gql, services, db code with the input given. Like a quickstart on developing modules (maybe not for a first PR but something to thin about)

backend/dataall/db/api/target_type.py
```diff
@@ -1,64 +1,43 @@
...
-        )
+    """Resolves the read/write permissions for different type of resources (target types)"""
+    _TARGET_TYPES = {}
+
+    def __init__(self, name, read_permission, write_permission):
+        self.name = name
+        self.read_permission = read_permission
+        self.write_permission = write_permission
+
+        TargetType._TARGET_TYPES[name] = self
```
dlpzx, line 17: [Clarification] - Smart way of creating the dic. Everytime we define a TargetType we are registering it in _TARGET_TYPES. With this we can avoid hardcoding the supported_target_types
I like using the read and write permissions instead of relying on indexes

backend/dataall/db/api/target_type.py
```diff
@@ -1,64 +1,43 @@
...
-            pipeline=models.DataPipeline,
-            redshift=models.RedshiftCluster,
-        )
-        return [k for k, v in target_types_map.items() if v == model_name][0]
+
+TargetType("dataset", permissions.GET_DATASET, permissions.UPDATE_DATASET)
+TargetType("environment", permissions.GET_ENVIRONMENT, permissions.UPDATE_ENVIRONMENT)
+TargetType("mlstudio", permissions.GET_SGMSTUDIO_NOTEBOOK, permissions.UPDATE_SGMSTUDIO_NOTEBOOK)
+TargetType("pipeline", permissions.GET_PIPELINE, permissions.UPDATE_PIPELINE)
+TargetType("redshift", permissions.GET_REDSHIFT_CLUSTER, permissions.GET_REDSHIFT_CLUSTER)
```
dlpzx, line 43: This will go to the modules once we create them right?

backend/dataall/db/models/Environment.py
```diff
@@ -1,5 +1,5 @@
 from sqlalchemy import Boolean, Column, String
-from sqlalchemy.orm import query_expression
+from sqlalchemy.orm import query_expression, relationship
```
dlpzx, line 2: We can remove this import right?

backend/dataall/modules/notebooks/services/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+The service layer is a layer where all business logic is aggregated
+"""
+
+from dataall.db.api import TargetType
+from dataall.modules.notebooks.services import services, permissions
+from dataall.modules.common.sagemaker import permissions as common_permission
+
+__all__ = ["services", "permissions", "common_permission"]
+
+TargetType("notebook", permissions.GET_NOTEBOOK, permissions.UPDATE_NOTEBOOK)
```
dlpzx, line 12: It will be important in the future that we document very well what a module needs to have. For example I see new joiners forgetting to add a TargetType

dlpzx: But this is a note to ourselves, implementation-wise I think it is what it should be

backend/dataall/core/permission_checker.py
```diff
@@ -0,0 +1,126 @@
...
+def has_tenant_permission(permission: str):
+    """
+    Decorator to check if a user has a permission to do some action.
+    All the information about the user is retrieved from RequestContext
+    """
+    def decorator(f):
+        fn, fn_decorator = _process_func(f)
+
+        def decorated(*args, **kwargs):
+            with _get_session() as session:
```
dlpzx, line 104: [Clarification] - since the session is not passed, we need to get it directly in the decorator. We check if there is an open session (current_session) and if not we create an scoped session. That is why we added a new tiny method in connection.py.

backend/dataall/modules/common/sagemaker/permissions.py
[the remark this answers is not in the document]
dlpzx: I also think each module should have separated permissions (idk if I already commented it). Because even if ML Studio and Notebooks share some IAM permissions, there will be other that they don't and for a case where only Notebooks are enabled, we would be granting additional permissions for ML Studio that are not needed

backend/dataall/modules/loader.py
nikpodsh: It can be easily achieved with a small changes to the current design:

1) With a change of the structure of config.json and how we read it.
2) Providing a better interface how to get a nested properties like `config.get_property("modules.notebooks.max_notebooks_per_user")`

If we 100% sure that we will need, I can add these changes in this PR

backend/dataall/modules/notebooks/db/models.py
```diff
@@ -21,4 +24,3 @@ class SagemakerNotebook(Resource, Base):
     SubnetId = Column(String, nullable=True)
     VolumeSizeInGB = Column(Integer, nullable=True)
     InstanceType = Column(String, nullable=True)
-    userRoleForNotebook = query_expression()
```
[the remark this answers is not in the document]
nikpodsh: It was not used. We didn't write any value of `userRoleForNotebook` into the database. 
Plus, I remember that we had a discussion that all user roles are leftovers and had been used before we started to use groups

backend/api_handler.py
nikpodsh: It's a part of a config. Config is something global and cdkproxyurl is a good match for a property. So I decided to put it there

backend/cdkproxymain.py
nikpodsh: Yes, we import `__init__.py` for cdk module, which imports all stacks and policies. Then stacks register themselves using `@stack` decorator (just re-used the current implementation)

backend/dataall/api/Objects/Stack/stack_helper.py
nikpodsh: It will definitely be part of the core and migrate out of gql
But I don't know about handlers. We placed the code that is processed async into handlers, yet this helper called from GraphQL for creation an async task. Maybe a `StackService`, but let's revisit it when we get to the point where we are migrating the rest code to the core part  :smile

backend/api_handler.py
nikpodsh: That's the goal)

backend/dataall/core/environment/models.py
nikpodsh: It's already the same way as you described. It confused me a little bit too at the beginning, but it's the way how SQLAlchemy creates composite keys.

backend/dataall/modules/loader.py
nikpodsh: That was my original approach, but it crashed on the rocks of reality 🙂 
Whenever we want to load all submodules of a module, the submodules will start the registration into the application.
In our infrastructure, we have three components that uses different modes: GraphQL Lambda, Lambda for async work, and ECS for deployment.  A user might want to load API in GraphQL Lambda and the lambda will launch the tasks, but it will delegate the run to ECS which was loaded with the CDK mode. 

When we load the whole module in ECS as an example, gql part will try to register it in ariadne framework and will crash.

backend/dataall/db/api/target_type.py
nikpodsh: Yes, it's correct!

backend/dataall/db/models/Environment.py
nikpodsh: Sure, that's a leftover)

backend/dataall/modules/notebooks/services/__init__.py
nikpodsh: Yes! Perhaps we can also come up with better-defined API for importing that will make you to pass it

tests/utils/clients/graphql.py
```diff
@@ -71,6 +74,7 @@ def graphql_server():
             debug=app.debug,
         )
 
+        dispose_context()
```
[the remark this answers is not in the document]
nikpodsh: Yes, it is. Even though it should be overwritten with a new request, just to avoid any side-effects (like try to write some code out of request scope ) I think it's nice to clean it up

backend/dataall/core/permission_checker.py
```diff
@@ -0,0 +1,126 @@
...
+def _get_session():
+    context: RequestContext = get_context()
+    db = context.db_engine
+
+    # trying to re-use the open session if there is one
+    if db.current_session():
+        yield db.current_session()
+    else:
+        yield db.scoped_session()
+
```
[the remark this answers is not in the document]
nikpodsh: I didn't see that we used "nested" sessions, so it might be a good spot)

backend/local_graphql_server.py
```diff
@@ -27,10 +30,12 @@
 es = connect(envname=ENVNAME)
 logger.info('Connected')
 # create_schema_and_tables(engine, envname=ENVNAME)
+load_modules(modes=[ImportMode.API, ImportMode.TASKS])
 Base.metadata.create_all(engine.engine)
 CDKPROXY_URL = (
     'http://cdkproxy:2805' if ENVNAME == 'dkrcompose' else 'http://localhost:2805'
 )
+config.set_property("cdk_proxy_url", CDKPROXY_URL)
```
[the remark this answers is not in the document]
nikpodsh: As I mentioned it [here](https://github.com/awslabs/aws-dataall/pull/359#discussion_r1144621772), It's more like a property, global and not associated only with one request
P.S. I'd like also put `db_engine` and `es` into it, but they are not properties, so we might want create a `GlobalContext` or just `Context` at some point in the future 🤔

frontend/src/views/Environments/EnvironmentEditForm.js
```diff
@@ -72,11 +74,16 @@ const EnvironmentEditForm = (props) => {
...
             warehousesEnabled: values.warehousesEnabled,
-            resourcePrefix: values.resourcePrefix
+            resourcePrefix: values.resourcePrefix,
+            parameters: [
+              {
+                key: "notebooksEnabled",
+                value: String(values.notebooksEnabled)
+              }
+            ]
           }
```
dlpzx, line 87: In an ideal world, the frontend would also be modular and we would not need to hardcode "notebooksEnabled" but that is something to think about when we start with the frontend

backend/dataall/modules/notebooks/gql/resolvers.py
```diff
@@ -0,0 +1,120 @@
...
+def create_notebook(context: Context, source: SagemakerNotebook, input: dict = None):
+    """Creates a SageMaker notebook. Deploys the notebooks stack into AWS"""
+    RequestValidator.validate_creation_request(input)
+    request = NotebookCreationRequest.from_dict(input)
+    return NotebookService.create_notebook(
+        uri=input["environmentUri"],
+        admin_group=input["SamlAdminGroupName"],
+        request=request
+    )
+
```
dlpzx, line 19: Can we remove context and source from the resolvers' definitions?

backend/dataall/modules/notebooks/services/services.py
```diff
@@ -0,0 +1,242 @@
...
+)
+from dataall.utils.slugify import slugify
+from dataall.modules.notebooks.db.models import SagemakerNotebook
+from dataall.modules.notebooks.services import permissions
+from dataall.modules.common.sagemaker.permissions import MANAGE_NOTEBOOKS, CREATE_NOTEBOOK
+from dataall.core.permission_checker import has_resource_permission, has_tenant_permission, has_group_permission
+
+logger = logging.getLogger(__name__)
+
+
```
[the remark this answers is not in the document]
dlpzx: But since it is a class used in the resolvers, I would place it with the resolvers. Validation of input

backend/dataall/modules/notebooks/gql/resolvers.py
```diff
@@ -0,0 +1,120 @@
...
+        if not data.get('label'):
+            raise exceptions.RequiredParameter('name')
+
+        required(data, "environmentUri")
+        required(data, "SamlAdminGroupName")
+
+    @staticmethod
+    def _required(data: dict, name: str):
+        if not data.get(name):
+            raise exceptions.RequiredParameter(name)
```
dlpzx, line 120: Could this be part of common? or of core? I feel like this piece is going to be repeated in every module and core component

backend/dataall/cdkproxy/stacks/policies/service_policy.py
wolanlu: I agree here with @dlpzx - we should keep the permissions inside the module. These are part of the business logic that is encapsulated inside module

backend/dataall/modules/loader.py
```diff
@@ -0,0 +1,64 @@
...
+
+    log.info("Loading %d modules that have been found in the config", len(modules))
+    for module in modules:
+        if not _check_if_module_exists(module):
+            raise ValueError(f"Couldn't find module {module} under modules directory")
+        for mode in modes:
+            _import_submodule(module, mode)
+
+
+def _check_if_module_exists(module):
```
wolanlu, line 49: This should be named check_and_load_if_exists ;-)

backend/dataall/modules/loader.py
dlpzx: I think we should first merge this one and then work on expanding the feature, but niiiiice

wolanlu: And in the end it will be hard to integrate those "forks". This is a good way to provide completely new modules but for extending existing there should be more granule extension points so that one can use composition or inheritance and modify only what is needed e.g. add additional field to db/api. If I copy the whole module it will be hard for me to follow upstream changes (even harder the it is now because those changes will be physically in other files). But probably this is smth to address as a next step

backend/dataall/modules/notebooks/db/repositories.py
dosiennik: ok I raised it up since I am just get used to create all the stuff for the lambda function once before the handler method gets executed to reduce the cold start time and for the case the lambda env get reused for the next calls to not instantiate objects once again

config.json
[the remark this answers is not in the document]
dosiennik: I think we may need to distinguish between disabling the feature vs disabling the module? If someone will disable a feature in the data.all UI (notebooksEnabled set to false in the db) it would mean disabling the module?

backend/dataall/core/permission_checker.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+    Decorator that check if a user has access to the resource.
+    The method or function decorated with this decorator must have a URI of accessing resource
+    Good rule of thumb: if there is a URI that accesses a specific resource,
+    hence it has URI - it must be decorated with this decorator
+    """
+    def decorator(f):
+        fn, fn_decorator = _process_func(f)
+
+        def decorated(*args, uri, **kwargs):
```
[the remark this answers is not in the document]
nikpodsh: Yeah, we can introduce this decorator here as well.

backend/dataall/core/environment/models.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, env_uri, key, value):
+        self.environmentUri = env_uri
+        self.key = key
+        self.value = value
+
+    def __repr__(self):
+        return f'EnvironmentParameter(paramKey={self.key}, paramValue={self.value})'
+
+
+class EnvironmentResource(Base):
```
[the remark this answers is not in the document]
nikpodsh: I remembered why I introduced EnvironmentResource. I was thinking if we created a notebook , then turned off the notebook module, and tried to deleted the environment, we should have got an error. This was the purpose of this table. If we use a common interface for deleting, then we will be able to delete the environment if some modules are just turned off. 
But we should think if we need to handle the first scenario at all.

backend/dataall/modules/loader.py
nikpodsh: I thought about your suggestion:
We can even override the way how the modules should be loaded. We can do this in the config file. If there are no overrides, then we use default loading.
It can potentially look like this:
```
{
    "modules": [
         {
                "module": "notebook",
                "enabled": "true",
                "loading": {
                       "API": "new_schema"
                }
                "parameters": {}
         }
    ]
}
```
This will allow us to provide a new entrypoint for an existing module. In the example above, we load `new_schema` submodule instead of `gql`. This allow us to add/delete submodules.`new_schema` can even load `gql` patching some methods/classes/functions. That will make our application more customizable. The small downside of this approach: 1) hard to learn how it works (can be solved with well-written documentation) 2) configuration overhead :)

config.json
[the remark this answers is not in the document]
nikpodsh: When we disable the module, we won't see`notebooksEnabled` in UI at all (not now, but it will be when we modularize frontend). But if the notebook module is enabled we still have a choice to disable it per environment. 
Basically, the rule is: 
if all your environments don't need notebooks/ you are not going to use it -> the notebook module should be disabled
If you are not sure or some/all your environments are planning to work with it -> the module should be enabled, but you will still have a choice to disable it for each env you want by `notebooksEnabled=false`

backend/dataall/modules/notebooks/aws/client.py
nikpodsh: I don't think that the cache is needed here, since there is a huge chance that lambda execution will be finished right after the request. Since Lambda is stateless in the nature, after adding a cache, we might end up in the situation where we cached a client, the lambda process finished with the request, and a new lambda will start on the next request without the cached client.
But I like this cache from functools :)

backend/dataall/modules/common/sagemaker/policies.py
```diff
@@ -1,9 +1,23 @@
-from .service_policy import ServicePolicy
+from dataall.cdkproxy.stacks.policies.service_policy import ServicePolicy
```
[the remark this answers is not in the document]
nikpodsh: - [x] Moved under cdk

backend/dataall/cdkproxy/stacks/policies/service_policy.py
nikpodsh: - [x] Migrated permissions. Now, ML studios have they own permissions and there are no interaction with notebook permissions

tests/modules/notebooks/test_sagemaker_notebook.py
nikpodsh: - [x] Clean up of parameters in test method

backend/dataall/modules/__init__.py
nikpodsh: - [x] Renamed gql to api as you suggested. After discussion with @dlpzx, we decide to leave `cdk` and `tasks` for a while. But we might revisit it in the later PRs :)

backend/dataall/modules/common/sagemaker/permissions.py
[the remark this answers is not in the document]
nikpodsh: - [x] Created more granular permissions

backend/migrations/versions/5fc49baecea4_add_enviromental_parameters.py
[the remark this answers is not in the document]
nikpodsh: - [x] Migrated all tables to use foreign keys and changed a check to use `IntegrityError` if there are resources that are associated with the env.

backend/dataall/modules/loader.py
nikpodsh: - [x] Added a more clear way to disable module in the config.json
- [x] Changed a little the structure of config.json.
- [x] Changed a way how the modules are loaded. Provided the interface to load module in the code, so it should suffice for the existing modules :)

backend/dataall/db/api/environment.py
nikpodsh: - [x] Removed counting. We are starting to use foreign keys for this check

backend/dataall/db/models/Environment.py
nikpodsh: - [x] Removed

backend/dataall/core/environment/models.py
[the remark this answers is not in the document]
nikpodsh: - [x] Removed EnvironmentResources. Added foreign keys.

backend/dataall/modules/loader.py
nikpodsh: - [x] Renamed

backend/dataall/core/permission_checker.py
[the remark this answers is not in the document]
nikpodsh: - [x] It turned out that scoped session is already reusing current session and we don't need this method

backend/dataall/modules/notebooks/services/__init__.py
nikpodsh: Moved it to `NotebookApiInterface` and documented in the code

PR #380, opened 2023-03-23
feat: generate url with dynamically domain name for quicksight embeded dashboards

wolanlu:
feat: generate url with dynamically domain name for quicksight embeded dashboards

### Feature or Bugfix
- Feature

### Detail
Previously one had to configure statically data.all domain name inside QuickSight to allow dashboards to be embeddable inside data.all. With new api it is possible to dynamically set in inside request. User no longer has to configure QuickSight before using it.

### Relates


dlpzx:
Nice one @wolanlu! This removes the need to go to Quicksight to configure the allowed domains! Quicksight team has been adding a lot of APIs lately, thank you for keeping an eye :) The only thing left is to update the documentation in the userguide.

wolanlu:
true - thanks for pointing that out. I will handle the change

PR #387, opened 2023-03-28
fix: dev docker images base

AmrSaber:
### Feature or Bug-fix
- Bug-fix

### Detail
The latest version of dev docker images for FE and BE no-longer has `amazon-linux-extras`, this update changes the based of the docker image to use tag `2` (which is consistent with the rest of the images) instead of `latest` (which is a bad practice anyway -- see 2.4 [here](https://sysdig.com/blog/dockerfile-best-practices))


dlpzx (approved):
lgtm thank you!

PR #394, opened 2023-03-30
frontend: make styling consistent and remove dead code

AmrSaber:
### Feature or Bug-fix
- Refactoring

### Details
This PR does 2 things:
- Makes the styling consistent across the project
- Removes dead code (unused variables, functions, and imports)

This is done through eslint rules, I noticed that the project has prettier rules and uses prettier, but the rules were not in sync with eslint and were not enforced, so the styling was in consistent. So for this matter I configured eslint to enforce prettier rules so that the styling is consistent. And that's what the first commit does, there are no code changes what so ever in it, just styling updates.

For the second commit, I removed all unused variables, functions, and imports, there is also no logic changes, only removing the unused parts.

The third commit removes `==` in favour of `===` as this may cause some unintentional bugs (see [this stackoverflow question](https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons)).

The last commit is the actual configuration enforcing all the rules -- this commit is at the end not the beginning because the configuration kept changing until I reached the final state, and I didn't want to pollute the PR with a lot of configuration commits.

As stated, the PR's commits are meaningful and I recommend reviewing the PR by going through each commit individually

---

### DIFF
--- a/frontend/src/views/Pipelines/PipelineEnvironmentEditForm.js
+++ b/frontend/src/views/Pipelines/PipelineEnvironmentEditForm.js
@@ -1,5 +1,4 @@
 import React, { useEffect, useState } from 'react';
-import { useSnackbar } from 'notistack';
 import {
   Box,
   Button,
@@ -37,7 +36,6 @@ const PipelineEnvironmentEditForm = (props) => {
     handleCountEnvironmentValid
   } = props;
   const dispatch = useDispatch();
-  const { enqueueSnackbar } = useSnackbar();
   const client = useClient();
   const [kvEnvs, setKeyValueEnvs] = useState([]);
   const [envsToRemove, setEnvsToRemove] = useState([]);
@@ -50,14 +48,14 @@ const PipelineEnvironmentEditForm = (props) => {
     { value: 'prod', label: 'prod' },
     { value: 'other', label: 'other' }
   ];
-  const [environmentOps, setEnvironmentOps] = useState(
+
+  const environmentOps =
     environmentOptions && environmentOptions.length > 0
       ? environmentOptions
       : [
           { environmentUri: 'someUri', label: 'some' },
           { environmentUri: 'someUri', label: 'some2' }
-        ]
-  );
+        ];
 
   useEffect(() => {
     if (client && pipeline) {
--- a/frontend/src/views/Pipelines/PipelineEnvironmentEditForm.js
+++ b/frontend/src/views/Pipelines/PipelineEnvironmentEditForm.js
@@ -59,8 +59,6 @@ const PipelineEnvironmentEditForm = (props) => {
 
   useEffect(() => {
     if (client && pipeline) {
-      console.log('useeffect');
-      console.log(pipeline);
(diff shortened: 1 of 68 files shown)


frontend/src/views/Shares/ShareOutboxListItem.js
```diff
@@ -80,51 +79,51 @@ const ShareOutboxListItem = (props) => {
...
-                  alignItems: 'center',
-                  display: 'flex',
-                  py: 0.5
-                }}
-            >
-              <Typography color="textPrimary" variant="body2">
-                {`Pending items: ${share.statistics.pendingItems}`}
-              </Typography>
-            </Box>
+          <Box
```
nikpodsh, line 82: This code is identical to [frontend/src/views/Shares/ShareInboxListItem.js](https://github.com/awslabs/aws-dataall/pull/394/files#diff-da6d97e0fad1548010c159d1fa44007b2b1316aaa3db2e3d711d0898ede7d402). Let's extract it into dedicated form

frontend/src/views/Pipelines/PipelineEnvironments.js
```diff
@@ -1,49 +1,38 @@
...
   const [environments, setEnvironments] = useState([]);
-  const [loading, setLoading] = useState(true);
 
   useEffect(() => {
-      if (client && pipeline) {
-        console.log("useeffect")
-        console.log(pipeline)
-        const environmentsSorted = pipeline.developmentEnvironments.nodes.sort((a, b) => {
+    if (client && pipeline) {
+      console.log('useeffect');
```
nikpodsh, line 25: Looks like the leftovers after debugging. It's not a great to print it in the prod env. Let's remove all console.log as well

frontend/src/views/Pipelines/PipelineEnvironmentEditForm.js
```diff
@@ -29,28 +28,46 @@ import listEnvironmentGroups from '../../api/Environment/listEnvironmentGroups';
...
+        ];
 
   useEffect(() => {
     if (client && pipeline) {
-      console.log("useeffect")
-      console.log(pipeline)
-      const environmentsSorted = pipeline.developmentEnvironments.nodes.sort((a, b) => {
-        return a.order - b.order;
-      });
+      console.log('useeffect');
```
nikpodsh, line 62: Same here

frontend/src/views/Shares/ShareOutboxListItem.js
AmrSaber: Refactoring and moving parts around will come in the subsequent PRs, I want this PR to be free from any logic so that it's easy to review and can be merged quickly. But I noted the comment and will apply it when working on the refactoring.

frontend/src/views/Pipelines/PipelineEnvironmentEditForm.js
AmrSaber: Thank you for the 2 comments, I added a rule to check for the logs and removed debugging logs

nikpodsh (approved):

dlpzx (approved):

PR #396, opened 2023-03-30
frontend: simplify dev dockerfile

AmrSaber:
### Feature or Bug-fix
- Refactoring

### Detail
For the development dockerfile in the frontend, instead of starting from amazon linux image and installing everything manually, it's a lot simpler to start from node image (that is also from ECR for security) and have all node-related environment pre-installed and just focus the dockerfile on project-related configuration.

This PR does not introduce any update in the logic, it just updates the development dockerfile and adds a `.dockerignore` file for frontend to ignore `node_modules` and `build` folders.

---

### DIFF
--- a/frontend/.dockerignore
+++ b/frontend/.dockerignore
@@ -1,0 +1,3 @@
+node_modules
+build
+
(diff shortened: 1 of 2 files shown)


dlpzx (approved):
Looks good. A probable dumb question that came to my mind. Right now the local image and the image that we use in prod are the same, so when we work locally we will be able to recreate/spot issues that might appear in prod right? To do this now we would need to just modify the docker-compose file right?

AmrSaber:
I don't think you will be able to use the production image without some updates, as it uses different configuration for nginx and it does not get the .env from `frontend/docker/dev/`

PR #406, opened 2023-04-06
Added value returned when working with locally with Docker

dlpzx:
### Feature or Bugfix
- Bugfix
When running docker compose in graphql we always get the error "secret not found" for the externalId secret.

#### Alternative 1 (implemented in the PR)
- Added return value for when a secret externalId is not found - for dkrcompose.
- linked environments need to create a pivotRole whose trust policy does not include externalId condition

Pros: does not need a deployment of data.all

#### Alternative 2
- Manually create a secret called dataall-externalId-dkrcompose --> add documentation in GitHub pages
- linked environments work with a pivotRole whose trust policy includes the externalId condition

Pros: No changes in the pre-reqs of linking an environment

#### Alternative 3
- Automatic creation of a secret called dataall-externalId-dkrcompose, as part of the backend stack
- linked environments work with a pivotRole whose trust policy includes the externalId condition

Pros: No changes in the pre-reqs of linking an environment. No manual steps.
Cons: additional secrets

### Conclusion
Alternative 1 can be implemented in conjunction with 2/3. I will document alternative 2 in gh-pages and 3 can be considered for next steps.

### Relates

### DIFF
--- a/backend/dataall/aws/handlers/sts.py
+++ b/backend/dataall/aws/handlers/sts.py
@@ -99,7 +99,7 @@ class SessionHelper:
         """
         return SessionHelper.get_secret(
             secret_name=f'dataall-externalId-{os.getenv("envname", "local")}'
-        )
+        ) or 'localNoExternalId'
 
     @classmethod
     def get_delegation_role_name(cls):


dosiennik:
Seems like currently I follow approach nr 2 so I simply created a secret as part of my own playground setup.

backend/dataall/aws/handlers/sts.py
```diff
@@ -99,7 +99,7 @@ def get_external_id_secret(cls):
         """
         return SessionHelper.get_secret(
             secret_name=f'dataall-externalId-{os.getenv("envname", "local")}'
-        )
+        ) or 'localNoExternalId'
```
dosiennik, line 102: I am personally not a big fan of mixing the production code with a code for local development.
If creating an own secret as a part of setting up a environment for playground is not desired I would think of making it configurable so the customer can decide whether it is needed to use external id additionally or not.

Then for the local setup we can set this parameter to false.

kukushking: I agree with the above.

dlpzx: Okey makey. I was also not very convinced :)  Closing this now

PR #411, opened 2023-04-13
frontend: opt into modularization structure

AmrSaber:
### Feature or Bug-fix

- Feature

### Detail
- I am working as we agreed on the planning mentioned below
- First step I took is to unify all the exports and make most of them into named exports (so it's easier to move things around and to have each directory's export aggregated to be imported from its main directory)
- I am trying to make the commit meaningful so that they can be reviewed individually, but that does not always work, as there is a lot of moving things around
- I created a branch out of `modularization-main` called `modularization/frontend` and made it the target of this PR so that FE updates are separated as we agreed

**Important note**: This PR does not contain any logic updates whatsoever, only moving things around, and updating their export/import with a minor exception of inlining the -previously- component `Topics` which is now updated to be just an array and moved to constants.

### Relates

---


frontend/src/App.js
```diff
@@ -1,20 +1,24 @@
...
-import { createMaterialTheme } from './theme';
 
-const App = () => {
+export const App = () => {
   const content = useRoutes(routes);
   const { settings } = useSettings();
   const auth = useAuth();
   useScrollReset();
 
+  console.log('hello');
```
nikpodsh, line 20: Should be removed

frontend/src/design/components/SearchInput.js
```diff
@@ -25,7 +25,7 @@ function SearchInput(props) {
             fullWidth
             onChange={onChange}
             onKeyUp={onKeyUp}
-            placeholder="Search"
+            placeholder="SearchIcon"
```
nikpodsh, line 28: Was it changed intentionally? Looks like it changed with the renaming.

nikpodsh: Ok, it seems not just text but the import of the icon. If so, please disregard the comment above

frontend/src/App.js
AmrSaber: Yes, my bad 😁

frontend/src/design/components/SearchInput.js
AmrSaber: Its description from the library says: `The short hint displayed in the input before the user enters a value` so I think you are right, it should be reverted. Probably got updated by mistake with a find and replace or something.

nikpodsh: If it's not importing, but just a placeholder then let's revert it for every change pair "Search" -> "SearchIcon" (I have come across this change in many places in the PR)

AmrSaber: Ok, will update that

frontend/src/App.js
AmrSaber: I updated the console rule to not allow any `console.log` only `error` and `info` as they are used in some places meaningfully.

frontend/src/design/components/SearchInput.js
AmrSaber: I reverted all the changes for the placeholders

frontend/src/modules/Misc/views/Glossaries/GlossaryManagement.js
```diff
@@ -227,7 +224,7 @@ const GlossaryManagement = (props) => {
             className={classes.root}
             defaultExpanded={['3']}
             defaultCollapseIcon={<ArrowDropDownIcon />}
-            defaultExpandIcon={<ArrowRightIcon />}
+            defaultExpandIcon={<ArrowRight />}
```
nikpodsh, line 227: I wonder why we changed it? Is this import equivalent to the previous?
I see that react docs suggest to import this as `import ArrowRightIcon from '@mui/icons-material/ArrowRight';` 
Just to check if this is an intentional change

frontend/src/modules/Misc/views/Organizations/OrganizationEnvironments.js
```diff
@@ -121,7 +121,7 @@ const OrganizationEnvironments = (props) => {
               }}
               onChange={handleInputChange}
               onKeyUp={handleInputKeyup}
-              placeholder="Search environments"
+              placeholder="SearchIcon environments"
```
nikpodsh, line 124: Last one)

frontend/src/index.js
```diff
@@ -22,9 +22,9 @@ ReactDOM.render(
           <LocalizationProvider dateAdapter={AdapterDateFns}>
             <SettingsProvider>
               <BrowserRouter>
-                <AuthProvider>
+                <LocalAuthProvider>
```
nikpodsh, line 25: Do we actually use it? It's weird that we have `LocalAuthProvider` in main `index.js`. It means to me that `CloudFront` is also using this

frontend/src/routes.js
```diff
@@ -12,152 +10,166 @@ const Loadable = (Component) => (props) =>
...
-  lazy(() => import('./views/Environments/EnvironmentCreateForm'))
+  lazy(() => import('./modules/Misc/views/Environments/EnvironmentCreateForm'))
 );
 const EnvironmentEditForm = Loadable(
-  lazy(() => import('./views/Environments/EnvironmentEditForm'))
+  lazy(() => import('./modules/Misc/views/Environments/EnvironmentEditForm'))
 );
 const EnvironmentView = Loadable(
-  lazy(() => import('./views/Environments/EnvironmentView'))
+  lazy(() => import('./modules/Misc/views/Environments/EnvironmentView'))
```
nikpodsh, line 41: I am pretty much sure that `Environment` will never be or need to be a module. It's a central piece of the application. `Organization` is too. There might be others. 
By this, I mean that there are some pieces that are unplugable. 

If we put `Environment` and `Organization` into modules we need to deal with a lot of inter-module dependencies

frontend/src/services/graphql/SagemakerNotebook/createSagemakerNotebook.js
```diff
@@ -1,6 +1,6 @@
 import { gql } from 'apollo-boost';
```
nikpodsh, line 1: I might be missing something, so please feel free correct me if I am wrong, but why we didn't put this code into modules? 
I thought we were going to localize the code as much as possible. So we could have `notebooks` folder and there will be the code only related to notebooks

frontend/src/constants.js
dlpzx, line 1: constants such as the list of topics are usually linked to a particular module. In this case topics is a property of Datasets. I am afraid customers with multiple Dataset modules or that customize this list of topics (very typical customization) do not find the constants or override some other customer constants

frontend/src/services/graphql/SagemakerNotebook/createSagemakerNotebook.js
dlpzx: Yes, I also thought that the modules services will be inside the modules. Otherwise we end up with the same situation than at the beginning, customers having to go to multiple places to develop a single feature

frontend/src/routes.js
AmrSaber: My opinion on this and Raul (senior SDE from my team) agreed with me here, is that we don't need to use file structure to force the users into some behaviour, we could agree on some kind of contribution guid, we could put a Github action to make sure they are not updating some part that they should not, and so on...

This part in the planning is left for future discussion, we don't need to finalise it now. We agreed to modularize the application gradually anyway. So I think in the future things will be more clear as to what works and what does not.

frontend/src/services/graphql/SagemakerNotebook/createSagemakerNotebook.js
AmrSaber: According to the planning (https://github.com/awslabs/aws-dataall/issues/398), the directory `src/services` is for the services that are used by several modules, and for any helpers related to the services.

What will happen (also according to the planning document) is that on each module that we update, we will move its related services inside that module -- By the end of this step, the only remaining services are the ones that are used in several modules.

frontend/src/modules/Misc/views/Organizations/OrganizationEnvironments.js
AmrSaber: My bad, updated it 😁

frontend/src/index.js
AmrSaber: You are correct, it should have been cognito auth. Fixed it

frontend/src/modules/Misc/views/Glossaries/GlossaryManagement.js
AmrSaber: Since this is a default export, we can assign any name we want upon importing it, so the name itself does not matter as long as you are importing it correctly.

That said, I have reverted it to the previous name to be more consistent with other icons import, and to be more descriptive.

frontend/src/constants.js
AmrSaber: I have updated the planning and split `src/constants.js` into `src/design/constants.js`, `src/utils/constants.js`, and `src/modules/constants.js` and later on the on in modules can be moved into its related module.

frontend/src/routes.js
nikpodsh: My biggest concerns here is only the folder where the changes placed. It seems strange for me that the environment is in modules and also in Misc (since it's a central part).
In the planning, I suggested using core for this reason. First, everything would be core and then we would split onto modules. Now, we have to do the other way around. 
Basically, that's not a blocker, so if the other are agree we can merge it

AmrSaber: it's all the same really, the FE won't be merged into main until the modules are finished anyway

Issue #412, opened 2023-04-13
Modularization of Datasets (backend)


nikpodsh:
A work-item for ongoing modularization work for backend  #295

PR #413, opened 2023-04-13
Dataset Modularization pt.1 

nikpodsh:
### Feature or Bugfix
- Refactoring (Modularization)

### Relates
- Related issues #295 and #412 

### Short Summary
First part of migration of `Dataset` (`DatasetTableColumn`) TL;DR :) 

### Long Summary 
Datasets are huge. It's one of the central modules that's spread everywhere across the application. Migrating the entire Dataset piece would be very difficult task and, more importantly, even more difficult to review. Therefore, I decided to break down this work into "small" steps to make it more convenient to review. 
Dataset's API consist of the following items:
* `Dataset`
* `DatasetTable`
* `DatasetTableColumn`
* `DatasetLocation`
* `DatasetProfiling`

In this PR, there is only creation of `Dataset` module and migration of `DatasetTableColumn` (and some related to it pieces). Why? Because the plan was to migrate it, to see what issues would come up along with it and to address them here. The refactoring of `DatasetTableColumn` will be in other PR.
The issues: 
1) Glossaries
2) Feed
3) Long tasks for datasets
4) Redshift

Glossaries rely on GraphQL UNION of different type (including datasets). Created an abstraction for glossary registration. There was an idea to change frontend, but it'd require a lot of work to do this

Feed: same as glossaries. Solved the similar way. For feed, changing frontend API is more feasible, but I wanted it to be consistent with glossaries

Long tasks for datasets. They migrated into tasks folder and doesn't require a dedicated loading for its code (at least for now). But there are two concerns:
1) The deployment uses a direct module folder references to run them  (e.g. `dataall.modules.datasets....`, so basically when a module is deactivated, then we shouldn't deploy this tasks as well). I left a TODO for it to address in future (when we migrate all modules), but we should bear in mind that it might lead to inconsistencies.
2) There is a reference to `redshift` from long-running tasks = should be address in `redshift` module 

Redshift: it has some references to `datasets`. So there will be either dependencies among modules or small code duplication (if `redshift` doesn't rely hard on `datasets`) = will be addressed in `redshift` module

Other changes:
Fixed and improved some tests 
Extracted glue handler code that related to `DatasetTableColumn`
Renamed import mode from tasks to handlers for async lambda.
A few hacks that will go away with next refactoring :) 

Next steps:
[Part2 ](https://github.com/nikpodsh/aws-dataall/pull/1) in preview :) 
Extract rest of datasets functionality (perhaps, in a few steps)
Refactor extractor modules the same way as notebooks
Extract tests to follow the same structure.


backend/dataall/modules/datasets/services/dataset_table.py
dlpzx, line 1: Very very tiny comment, but can we rename the "sync" function to something more meaningful? I have intended to do this for a while

deploy/stacks/container.py
```diff
@@ -174,12 +175,13 @@ def __init__(
             update_bucket_policies_task.task.security_groups
         )
```
dlpzx, line 177: This is a #TODO but I think it is not that complicated to add parameters or read parameters. My concern is, how do we design something that scales? Are we going to have deployment changes/architectural components that change or are added based on modules selected?

nikpodsh: Yes, it shouldn't be difficult to implement, but I'd rather implement it in a dedicated PR after migrating the other dataset/sharing tasks 

I think that we should not replicate changes between config.json and cdk.json, otherwise it'd be mess of configuration. So, I suggest placing all information about modules (even if it can impact deployment) to config.json and consulting with it during the deployment. 
Of course, this approach has its own drawbacks, but it should be easy to implement and, more importantly, change if we decide to go this the other approach.
What do you think?

backend/dataall/modules/datasets/handlers/glue_column_handler.py
```diff
@@ -0,0 +1,141 @@
...
+                            f'Column {column.name} updated successfully: {response}'
+                        )
+                return True
+
+            except ClientError as e:
+                log.error(
+                    f'Failed to update table column {column.name} description: {e}'
+                )
+                raise e
+
```
dlpzx, line 103: This method does not make much sense here. we have the glue_table and the glue_column handlers and this is a function is a lf_table function that is inside the glue_column handler.

In total we have 2 functionalities:
- glue (tables and columns) schema handler
- lake formation permissions handler

backend/dataall/modules/datasets/services/dataset_table.py
```diff
@@ -224,7 +224,7 @@ def get_dataset_table_by_uri(session, table_uri):
         return table
 
     @staticmethod
-    def sync(session, datasetUri, glue_tables=None):
+    def sync_existing_tables(session, datasetUri, glue_tables=None):
```
dlpzx, line 227: :)

backend/dataall/modules/loader.py
```diff
@@ -19,9 +19,9 @@ class ImportMode(Enum):
     of functionality to be loaded, there should be different loading modes
     """
 
-    API = "api"
-    CDK = "cdk"
-    TASKS = "tasks"
+    API = auto()
+    CDK = auto()
+    HANDLERS = auto()
```
dlpzx, line 25: I like this, it defaults to the lowercase right?

backend/dataall/modules/loader.py
```diff
@@ -57,7 +57,7 @@ def load_modules(modes: List[ImportMode]) -> None:
             log.info(f"Module {name} is not active. Skipping...")
             continue
```
dlpzx, line 59: active does not need to be checked because we are checking it in 56

backend/dataall/modules/loader.py
dlpzx, line 1: We will always have a loader per module right?

dlpzx (approved):

backend/dataall/modules/loader.py
nikpodsh: No, we have only one loader that loads all modules consequently depending on the mode.

backend/dataall/modules/loader.py
nikpodsh: It's used to apply default values to enums. It can be often seen when the value is not important

backend/dataall/modules/datasets/services/dataset_table.py
nikpodsh: Tried my best :)

backend/dataall/modules/datasets/handlers/glue_column_handler.py
nikpodsh: Agree! I was planning to split the handlers code into two pieces: `handlers` and `clients`. Handlers is the code that's need for  `@Worker.handler` and `clients` will contain API to make calls to AWS. I wanted to do it in the refactoring PR, but since you pointed it out, I will do it here at least for LF

PR #414, opened 2023-04-13
frontend: notebook modularization

AmrSaber:
### Feature or Bugfix
- Refactoring

### Detail
- Modularize notebooks
- Move notebooks views into notebooks module
- Move services that were only used by notebooks into notebooks module
- Move notebooks components into notebooks module

### Relates

---


nikpodsh (approved):

frontend/src/modules/Misc/views/Notebooks/NotebookEditForm.js
dlpzx, line 1: This view is not used at the moment because there is not an edit button in notebooks to edit properties such as description of the notebook. I think that the edit functionality should be there, but since it is not there at the moment is better to remove it all in all and redesign it in a different task.

PR #415, opened 2023-04-14
frontend: enable/disable modules

AmrSaber:
### Feature or Bug-fix
- Feature

### Detail
- This PR implements the feature for enabling/disabling modules, and applies that for notebooks module.
- I have tested dockerfile configuration locally for the dev dockerfile and have applied the same changes to the production version but I could not test it.

### Relates
- https://github.com/awslabs/aws-dataall/issues/410

---


frontend/package.json
```diff
@@ -6,6 +6,10 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
+    "copy-config": "mkdir -p ./src/generated; cp ../config.json ./src/generated/",
+    "postinstall": "yarn copy-config",
+    "prestart": "yarn copy-config",
+    "prebuild": "yarn copy-config",
```
dlpzx, line 12: 1. This is the first step of the PR, here we are adding the `copy-config` script that will be executed when we start or build the application

dlpzx: In the postinstall we add it to avoid importing issues, issues of the type notFound because it does not find the config.json

frontend/src/routes.js
```diff
@@ -1,6 +1,7 @@
 import { lazy, Suspense } from 'react';
 import { AuthGuard, GuestGuard } from './authentication';
 import { DefaultLayout, LoadingScreen } from './design';
+import config from './generated/config.json';
```
dlpzx, line 5: 2. Now that the file is in the source we can read the config and enable or disable the routes of a module if it is disabled (conditional rendering below)

frontend/src/design/components/layout/DefaultSidebar.js
```diff
@@ -12,133 +7,144 @@ import {
...
+      path: '/console/dashboards',
+      icon: <MdShowChart size={15} />
+    };
+
+    const notebooksSection = {
+      title: 'Notebooks',
+      path: '/console/notebooks',
+      icon: <SiJupyter size={15} />,
+      active: config.modules.notebooks.active
+    };
```
dlpzx, line 74: 3. This 'active' flag enables/disable the module from the side pane. See line 139 where we do the filter to show only active modules. In the future we will need to add this parameter to all modules

docker-compose.yaml
```diff
@@ -91,8 +91,8 @@ services:
 
   frontend:
     build:
-      context: ./frontend
-      dockerfile: docker/dev/Dockerfile
+      context: .
```
dlpzx, line 94: 5. The context of the dockerfile needs to be updated because we need to read the config.json file

frontend/docker/dev/Dockerfile
```diff
@@ -2,13 +2,18 @@ FROM public.ecr.aws/docker/library/node:18-slim
 
 WORKDIR /app
```
dlpzx, line 4: 6. Because of changes in 5 we need to update the dockerfiles paths

dlpzx (approved):

PR #409, opened 2023-04-11
V1.5.0 Features

dlpzx:
### Feature or Bugfix
- V1.5.0 Features. Check each PR for a complete description of the feature.

### Detail
- #292 
- #355 
- #337 
- #427 
- #431 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #416, opened 2023-04-18
Fix issue with duplicate resources - backwards compatibility

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
- In v1.5.0 SageMaker Studio base infra is moved out in a nested stack. As a consequence sometimes there are conflicts with already existing AWS resources that the nested stack is trying to re-create. But the issue is bigger than that, the new nested stack would create a new Sagemaker domain and delete the previous one. If there are existing Sagemaker studio users this results in errors because it cannot delete a domain with existing users. We have 2 alternatives: 1) migrate existing users to the new profile 2) revert the changes and put the Sagemaker resources inside the environment stack, this will keep the previous resources.

Alternative 1 is quite complex (and I am not sure it is even possible). The benefit that it brings, which is "logical separation of environment-module resources" is not worth the effort.

### Relates
#409 

### DIFF
--- a/backend/dataall/cdkproxy/stacks/sagemakerstudio.py
+++ b/backend/dataall/cdkproxy/stacks/sagemakerstudio.py
@@ -42,7 +42,7 @@ class SageMakerDomain(NestedStack):
             )
             dataall_created_domain = ParameterStoreManager.client(
                 AwsAccountId=environment.AwsAccountId, region=environment.region, role=cdk_look_up_role_arn
-            ).get_parameter(Name=f'/dataall/{environment.environmentUri}/sagemaker/sagemakerstudio/domainId')
+            ).get_parameter(Name=f'/dataall/{environment.environmentUri}/sagemaker/sagemakerstudio/domain_id')
             return False
         except ClientError as e:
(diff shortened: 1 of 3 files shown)


AmrSaber (approved):

AmrSaber (approved):

backend/dataall/cdkproxy/stacks/sagemakerstudio.py
```diff
@@ -1,126 +1,19 @@
...
-from ...aws.handlers.sagemaker_studio import (
-    SagemakerStudio,
-)
 from ...utils.cdk_nag_utils import CDKNagUtil
 from ...utils.runtime_stacks_tagging import TagsUtil
 
 logger = logging.getLogger(__name__)
 
 
-class SageMakerDomain(NestedStack):
```
nikpodsh, line 34: Why should we delete `SageMakerDomain` ? If we decided not to go with the nested stack and rollback it, I'd still leave this class, remove only `NestedStack, changed init to `create_ml_studio_part` (or something like that) and keep using it :)

dlpzx:
I think I am going to start clean in another branch with your ideas, with no rush. Thanks @nikpodsh

PR #418, opened 2023-04-19
Fixed unresolved reference for ShareObjectPermission

nikpodsh:
### Feature or Bugfix
- Bugfix

### Detail
Working on datasets, I found that there was no `ApproversAndRequesters` value in `ShareObjectPermission`. Whenever the code reaches this point there will be an error. 

I added the other value to avoid an exception. But this is just for awareness.
We might need to add `ApproversAndRequesters` to `ShareObjectPermission`
I also considered that It might be dead code that we could delete.

### DIFF
--- a/backend/dataall/api/Objects/ShareObject/resolvers.py
+++ b/backend/dataall/api/Objects/ShareObject/resolvers.py
@@ -231,16 +231,6 @@ def resolve_user_role(context: Context, source: models.ShareObject, **kwargs):
             or dataset.SamlAdminGroupName in context.groups
         ):
             return ShareObjectPermission.Requesters.value
-        if (
-            dataset and dataset.stewards in context.groups
-            and (
-                source.owner == context.username
-                or source.principalId in context.groups
-                or dataset.owner == context.username
-                or dataset.SamlAdminGroupName in context.groups
-            )
-        ):
-            return ShareObjectPermission.Approvers.value
         else:
             return ShareObjectPermission.NoPermission.value
 
--- a/backend/dataall/api/Objects/ShareObject/resolvers.py
+++ b/backend/dataall/api/Objects/ShareObject/resolvers.py
@@ -222,13 +222,15 @@ def resolve_user_role(context: Context, source: models.ShareObject, **kwargs):
         return None
     with context.engine.scoped_session() as session:
         dataset: models.Dataset = db.api.Dataset.get_dataset_by_uri(session, source.datasetUri)
-        if dataset and dataset.stewards in context.groups:
-            return ShareObjectPermission.Approvers.value
-        if (
-            source.owner == context.username
-            or source.principalId in context.groups
+        if dataset and (
+            dataset.stewards in context.groups
             or dataset.owner == context.username
             or dataset.SamlAdminGroupName in context.groups
+        ):
+            return ShareObjectPermission.Approvers.value
+        if dataset and (
+            source.owner == context.username
+            or source.principalId in context.groups
         ):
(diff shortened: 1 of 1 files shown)


backend/dataall/api/Objects/ShareObject/resolvers.py
```diff
@@ -240,7 +240,7 @@ def resolve_user_role(context: Context, source: models.ShareObject, **kwargs):
                 or dataset.SamlAdminGroupName in context.groups
             )
         ):
-            return ShareObjectPermission.ApproversAndRequesters.value
+            return ShareObjectPermission.Approvers.value
```
dlpzx, line 243: I think this logic is repeated in line 225-226. Maybe we can have a single case that returns Approvers

nikpodsh: It seems so. I will remove the check

backend/dataall/api/Objects/ShareObject/resolvers.py
```diff
@@ -231,16 +231,6 @@ def resolve_user_role(context: Context, source: models.ShareObject, **kwargs):
...
-            dataset and dataset.stewards in context.groups
-            and (
-                source.owner == context.username
-                or source.principalId in context.groups
-                or dataset.owner == context.username
-                or dataset.SamlAdminGroupName in context.groups
-            )
-        ):
-            return ShareObjectPermission.ApproversAndRequesters.value
         else:
```
dlpzx, line 234: If I look at the logic of this, we are classifying users into Approvers, Requesters and NoPermissions.
Approvers should be = dataset owners and stewards
Requesters = share owners
NoPermissions = none of the above

Following this definitions, we should have something like:
```
       if dataset and (
                dataset.stewards in context.groups 
                or dataset.SamlAdminGroupName in context.groups
                or dataset.owner == context.username
       ):
            return ShareObjectPermission.Approvers.value
        if (
            source.owner == context.username
            or source.principalId in context.groups
        ):
            return ShareObjectPermission.Requesters.value
```

We probably did not spot this issue because we were not using additional stewards (by default if no stewards are selected it defaults to the dataset.SamlAdminGroupName)

nikpodsh: We didn't spot the issue because we have two checks that are almost identical. There is no way that the second check will be true, but first not.
The first check that returns Approvers.value: 
```
dataset and dataset.stewards in context.groups 
```
Second check that should retrun broken value ApproversAndRequesters:
```
dataset and dataset.stewards in context.groups and (more checks....)
```
So basically, this is a dead code) 

If you are still sure that the `Approvers` logic should be changed as you suggested, please let know I'll do it :)

dlpzx: Let's see it with an example, because I am not convinced hahahah
Dataset: DatasetA1
- owner: userA
- SamlAdminGroupName: groupA
- Stewards: None --> it defaults to groupA
ShareRequest: ShareB
- owner: userB
- principalId: groupB

groupA should always be classified as Approvers and groupB as Requesters



Dataset: DatasetA2
- owner: userA
- SamlAdminGroupName: groupA
- Stewards: groupA2
ShareRequest: ShareB
- owner: userB
- principalId: groupB

nikpodsh: Got it! I thought that we were discussing about why the check is not hit, but you are suggesting to change the logic in the check. I'll gladly do that :)

dlpzx (approved):
Looks perfect now

Issue #352, opened 2023-03-06
SageMaker Studio long start-up time


atzev:
### Describe the bug

As a user, when I open SageMaker Studio from the data.all portal I have to wait several minutes before I can start using Studio. 

### How to Reproduce

```
*P.S. Please do not attach files as it's considered a security risk. Add code snippets directly in the message body as much as possible.*
```
New user:
1. Log in to data.all with a user who has access to create SageMaker Studio profiles
2. Go to ML Studio and create new ML Studio profile
3. Click on launch JupyterLab
4. Amazon SageMaker Studio creates the app
5. Studio starts loading and displays "Loading... the loading screen is taking a long time. Would you like to clear the workspace or keep waiting?"
6. The user has to click Keep Waiting or clear workspace for 2-3 times until Studio opens probably 10 or more minutes later

Returning user:
1. Log in to data.all with a user who has access to SageMaker Studio profiles
3. Go to ML Studio and click on launch JupyterLab
5. Studio starts loading and displays "Loading... the loading screen is taking a long time. Would you like to clear the workspace or keep waiting?"
6. The user has to click Keep Waiting or clear workspace for 2-3 times until Studio opens probably 10 or more minutes later

### Expected behavior

For existing users, Studio load within seconds after the use clicks to launch Jupyter Lab icon.
For new users,  Studio should launch as soon as the app is created.

### OS

Mac

### Python version

3.8

### AWS data.all version

1.4.1

dlpzx:
Hi @atzev, I can confirm this behavior. I will pass this information internally to see if we can pick up this task and improve the user experience when using ML Studio from data.all

tasneemfazl:
Hello Team, do we have any fix or workaround for this issue?

dlpzx:
A colleague has done some investigation and has a explanation for the issue. Basically in the CDK code, the SageMaker Domain is defined with “VPCOnly” together with the default VPC (which has only public subnets):
https://github.com/awslabs/aws-dataall/blob/main/backend/dataall/cdkproxy/stacks/environment.py#L581
SageMaker Domain doesn’t support this (VPCOnly + public subnet):
[Doc link stating the requirements (no 1.) that when using VPCOnly mode the Subnets must be private](https://docs.aws.amazon.com/sagemaker/latest/dg/studio-notebooks-and-internet-access.html#studio-notebooks-and-internet-access-vpc#studio-notebooks-and-internet-access-vpc-requirements)
[...]

PR #420, opened 2023-04-19
Backwards compatibility of SageMaker Studio infra + add VPC creation for SageMaker domain

dlpzx:
### Feature or Bugfix
- Feature
- Bugfix

### Detail
- Instead of creating the SageMaker Studio domain as a nested stack we create it as part of the environment stack. To clearly show that the resources created for SageMaker are part of the ML Studio functionality they `check_existing_sagemaker_studio_domain` and `create_sagemaker_domain_resources` are class methods of `SageMakerDomain` placed in `backend/dataall/cdkproxy/stacks/sagemakerstudio.py`.
- As reported in #352 data.all uses the default VPC of the account, which does not fill the requirements for SM Studio. This results in long start times. This PR also adds the creation of a dedicated VPC that solves the issue of slow starts.

### Relates
- #409 
- #352 


backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -555,33 +555,25 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
             )
 
         # Create or import SageMaker Studio domain if ML Studio enabled
-        if self._environment.mlStudiosEnabled:
+        self.existing_sagemaker_domain = SageMakerDomain(
```
nikpodsh, line 558: Minor. We could create SageMakerDomain first and then we could re-use it.

```
domain = SageMakerDomain(
     stack=self,
     id='SageMakerDomain',
     environment=self._environment
)
has_existing_domain = domain..check_existing_sagemaker_studio_domain()
.....
domain.create_sagemaker_domain_resources(sagemaker_principals=[default_role] + group_roles)

```

dlpzx: good call, thanks!

nikpodsh (approved):

PR #429, opened 2023-04-21
Rename glue-handler to previous name to avoid CloudFormation unexpected deletions

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
Giving a more descriptive name to the logical ID of the glue-handler custom resource results in CloudFormation recreating it. That means that CFN:
1) creates new custom resource and tries to create db and grant permissions (db already exists, so it skips)
2) deletes old custom resource and in this action it deleted the db --> we cannot allow this as it is removing the database and tables from Glue which will break our sharing and LF permissions

### Relates
- #409 


dosiennik (approved):
LGTM

PR #431, opened 2023-04-23
Fix handlers v1m5m0 + documentation

dlpzx:
### Feature or Bugfix
- Bugfix
- Refactoring

### Detail
- Reverted LakeFormation registering location to a CloudFormation resource instead of using a custom resource. There was a bug on the CloudFormation LakeFormation resource and a storage location could not be updated. The bug has been fixed and we can now update the LakeFormation resources (which was needed to update from PivotRole to PivotRole-cdk)
- Readded the original custom resource that creates the dataset Glue database and grants permissions in Lake Formation. We wanted to replace the custom resource service token because it was causing Lambda policies to reach their maximum size; but it is not possible to update the service token of a custom resource, one has to recreate it. The issue is that when we directly replace the custom resource it deletes the database (on_delete) when the CloudFormation stack cleans up the resources. Deleting the Glue database does not delete the S3 data, but it deletes all Lake Formation permissions, which is critical for production use-cases. We cannot introduce this risk, so we have updated the old custom resource Lambda to not delete anything on the on_delete event and we will leave the deprecated custom resource until the next minor release where we will safely remove it from the dataset stack. In V1.6.0 we will ensure to notify that upgrading to V1.5.0 is a requirement before upgrading to V1.6.0.
- Added documentation for ML Studio networking requirements and new features
- Added UserGuide as pdf for V1.5.0

### Relates
- #409 

### DIFF
--- a/backend/dataall/aws/handlers/lakeformation.py
+++ b/backend/dataall/aws/handlers/lakeformation.py
@@ -6,7 +6,7 @@ from botocore.exceptions import ClientError
 from .sts import SessionHelper
 
 log = logging.getLogger('aws:lakeformation')
-
+PIVOT_ROLE_NAME_PREFIX = "datallPivotRole"
 
 class LakeFormation:
     def __init__(self):
@@ -25,7 +25,7 @@ class LakeFormation:
             registered_role_name = response['ResourceInfo']['RoleArn'].lstrip(f"arn:aws:iam::{accountid}:role/")
             pivot_role_name = SessionHelper.get_delegation_role_name()
             log.info(f'LF data location already registered: {response}, registered with role {registered_role_name}')
-            if registered_role_name[:11] == pivot_role_name[:11]:
+            if registered_role_name[:14] == PIVOT_ROLE_NAME_PREFIX:
                 log.info('The existing data location was created as part of the dataset stack. There was no pre-existing data location.')
                 return False
             return response['ResourceInfo']
--- a/backend/dataall/aws/handlers/lakeformation.py
+++ b/backend/dataall/aws/handlers/lakeformation.py
@@ -13,18 +13,18 @@ class LakeFormation:
         pass
 
     @staticmethod
-    def describe_resource(resource_arn, role_arn, accountid, region):
+    def check_existing_lf_registered_location(resource_arn, accountid, region):
         """
         Describes a LF data location
         """
         try:
             session = SessionHelper.remote_session(accountid)
             lf_client = session.client('lakeformation', region_name=region)
-
             response = lf_client.describe_resource(ResourceArn=resource_arn)
-
-            log.info(f'LF data location already registered: {response}, checking if data.all registered it ...')
-            if response['ResourceInfo']['RoleArn'] == role_arn:
+            registered_role_name = response['ResourceInfo']['RoleArn'].lstrip(f"arn:aws:iam::{accountid}:role/")
+            pivot_role_name = SessionHelper.get_delegation_role_name()
+            log.info(f'LF data location already registered: {response}, registered with role {registered_role_name}')
+            if registered_role_name[:11] == pivot_role_name[:11]:
                 log.info('The existing data location was created as part of the dataset stack. There was no pre-existing data location.')
                 return False
             return response['ResourceInfo']
--- a/backend/dataall/aws/handlers/lakeformation.py
+++ b/backend/dataall/aws/handlers/lakeformation.py
@@ -23,9 +23,8 @@ class LakeFormation:
             lf_client = session.client('lakeformation', region_name=region)
             response = lf_client.describe_resource(ResourceArn=resource_arn)
             registered_role_name = response['ResourceInfo']['RoleArn'].lstrip(f"arn:aws:iam::{accountid}:role/")
-            pivot_role_name = SessionHelper.get_delegation_role_name()
             log.info(f'LF data location already registered: {response}, registered with role {registered_role_name}')
-            if registered_role_name[:14] == PIVOT_ROLE_NAME_PREFIX:
+            if registered_role_name.startswith(PIVOT_ROLE_NAME_PREFIX):
                 log.info('The existing data location was created as part of the dataset stack. There was no pre-existing data location.')
                 return False
             return response['ResourceInfo']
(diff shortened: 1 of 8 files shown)


backend/dataall/aws/handlers/lakeformation.py
```diff
@@ -13,26 +13,25 @@ def __init__(self):
...
             lf_client = session.client('lakeformation', region_name=region)
-
             response = lf_client.describe_resource(ResourceArn=resource_arn)
-
-            log.info(f'LF data location already registered: {response}, checking if data.all registered it ...')
-            if response['ResourceInfo']['RoleArn'] == role_arn:
+            registered_role_name = response['ResourceInfo']['RoleArn'].lstrip(f"arn:aws:iam::{accountid}:role/")
+            pivot_role_name = SessionHelper.get_delegation_role_name()
+            log.info(f'LF data location already registered: {response}, registered with role {registered_role_name}')
+            if registered_role_name[:11] == pivot_role_name[:11]:
```
nikpodsh, line 28: Why 11? The default name is `dataallPivotRole` and we can also have `dataallPivotRole-cdk`. 
+ Let's create a constant with the proper name for this

dlpzx: Miscounted that one, but yes you are right it better be in a constant

backend/dataall/aws/handlers/lakeformation.py
```diff
@@ -6,33 +6,32 @@
...
             lf_client = session.client('lakeformation', region_name=region)
-
             response = lf_client.describe_resource(ResourceArn=resource_arn)
-
-            log.info(f'LF data location already registered: {response}, checking if data.all registered it ...')
-            if response['ResourceInfo']['RoleArn'] == role_arn:
+            registered_role_name = response['ResourceInfo']['RoleArn'].lstrip(f"arn:aws:iam::{accountid}:role/")
+            pivot_role_name = SessionHelper.get_delegation_role_name()
+            log.info(f'LF data location already registered: {response}, registered with role {registered_role_name}')
+            if registered_role_name[:14] == PIVOT_ROLE_NAME_PREFIX:
```
nikpodsh, line 28: This will always be false. (`registered_role_name[:14]` is `dataallPivotRo`)
Let's better use `registered_role_name.startswith(PIVOT_ROLE_NAME_PREFIX)` :)

backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -325,13 +325,10 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
             string_value=lf_default_settings_custom_resource.function_name,
             parameter_name=f'/dataall/{self._environment.environmentUri}/cfn/lf/defaultsettings/lambda/name',
         )
-
-        # Glue database custom resource
-        # This Lambda is triggered with the creation of each dataset, it is not executed when the environment is created
+        # Glue database custom resource - Old, to be deleted in future release
```
nikpodsh, line 328: I didn't catch this. Why do we consider this `no_delete` as old? In the main branch, we do some actions on `on_delete`

kukushking: As I understand that is addressed in the PR notes:
```
Deleting the Glue database does not delete the S3 data, but it deletes all Lake Formation permissions, which is critical for production use-cases. We cannot introduce this risk, so we have updated the old custom resource Lambda to not delete anything on the on_delete event and we will leave the deprecated custom resource until the next minor release where we will safely remove it from the dataset stack.
```

kukushking: > In V1.6.0 we will ensure to notify that upgrading to V1.5.0 is a requirement before upgrading to V1.6.0.

I wonder though if there is a way to avoid this

backend/dataall/aws/handlers/lakeformation.py
dlpzx: Yes, I don't know why I read that startswith was not available in 3.8 (our Lambdas Python version, yes we should upgrade them since Lambda Python 3.10 was released last week), but checking again it seems to be there, so I made the changes

dlpzx: Thank you!!

backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -357,30 +354,73 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
+            parameter_name=f'/dataall/{self._environment.environmentUri}/cfn/custom-resources/lambda/name',
+        )
+        # Glue database custom resource - New
+        # This Lambda is triggered with the creation of each dataset, it is not executed when the environment is created
+        entry_point = str(
+            pathlib.PosixPath(os.path.dirname(__file__), '../assets/gluedatabasecustomresource').resolve()
+        )
+
+        gluedb_lf_cr_dlq = self.set_dlq(f'{self._environment.resourcePrefix}-gluedb-lf-cr-{self._environment.environmentUri}')
+        gluedb_lf_custom_resource = _lambda.Function(
```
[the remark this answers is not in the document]
dlpzx: But yes, you are right, this is not the prettiest setup :)

backend/dataall/cdkproxy/assets/gluedatabasecustomresource_nodelete/index.py
```diff
@@ -0,0 +1,118 @@
...
+    if request_type == 'Create':
+        return on_create(event)
+    if request_type == 'Update':
+        return on_update(event)
+    if request_type == 'Delete':
+        return on_delete(event)
+    raise Exception('Invalid request type: %s' % request_type)
+
+
+def on_create(event):
```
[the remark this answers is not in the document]
dlpzx: Same as above, it is quite complicated (I am not sure if it is even possible) to do it that way. We would need to create one single Lambda with all the methods, on_create, on_update, on_delete and on_delete with no deletion and somehow redirect the events of one customer resource provider to the on_create, on_update, on_delete and for the other custom resource provider to on_create, on_update and on_delete with no deletion. Or create 2 Lambdas from multiple entries, or with a common layer

backend/dataall/cdkproxy/stacks/environment.py
dlpzx: Exactly, but as I explain below we might choose to have `no_delete` as the general behavior because deleting the database is a quite dangerous action that customers should be fully aware of

kukushking (approved):

PR #442, opened 2023-05-02
Datasets modularization pt.5

nikpodsh:
### Feature or Bugfix
- Refactoring

### Detail
Refactoring of the Dataset entity and related to it code.
Refactoring for Votes
Introduced DataPolicy (the same way as ServicePolicy was used used)
Extracted dataset related permissions. 
Used new `has_tenant_permission` instead of `has_tenant_perm` that allows not to pass unused parameters

### Relates
#412 and #295


backend/dataall/api/Objects/Vote/resolvers.py
```diff
@@ -49,3 +51,6 @@ def get_vote(context: Context, source, targetUri: str = None, targetType: str =
             data={'targetType': targetType},
             check_perm=True,
         )
+
+
+add_vote_type("dashboard", DashboardIndexer)
```
dbalintx, line 56: I'd leave here a TODO note to move this to the Dashboard module as soon as it gets created

nikpodsh: Sure! Why not :)

backend/dataall/cdkproxy/stacks/policies/data_policy.py
```diff
@@ -102,7 +103,7 @@ def generate_data_access_policy(self) -> iam.Policy:
 
         return policy
 
-    def get_statements(self):
+    def get_statements(self, session):
```
dbalintx, line 106: session is only used in the child class method, maybe replace it on this abstract level with args/kwargs?

backend/dataall/modules/datasets/services/share_notification_service.py
```diff
@@ -0,0 +1,101 @@
+from dataall.db import models
+from dataall.db.api import Notification
+from dataall.modules.datasets.db.models import Dataset
+
+
+class ShareNotificationService:
```
dbalintx, line 6: As this class only has static methods, shouldn't these functions be on module level? (without a class)

I'd also rename the file to follow the pattern in the folder: dataset_ ...

nikpodsh: I don't think that we will gain more benefits by removing the class. Yeah, it doesn't carry any functional role here, but it groups the code with the same functionality.

About the renaming: yes, I completely agree with you. This class will be changed when the sharing module comes up :)

backend/dataall/modules/datasets/services/permissions.py
dbalintx, line 1: I'd rename the file to dataset_permissions.py

backend/dataall/modules/datasets/tasks/subscription_service.py
dbalintx, line 1: Would it make sense to rename the file and the class to avoid confusion? It is scheduled as a task, but named as a service for which we have now a separate folder.

backend/dataall/modules/datasets/services/dataset_group_resource.py
```diff
@@ -0,0 +1,8 @@
+from dataall.core.group.services.group_resource_manager import GroupResource, GroupResourceManager
+from dataall.modules.datasets.db.dataset_repository import DatasetRepository
+
+
+class DatasetGroupResource(GroupResource):
```
dbalintx, line 5: This could be moved to dataset_repository

backend/dataall/core/group/services/group_resource_manager.py
```diff
@@ -0,0 +1,24 @@
+from typing import Protocol, List
+
+
+class GroupResource(Protocol):
```
dbalintx, line 4: As we agreed: this could be abstract class instead of a Protocol

backend/dataall/modules/datasets/services/dataset_service.py
```diff
@@ -623,3 +619,82 @@ def count_dataset_tables(session, dataset_uri):
             .filter(DatasetTable.datasetUri == dataset_uri)
             .count()
         )
+
+    @staticmethod
+    def query_environment_group_datasets(session, envUri, groupUri, filter) -> Query:
```
dbalintx, line 624: These query functions should be in dataset_repository (following your design pattern), but as we discussed this could also be in a followup :)

backend/dataall/modules/datasets/tasks/subscription_service.py
nikpodsh: Yes, let's rename it!

backend/dataall/modules/datasets/services/dataset_service.py
nikpodsh: Yes, I will extract this methods after migrating the sharing.

backend/dataall/api/Objects/Stack/stack_helper.py
```diff
@@ -84,15 +83,6 @@ def deploy_stack(targetUri):
         return stack
 
 
-def deploy_dataset_stack(dataset: models.Dataset):
```
dlpzx, line 87: I agree on moving this into the resolvers, it did not make much sense here

backend/dataall/modules/datasets/tasks/bucket_policy_updater.py
dlpzx, line 1: Unrelated to modularization: this task needs to be reviewed. It is triggered every 15 mins and it was originally conceived to add statements to the pre-existing bucket policy of imported datasets. Through this task we can enforce more restrictive access to imported buckets

nikpodsh: Yes, it looks strange. The task runs every 15 mins to update  policies for imported datasets. Most of the time it will do nothing, but just incur cost for running. 
I wonder why we can't update the policies during the importing of datasets?

backend/dataall/modules/datasets/aws/lf_table_client.py
```diff
@@ -22,21 +24,36 @@ def grant_pivot_role_all_table_permissions(self):
         :param table:
```
dlpzx, line 24: :param table: not needed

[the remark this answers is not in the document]
dlpzx: Looking at the `clients` I see your logic of dividing the clients by service (glue, lf) and by dataset-object (dataset, table, storage_location...). At first I thought that it would be easier to have a single service client, but I think this approach by service+ by object makes it easier to follow and allows us to scale more easily. The only duplicated lines would be the "creating the client" lines from 13 to 16 if there are other lf_something_clients, but we save on parameters as having the _table as part of the class, so I would say it pays off. 

The only thing to add, there is some sort of `lf_dataset_client` in the creation of datasets. In `backend/dataall/cdkproxy/assets/gluedatabasecustomresource` This is a custom resource Lambda that uses some LakeFormation methods.

backend/dataall/modules/datasets/handlers/glue_column_handler.py
```diff
@@ -38,7 +38,7 @@ def update_table_columns(engine, task: models.Task):
 
             aws_session = SessionHelper.remote_session(table.AWSAccountId)
```
dlpzx, line 40: minor suggestion: define params`table=table, aws_session=aws_session` to avoid mistakes

backend/dataall/modules/datasets/handlers/glue_dataset_handler.py
dlpzx, line 1: If it is going to contain only the crawler maybe we can rename the file to something that indicates that it is containing the Crawling handling

backend/dataall/modules/datasets/handlers/glue_table_handler.py
```diff
@@ -18,7 +17,7 @@ class DatasetColumnGlueHandler:
     @Worker.handler(path='glue.dataset.database.tables')
     def list_tables(engine, task: models.Task):
```
dlpzx, line 18: This method has a very bad name, not a problem of modularization but the function is actually syncing tables from glue catalog to data.all. And the handler should be called something like glue_table_sync_handler

backend/dataall/modules/datasets/handlers/sns_dataset_handler.py
dlpzx, line 1: In line 30 and 40 we are creating and using the sns client, would this not be part of clients? Something like 'sns_dataset_client' and then in the handlers we would have something like dataset_sns_subscription_handler

backend/dataall/modules/datasets/handlers/s3_location_handler.py
dlpzx, line 1: Same as other handlers, if we are creating separate files I would go for more precise names. This is related to creation of folders

backend/dataall/api/Objects/Environment/resolvers.py
```diff
@@ -370,23 +370,6 @@ def list_environment_group_permissions(
...
-            session=session,
-            username=context.username,
-            groups=context.groups,
-            uri=environmentUri,
-            data=filter,
-            check_perm=True,
-        )
-
-
 def list_shared_with_environment_data_items(
```
dlpzx, line 373: I guess that `list_shared_with_environment_data_items` will also be deleted from here and replaced by the `count` in GroupResources once shares are modularized right?

backend/dataall/aws/handlers/glue.py
```diff
@@ -370,142 +368,6 @@ def batch_delete_tables(**data):
...
-                    'crawler_name': f'{dataset.GlueDatabaseName}-{location}'[:52],
-                    'region': dataset.region,
-                    'accountid': dataset.AwsAccountId,
-                    'database': dataset.GlueDatabaseName,
-                    'location': location or f's3://{dataset.S3BucketName}',
-                }
-            )
-
-    @staticmethod
-    def create_glue_crawler(**data):
```
dlpzx, line 392: What happened with this method? I do not see it in glue_dataset_client, was it unused?

backend/dataall/cdkproxy/stacks/policies/data_policy.py
```diff
@@ -86,11 +84,14 @@ def generate_admins_data_access_policy(self) -> iam.Policy:
 
         return policy
 
-    def generate_data_access_policy(self) -> iam.Policy:
+    def generate_data_access_policy(self, session) -> iam.Policy:
         """
```
dlpzx, line 88: why do we need to pass the session?

backend/dataall/db/api/share_object.py
```diff
@@ -561,7 +563,9 @@ def submit_share_object(
 
         Share_SM.update_state(session, share, new_share_state)
```
dlpzx, line 565: ok!

backend/dataall/modules/datasets/cdk/dataset_data_policy.py
```diff
@@ -0,0 +1,33 @@
+from typing import List
+from aws_cdk import aws_iam as iam
+
+from dataall.cdkproxy.stacks.policies.data_policy import DataPolicy
+from dataall.modules.datasets.db.models import Dataset
+from dataall.modules.datasets.services.dataset_service import DatasetService
+
+
+class DatasetDataPolicy(DataPolicy):
```
dlpzx, line 9: Maybe we can rename it S3 policy as it is what is actually doing and then everything is a service policy

dlpzx: Right now "data" can be anything

backend/dataall/modules/datasets/handlers/sns_dataset_handler.py
nikpodsh: Yep, I had thought to migrate it later, but since you have pointed it out I will do it here

backend/dataall/aws/handlers/glue.py
nikpodsh: This was the dead code. It should have been been invoked by sending SQS action `glue.dataset.crawler.create`, but there was no `glue.dataset.crawler.create` anymore in the project, I deleted it

tests/api/test_dataset_location.py
```diff
@@ -70,10 +73,8 @@ def test_get_dataset(client, dataset1, env1, user, group):
```
dlpzx, line 74: What is the advantage of using MagicMock?

nikpodsh: Honestly? it's just working :) 
I couldn't find a better way to mock _instance_ of class. Tried a few other ways to patch an instance, but got only `FAILED`. This one worked)

dlpzx (commented):
thank you!

nikpodsh:
Added a new commits:
Most of them were review remarks, but I also returned triggering the alarms as we discussed (+ added a line to trigger alarm not only create a message) @dlpzx 
Extracted DatasetRole (kudos to @dbalintx for noticing it)
Created a way to extend `EnvironmentSetup` (which is an environment stack) and extracted glue profiler since it looks related to datasets. I didn't do that for the others like`lakeformationdefaultsettings` and `gluedatabasecustomresource` since wasn't sure if there are only dataset-specific.

PR #446, opened 2023-05-09
Modify docker-compose yaml to read region and default region from env…

dlpzx:
### Feature or Bugfix
- Bugfix
- Refactoring

### Detail
Added `AWS_REGION` to the environment variables of the Docker containers for local development. Set both`AWS_DEFAULT_REGION` and `AWS_REGION` to their values set on the terminal where `docker-compose up` is run.
If these values are not set, `eu-west-1` is used as default
Another PR with better instructions to the github pages documentation (deploy locally) will follow.

### Relates

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -16,7 +16,8 @@ services:
       - db
     environment:
       envname: 'dkrcompose'
-      AWS_DEFAULT_REGION: "eu-west-1"
+      AWS_REGION: "${AWS_REGION}"
+      AWS_DEFAULT_REGION: "${AWS_DEFAULT_REGION}"
     volumes:
       - ./backend:/code
       - $HOME/.aws/credentials:/root/.aws/credentials:ro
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -16,7 +16,8 @@ services:
       - db
     environment:
       envname: 'dkrcompose'
-      AWS_DEFAULT_REGION: "eu-west-1"
+      AWS_REGION: "${AWS_REGION}"
```
nikpodsh, line 19: We could consider specify the default values if the env variables are not specify. e.g.:
```
AWS_REGION: "${AWS_REGION:-eu-west-1}"
AWS_DEFAULT_REGION: "${AWS_DEFAULT_REGION:-eu-west-1}"
```
Otherwise, we have to provide them all the time from the console

dlpzx:
@nikpodsh yesterday GitHub faced some issues with their APIs and I had some issues commiting your suggestion. But it is there now

nikpodsh (approved):

PR #447, opened 2023-05-10
Module dependencies

nikpodsh:
### Feature or Bugfix
- Refactoring

### Detail
Creation of the mechanism to import dependencies for modules. It should help easier extract common code between modules.
Now, one module can refer to another module in order to import it. 

Deleted `common` folder since it has been replaced replaced.
Added some checks if the module was unintentionally imported + logging


backend/dataall/cdkproxy/stacks/policies/mlstudio.py
```diff
@@ -1,7 +1,7 @@
 from dataall.cdkproxy.stacks.policies.service_policy import ServicePolicy
 
 from dataall.db import permissions
-from dataall.modules.common.sagemaker.cdk.statements import create_sagemaker_statements
+from dataall.modules.sagemaker_base.cdk.statements import create_sagemaker_statements
```
dlpzx, line 5: I think we will discuss this today, but I am more inclined to separate notebooks and ML Studio permissions to follow the least-privilege principle. Even if that means to have some duplicated permissions

backend/dataall/modules/dataset_sharing/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+log = logging.getLogger(__name__)
+
+
+class SharingApiModuleInterface(ModuleInterface):
+    @staticmethod
+    def is_supported(modes: List[ImportMode]) -> bool:
+        return ImportMode.API in modes
+
+    def __init__(self):
+        log.info("API pf dataset sharing has been imported")
```
dlpzx, line 16: small typo: API of dataset sharing has been imported

backend/dataall/modules/loader.py
```diff
@@ -29,49 +31,187 @@ class ModuleInterface(ABC):
...
+    @staticmethod
+    def depends_on() -> List[Type['ModuleInterface']]:
+        """
+        It describes on what modules this ModuleInterface depends on.
+        It will be used to eventually load these module dependencies. Even if a dependency module is not active
+        in the config file.
+
+        The default value is no dependencies
+        """
+        return []
```
dlpzx, line 59: We are changing `is_supported` and `depends_on` from @classmethods to @staticmethods because they do not reference anything about the class itself right? Or am I missing something else?

dlpzx: Maybe we can discuss it today as well, but could we have something more like.... I don't understand why we need to repeat the is_supported in every module
```
class MLStudioApiModuleInterface(ModuleInterface):
    """Implements ModuleInterface for MLStudio GraphQl lambda"""

    def __init__(self):
        import dataall.modules.mlstudio.api
        self.import_mode = ImportMode.API
        log.info("API of sagemaker mlstudio has been imported")

```
And then implement the `is_supported` in the loader?
```
class ModuleInterface(ABC):
    """
    An interface of the module. The implementation should be part of __init__.py of the module
    Contains an API that will be called from core part
    """
    @classmethod
    def is_supported(cls, modes: List[ImportMode]):
        if cls.import_mode in modes:
```

backend/dataall/modules/loader.py
```diff
@@ -29,49 +31,187 @@ class ModuleInterface(ABC):
...
+    expected_load = set(in_config)
+    for module in _all_modules():
+        for dependency in module.depends_on():
+            expected_load.add(dependency.name())
+
+    for module in _all_modules():
+        if module.is_supported(modes) and module.name() not in expected_load:
+            raise ImportError(
+                f"ModuleInterface has not been initialized for module {module.name()}. "
+                "Declare the module in depends_on"
```
dlpzx, line 183: I am not sure that I follow all the logic here. In line 174 we are getting all the modules defined in_config (with active set to True or False) from `_load_modules`, shouldn't be only those that are active? I only see the `inactive` begin used in the `_describe_loading` function

dlpzx (commented):
I have some questions

tests/modules/test_loader.py
```diff
@@ -0,0 +1,145 @@
...
+    assert CModule not in correct_order
+
+
+def test_complex_loading(mocker):
+    patch_loading(mocker, [
+        AModule, BModule, CModule, DModule, EModule, FModule, GModule, IModule, JModule, KModule
+    ], {AModule, CModule, JModule})
+
+    loader.load_modules([ImportMode.API])
+    assert order == [AModule, JModule, BModule, DModule, EModule, GModule, FModule, IModule, KModule]
```
dbalintx, line 137: Note: I first looked at this test file without peeking into the implementation details.

In this test you patch all the modules as if they were found.
In the config, you only have A, C and J.
C is a CDK module, so we ignore it.
A has no dependency, and J also has no dependency.

My question is: why do we expect the rest of the modules besides A and J in "order"? (i.e. expect to load them)

backend/dataall/cdkproxy/stacks/policies/mlstudio.py
nikpodsh: Yes, I also think that a module just for one file is too much :) 
I believe we can merge this and rework it in ML studio or in other PR. But I will add a todo to delete it

backend/dataall/modules/loader.py
nikpodsh: Unfortunately, if we enter the constructor of `ModuleInterface` to set modes we will also load `import dataall.modules.mlstudio.api` and we want ot avoid it if the mode is not supported

tests/modules/test_loader.py
nikpodsh: Good catch! It was copy-paste and the test is in correct. But it works because the check is in correct:
I was wondering how it was missed that `in_config` has not only active modules and found that I am not checking properly. Will fix it!

dbalintx: also, the true question is: where did module H disappear? :)

backend/dataall/modules/loader.py
```diff
@@ -29,49 +31,199 @@ class ModuleInterface(ABC):
...
+        if module.name() in in_config:
+            expected_load.add(module)
+
+    to_add = list(expected_load)
+    while to_add:
+        new_to_add = []
+        while to_add:
+            module = to_add.pop()
+            for dependency in module.depends_on():
+                if dependency not in expected_load:
```
dbalintx, line 185: I believe here we should also check if the dependency (the module) supports the input ImportModes, and only in that case add it to the set/list.

More generic question which came up for me due to this case: is it allowed (and fine) to declare module dependencies with different ImportMode support?

PR #449, opened 2023-05-10
Worksheets modularization

dbalintx:
Modularization of Worksheets

- Moved Worksheet related code to its own new module
- Merged AthenaQueryResult object into Worksheets
- Worksheet related permissions moved to new module
- Removed worksheet sharing related (unused) code from the entire repo


backend/dataall/core/group/services/group_resource_manager.py
dlpzx, line 1: Why did you rename the environment_uri to just environment? I think it makes it clearer to know that we are talking about the unique identifier and not the whole environment object

backend/dataall/modules/datasets/db/dataset_repository.py
```diff
@@ -15,12 +15,12 @@ def get_dataset_by_uri(session, dataset_uri) -> Dataset:
             raise exceptions.ObjectNotFound('Dataset', dataset_uri)
         return dataset
 
-    def count_resources(self, session, environment_uri, group_uri) -> int:
+    def count_resources(self, session, environment, group_uri) -> int:
```
dlpzx, line 18: I see now! Ignore my previous comment on the environment_uri

backend/dataall/modules/worksheets/db/repositories.py
```diff
@@ -0,0 +1,70 @@
...
+            .filter(
+                WorksheetQueryResult.AwsAccountId == environment.AwsAccountId
+            )
+            .count()
+        )
+
+    def find_worksheet_by_uri(session, uri) -> Worksheet:
+        return session.query(Worksheet).get(uri)
+    
+    def find_worksheet_share_by_uri(session, uri) -> WorksheetShare:
```
dlpzx, line 29: the sharing piece is unused right now, we should clean it up

backend/dataall/modules/worksheets/services/worksheet_services.py
```diff
@@ -0,0 +1,220 @@
...
+            group=worksheet.SamlAdminGroupName,
+            resource_uri=uri,
+            resource_type=Worksheet.__name__,
+        )
+        return True
+
+    @staticmethod
+    def run_sql_query(session, username, groups, environmentUri, worksheetUri, sqlQuery):
+
+        ResourcePolicy.check_user_resource_permission(
```
dlpzx, line 175: Shouldn't this be executed as a decorator?

backend/dataall/modules/worksheets/api/schema.py
dlpzx, line 1: this file should be named "types"

backend/dataall/modules/worksheets/db/repositories.py
dbalintx: done, as agreed I removed the worksheet sharing functionality along the whole repo (BE & FE)

backend/dataall/modules/worksheets/services/worksheet_services.py
```diff
@@ -0,0 +1,161 @@
...
+        worksheet = WorksheetRepository.find_worksheet_by_uri(session, uri)
+        if not worksheet:
+            raise exceptions.ObjectNotFound('Worksheet', uri)
+        return worksheet
+
+
+    @staticmethod
+    @has_tenant_permission(MANAGE_WORKSHEETS)
+    def create_worksheet(
+        session, username, groups, uri, data=None, check_perm=None
```
nikpodsh, line 33: If we start using `has_tenant_permission` (instead of `has_tenant_perm`) we can clean up the method parameters. It's not related to modularization, but we could change it if we are still re-working on this part

backend/dataall/modules/worksheets/services/worksheet_services.py
```diff
@@ -0,0 +1,161 @@
...
+    ) -> Worksheet:
+        if not data:
```
nikpodsh, line 35: Also, not related to modularization. 
Here is a strange pattern that we validate params after the creation of the session. We could at least move it to resolvers

backend/dataall/modules/worksheets/services/worksheet_services.py
nikpodsh: Yes, we could use `@has_resource_permission`, but we need to rename the parameter `environmentUri` to `uri`

backend/dataall/modules/worksheets/services/worksheet_services.py
```diff
@@ -0,0 +1,161 @@
...
+        for row in cursor:
+            record = {'cells': []}
+            for col_position, column in enumerate(columns):
+                cell = {}
+                cell['columnName'] = column['columnName']
+                cell['typeName'] = column['typeName']
+                cell['value'] = str(row[col_position])
+                record['cells'].append(cell)
+            rows.append(record)
+        return {
```
nikpodsh, line 148: Very minor: probably, this code should be moved to `AthenaClient`

backend/migrations/versions/bd271a2780b2_init_database.py
```diff
@@ -1275,7 +1263,6 @@ def upgrade():
 
 def downgrade():
     # ### commands auto generated by Alembic - please adjust! ###
-    op.drop_table('worksheet_share')
```
nikpodsh, line 1278: If we are deleting everything related to worksheet shares, should we delete the table as well? 
I mean put this line `op.drop_table('worksheet_share')` into modularization migration script

backend/dataall/modules/worksheets/db/repositories.py
```diff
@@ -0,0 +1,53 @@
...
+    def count_resources(self, session, environment, group_uri) -> int:
+        return (
+            session.query(WorksheetQueryResult)
+            .filter(
+                WorksheetQueryResult.AwsAccountId == environment.AwsAccountId
+            )
+            .count()
+        )
+
+    def find_worksheet_by_uri(session, uri) -> Worksheet:
```
nikpodsh, line 26: `session` will become `self` since the methods are not static

nikpodsh:
We can also move/remove `WorksheetRole` from `db.models.WorksheetRole`

tests/api/test_worksheet.py
```diff
@@ -1,5 +1,5 @@
 import pytest
-from dataall.api.constants import WorksheetRole
+from dataall.modules.worksheets.api.resolvers import WorksheetRole
```
nikpodsh, line 2: Let's move this test under `test.modules.worksheets`. It will allow us to turn on / turn off tests based on the config file (there is no such feature yet, but we are planning to add it :) ). If we don't do it we will face errors during the deployment. 
'+ There are other references e.g. `test_feed` uses `target_type = Worksheet` and if worksheet turned off we face issue

backend/dataall/modules/worksheets/services/worksheet_services.py
dbalintx: good point, I'll do it, for the other functions as well

backend/dataall/modules/worksheets/services/worksheet_services.py
dbalintx: Agree, I'll move it to the resolver

backend/migrations/versions/bd271a2780b2_init_database.py
dbalintx: I moved it

dbalintx:
@dlpzx @nikpodsh  I addressed the review comments, could you please double-check the content and approve/provide feedback? :) Ty!

backend/dataall/modules/worksheets/services/worksheet_services.py
```diff
@@ -0,0 +1,124 @@
...
+    @has_resource_permission(RUN_ATHENA_QUERY)
+    def run_sql_query(session, uri, worksheetUri, sqlQuery):
+        environment = db.api.Environment.get_environment_by_uri(session, uri)
+        worksheet = WorksheetService.get_worksheet_by_uri(session, worksheetUri)
+
+        env_group = db.api.Environment.get_environment_group(
+            session, worksheet.SamlAdminGroupName, environment.environmentUri
+        )
+
+        base_session = SessionHelper.remote_session(accountid=environment.AwsAccountId)
```
nikpodsh, line 113: Minor: seems like that can be moved to `AthenaClient`

tests/api/test_worksheet.py
dbalintx: I created a testhelper for Feed which is not dependent on any object, and added the existing tests to worksheet tests using this testhelper (feed tests only used worksheet as of now as target type)

nikpodsh (approved):

backend/migrations/versions/bd271a2780b2_init_database.py
```diff
@@ -1253,18 +1253,6 @@ def upgrade():
                 sa.Column('created', sa.DateTime(), nullable=True),
                 sa.PrimaryKeyConstraint('AthenaQueryId'),
             )
-            op.create_table(
```
dlpzx, line 1256: We cannot modify pre-existing alembic migration scripts. If we are going to delete a table we need to create a new script deleting it otherwise nothing gets deleted in the pre-existing deployments

dlpzx: In a new deployment, all migration scripts are run in order, so the changes will work. But let's say we have a deployment in V1.3, when we upgrade the CodeBuild dbmigration stage runs `alembic upgrade` (or something similar) and executes the new migration scripts only

dlpzx (commented):
We need to modify the migration scripts before merging

backend/migrations/versions/bd271a2780b2_init_database.py
dbalintx: as we discussed, I removed the unnecessary changes from the former migration scripts, and fixed the changes in the [latest one](https://github.com/awslabs/aws-dataall/pull/449/files#diff-885ec94a722c02b6bbdedc29b9e758a8076b9517ff6d96f856ee38a41fac6634)

dlpzx: Looks good, I will review the comments in mlstudio and see if I have time to test this one

Issue #443, opened 2023-05-03
Encrypted Secrets in secret manager should be rotated atleast annually and encrypted with a KMS key


manjulaK:
### Describe the bug

Encrypted Secrets in secret manager should be rotated atleast annually and encrypted with a KMS key. If the rotation is not an option then consider storing the secret under systems manager parameter store. 

### How to Reproduce

```
*P.S. Please do not attach files as it's considered a security risk. Add code snippets directly in the message body as much as possible.*
```

### Expected behavior

Encrypted Secrets in secret manager should be rotated atleast annually and encrypted with a KMS key. If the rotation is not an option then consider storing the secret under systems manager parameter store.

### OS

All

### Python version

3.1

### AWS data.all version

v1.3,v1.4,v1.5

### Additional context

As per security best practice for secrets management the above need to be met.

dlpzx:
Hi @manjulaK,

I have been looking at this issue, I think you are referring to `dataall-pivot-role-name-<envname>` and `dataall-externalId-<envname>`. They are both encrypted with a different KMS key see `deploy/stacks/secrets_stack.py` 

They are constant values that are referenced in the code and in the environments stack, that's why we do not rotate them.

For the case of `dataall-pivot-role-name-<envname>`, it is more a parameter than a secret, so I will take your suggestion and replace it by an SSM parameter

For `dataall-externalId-<envname>` I am considering different options as it takes advantage of the aws secrets manager `SecretStringGenerator` to get created.

noah-paige:
Closing issue - implemented in v1.6

PR #455, opened 2023-05-15
Move parameters from secrets manager to ssm

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail
Replaced externalId and PivotRoleName secrets in SecretsManager by SSM parameters:
- modified parameter_stack and secrets_stack to create the parameters and remove the secrets.
- the new externalId parameter uses a pre-existing secret if it exists, otherwise it generates a random sequence of numbers and letters
- modified the sts handler to retrieve the pivotRoleName and externalId from SSM

### Relates
- #443 

### DIFF
--- a/backend/dataall/api/Objects/Environment/resolvers.py
+++ b/backend/dataall/api/Objects/Environment/resolvers.py
@@ -731,6 +731,6 @@ def get_pivot_role_name(context: Context, source, organizationUri=None):
         if not pivot_role_name:
             raise exceptions.AWSResourceNotFound(
                 action='GET_PIVOT_ROLE_NAME',
-                message='Pivot role name could not be found on AWS Secretsmanager',
+                message='Pivot role name could not be found on AWS Systems Manager - Parameter Store',
             )
         return pivot_role_name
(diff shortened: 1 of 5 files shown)


dlpzx:
For this feature I have tested 2 scenarios:
- Scenario 1: deployment from scratch -> when using this branch directly to deploy data.all to a clean account, a random externalId is generated and a parameter with this value is stored in SSM Parameter Store.
- Scenario 2: backwards compatibility for pre-existing secret in Secrets Manager -> when upgrading to this branch, the pre-existing secret value is used to assign a value to the new parameter in SSM Parameter Store and then the secret is deleted.

For both I have performed several actions in the deployed UI:

- [ ] - List objects in Catalog --> ensure correct configuration of OpenSearch and Lambda
- [ ] - List objects in any other view --> ensure correct configuration of GraphQL API
- [ ] - Link environment and Create Dataset --> ensure correct configuration of cdkproxy
- [ ] - Crawl and sync tables --> ensure correct configuration of short async tasks (IMPORTANT they use the pivotRole and externalId)
- [ ] - Table sharing --> ensure correct configuration of sharing task (IMPORTANT they use the pivotRole and externalId)

PR #464, opened 2023-05-17
Fixes issue with existing cognito callbacks

gmuslia:
### Feature or Bugfix
- BugFix

### Detail
- In line (https://github.com/awslabs/aws-dataall/blob/13a2fc082694600a0dacaa7e88d0d61ec950d753/deploy/configs/cognito_urls_config.py#L61)

It checks for example.com where instead the right callback to check is ```https://example.com``` and that's why it doesn't get replaced during the configuration phase.

### Relates
- https://github.com/awslabs/aws-dataall/issues/454


dlpzx (approved):
lgtm

PR #466, opened 2023-05-17
465 - Update Aurora default Parameter Group to 'default.aurora-postgresql11'.

rbernotas:
### Feature or Bugfix
- Bugfix

### Detail
- Update Aurora default Parameter Group to 'default.aurora-postgresql11'.  Fixes an issue where the Aurora nested stack deploy in the data.all backend deploy would fail and/or block indefinitely due to 'default.aurora-postgresql10' mismatch with version 11 of the Aurora database engine.

### Relates
- https://github.com/awslabs/aws-dataall/issues/465

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #471, opened 2023-05-23
fix: Upgrade aurora engine version to 11.16

kimengu-david:
### Feature or Bugfix

- Bugfix

### Detail
Update Aurora engine version to 11.16. Fixes an issue where the Aurora nested stack deployment in the data.all backend which goes  to deployment account would fail  as AuroraPostgresEngineVersion.VER_10_18  is not compatible with parameter group default.aurora-postgresql11
### Relates
- https://github.com/awslabs/aws-dataall/pull/466

### DIFF
--- a/deploy/stacks/aurora.py
+++ b/deploy/stacks/aurora.py
@@ -120,7 +120,7 @@ class AuroraServerlessStack(pyNestedClass):
             self,
             f'AuroraDatabase{envname}',
             engine=rds.DatabaseClusterEngine.aurora_postgres(
-                version=rds.AuroraPostgresEngineVersion.VER_10_18
+                version=rds.AuroraPostgresEngineVersion.VER_11_15
             ),
             deletion_protection=True,
             cluster_identifier=f'{resource_prefix}-{envname}-db',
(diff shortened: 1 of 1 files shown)


dlpzx:
Hi @kimengu-david thanks for opening the PR! I tested the upgrade to 11.15 on a pre-existing deployment and it failed with the following:
image

Which is a little strange given that the [documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html#USER_UpgradeDBInstance.PostgreSQL.MajorVersion) states that we can upgrade from 10.18 to 11.15. In any case, I tested upgrading to 11.16 and it worked fine, but I think that we still need to understand this behavior a little better.

I have a deployment of data.all V1.4  deployed a couple of months ago that has never been updated. RDS automatically performed the upgrade from engine 10.18 to 11.16 on May 15th. The same has happened with other deployments that initially set the engine version to 10.18.

In conclusion, my guess is that the PR #466 solved a particular issue in which the upgrade must have gotten stuck. We can either:
- revert the changes in that PR and deploy with engine in major version 10 - if deployments are automatically upgraded to version 11 there must be a good reason, so staying in version 10 does not seem like the best idea
- upgrade to major version 11 (using 11.16) - we need to be careful to not get into backwards compatibility issues as the one in the screenshot

What are your thoughts?

dlpzx (changes requested):
Change to ver_11_16

kimengu-david:
I would also go for the second option. Okay updating that now.

dlpzx (approved):

PR #476, opened 2023-05-24
Updated CDK Version to fix issue with cdkproxy/ dataset stack creations

gmuslia:
### Feature or Bugfix
- BugFix

### Detail
- cdkproxy was using an outdated version of aws-cdk-lib which uses NODEJS_12_X for the AWS Custom Resources Lambda Functions, which are now not anymore supported in the AWS Accounts and causes failure of the creation of CloudFormation stacks in the case when you create a new DataSet Stack
- The version change also triggered a minor type enforcement for the AccountPrincipal AccountId to be explicitly ```string```

### Relates
- https://github.com/awslabs/aws-dataall/issues/475


dlpzx:
I just tested it with 2.61.1 and Environment as well as Dataset deployment work as expected
image

Issue #478, opened 2023-05-24
Feature request: disable profiling results view to secret datasets


dlpzx:
**Is your idea related to a problem? Please describe.**
Since datasets preview is disabled for "secret" datasets it also seems logical to restrict the view of the profiling results.

**Describe the solution you'd like**
If a dataset has a confidentiality=secret, profiling results are not visible

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

PR #482, opened 2023-05-26
feat: Disabling profiling results from "secret" and "official" datasets

dlpzx:
### Feature or Bugfix
- Feature

### Detail
- For datasets that are classified as "Secret", data preview is disabled. In the same way, data profiling results should alse be denied.
image

### Relates
- #478 

### DIFF
--- a/backend/dataall/api/Objects/DatasetProfiling/resolvers.py
+++ b/backend/dataall/api/Objects/DatasetProfiling/resolvers.py
@@ -1,6 +1,7 @@
 import json
 import logging
 
+from .... import db
 from ....api.context import Context
 from ....aws.handlers.service_handlers import Worker
 from ....aws.handlers.sts import SessionHelper
@@ -89,6 +90,21 @@ def get_profiling_run(context: Context, source, profilingRunUri=None):
 
 def get_last_table_profiling_run(context: Context, source, tableUri=None):
     with context.engine.scoped_session() as session:
+        table: models.DatasetTable = db.api.DatasetTable.get_dataset_table_by_uri(
+            session, tableUri
+        )
+        dataset = db.api.Dataset.get_dataset_by_uri(session, table.datasetUri)
+        if (
+                dataset.confidentiality
+                != models.ConfidentialityClassification.Unclassified.value
+        ):
+            ResourcePolicy.check_user_resource_permission(
+                session=session,
+                username=context.username,
+                groups=context.groups,
+                resource_uri=table.tableUri,
+                permission_name=permissions.PREVIEW_DATASET_TABLE,
+            )
         run: models.DatasetProfilingRun = (
             api.DatasetProfilingRun.get_table_last_profiling_run(
                 session=session, tableUri=tableUri
--- a/backend/dataall/api/Objects/DatasetProfiling/resolvers.py
+++ b/backend/dataall/api/Objects/DatasetProfiling/resolvers.py
@@ -79,21 +79,7 @@ def get_dataset_table_profiling_run(context: Context, source, tableUri=None):
     For datasets "Secret" or "Official", only users with PREVIEW_DATASET_TABLE permissions can perform this action.
     """
     with context.engine.scoped_session() as session:
-        table: models.DatasetTable = db.api.DatasetTable.get_dataset_table_by_uri(
-            session, tableUri
-        )
-        dataset = db.api.Dataset.get_dataset_by_uri(session, table.datasetUri)
-        if (
-                dataset.confidentiality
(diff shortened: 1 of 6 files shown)


backend/dataall/api/Objects/DatasetProfiling/resolvers.py
```diff
@@ -89,6 +90,21 @@ def get_profiling_run(context: Context, source, profilingRunUri=None):
 
 def get_last_table_profiling_run(context: Context, source, tableUri=None):
     with context.engine.scoped_session() as session:
+        table: models.DatasetTable = db.api.DatasetTable.get_dataset_table_by_uri(
```
nikpodsh, line 93: I wonder if we should start profiling for secret datasets at all?

If so, we should add more checks in `list_table_profiling_runs` , `get_profiling_run`, `list_profiling_runs`, ( + `queue_profiling_run`, `update_profiling_run_results` also don't have any permission check) 😃

dlpzx: @nikpodsh thanks for reviewing! You are right on reviewing this. I have taken a closer look at the resolvers of DataserProfiling. 
- I deleted those that were not used in the frontend: `update_profiling_run_results`, `get_profiling_run`, `list_profiling_runs`
- Added check permissions in `list_table_profiling_runs` I think you are right, enforcing permission checks is always a good enhacement.
- Renamed `get_profiling_results` --> `resolve_profiling_results` and `resolve_profiling_run_status` to remark that they are part of the Profiling schema and not APIs
- Renamed `get_profiling_results_from_s3` to `_get_profiling_results_from_s3` to indicate that it is an internally used function.
- Some that you mention I was not able to find: `queue_profiling_run`, `update_profiling_run_results`

dlpzx:
@nikpodsh I just tested the 3 scenarios: secret, official and unclassified and it works as expected:
With the following error appearing for secret and official datasets (tables):
image

backend/dataall/api/Objects/DatasetProfiling/resolvers.py
```diff
@@ -141,7 +146,27 @@ def get_profiling_results_from_s3(environment, dataset, table, run):
 
 
 def list_table_profiling_runs(context: Context, source, tableUri=None):
+    """
+    Lists the runs of a profiling job on a Table.
+    For datasets "Unclassified" all users can perform this action.
+    For datasets "Secret" or "Official", only users with PREVIEW_DATASET_TABLE permissions can perform this action.
+    """
     with context.engine.scoped_session() as session:
+        table: models.DatasetTable = db.api.DatasetTable.get_dataset_table_by_uri(
```
nikpodsh, line 155: Let's create a method to avoid duplication:
```
def _check_table_permission_if_needed(table_uri)
```

backend/dataall/api/Objects/DatasetProfiling/resolvers.py
```diff
@@ -141,7 +146,27 @@ def get_profiling_results_from_s3(environment, dataset, table, run):
...
+        dataset = db.api.Dataset.get_dataset_by_uri(session, table.datasetUri)
+        if (
+                dataset.confidentiality
+                != models.ConfidentialityClassification.Unclassified.value
+        ):
+            ResourcePolicy.check_user_resource_permission(
+                session=session,
+                username=context.username,
+                groups=context.groups,
+                resource_uri=table.tableUri,
```
nikpodsh, line 167: Hm, I thought that we set permission on dataset level. For example, if we want to update table in `DatasetTable#update_dataset_table` we pass dataset uri

dlpzx: I am also double checking this as I did not implement the table preview checks. I also think that we need to add some tests to test the "secret" unauthorized datasets.

dlpzx: I can confirm that the permission `PREVIEW_DATASET_TABLE` is applied on tables and is added at the creation of the table as `DATASET_TABLE_READ` 
```
            ResourcePolicy.attach_resource_policy(
                session=session,
                group=group,
                permissions=permissions.DATASET_TABLE_READ,
                resource_uri=table.tableUri,
                resource_type=models.DatasetTable.__name__,
            )
```

I have added several tests in my last commit. I think the PR is ready for a re-review :)

nikpodsh: Thanks for the update and checking it!

nikpodsh (approved):

Issue #461, opened 2023-05-17
Limit Dataset Role permissions


dlpzx:
**Is your idea related to a problem? Please describe.**
After analyzing IAM roles in data.all we noticed that the IAM role created as part of the dataset stack has some permissions that are too open (e.g. lakeformation* on *)

**Describe the solution you'd like**
Review and harden permissions for Dataset role

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

noah-paige:
Closing issue - implemented in v1.6

PR #497, opened 2023-06-06
feat: limit dataset role permissions

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail
The resulting IAM policy can:
- list all buckets
- read and write objects to the dataset Bucket which is encrypted
- read S3 access points in the dataset Bucket
- putLogs in dataset Glue crawler log group
- read dataset Glue database, read and write tables in the dataset Glue database. This is not strictly necessary as in data.all permission to data is handled using Lake Formation. But restricting the IAM-based data permissions we ensure that any Glue resource that is not protected using Lake Formation is not accessible by this role
- WIP - read objects to the `/profiling/code` prefix in the environment bucket
- WIP - read and write objects to the `/profiling/code/results/datasetUri/` prefix in the environment bucket

IMPORTANT: I found a bug related to profiling jobs that prevented me to test the profiling jobs. A separate [issue](https://github.com/awslabs/aws-dataall/issues/506) has been opened for it. For this reason the profiling permissions are a work in progress and might require changes. e.g. additional KMS permissions.

It cannot:
- read or write to any other S3 Bucket
- use any KMS key different from the dataset KMS key
- read or write any other Glue database/tables

In addition, the Glue crawler and the profiling Job of the dataset have been modified to always use the dataset role and not the PivotRole to break down the "super permissions" of the pivot role and distribute responsibilities. As a result, the dataset role can be assumed:
- by the pivotRole -> used whenever users are assuming the role from data.all UI
- by Glue -> to run Glue profiling jobs and Glue crawler

### Relates
- #461 

### DIFF
--- a/backend/dataall/cdkproxy/stacks/dataset.py
+++ b/backend/dataall/cdkproxy/stacks/dataset.py
@@ -247,6 +247,36 @@ class Dataset(Stack):
                     ],
                 ),
                 iam.PolicyStatement(
+                    sid="GlueAccessCrawler",
+                    actions=[
+                        "glue:GetDatabase",
+                        "glue:GetTableVersion",
+                        "glue:CreateTable",
+                        "glue:GetTables",
+                        "glue:GetTableVersions",
+                        "glue:UpdateTable",
+                        "glue:DeleteTableVersion",
+                        "glue:DeleteTable",
+                        "glue:GetTable"
+                    ],
+                    effect=iam.Effect.ALLOW,
+                    resources=[
+                        f"arn:aws:glue:*:{dataset.AwsAccountId}:catalog",
+                        f"arn:aws:glue:{dataset.region}:{dataset.AwsAccountId}:database/{dataset.S3BucketName}",
+                        f"arn:aws:glue:{dataset.region}:{dataset.AwsAccountId}:table/{dataset.S3BucketName}/*"
+                    ]
+                ),
+                iam.PolicyStatement(
+                    sid="LoggingGlueCrawler",
+                    actions=[
+                        'logs:PutLogEvents',
+                    ],
+                    effect=iam.Effect.ALLOW,
+                    resources=[
+                        f'arn:aws:logs:{dataset.region}:{dataset.AwsAccountId}:log-group:/aws-glue/crawlers:log-stream:{dataset.GlueCrawlerName}',
+                    ],
+                ),
+                iam.PolicyStatement(
                     actions=['s3:ListEnvironmentBucket'],
                     resources=[f'arn:aws:s3:::{env.EnvironmentDefaultBucketName}'],
                     effect=iam.Effect.ALLOW,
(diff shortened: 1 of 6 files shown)


backend/dataall/cdkproxy/stacks/dataset.py
```diff
@@ -194,91 +194,114 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
+                        f"arn:aws:glue:{dataset.region}:{dataset.AwsAccountId}:table/{dataset.GlueDatabaseName}/*"
+                    ]
                 ),
                 iam.PolicyStatement(
+                    sid="LoggingGlueCrawler",
                     actions=[
-                        'logs:CreateLogGroup',
-                        'logs:CreateLogStream',
                         'logs:PutLogEvents',
                     ],
```
noah-paige, line 273: Confirming that on new dataset creation and first run of a new dataset crawler - does it not need `logs:CreateLogStream` permissions?

noah-paige (commented):
looks good to me - just the 1 question on logs permissions

backend/dataall/cdkproxy/stacks/dataset.py
dlpzx: Let me check it!

dlpzx: Incredible catch! Thank you @noah-paige! The issue was that previously whenever we run the crawler the pivot role updated the crawler to use the pivotRole as execution role. So when I was testing I removed this behavior after the log group was created, damm

image

dlpzx: I implemented and tested the permission problem in my last commit

dlpzx:
@noah-paige one last comment. I added a clarification on the Glue permissions on the PR description.

```
read dataset Glue database, read and write tables in the dataset Glue database. This is not strictly necessary as in data.all permission to data is handled using Lake Formation. But restricting the IAM-based data permissions we ensure that any Glue resource that is not protected using Lake Formation is not accessible by this role
```
Do you agree on this? or am I going to far?

noah-paige (commented):
Reviewed the revised policies and all looks good. Tested the following:

- [ ] Update an existing Environment

    - [ ] Env Admin and Env Group Roles Updated (not re-created)
- [ ] Import Dataset with Specified KMS Key Id
- [ ] Env Admin S3 Permissions to Env Bucket Contents and Dataset Buckets its team owns ONLY
- [ ] Glue Create Job with proper Tag and resource prefix on Job Name Only (otherwise access denied)
- [ ] Lambda Create Function with proper resource prefix in name Only (otherwise access denied)


Couple of Questions/Notes From Testing:
- Should we add a condition to Lambda Permissions for Team Tag Required (similar to other IAM Policy Statements)?
- In order for Crawler to run against data that is nested in folders within the S3 Bucket I needed `glue:BatchGetPartition` and `glue:BatchCreatePartition` permissions added to the dataset IAM Role
    - Is this handled already in the PR [#497 ]

Issue #462, opened 2023-05-17
Limit Environment Roles permissions


alopezlopez4:
Is your idea related to a problem? Please describe.
After analyzing IAM roles in data.all we noticed that the IAM roles created as part of the environment stack has some permissions that are too open.

Describe the solution you'd like
Review and harden permissions for Environment roles

P.S. Don't attach files. Please, prefer add code snippets directly in the message body.

noah-paige:
Closing issue - implemented in v1.6

PR #515, opened 2023-06-13
feat: limit environment roles permissions + import KMS key in imported datasets

dlpzx:
### Feature or Bugfix
- Feature

### Detail
With this PR, the IAM policies for the environment admin group and for environment invited groups, are generated in the same way. We remove unnecessary duplicities and use a single method to create both.

The resulting role has been hardened and its policies restricted. The following sheet contains the analysis of the 2 types of roles and the measures taken to restrict the policies.

[DataallEnvRolesAnalysis.xlsx](https://github.com/awslabs/aws-dataall/files/11732639/DataallEnvRolesAnalysis.xlsx)

As a summary:
- restricted access to S3 buckets and KMS keys to the Team-owned datasets only
- restricted access to environment s3 bucket, athenaqueries result of the team only
- restricted trust policies to Glue and iam:PassRole to Lambda, Glue, StepFunctions, Databrew and SageMaker
- restricted creation and access to resources of any type: all must be tagged with tag `Team=teamname` and prefixed by the environment specified `resource_prefix`
- removed unused Redshift policies and created new service policies (ssm, secretsmanager, athena) to clearly differenciate policies.

To restrict access to KMS keys, we had to introduce a KMS Alias field in the importing dataset form, allowing customers to import existing KMS keys. The necessary database and graphql fields were already in place, we just made it available as input of the imported dataset. It adds a check to import the KMS resource, in case the KMS key does not exists the stack deployment will fail. In this PR we also modify the frontend views to show the KMS alias in imported datasets.

image

Userguide documentation has been updated:
1) add KMS import information in datasets
2) description of dataset IAM role - usage, policies, data access
3) description of env roles - usage, policies, data access

Finally, the CloudFormation output giving the pivotRole name has been updated to be more specific to the environment and allow multiple environments being deployed in the same account/region (not recommended, but might be needed in some customers)

### Relates
- #462 


dlpzx:
Hi @noah-paige! I am aware of the fields of the dataset model being a bit messy at the moment with the `KmsAlias` column being Alias for created datasets and Key for imported datasets. My intention is to do a real cleanup after modularization so that in V1.6 we do not introduce a lot of changes that need to be considered for v2.0. I have made a note in the modularization clean-up tasks

backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -734,7 +675,7 @@ def create_group_environment_role(self, group):
 
         group_role = iam.Role(
             self,
-            f'{group.environmentIAMRoleName}',
+            id,
```
dlpzx, line 678: If you wonder why I added this --> it is for backwards compatibility. 

Original environment - before this PR
EnvironmentA with Admin = Scientists and InvitedTeam = Engineers. In the environment stack we will create 2 roles `...-environmentA-...` CloudFormation Id ="EnvAdminRole" for the Scientists team and `...-engineers-..` with CloudFormation Id  ...-engineers-..` for Engineers.

dlpzx: Upgraded environment - after this PR
We do not want IAM roles to be renamed, because that would delete roles and create new ones, which will break previous Lake Formation and S3 permissions. But also the CloudFormation resource Id needs to be the same as the original one, otherwise CloudFormation will try to create a new resource and that will cause issues with the old role that gets cleaned up after the CloudFormation deployment

backend/dataall/cdkproxy/stacks/policies/quicksight.py
```diff
@@ -7,6 +7,8 @@ class QuickSight(ServicePolicy):
     def get_statements(self):
         return [
             iam.PolicyStatement(
```
dlpzx, line 9: Quicksight resources cannot be tagged, that's why I added the teamname in the resource string

noah-paige (commented):
Reviewed the revised policies and tested the following:

- [x] Update an existing Environment
     - [x] Env Admin and Env Group Roles Updated (not re-created)
- [x] Import Dataset with Specified KMS Key Id
- [ ] Env Admin S3 Permissions to Env Bucket Contents and Dataset Buckets its team owns ONLY
- [ ] Glue Create Job with proper Tag and resource prefix on Job Name Only (otherwise access denied)
- [ ] Lambda Create Function with proper resource prefix in name Only (otherwise access denied)

Couple of Questions/Notes From Testing:

- Should we add a condition to Lambda Permissions for Team Tag Required (similar to other IAM Policy Statements)?
- In order for Crawler to run against data that is nested in folders within the S3 Bucket I needed `glue:BatchGetPartition` and `glue:BatchCreatePartition` permissions added to the dataset IAM Role
        Is this handled already in the PR [ #497 ]

backend/dataall/cdkproxy/stacks/policies/_lambda.py
```diff
@@ -30,5 +30,18 @@ def get_statements(self):
                     f'arn:aws:lambda:{self.region}:{self.account}:layer:{self.resource_prefix}*:*',
                 ],
```
noah-paige, line 31: Could we add a condition similar to other policies here

```
                conditions={
                    'StringEquals': {
                        f'aws:resourceTag/{self.tag_key}': [self.tag_value]
                    }
                },
```

dlpzx:
Hi @noah-paige, after discussing with @chamcca, he suggested to change the KMS key ID for the Alias. It is true that for rotating keys it works better and we can always get the KMS key id from the alias in the backend. In my last commit I implemented this logic, I will be testing it today

backend/dataall/cdkproxy/stacks/policies/_lambda.py
dlpzx: Yes we can (https://docs.aws.amazon.com/service-authorization/latest/reference/list_awslambda.html#awslambda-policy-keys) let me add it at least to the functions. Good catch

dlpzx: I think we can further enhance the policies. Some only include conditions on `aws:ResourceTag`, but users can modify the tag of their resources once they are created. To ensure that they can only tag resources with their own tag we need to add a `aws:RequestTag` condition in those policies where is not already there

dlpzx:
#### Testing
##### DataPolicy



##### Lambda

##### Athena

##### AWS_CICD

##### CloudFormation

##### DataBrew

##### Glue

##### Quicksight

##### SageMaker

##### Secrets

##### SystemsManager

##### StepFunctions

dlpzx:
## Testing
##### DataPolicy

- [ ] Check that role can see, upload and download data from owned Dataset S3 Bucket - created by data.all
- [ ] Check that role can see database and tables from owned Dataset Glue Database - created by data.all
- [ ] Check that role can see, upload and download data from owned Dataset S3 Bucket - imported
- [ ] Check that role can see database and tables from owned Dataset Glue Database - imported
- [ ] Check that role gets AccessDenied to not owned Buckets
- [ ] Check that role does not see not owned Glue databases and tables in console and cannot access them

##### Lambda
- [x] Create Function only if Lambda function name with proper resource prefix + tagged with team tag (otherwise AccessDenied) 
- [ ] 
##### Athena

##### AWS_CICD

##### CloudFormation

##### DataBrew

##### Glue

##### Quicksight

##### SageMaker

##### Secrets

##### SystemsManager

##### StepFunctions

noah-paige:
For SageMaker Studio Permissions:

- 🟡 I am only able to start the JupyterServer App for my user profile when I edit by Env Role for `sagemaker:CreateApp` on `"arn:aws:sagemaker:REGION:ACCOUNT:app/*"` and remove the Condition `'StringEquals': {f'aws:RequestTag/{self.tag_key}': [self.tag_value]}`

    - Domain Id is populated by AWS and is does not contain `resource_prefix` 
    -  Currently we let SageMaker Create the Jupyter Server App for us when loading the presigned URL rather than calling `create_app()` API Call so no AWS Resource Tags Added to the JupyterServer App (as per my understanding)


For QuickSight Permissions:
- Is the questions regarding whether we should remove most of the QuickSight Permissions from Env Roles since Quicksight has its own permission model to manage QuickSight Users/Groups and we already manage LF Permissions for those QuickSight users if enabled?
  - If so I imagine that would limit the ability of users creating QS Resources programmatically via API calls and we would recommend only doing so via QuickSight Sessions started through the dataall UI, right?

dlpzx:
Hi @noah-paige,
For SageMaker Studio Permissions: I forgot to push my commits  :'( 
For Quicksight: it is not very harmful but to be useful the naming convention needs to be followed by QS users (so at least well documented)

noah-paige (commented):
1. I think we are still missing SQS permissions for the env roles

2. I am running into issues when running Glue Profiling Job now with the following error:
`Unable to verify existence of default database: com.amazonaws.services.glue.model.AccessDeniedException`

Please let me know if you can replicate as well?

3. Lastly, I want to do some additional testing with SageMaker Studio. I am able to create the Domain, User Profile, and App with the permissions in this PR but am currently having trouble launching a Studio Notebook or having any of the `Notebooks and compute resources` Section load - I want to make sure my VPC networking configuration is proper as I imagine this may be causing the issue and not the permissions

dlpzx:
Thanks Noah! :) 
1. I am working on the SQS permissions and I will test them this morning
2. I will retry Glue Profiling Job
3. I will review SageMaker in a clean environment

dlpzx:
### Testing 2 --> Only modified policies

##### Lambda 🟢 
- [ ] Create Function only if Lambda function name with proper resource prefix + tagged with team tag (otherwise AccessDenied) 
- [ ]  Access, manage, update Lambda functions with proper resource prefix + tagged with team tag (otherwise AccessDenied) 
- [ ] CANNOT modify the team tag value of the Lambda function
- [ ] Publish Layers with proper resource prefix (otherwise AccessDenied) 
- [ ] Delete Layers with proper resource prefix (otherwise AccessDenied) 
- [ ] Verify logs are logged in CloudWatch

##### Athena 🟢 
- [X] Assume team workgroup only (otherwise AccessDenied when executing the query)
- [X] Run query on team dataset, only those databases and tables appear on the data window
- [X] Role can access the folder athenaqueries/teamname in the environment bucket but gets access denied to any other subfolder in athenaqueries/
- [X] Cannot modify the workgroup or update the query results location

##### AWS_CICD 🟢

- [ ]  Create CodeCommit repo only if named with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] Access, manage, update repositories with proper resource prefix + tagged with team tag (otherwise AccessDenied) --> note: it takes around 1 minute until permissions work. At first we get AccessDenied
- [ ] CANNOT modify the team tag value of the CodeCommit repo
- [ ] Access, manage, update CodePipeline Pipelines with proper resource prefix + tagged with team tag (otherwise AccessDenied) --> note: it takes around 1 minute until permissions work. At first we get AccessDenied
- [ ] CANNOT modify the team tag value of the CodePipeline Pipeline
- [ ] Create CodeBuild resources only if named with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] Access, manage, update CodeBuild resourcess with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] CANNOT modify the team tag value of the CodeBuild resources

##### CloudFormation 🟢 
I removed update stack permissions, most stacks should be created or modified from data.all or from pipelines. But in the case an additional template is needed or a cleanup is needed I left these 2 permissions
- [ ] Delete stack only if named with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] CANNOT modify the team tag value of the stack


##### DataBrew 🟢 
- [ ] Create dataset only if named with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] Access, manage, update datasets with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] CANNOT modify the team tag value of the dataset


##### Glue 🟢 
- [ ] Create job only if named with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] Access, manage, update jobs with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] CANNOT modify the team tag value of the job
- [ ] Create crawler only if named with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] Access, manage, update crawlers with proper resource prefix + tagged with team tag (otherwise AccessDenied) - > we can get the details of crawlers and jobs but we cannot edit or run other team's crawlers
- [ ] CANNOT modify the team tag value of the crawler
- [ ] Create trigger only if named with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] Access, manage, update triggers with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] CANNOT modify the team tag value of the trigger

##### SageMaker 🟢 ⚠️ 
- [ ] Create notebook only if named with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] Access, manage, update notebooks with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] Start Notebook in Jupyter
- [ ] CANNOT modify the team tag value of the notebook --> it shows an error and then it says success but it does not update the tags

- [ ] Access, manage, update domains tagged with team tag (otherwise AccessDenied) 
- [ ] Access, manage, update users tagged with team tag (otherwise AccessDenied) 


##### Secrets 🟢 
- [ ] Create secret only if named with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] Access, manage, update secrets with proper resource prefix + tagged with team tag (otherwise AccessDenied)
- [ ] CANNOT modify the team tag value of the secret

frontend/src/views/Datasets/DatasetConsoleAccess.js
```diff
@@ -47,16 +47,14 @@ const DatasetConsoleAccess = (props) => {
...
       </CardContent>
-      {!dataset.imported && (
-        <CardContent>
+      <CardContent>
           <Typography color="textSecondary" variant="subtitle2">
             KMS alias
           </Typography>
           <Typography color="textPrimary" variant="body2">
             {`arn:aws:kms:${dataset.region}:${dataset.AwsAccountId}/alias:${dataset.KmsAlias}`}
           </Typography>
```
dlpzx, line 56: I just checked that the dataset.KmsAlias will never be a None value, by default we write `KmsAlias='undefined'`

backend/dataall/cdkproxy/stacks/policies/aws_cicd.py
```diff
@@ -65,18 +96,23 @@ def get_statements(self):
...
+                    's3:ListBucket',
+                    's3:PutBucketPublicAccessBlock',
+                    's3:GetObject',
+                    's3:PutObject',
+                    's3:DeleteObject'
+                ],
                 resources=[
-                    f'arn:aws:codebuild:{self.region}:{self.account}:project/{self.resource_prefix}*',
-                    f'arn:aws:codebuild:{self.region}:{self.account}:report-group/{self.resource_prefix}*',
+                    "arn:aws:s3:::codepipeline-*",
```
chamcca, line 110: who creates these buckets? these resources should probably be scoped down more. do the buckets have account/region identifiers in their names? something to make this more specific?

backend/dataall/cdkproxy/stacks/policies/glue.py
```diff
@@ -64,13 +59,80 @@ def get_statements(self):
...
+            ),
+            iam.PolicyStatement(
+                #sid="GlueCreateS3Bucket",
+                effect=iam.Effect.ALLOW,
+                actions=[
+                    's3:CreateBucket',
+                    's3:ListBucket',
+                    's3:PutBucketPublicAccessBlock'
+                ],
+                resources=["arn:aws:s3:::aws-glue-*"],
```
chamcca, line 113: can we limit this scope anymore?

backend/dataall/cdkproxy/stacks/policies/glue.py
```diff
@@ -64,13 +59,80 @@ def get_statements(self):
...
+            iam.PolicyStatement(
+                #sid="GlueReadWriteS3Bucket",
+                actions=[
+                    's3:GetObject',
+                    's3:PutObject',
+                    's3:DeleteObject'
+                ],
+                effect=iam.Effect.ALLOW,
+                resources=[
+                    'arn:aws:s3:::aws-glue-*/*',
```
chamcca, line 124: can we limit this scope anymore?

backend/dataall/cdkproxy/stacks/policies/glue.py
```diff
@@ -82,36 +144,52 @@ def get_statements(self):
...
                 },
             ),
+            iam.PolicyStatement(
+                #sid="SupportGluePermissions",
+                effect=iam.Effect.ALLOW,
+                actions=[
+                    'glue:*Classifier',
+                    'glue:CreateScript',
+                    'glue:CreateSecurityConfiguration',
+                    'glue:DeleteSecurityConfiguration',
```
chamcca, line 177: can we limit the scope for the create/delete security configuration operations?

chamcca (commented):
comments added on some resource scoping

backend/dataall/cdkproxy/stacks/policies/aws_cicd.py
dlpzx: Yes, we can! Thank you!

backend/dataall/cdkproxy/stacks/policies/sagemaker.py
```diff
@@ -15,89 +36,124 @@ def get_statements(self):
...
-                    f'arn:aws:sagemaker:{self.region}:{self.account}:notebook-instance/*',
-                    f'arn:aws:sagemaker:{self.region}:{self.account}:monitoring-schedule/*',
-                    f'arn:aws:sagemaker:{self.region}:{self.account}:pipeline/*',
-                    f'arn:aws:sagemaker:{self.region}:{self.account}:training-job/*',
-                    f'arn:aws:sagemaker:{self.region}:{self.account}:processing-job/*',
-                    f'arn:aws:sagemaker:{self.region}:{self.account}:hyper-parameter-tuning-job/*',
-                    f'arn:aws:sagemaker:{self.region}:{self.account}:transform-job/*',
-                    f'arn:aws:sagemaker:{self.region}:{self.account}:automl-job/*'
+                    f'arn:aws:sagemaker:{self.region}:{self.account}:domain/*',
+                    f'arn:aws:sagemaker:{self.region}:{self.account}:user-profile/*/*',
```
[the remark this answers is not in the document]
dlpzx: It is scoped down to domains and users tagged with the team tag:
```
conditions={
                    'StringEquals': {
                        f'aws:ResourceTag/{self.tag_key}': [self.tag_value]
                    }
                },
```
When it was first implemented the domain and the users created by data.all backend were not prefixed with any particular prefix (which happens for all other resources). In the case of a SageMaker domain, updating the prefix means re-creating the domain = deleting all users and apps that they have = break whatever they have. That is why I did not add the prefix here, but I am more than open to ideas to break from this legacy issue

backend/dataall/cdkproxy/stacks/policies/glue.py
dlpzx: It is not possible to scope down this action ([docs](https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsglue.html#awsglue-resources-for-iam-policies)). Our alternatives are:
1. Allow only creation, so that we do not delete any other-team's security configuration
2. Do not allow anything and let the AWS Account admins manage the security configurations, we can then use them in our Glue Jobs....

What do you think?

backend/dataall/cdkproxy/stacks/policies/sagemaker.py
```diff
@@ -15,89 +36,124 @@ def get_statements(self):
...
+                effect=iam.Effect.ALLOW,
+                actions=[
+                    'sagemaker:Delete*',
+                    'sagemaker:Update*'
+                ],
+                not_resources=[
+                    f'arn:aws:sagemaker:{self.region}:{self.account}:notebook-instance/{self.resource_prefix}*',
+                    f'arn:aws:sagemaker:{self.region}:{self.account}:domain/*',
+                    f'arn:aws:sagemaker:{self.region}:{self.account}:user-profile/*/*',
+                    f'arn:aws:sagemaker:{self.region}:{self.account}:app/*/*',
```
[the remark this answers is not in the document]
dlpzx: I added a "resource-prefix" condition and then an explicit DENY in the domain and user actions. I think this should solve the issue

backend/dataall/cdkproxy/stacks/policies/glue.py
chamcca: if Creation is mandatory, then we should consider removing the Delete permission. if we can support using configurations created outside of dataall, then that would be safest.

backend/dataall/cdkproxy/stacks/policies/sagemaker.py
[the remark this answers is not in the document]
chamcca: got it. at least it is scoped to domains and users with the team tag. i missed that before

backend/dataall/cdkproxy/stacks/policies/sagemaker.py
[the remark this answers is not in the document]
chamcca: +1

dlpzx:
@noah-paige reaady for final review!

backend/dataall/cdkproxy/stacks/policies/aws_cicd.py
```diff
@@ -2,26 +2,52 @@
...
                     'codecommit:CreateApprovalRuleTemplate',
                     'codecommit:UpdateApprovalRuleTemplateName',
                     'codecommit:GetApprovalRuleTemplate',
-                    'codecommit:ListApprovalRuleTemplates',
                     'codecommit:DeleteApprovalRuleTemplate',
-                    'codecommit:ListRepositories',
                     'codecommit:UpdateApprovalRuleTemplateContent',
                     'codecommit:UpdateApprovalRuleTemplateDescription',
                 ],
                 resources=['*'],
```
nikpodsh, line 26: Shouldn't it be a dataall/resource specific? 
A malicious user can use data.all to compromise all repos in the account.

backend/dataall/cdkproxy/stacks/dataset.py
```diff
@@ -249,19 +251,17 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
                 iam.PolicyStatement(
                     sid="GlueAccessCrawler",
                     actions=[
-                        "glue:GetDatabase",
-                        "glue:GetTableVersion",
+                        "glue:Get*",
```
nikpodsh, line 254: I see that we replaced all get actions with a wildcard get* and I wonder if we really need all new permissions? I'm asking because get* allows us to use many new methods and can be misused + violates least privilege principle 
This question is not only for glue, but also for other service policies in this PR like codecommit

backend/dataall/cdkproxy/stacks/policies/secretsmanager.py
```diff
@@ -0,0 +1,51 @@
...
+    -
+    """
+
+    def get_statements(self):
+        statements = [
+            aws_iam.PolicyStatement(
+                #sid="SecretsReadAll",
+                effect=aws_iam.Effect.ALLOW,
+                actions=["secretsmanager:ListSecrets"],
+                resources=["*"],
```
nikpodsh, line 18: Just to clarify: we set resources * for ListSecrets, listQueues, and etc. because it raises errors is we try to get dataall the secret values, doesn't it?
I am interested if there is a way to get a dataall secret value (when we know the name of secret) without trying to list all secrets?

backend/dataall/cdkproxy/stacks/policies/sagemaker.py
```diff
@@ -15,126 +36,170 @@ def get_statements(self):
...
                     'ecr:GetDownloadUrlForLayer',
-                    'ecr:BatchGetImage'],
-                resources=[
-                    '*'
-                ]
+                    'ecr:BatchGetImage',
+                    'servicecatalog:ListAcceptedPortfolioShares',
+                    'servicecatalog:ListPrincipalsForPortfolio',
+                ],
+                resources=['*']
```
nikpodsh, line 202: Not sure if we should set wildcard here. Especially, I am concerned about `GetAuthorizationToken`

backend/dataall/cdkproxy/stacks/policies/service_policy.py
```diff
@@ -30,88 +36,95 @@ def __init__(
...
-                            'tag:UntagResources',
                             'tag:GetTagValues',
                             'tag:GetTagKeys',
                         ],
                         resources=['*'],
                     ),
                     aws_iam.PolicyStatement(
+                        sid="CreateServiceRole",
+                        actions=[
+                            'iam:CreateRole',
```
nikpodsh, line 91: Can't we create all roles we need and then use assume role for services?

backend/dataall/cdkproxy/stacks/policies/aws_cicd.py
dlpzx: Some actions (usually List actions) cannot be filtered by resource

backend/dataall/cdkproxy/stacks/policies/service_policy.py
dlpzx: This one is arguable. The reason to have it is that some aws services require a service role or work easily with a service role autocreation

backend/dataall/cdkproxy/stacks/policies/aws_cicd.py
nikpodsh: That might be minor, but we could extract all update/delete/add operations like `UpdateApprovalRuleTemplateName`, `UpdateApprovalRuleTemplateContent` and etc from `*`

noah-paige (approved):
My one minor comment...

If we are intentionally not assigning `sid="STATEMENT_ID"` for the service policies (I imagine this is to reduce characters in the attached role policies) - then should we make these lines comments on top of each `PolicyStatement` so it does not appear to be unintentional commented out code, meaning:

Instead of:
```
            iam.PolicyStatement(
                #sid="ListLambda",
                actions=[
                    'lambda:List*',
                    ...
```
Have it as:
```
            # List Lambda Permissions
            iam.PolicyStatement(
                actions=[
                    'lambda:List*',
                    ...
```

Either way the commented out lines help to understand what each Policy Statement is doing in terms of allowing permissions so I think they are useful to keep in some form.

Let me know your thoughts @dlpzx - still approving this PR as the above is a layout preference and non-functional... I think the changes look great!

PR #522, opened 2023-06-16
modularization: Disable and skip module test directories for modules that are inactive

nikpodsh:
Enabled the feature to turn off tests whenever a module if inactive. 
More info: https://docs.pytest.org/en/7.3.x/example/pythoncollection.html#customizing-test-collection

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,27 @@
 import os
 import pytest
 import dataall
+from dataall.core.config import config
 from dataall.modules.loader import load_modules, ImportMode
 
 load_modules(modes=[ImportMode.HANDLERS, ImportMode.API, ImportMode.CDK])
 ENVNAME = os.environ.get('envname', 'pytest')
+
+collect_ignore_glob = []
+
+
+def ignore_module_tests_if_not_active():
+    """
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1,11 +1,28 @@
...
+
+def ignore_module_tests_if_not_active():
+    """
+    Ignores tests of the modules that are turned off.
+    It uses the collect_ignore_glob hook
+    """
+    modules = config.get_property('modules', {})
+    for name, props in modules.items():
+        if not props["active"]:
+            collect_ignore_glob.append(os.path.join('modules', f'{name}*'))
```
dlpzx, line 21: For the very rare case in which a module name is part of another module name, for example, let's say one customer implemented a module called `dataset_custom` and deactivated the original `dataset` module; with the appending of 'modules/dataset*` in the ignore_glob we would ignore tests for dataset_custom right?

nikpodsh: Yes, it should deactivate it. 
I can add a filesystem separator to avoid this case 👍

dlpzx (approved):
lgtm, thank you for the tiny change

PR #523, opened 2023-06-16
feat: Limiting read-only access to root file systems in ECS

dbalintx:
Limiting read-only access for ECS tasks deployed by data.all, reasoning detailed in: https://github.com/awslabs/aws-dataall/issues/426

Out of the 7 ECS tasks that gets deployed, only CDKProxy performs multiple write operations to the root filesystem.
The workaround is to mount [bind volumes](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/bind-mounts.html) to the proper paths in the filesystem:

- **/dataall:** required for cdk deploy write operations (cdk.out, cdk.context.json) and further file write operations invoked through dataa.all business logic like archiving objects for the Glue profiling job
- **/tmp:** required since by upon importing aws_cdk libraries a write operation happens to the /tmp folder

Since the [currently used CDK class](https://docs.aws.amazon.com/cdk/api/v1/python/aws_cdk.aws_ecs/FargateTaskDefinition.html) for the Fargate task definition doesn't allow the definition of mount points, I had to replace the it with the [CFN-style class.](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_ecs/CfnTaskDefinition.html)

**[Testing]**
I've created 2 environments and a dataset, and performed the sharing of the dataset between the 2 environments.
I've verified, that:

- the newly created CDKProxy task definition has the same attributes as the old one (with the further addition of the ReadOnlyRootFileSystem=True flag and the 2 new bind volumes)
- the other 6 task definitions have ReadOnlyRootFileSystem=True enabled
- all 7 tasks were executed without failure with the new setting
- the security alert in security hub got archived

### DIFF
--- a/deploy/stacks/container.py
+++ b/deploy/stacks/container.py
@@ -49,49 +49,88 @@ class ContainerStack(pyNestedClass):
         self.task_role = self.create_task_role(envname, resource_prefix, pivot_role_name)
         self.cicd_stacks_updater_role = self.create_cicd_stacks_updater_role(envname, resource_prefix, tooling_account_id)
 
-        cdkproxy_task_definition = ecs.FargateTaskDefinition(
-            self,
-            f'{resource_prefix}-{envname}-cdkproxy',
-            cpu=1024,
-            memory_limit_mib=2048,
-            task_role=self.task_role,
-            execution_role=self.task_role,
-            family=f'{resource_prefix}-{envname}-cdkproxy',
+        cdkproxy_container_name = f'container'
+        cdkproxy_log_group = self.create_log_group(
+            envname, resource_prefix, log_group_name='cdkproxy'
+        )
+        cdkproxy_image = ecs.ContainerImage.from_ecr_repository(
+            repository=ecr_repository,
+            tag=cdkproxy_image_tag
         )
 
-        cdkproxy_container = cdkproxy_task_definition.add_container(
-            f'ShareManagementTaskContainer{envname}',
-            container_name=f'container',
-            image=ecs.ContainerImage.from_ecr_repository(
-                repository=ecr_repository, tag=cdkproxy_image_tag
-            ),
-            environment={
-                'AWS_REGION': self.region,
-                'envname': envname,
-                'LOGLEVEL': 'DEBUG',
-            },
-            command=['python3.8', '-m', 'dataall.tasks.cdkproxy'],
-            logging=ecs.LogDriver.aws_logs(
-                stream_prefix='task',
-                log_group=self.create_log_group(
-                    envname, resource_prefix, log_group_name='cdkproxy'
+        cdkproxy_task_definition = ecs.CfnTaskDefinition(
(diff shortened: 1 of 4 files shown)


dlpzx (approved):
lgtm, the 7 tasks (2 on-demand and 5 scheduled) have the parameter readonly_root_filesystem set to true.
How have you tested it?

deploy/stacks/container.py
```diff
@@ -49,48 +49,88 @@ def __init__(
         self.task_role = self.create_task_role(envname, resource_prefix, pivot_role_name)
         self.cicd_stacks_updater_role = self.create_cicd_stacks_updater_role(envname, resource_prefix, tooling_account_id)
 
-        cdkproxy_task_definition = ecs.FargateTaskDefinition(
+        cdkproxy_container_name = f'container'
+        cdkproxy_log_group = self.create_log_group(
```
nikpodsh, line 53: The logging has not been added to the task definition

nikpodsh: Ok, I see that we created `LogConfigurationProperty` . Then we should removed this log group as it's not used

srinivasreddych:
LGTM from docker bind mounts perspective

deploy/stacks/container.py
dbalintx: The `LogConfigurationProperty` in its current form doesn't create the log group itself, it is just a parametrization, which tells which already existing log group should be used. The creation of the log group itself is happening here via `self.create_log_group`.

I checked it, and according to [this documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_awslogs.html), you could specify within the options of the `LogConfigurationProperty` the following: "awslogs-create-group": "true"
There are 2 problems with this:
1. I tried deploying it, and it doesn't work. No errors, it's visible in the JSON description of the ECS task definition, but the log group isn't created within Cloudwatch.
2. Even if it would work, you can't specify the 2 further attributes which were defined [here](https://github.com/awslabs/aws-dataall/blob/main/deploy/stacks/container.py#L479) via the already existing method.

Because of this, I'd keep the log group creation as it is done currently.

nikpodsh: Got it, thanks for the clarification! I was confused when I found that `cdkproxy_log_group` created, but it wasn't used. I thought that we should have deleted as it seemed like a leftover to me.

Then I have a minor suggestion which we could implement to avoid the confusion. I see that `LogConfigurationProperty` and `LogGroup` are implicitly bound by log_group_name which is the same.  We could make it more explicit by:
```
log_configuration=ecs.CfnTaskDefinition.LogConfigurationProperty(
                    log_driver="awslogs",
                    options={
                        "awslogs-group": cdkproxy_log_group.log_group_name

```
It should emphasize that there is a connection between those two. 
What do you think? :)

nikpodsh (approved):

Issue #512, opened 2023-06-12
Limit KMS keys policies


dlpzx:
**Is your idea related to a problem? Please describe.**
As an additional layer of security besides restrictive IAM policies, I would like more specific KMS policies that restrict access to the KMS keys to the minimum.

**Describe the solution you'd like**
Define a KMS policy in CDK that restricts access to the different KMS keys to the IAM roles that use them in each case. Here is a list of keys that need an additional policy:
-TBD
-TBD

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

noah-paige:
Leaving open until v1m6m0 release and merge to main

noah-paige:
Implemented in v1.6

PR #524, opened 2023-06-16
feat: Restrict kms keys

noah-paige:
### Feature or Bugfix

- Feature

### Detail
- Restricting KMS Key Policies

This PR restricts the KMS Keys created in Data.all Environments (i.e. linked AWS Accounts), including 
- Environment KMS Keys (Used for subscriptions and Custom Resources)
- Dataset KMS Keys (To Encrypt Data in S3)
- Notebook KMS Keys (To Encrypt SageMaker Notebook Instance Storage Volumes)
- SageMaker Studio KMS Keys (To Encrypt SageMaker Studio Storage Volumes)

Additional as part of this PR:
- Consolidate Environment Keys 
  - Reduce 4 KMS Keys to 1 Key to Encrypt all SQS DLQs used for Lambda Custom Resource Failed Invocations
  - Reduce 3 KMS Keys to 1 Key to Encrypt SNS and SQS Resources related to Environment Subscriptions

- Remove Unused Custom Resources
  - Remove Glue DB No Delete Custom Resource (this is no longer used and is replaced by Glue DB Custom Resource which additionally handles deletes of Glue DBs)
  - Remove Data Location CustomResource (this is no longer used and is replaced by CfnResource used to register S3 location in LakeFormation)

- Remove KMS Key Permissions in Environment Roles where applicable
  - Since we are restricting KMS Key Permissions we do not need to explicitly give KMS Permissions to IAM Roles for Env Group Roles where the dataset is created by data.all (if the dataset is imported we still add KMS Permissions to the IAM Role)

### Relates
- [#512 ]


noah-paige:
Tested The Following:
Environment
- [x] Create a New Environment with restricted KMS Key Policy
    - [x] Failed Lambda Custom Resource Sends Message to Encrypted DLQ
    - [x] Able to poll message from encrypted SQS Queue with Env Role (only if sqs permissions added too to the role)


Create New Dataset
- [x] Create a New Dataset with restricted KMS Key Policy
    - [x] Access S3 Bucket and Query Data as Dataset IAM Role 
    - [x] Access S3 Bucket and Query Data as Env Group IAM Role
    - [x] Dataset Registered in LF
    - [x] Run Data Profiling Job
    - [x] Run Data Upload and Crawler + Table Sync

Folder Sharing
- [x] Sharing Dataset Folder Same Account
- [x] Sharing Dataset Folder Different Account

SM Studio
- [x] Enable ML Studio on Environment and Create User Profile with restricted KMS Key Policy
    - [ ] Can Access S3 Data Owned by User Profile Team

Notebook
- [x] Create SageMaker Notebook with restricted KMS Key Policy
    - [x] Can Access S3 Data Owned by Notebook Creator Team
    - [x] Access Denied to Other S3 Buckets

Pipelines:
- [x] Created Gitflow Code Pipeline with restricted Build Project KMS Policy
    - [x] Pipeline Exeuction Successful
- [x] Created Trunk Based CodePipeline with restricted Build Project KMS Policy
    - [x] Pipeline Exeuction Successful

backend/dataall/aws/handlers/sts.py
```diff
@@ -168,6 +168,17 @@ def get_cdk_look_up_role_arn(cls, accountid, region):
...
 
+    @classmethod
+    def get_cdk_exec_role_arn(cls, accountid, region):
+        """Returns the name that will be assumed to perform IAM actions on a given AWS accountid using CDK Toolkit role
+        Args:
+            accountid(string) : aws account id
+        Returns:
+                string : arn of the CDKToolkit role on the target aws account id
+        """
+        log.info(f"Getting CDK exec role: arn:aws:iam::{accountid}:role/cdk-hnb659fds-lookup-role-{accountid}-{region}")
```
dlpzx, line 179: Very minor - change the log.info to make sure we know that we get the exec role

backend/dataall/cdkproxy/stacks/policies/data_policy.py
```diff
@@ -65,6 +65,8 @@ def get_statements(self):
                     "s3:ListAllMyBuckets",
                     "s3:ListAccessPoints",
                     "s3:GetBucketLocation",
+                    'kms:ListAliases',
+                    'kms:ListKeys',
```
dlpzx, line 69: nice

backend/dataall/cdkproxy/stacks/dataset.py
```diff
@@ -360,45 +419,6 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
-                    ),
-                    'LocationUri': f's3://{dataset.S3BucketName}/',
-                    'Name': f'{dataset.GlueDatabaseName}',
-                    'CreateTableDefaultPermissions': [],
-                },
-                'DatabaseAdministrators': dataset_admins,
-            },
-        )
-
         # Get the Provider service token from SSM, the Lambda and Provider are created as part of the environment stack
```
dlpzx, line 422: very minor, remove imports from the import section

backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -775,22 +718,63 @@ def zip_code(assetspath, s3_key='profiler'):
...
+                        "kms:List*",
+                        "kms:GetKeyPolicy",
+                    ],
+                    effect=iam.Effect.ALLOW,
+                    principals=[
+                        self.default_role,
+                    ] + group_roles,
+                    resources=["*"],
+                )
+            ]
```
dlpzx, line 754: Just for clarification. Here we are creating the key for the custom resources DLQ. Since this is an SQS queue, only via SQS the messages should get encrypted and decrypted. The CRs use the pivot role right? so it needs to use the key, but in addition we add the groups to the policy so that they can check the key

backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -414,84 +379,61 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
+                statements=[
+                    iam.PolicyStatement(
+                        actions=[
+                            "kms:Encrypt",
+                            "kms:Decrypt",
+                            "kms:ReEncrypt*",
+                            "kms:GenerateDataKey*",
+                        ],
+                        effect=iam.Effect.ALLOW,
+                        principals=[self.default_role] + group_roles,
```
dlpzx, line 395: At the moment subscriptions are a bit unused. It is just an SNS topic created for users to use. But, do you think it could happen that we send sharing or dataset update notifications from data.all backend? In that case I think the pivotRole will be added or not?

backend/dataall/cdkproxy/stacks/notebook.py
```diff
@@ -62,26 +63,47 @@ def __init__(self, scope, id: str, target_uri: str = None, **kwargs) -> None:
 
         env_group = self.get_env_group(notebook)
 
+        cdk_exec_role = SessionHelper.get_cdk_exec_role_arn(notebook.AWSAccountId, notebook.region)
+
```
dlpzx, line 67: Couldn't we use environment.CDKRoleArn and avoid this function?

backend/dataall/cdkproxy/stacks/sagemakerstudio.py
```diff
@@ -146,14 +146,41 @@ def create_sagemaker_domain_resources(self, sagemaker_principals):
...
+                            "kms:Encrypt",
+                            "kms:Decrypt",
+                            "kms:ReEncrypt*",
+                            "kms:GenerateDataKey*",
+                            "kms:CreateGrant"
+                        ],
+                        effect=iam.Effect.ALLOW,
+                        principals=[
+                            sagemaker_domain_role,
+                            iam.ArnPrincipal(self.environment.CDKRoleArn)
```
dlpzx, line 166: Do we need the cdk role here or just as keyadmin?

backend/dataall/tasks/data_sharing/share_managers/s3_share_manager.py
```diff
@@ -349,7 +349,7 @@ def delete_target_role_access_policy(
         )
         access_point_name = S3ShareManager.build_access_point_name(share)
         existing_policy = IAM.get_role_policy(
-            dataset.AwsAccountId,
+            target_environment.AwsAccountId,
```
dlpzx, line 352: good catch!

dlpzx (commented):
I am testing it locally and when I try to update an existing environment I get the following in the cdkproxy logs
image

backend/dataall/aws/handlers/sts.py
noah-paige: good catch

backend/dataall/cdkproxy/stacks/dataset.py
noah-paige: will clean up

backend/dataall/cdkproxy/stacks/dataset.py
```diff
@@ -346,7 +405,7 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
                 type='AWS::LakeFormation::Resource',
                 properties={
                     'ResourceArn': f'arn:aws:s3:::{dataset.S3BucketName}',
-                    'RoleArn': f'arn:aws:iam::{env.AwsAccountId}:role/{self.pivot_role_name}',
+                    'RoleArn': dataset.IAMDatasetAdminRoleArn,
```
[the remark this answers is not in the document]
noah-paige: I did a few tests registering with the dataset admin role and all worked fine... to be honest I am a little surprised now that that was the case since the requirements in the [docs](https://docs.aws.amazon.com/lake-formation/latest/dg/registration-role.html) for the user-defined register role are:
- S3 Permissions to S3 Location - YES
- KMS Permissions - YES
- Trust Relationship to both Glue and LF - NO (only with Glue for dataset admin role)

I see your point this may have unintentional implications and can be seen as out of scope for this PR - will revert back to pivotRole for registering. At a later time we should look into removing PivotRole permissions for all datasets and instead use dataset roles to do the heavy lifting

backend/dataall/cdkproxy/stacks/environment.py
noah-paige: Yes that is exactly correct - otherwise the `onFailure` messages sent from Lambda to SQS (via the PivotRole) would not be able to be polled off the queue by the environment user. Meaning if a envUser A creates a dataset and the CR Fails to add LF Permissions and the stack fails - envUserA is able to debug by polling message off of the Queue. This is restricted to Encrypting/Decrypting only via SQS!

Also I gave the envUsers the ability to Describe the Key and only GET the Key Policy so these users can understand the permissions they have for this key

backend/dataall/cdkproxy/stacks/environment.py
noah-paige: Yes it is just an SQS Queue subscribing to an SNS Topic - it depends on what user/role would be sending the messages to the SQS Queue

I would opt to leave this as is and work on subscriptions in a separate PR adding the Pivot Role as an additional key user there if needed. At this current state, I think we are good to leave it as is to keep it as least permissive as possible but what do you think?

backend/dataall/cdkproxy/stacks/notebook.py
noah-paige: Yes we could also do it that way - currently notebooks stack has no helper function to get the environment, only to get the env group. So either it would be adding that helper function in the Notebook Stack to get the environment and the `environment.CDKRoleArn` property or using the `get_cdk_exec_role_arn()` defined above

This is the only use of `get_cdk_exec_role_arn()` but in future it could be used more if needed - similar to how we have `get_cdk_look_up_role_arn()`. The value is rather static where the IAM role ARN would only change if we had custom synthesizers in CDK changing the qualifier - which I see little risk for. What are your thoughts?

backend/dataall/cdkproxy/stacks/sagemakerstudio.py
noah-paige: The `CreateDomain` Operation requires the creator to have the following permissions on the CMK Key:
- kms:CreateGrant
- kms:Decrypt
- kms:DescribeKey
- kms:GenerateDataKeyWithoutPlainText

Similarly `kms:CreateGrant` is needed along with Encrypt/Decrypt to start a number of ML related Jobs in Studio if specifying the KMS Key

backend/dataall/cdkproxy/stacks/notebook.py
dlpzx: 2 things. 
- Independently of how we get the role arn, we need to be consistent across modules. Either we use the environment property or the method but I would not have a mix of both.
- I am more inclined to use a method exactly because of what you pointed out, custom qualifiers in CDK. Let's say that one customer wants to implement something like this. It would be relatively easy to add an SSM parameter with the qualifier and then get it in the `get_cdk_exec_role_arn()` method. Plus, we can get rid of an unnecessary column in RDS

backend/dataall/cdkproxy/stacks/environment.py
```diff
@@ -227,7 +228,7 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
         )
 
         # Create or import team IAM roles
-        self.default_role = self.create_or_import_environment_admin_group_role()
```
dlpzx: good catch!

dlpzx (approved):
I will do a final testing as well, but it looks good

PR #529, opened 2023-06-21
Enhance security: Cloudfront distributions should use security response headers

nikpodsh:
### Feature or Bugfix

- Feature

### Detail
To avoid the following attacks: cross site scripting, click-hijacking, + also MIME type sniffing,
the security response headers have been added to all cloudfront distribution.

Since cloudfront's `CloudFrontWebDistribution` is not supporting response policies (it can only be done by escape hatches), all distribution have been migrated to `Distribution` interface. Along with that, the legacy caching was migrated to `CACHE_OPTIMIZED` which is recommended for S3 origins.
The headers in the documentation distribution:


nikpodsh:
The headers in the documentation app
Headers for documentation
The headers in the data.all app
Headers for data all app
The policy is associated with the distributions:
The policy is associated

dlpzx:
Hi thanks @nikpodsh!!! I have checked the PR and I deployed it in an existing deployment (merging your branch in my deployment branch)The headers include the secure headers policy controls 👍 

The only issue that I saw is that the PR creates a new distribution instead of updating the existing one. This means that the domain name that customers were using would be invalid with this PR and their administrators will need to communicate end-users the new domain name (in the case they are using a CloudFront generated domain). I checked that in the new domain the connection with the Cognito user pool works and that all resources that were previously created still work, so it is just a communication challenge.
image

In the deployment, instead of using CloudFront ugly-domain names, customers can use their own custom domains by specifying this option in the `cdk.json` file

nikpodsh:
Changed the target branch to v1.6.0. Three are a few commits that have been merged to main but we don't have them in v1.6.0. Since [awslabs:v1m6m0](https://github.com/awslabs/aws-dataall/tree/v1m6m0) is protected and can't be rebased to a fresh main, I will leave them in this PR. There should not be any collision then we will try to merge v1.6 -> main

dlpzx (approved):
I see that it also includes the latest changes in main, good :)

PR #532, opened 2023-06-23
Sanitize the string to avoid a connection string injection

nikpodsh:
### Feature or Bugfix
Feature

### Detail
We should avoid possible connection injection. To do so, we have to sanitize and validate inputs for the database

### DIFF
--- a/backend/dataall/db/connection.py
+++ b/backend/dataall/db/connection.py
@@ -29,19 +29,19 @@ class Engine:
             dbconfig.url,
             echo=False,
             pool_size=1,
-            connect_args={'options': f"-csearch_path={dbconfig.params['schema']}"},
+            connect_args={'options': f"-csearch_path={dbconfig.schema}"},
         )
         try:
             if not self.engine.dialect.has_schema(
-                self.engine, dbconfig.params['schema']
+                self.engine, dbconfig.schema
             ):
                 log.info(
-                    f"Schema not found - init the schema {dbconfig.params['schema']}"
+                    f"Schema not found - init the schema {dbconfig.schema}"
                 )
                 self.engine.execute(
-                    sqlalchemy.schema.CreateSchema(dbconfig.params['schema'])
+                    sqlalchemy.schema.CreateSchema(dbconfig.schema)
                 )
-            log.info('-- Using schema: %s --', dbconfig.params['schema'])
+            log.info('-- Using schema: %s --', dbconfig.schema)
         except Exception as e:
             log.error(f'Could not create schema: {e}')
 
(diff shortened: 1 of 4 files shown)


backend/dataall/db/connection.py
```diff
@@ -124,10 +124,12 @@ def get_engine(envname=ENVNAME):
...
         pwd = creds['password']
+        host = param_store.get_parameter(env=envname, path='aurora/hostname')
+        database = param_store.get_parameter(env=envname, path='aurora/db')
+
         db_params = {
-            'host': param_store.get_parameter(env=envname, path='aurora/hostname'),
-            'port': param_store.get_parameter(env=envname, path='aurora/port'),
-            'db': param_store.get_parameter(env=envname, path='aurora/db'),
+            'host': host,
+            'db': database,
```
dlpzx, line 132: Hello @nikpodsh I see that you removed the port. Do we need it to make a connection or was it unnecessary?

nikpodsh: Yeah, I double checked and it turned out that the port was used. Thanks for bringing this up!
We don't use the port while creating the database connection in the backend. So it should always roll back to the default value 5432. 
I can re-add it and this time make the port be configurable on the backend or delete the usages of the port (two in the local env and one `quicksight#create_analysis` ). What do you think would be better ?

backend/dataall/db/dbconfig.py
```diff
@@ -1,23 +1,55 @@
...
+        lines.append(' '.join(['|', "user".ljust(10), '|', self.user.ljust(65), '|']))
+        lines.append(' '.join(['|', "pwd".ljust(10), '|', "*****".ljust(65), '|']))
 
         hr = ' '.join(['+', ''.ljust(10, '-'), '+', ''.ljust(65, '-'), '+'])
         lines.append(hr)
         return '\n'.join(lines)
+
+    @staticmethod
+    def _sanitize(regex, string: str) -> str:
+        return re.sub(regex, "", string)
```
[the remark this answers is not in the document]
nikpodsh: Thanks! I will raise an error if the string is not the same after sanitizing.

backend/dataall/db/connection.py
dlpzx: I don't think adding the port as a configuration adds any value for the user, just more "work" for them. When you say deleting the port you mean hardcoding it in local and quicksight?

nikpodsh: Yes, just hardcoding. By deleting I meant the deleting a configurable parameter (which is configurable only in a few places)

dlpzx (approved):
lgtm

backend/dataall/db/connection.py
dlpzx: Looks good

Issue #487, opened 2023-05-31
Restrict NACL rules for backend VPC


dlpzx:
Current VPC deployed by data.all defines inbound and outbound rules open to all traffic. 
What we would like is to have the rules defined as restrictive as possible, allowing only the needed traffic.

dlpzx:
After testing the solution, I was able to confirm that data.all works correctly with the following NACL rules:
![image](https://github.com/awslabs/aws-dataall/assets/71252798/af66cd52-6194-4cdf-b00e-a84cea0135f1)

![image](https://github.com/awslabs/aws-dataall/assets/71252798/60332fb7-7e9e-4fd6-9488-9aff9400fb10)

In addition we need to modify STS boto3 calls because for global VPC endpoints without the region it is not possible to resolve it (related to  https://github.com/awslabs/aws-dataall/issues/144)
```
            sts = base_session.client(
                'sts',
                config=Config(user_agent_extra=f'{__pkg_name__}/{__version__}'),
                region_name=region,
                endpoint_url=f"https://sts.{region}.amazonaws.com"
            )
```

noah-paige:
Implemented as part of v1.6

PR #543, opened 2023-06-30
feat: fine-grained nacls for backend vpc creation 

noah-paige:
### Feature or Bugfix
- Feature

### Detail
- Replace NACL Rules on VPC Subnet with individual security groups defined for the backend VPC as restrictive as possible, allowing only the needed traffic.

### Relates
- #487 


noah-paige:
I have tested the following for the security groups defined in this PR:

ORGANIZATIONS
- [x]  Create Organization
- [x]  Edit Organization
- [x]  Link Environment
- [x]  Invite Team


ENVIRONMENTS
- [x]  Create Environment
- [x]  Edit Environment
- [x]  Invite Team to Environment
- [x]  All Play Features Enabled


ENVIRONMENT PLAY FEATURES
- [x]  New SageMaker Studio User Profile Created
- [x]  New SageMaker Notebook Created
- [x]  New Pipeline (CDK Type) Created
- [x]  New Pipeline (CodePipeline Type) Created 
- [x]  Worksheet Query Run


DATASETS
- [x]  Create Dataset
- [x]  Edit Dataset
- [x]  Upload File
- [x]  Crawler + Sync Tables
- [x]  Glue Profile Job


SHARING
- [x] Folder Share
- [x]  Dataset Table Share

noah-paige:
I think the remaining item to test is a deployment using OpenSearch Serverless to ensure the network configuration set up in this PR also works for that type of data.all deployment (the above testing scenarios is using a regular OpenSearch Domain Instance by default)

noah-paige:
Also, in this PR we introduce an additional 6 VPC Endpoints, having a total of 26 endpoints now deployed for the solution). If we make Backend VPC Security Group Restrictions a configurable option in the `cdk.json` we can explore removing most (if not all) of these endpoints if configured and have the private subnets in the VPC be configured with internet connectivity via NAT Gateway

This is an additional item we could add to the PR and have to test all of the scenarios but first wanted to hear your all thoughts on the above idea/approach?

backend/dataall/api/Objects/Dataset/resolvers.py
```diff
@@ -15,6 +15,7 @@
 from ....aws.handlers.sts import SessionHelper
 from ....aws.handlers.sns import Sns
 from ....aws.handlers.quicksight import Quicksight
+from ....aws.handlers.parameter_store import ParameterStoreManager
```
dlpzx, line 18: This is not needed, but I maybe it is part of unrelated commits

backend/dataall/api/Objects/DatasetTable/resolvers.py
```diff
@@ -12,7 +12,7 @@
 from ....db import permissions, models
 from ....db.api import ResourcePolicy, Glossary
 from ....searchproxy import indexers
-from ....utils import json_utils
+from ....utils import json_utils, sql_utils
```
dlpzx, line 15: I see that this feature branch has been created from main instead than from v1.6 branch. maybe we get some issues when merging back to main

deploy/stacks/container.py
```diff
@@ -309,6 +301,80 @@ def __init__(
             subscriptions_task.task_definition,
         ]
 
+    def create_ecs_security_groups(self, envname, resource_prefix, vpc, vpce_connection, s3_prefix_list, lambdas):
+        scheduled_tasks_sg = ec2.SecurityGroup(
+            self,
+            f'ScheduledTasksSG{envname}',
```
dlpzx, line 307: It is for scheduled tasks and for cdkproxy tasks right?

deploy/stacks/codeartifact.py
```diff
@@ -36,6 +36,27 @@ def __init__(
...
+                            "codeartifact:ListPackageVersions",
+                            "codeartifact:ListPackages",
+                            "codeartifact:ReadFromRepository"
+                        ],
+                        principals=principals,
+                        resources=['*'],
+                    )
+                ]
+            )
+
```
dlpzx, line 59: Just for my understanding. This block is needed for the cross-account permissions needed for installing packages using this repo from the DBMigration CodeBuild stage in the infra account right?

backend/dataall/cdkproxy/stacks/dataset.py
```diff
@@ -98,8 +98,7 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
 
         quicksight_default_group_arn = None
         if env.dashboardsEnabled:
-            quicksight_default_group = Quicksight.create_quicksight_group(AwsAccountId=env.AwsAccountId)
-            quicksight_default_group_arn = quicksight_default_group['Group']['Arn']
+            quicksight_default_group_arn = f"arn:aws:quicksight:{dataset.region}:{dataset.AwsAccountId}:group/default/{Quicksight._DEFAULT_GROUP_NAME}"
```
dlpzx, line 101: We cannot assume that the QS default group is already created, if dashboards was just enabled in the environment the QS group won't exist

deploy/stacks/pipeline.py
```diff
@@ -634,13 +686,13 @@ def set_stacks_updater_stage(
                     'aws sts get-caller-identity --profile buildprofile',
                     f"export cluster_name=$(aws ssm get-parameter --name /dataall/{target_env['envname']}/ecs/cluster/name --profile buildprofile --output text --query 'Parameter.Value')",
                     f"export private_subnets=$(aws ssm get-parameter --name /dataall/{target_env['envname']}/ecs/private_subnets --profile buildprofile --output text --query 'Parameter.Value')",
-                    f"export security_groups=$(aws ssm get-parameter --name /dataall/{target_env['envname']}/ecs/security_groups --profile buildprofile --output text --query 'Parameter.Value')",
+                    f"export security_groups=$(aws ssm get-parameter --name /dataall/{target_env['envname']}/ecs/cdkproxy_security_groups --profile buildprofile --output text --query 'Parameter.Value')",
```
dlpzx, line 689: I think this name does not exist in container.py we define `f'/dataall/{envname}/ecs/security_groups',`

deploy/stacks/vpc.py
```diff
@@ -108,7 +112,7 @@ def __init__(
                 description=f'{resource_prefix}-{envname}-cidrBlock',
             )
 
-    def create_new_vpc(self, cidr, envname, resource_prefix, restricted_nacl):
+    def create_new_vpc(self, cidr, envname, resource_prefix, restricted_nacl, backend_vpc, tooling_region):
```
dlpzx, line 115: why do we need those 2 additional parameters?

deploy/stacks/vpc.py
```diff
@@ -23,26 +24,29 @@ def __init__(
         cidr=None,
         resource_prefix=None,
         restricted_nacl=False,
+        backend_vpc=False,
+        tooling_region=None,
```
dlpzx, line 28: why do we need these params?

deploy/stacks/backend_stack.py
```diff
@@ -146,12 +155,18 @@ def __init__(
...
             resource_prefix=resource_prefix,
             vpc=vpc,
-            vpc_endpoints_sg=vpc_endpoints_sg,
+            vpce_connection=vpce_connection,
             ecr_repository=repo,
             image_tag=image_tag,
             prod_sizing=prod_sizing,
             pivot_role_name=self.pivot_role_name,
             tooling_account_id=tooling_account_id,
+            s3_prefix_list=self.s3_prefix_list,
```
dlpzx, line 164: NIce, since there are different sg for each lambda now we can ser all sg here, but in aurora only the api and aws handlers will get access

dlpzx (commented):
Thank you for this PR! I added my comments and will do some additional testing :)

dlpzx:
> 

Yes, I was thinking of having some alternative to VPC endpoints. The ones introduced in this PR are also needed only for particular features which means that can be avoided if a module is disabled in modularization. So I would wait for this change and implement it directly in the modularization architecture

dlpzx:
Hi @noah-paige! I have tested the backwards compatibility of this feature and I run into the following issue when merging this feature branch in a pre-existing deployment.
The deployment fails in the `CDK Synth` stage throwing the exception`botocore.exceptions.ClientError: An error occurred (UnauthorizedOperation) when calling the DescribePrefixLists operation: You are not authorized to perform this operation.`
The issue is that the role used in Codebuild does not have this permission before being updated in the `SelfMutate` stage.
image

What I did is adding the permission manually to the role that is created for CodeBuild based on the IAM policy(a role which will be deleted in CDK Synth and replaced by a role instead of a policy). After this it succeeds and does not throw any other error.
image

We have 2 options:
1) update the IAM policy of the CodeBuild stage and add the `ec2:DescribePrefixLists` in a PR include it in v1.5.6 minor release > Force customers to upgrade to v1.5.6 before upgrading to v1.6.0
2) include instructions in the v1.6.0 release notes where customers add that permission manually to CodeBuild

backend/dataall/cdkproxy/main.py
```diff
@@ -56,7 +56,12 @@ def up(response: Response):
 def check_creds(response: Response):
     logger.info('GET /awscreds')
     try:
-        sts = boto3.client('sts', region_name=os.getenv('AWS_REGION', 'eu-west-1'))
+        region = os.getenv('AWS_REGION', 'eu-west-1')
```
nikpodsh, line 59: To consider: Looks like we reach the same behavior if we specify `AWS_STS_REGIONAL_ENDPOINTS=regional`. boto3 sts client will make a call to a regional sts endpoint. This way we don't have to worry about forgetting to put region to sts client

backend/dataall/cdkproxy/stacks/dataset.py
noah-paige: Yes exactly - as part of Dataset's `create_dataset()` resolver we check the QuickSight group and create the group if it does not exist already. It would be redundant to have it both there and in the cdkproxy Dataset stack

The reason it is included in this PR is because the cdkproxy would ONLY need internet access via NAT to communicate with QuickSight. If we remove here - the security group for cdkproxy can speak to the remaining services via PrivateLink and VPC Endpoints with no traffic via the internet

deploy/stacks/codeartifact.py
noah-paige: Yes exactly - similar to the way we allow cross account access to the codeartifact domain we need to allow cross account access specifically to the PyPi Repo a part of the domain so that the DBMigration CodeBuild can install packages directly from there

deploy/stacks/container.py
noah-paige: Yes - originally I had cdkproxy with it's own SG but ultimately I found a way to make cdkproxy ECS tasks work with no traffic over the internet so I removed the additional `cdkproxy_security_group` and now use `scheduled_tasks_sg` for all scheduled tasks and cdkproxy while only shares are handled separately with their own SG

deploy/stacks/pipeline.py
```diff
@@ -171,23 +111,10 @@ def __init__(
...
-            ),
-        )
-        for policy in self.codebuild_policy:
-            self.pipeline_iam_role.add_to_policy(policy)
-
         if self.source == 'github':
             source = CodePipelineSource.git_hub(
                 repo_string='awslabs/aws-dataall',
-                branch=self.git_branch,
+                branch=git_branch,
```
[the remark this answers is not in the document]
noah-paige: Good catch! This was a local change I made only when deploying as to not recreate a bunch of new resources with a new branch name when switching the `git_branch` from `main` to `487-restrict-nacls-backend-vpc`.

When reverting back to open the PR I must've reverted this one back incorrectly - I will fix back to `self.git_branch` now 😄

deploy/stacks/pipeline.py
noah-paige: Good catch - will fix

deploy/stacks/pipeline.py
```diff
@@ -209,14 +136,14 @@ def __init__(
...
+                    f'aws codeartifact login --tool npm --repository {self.codeartifact.codeartifact_npm_repo_name} --domain {self.codeartifact.codeartifact_domain_name} --domain-owner {self.codeartifact.domain.attr_owner}',
                     'npm install -g aws-cdk',
-                    f'aws codeartifact login --tool pip --repository {self.codeartifact.pip_repo.attr_name} --domain {self.codeartifact.domain.attr_name} --domain-owner {self.codeartifact.domain.attr_owner}',
+                    f'aws codeartifact login --tool pip --repository {self.codeartifact.codeartifact_pip_repo_name} --domain {self.codeartifact.codeartifact_domain_name} --domain-owner {self.codeartifact.domain.attr_owner}',
                     'pip install -r deploy/requirements.txt',
                     'cdk synth',
                     'echo ${CODEBUILD_SOURCE_VERSION}'
                 ],
-                role_policy_statements=self.codebuild_policy,
+                role=self.pipeline_iam_role,
```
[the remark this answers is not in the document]
noah-paige: This goes back to the previous code where we had the following configuration:

Baseline CodeBuild Policies were assigned to the following CodeBuild Steps:
- Synth
- ALL Quality Gate CodeBuild Steps
- Lambda Image
- ECS Image

Baseline CodeBuild Policies + select sts/cloudfront permissions were assigned to:
- DB Migration
- Stack Updater
- Deploy FrontEnd
- Update Documentation
- Configure RUM
- Configure Cognito
- ALBFrontEndImage
- UserGuideImage

Baseline CodeBuild Policies + sts/cloudfront + select codecommit permissions were assigned to:
- GitRelease

I created 3 roles to align with the above 3 sets of permissions, `pipeline_iam_role`, `codebuild_role`, and `git_project_role` keeping the permissions the same and refactoring to focus on improving management of these permission sets. 

I will rename the roles to be `baseline_codebuild_role`, `expanded_codebuild_role`, and `git_project_role` to avoid confusion with naming. 

I think an optimization here would be to define individual permission sets for each CodeBuild Role (or each logical grouping of CodeBuild Steps) but would opt to leave that as a separate PR unless others think differently?

deploy/stacks/pipeline.py
```diff
@@ -301,6 +206,173 @@ def __init__(
...
+            id=f'CobdeBuildRole{self.git_branch}',
+            role_name=f'{self.resource_prefix}-{self.git_branch}-codebuild-role',
+            assumed_by=iam.ServicePrincipal('codebuild.amazonaws.com'),
+        )
+
+        self.pipeline_iam_policy = iam.Policy(
+            self,
+            'CDKPipelinePolicy',
+            policy_name=f'{self.resource_prefix}-{self.git_branch}-cdkpipelines-policy',
+            roles=[self.pipeline_iam_role, self.codebuild_role],
```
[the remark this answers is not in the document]
noah-paige: Refer to above comment

deploy/stacks/pipeline.py
```diff
@@ -362,15 +423,15 @@ def set_quality_gate_stage(self):
...
+                        f'aws codeartifact login --tool pip --repository {self.codeartifact.codeartifact_pip_repo_name} --domain {self.codeartifact.codeartifact_domain_name} --domain-owner {self.codeartifact.domain.attr_owner}',
                         f'export envname={self.git_branch}',
                         f'export schema_name=validation',
                         'python -m venv env',
                         '. env/bin/activate',
                         'make drop-tables',
                         'make upgrade-db',
                     ],
-                    role_policy_statements=self.codebuild_policy,
+                    role=self.pipeline_iam_role,
```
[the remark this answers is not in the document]
noah-paige: Refer to above comment

deploy/stacks/pipeline.py
```diff
@@ -395,17 +456,17 @@ def set_quality_gate_stage(self):
...
                         '. env/bin/activate',
                         'make lint',
                         'cd frontend',
-                        f'aws codeartifact login --tool npm --repository {self.codeartifact.npm_repo.attr_name} --domain {self.codeartifact.domain.attr_name} --domain-owner {self.codeartifact.domain.attr_owner}',
+                        f'aws codeartifact login --tool npm --repository {self.codeartifact.codeartifact_npm_repo_name} --domain {self.codeartifact.codeartifact_domain_name} --domain-owner {self.codeartifact.domain.attr_owner}',
                         'npm install',
                         'npm run lint',
                     ],
-                    role_policy_statements=self.codebuild_policy,
+                    role=self.pipeline_iam_role,
```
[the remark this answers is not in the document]
noah-paige: Refer to above comment

deploy/stacks/pipeline.py
```diff
@@ -440,7 +501,7 @@ def set_quality_gate_stage(self):
                         )
                     ),
                     commands=[],
-                    role=it_project_role,
+                    role=self.pipeline_iam_role,
```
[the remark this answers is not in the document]
noah-paige: Refer to above comment

deploy/stacks/pipeline.py
```diff
@@ -455,24 +516,12 @@ def set_quality_gate_stage(self):
                         'cd source_build/ && zip -r ../source_build/source_build.zip *',
                         f'aws s3api put-object --bucket {self.pipeline_bucket.bucket_name}  --key source_build.zip --body source_build.zip',
                     ],
-                    role_policy_statements=self.codebuild_policy,
+                    role=self.pipeline_iam_role,
```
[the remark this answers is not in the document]
noah-paige: Refer to above comment

deploy/stacks/vpc.py
noah-paige: good catch - we do not anymore will remove

deploy/stacks/vpc.py
noah-paige: good catch! Will remove

backend/dataall/api/Objects/Dataset/resolvers.py
noah-paige: Added by my changes but needed - will remove

deploy/stacks/pipeline.py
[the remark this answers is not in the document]
dlpzx: My bad for not looking into each role policies! I see now. And yes and additional improvement would be to define each role/set of roles separatedly. But I agree, this can be part of a different PR

backend/dataall/cdkproxy/main.py
noah-paige: Relates to #561

dlpzx (approved):
Looks good, thanks @noah-paige

PR #538, opened 2023-06-28
Add generic way to toggle data.all features

blitzmohit:
### Feature or Bugfix
- Feature

### Detail
- Ability to turn disable features via a single configuration
- Define feature boundaries for 
  - AWS token/console actions on both dataset & env
  - File uploads/actions

### Relates
- #473 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #545, opened 2023-07-03
Change the pipeline for the modularization

nikpodsh:
A way to change the deployment pipeline based on the configuration file has been created in this PR.
Used the same approach with the feature flags as in #538 ( all kudos to @blitzmohit ) since it doesn't require a dedicated module loader (would be too much at least at the current stage for just toggling 4 tasks). Decided to use a bit different naming as it's not about futures/API, though the idea is completely the same. 

The config and decorator `run_if` can be used in many places, not only related to modularization 

I had to copy the config file from the backend since the file is not reachable from the deployment code base (except changing `PYTHONPATH` on the fly, but it'd bring other complications with backend loading)

The open question that's out of this PR scope: should all / some / none properties be migrated from `cdk.json` -> `config.json to` keep them centralize?

### DIFF
--- a/deploy/stacks/container.py
+++ b/deploy/stacks/container.py
@@ -32,11 +32,15 @@ class ContainerStack(pyNestedClass):
     ):
         super().__init__(scope, id, **kwargs)
         self._envname = envname
+        self._resource_prefix = resource_prefix
 
         if self.node.try_get_context('image_tag'):
             image_tag = self.node.try_get_context('image_tag')
 
-        cdkproxy_image_tag = f'cdkproxy-{image_tag}'
+        self._cdkproxy_image_tag = f'cdkproxy-{image_tag}'
+        self._ecr_repository = ecr_repository
+        self._vpc = vpc
+        self._prod_sizing = prod_sizing
 
         self.ecs_security_groups: [aws_ec2.SecurityGroup] = []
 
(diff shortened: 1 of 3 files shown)


blitzmohit:
This looks good @nikpodsh, I believe this can be later be enhanced to include preconfigured contexts (such as an authcontext) to handle more types of complex operations.

deploy/stacks/container.py
```diff
@@ -163,129 +147,161 @@ def __init__(
...
+            cdkproxy_task_definition,
+            catalog_indexer_task.task_definition,
+        ]
+
+        self.add_sync_dataset_table_task()
+        self.add_bucket_policy_updater_task()
+        self.add_subscription_task()
+        self.add_share_management_task()
+
+    @run_if("modules.datasets.active")
```
dbalintx, line 180: Is it expected that we have further module-dependent ECS tasks in the near future?
If the answer is yes, I'd think about generalizing further your solution:
1. keeping a mapping between: modules -> module related ECS tasks (eg. in config.json)
2. iterating through the active modules (maybe taken/saved from the module loader?) and spawning its ECS tasks here

nikpodsh: I believe at this point, no one can say for sure whether we have a new ECS tasks or not.
As of today, only datasets are using deployment changes and we can't predict how things will evolve in the future. So, we should stay as flexible as we can :) 

Plus, the potential changes can be everywhere and not only related to modularization. Some can be small (say like existing  `enable_cw_canaries`) and some big ( `opensearch serverless`). Of course, if we decide not to use `cdk.json` for configuring those properties

dbalintx (approved):

PR #550, opened 2023-07-04
Refactoring of aws calls

nikpodsh:
AWS calls should be done only from clients (or old handlers). 
I found some missing handler/client that weren't migrated during the modularization. Moved a couple of calls to AWS from the resolvers/tasks to clients  


backend/dataall/db/connection.py
```diff
@@ -11,7 +11,7 @@
 from .. import db
 from ..db import Base
 from ..db.dbconfig import DbConfig
-from ..utils import Parameter, Secrets
```
dbalintx, line 14: I guess usage should also be removed from [line 118](https://github.com/awslabs/aws-dataall/pull/550/files#diff-0674a6fd89dc34e30247435f9c6e4db20db0934e3171aa94f3acf22a592f14eaL118).

There is also a boto3 call in line 120 - if you want to outsource that as well.

backend/dataall/modules/dataset_sharing/services/share_managers/lf_share_manager.py
```diff
@@ -10,7 +10,7 @@
 from dataall.modules.dataset_sharing.aws.lakeformation_client import LakeFormationClient
 from dataall.aws.handlers.quicksight import QuicksightClient
 from dataall.aws.handlers.sts import SessionHelper
-from dataall.aws.handlers.ram import Ram
+from dataall.modules.dataset_sharing.aws.ram_client import RamClient
```
dbalintx, line 13: import is not needed

backend/dataall/modules/datasets/aws/s3_dataset_client.py
```diff
@@ -32,3 +35,80 @@ def get_file_upload_presigned_url(self, data):
             return json.dumps(response)
         except ClientError as e:
             raise e
+
+
+class S3BucketPolicyClient:
```
dbalintx, line 40: I'd include maybe `Dataset` in the class name to better differentiate it from `S3Client` class.

Or, as I see in this class only the `dataset.S3BucketName`, `dataset.datasetUri ` and `dataset.AwsAccountId `members are used from Dataset. If you would like to, you could generalize these 2 methods and put them in the `S3Client` class itself, then use them within Datasets providing the above mentioned members.

backend/dataall/modules/datasets/tasks/dataset_subscription_task.py
```diff
@@ -163,17 +161,6 @@ def publish_sns_message(
                             f'Failed to deliver message {message} due to: {e}'
                         )
 
-    @staticmethod
-    def sns_call(message, environment):
-        aws_session = SessionHelper.remote_session(environment.AwsAccountId)
-        sns = aws_session.client('sns', region_name=environment.region)
-        response = sns.publish(
-            TopicArn=f'arn:aws:sns:{environment.region}:{environment.AwsAccountId}:{environment.subscriptionsConsumersTopicName}',
```
dbalintx, line 171: In SnsDatasetClient - which you used to replace this method - the TopicArn differs, there it's:
`f'arn:aws:sns:{environment.region}:{environment.AwsAccountId}:{environment.subscriptionsProducersTopicName}'`

backend/dataall/modules/datasets/aws/s3_dataset_client.py
nikpodsh: I have added `Dataset` to the name.
I have split those two `S3Client` and `S3DatasetBucketPolicyClient`, because they use different boto3 clients:
S3Client uses an additional param `config=Config(signature_version='s3v4', s3={'addressing_style': 'virtual'}),`. it might have been nothing and `S3DatasetBucketPolicyClient` could use with config without issues, but I was not sure and created a new class

backend/dataall/modules/datasets/tasks/dataset_subscription_task.py
nikpodsh: Nice catch! Updated the topic

dbalintx (approved):

PR #552, opened 2023-07-05
Converting data policy from inline policy to managed policy

arjun-chopra:
### Feature or Bugfix
- Enhancement

### Detail
- Data.all adds an inline policy called `dataPolicy` to the IAM roles created in the environment stack. The inline statements contain permissions to datasets and kms keys owned by each team. If a team owns a big number of datasets the inline policy will grow accordingly.
- IAM restricts inline policy to 10k length combined and which with the current approach it is probable to cause the CloudFormation environment stack to fail when a team owns a certain number of datasets. To mitigate that dataPolicy needs to be implemented using managed policy instead on inline policy.


arjun-chopra:
#553

dlpzx:
As we discussed in our call today @arjun-chopra, this PR won't solve your issue. This PR modified the environment roles, while you want to reduce the inline policy size of the pivotRole. Do you want me to close this PR? or do you want to keep working on it? Let me know what works best for you

arjun-chopra:
@dlpzx This wont fix my issues i am having but its good to have managed policies instead of inline. If you are inline, we can move ahead or close this.

PR #560, opened 2023-07-10
Cdk exec policy link env

mourya-33:
### Detail
Add Least privilege policy cloudformation stack YAML file on the linked environments screen to allow users choose a custom IAM policy for CDK bootstrapping instead of the default Administrator Access policy

### Relates


deploy/cdk_exec_policy/cdkExecPolicy.yaml
```diff
@@ -0,0 +1,180 @@
...
+              - 'ecr:DeleteRepository'
+            Resource: !Sub 'arn:aws:ecr:${AWS::Region}:${AWS::AccountId}:repository/cdk-*'
+          - Sid: SSMTwo
+            Effect: Allow
+            Action:
+              - 'ssm:GetParameter'
+              - 'ssm:PutParameter'
+              - 'ssm:DeleteParameter'
+            Resource: !Sub 'arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/cdk-bootstrap/*'
+          - Sid: CloudformationTwo
```
noah-paige, line 167: For the `CloudfromationTwo` Statement - if my understanding is correct it appears that this statement giving the execution role access to ALL actions as long as the role is being used by CloudFormation

If so is this not the same as the default behavior of adding AdministratorAccess policy to the CDK Execution role? Can we possible restrict this further?

PR #562, opened 2023-07-11
Cdk exec policy linked envs

mourya-33:
### Feature 
- Feature

### Detail
- UI changes to provide CloudFormation stack Yaml file to create custom IAM policy with least privileges for bootstrapping CDK in linked environments.


backend/dataall/api/Objects/Environment/resolvers.py
```diff
@@ -699,7 +699,51 @@ def get_pivot_role_template(context: Context, source, organizationUri=None):
...
+def get_cdk_exec_policy_template(context: Context, source, organizationUri=None):
+    from ....utils import Parameter
 
+    with context.engine.scoped_session() as session:
+        ResourcePolicy.check_user_resource_permission(
+            session=session,
+            username=context.username,
+            groups=context.groups,
+            resource_uri=organizationUri,
+            permission_name=permissions.GET_ORGANIZATION,
```
dlpzx, line 711: I am not sure about this permission here. Permissions in data.all apply to data.all objects. For example "`DeleteEnvironment` applies to resource: `EnvironmentA`" or "`CreateEnvironment` applies to resource: OrganizationA"

In this case what you have defined is that any user with access to an Organization can download the template. 

I would say that we have 2 options:
1) permissions are not needed as this template is publicly available in GitHub and they do not contain any sensitive info. There is a permission that restricts the creation of environments in the organization
2) use the `LINK_ENVIRONMENT` permission instead. It is more appropiate as this is a pre-step for linking an environment

backend/dataall/api/Objects/Environment/resolvers.py
```diff
@@ -699,7 +699,51 @@ def get_pivot_role_template(context: Context, source, organizationUri=None):
             )
             raise e
 
+def get_cdk_exec_policy_template(context: Context, source, organizationUri=None):
```
dlpzx, line 702: Instead of using the utils.Parameter class could you re-use the ParameterStore that is imported globally? This is to centralize the boto3 calls (Parameter will be removed in future releases)

backend/dataall/api/Objects/Environment/resolvers.py
```diff
@@ -699,7 +699,51 @@ def get_pivot_role_template(context: Context, source, organizationUri=None):
...
+            groups=context.groups,
+            resource_uri=organizationUri,
+            permission_name=permissions.GET_ORGANIZATION,
+        )
+        cdk_exec_policy_bucket = Parameter().get_parameter(
+            env=os.getenv('envname', 'local'), path='s3/resources_bucket_name'
+        )
+        cdk_exec_policy_bucket_key = Parameter().get_parameter(
+            env=os.getenv('envname', 'local'), path='s3/cdk_exec_policy_prefix'
+        )
```
dlpzx, line 718: The parameter path needs to be prefixed, something like `parameter_path=f"/dataall/{os.getenv('envname', 'local')}/s3/...` otherwise the pivotRole won't have permissions to read it

backend/dataall/api/Objects/Environment/resolvers.py
```diff
@@ -699,7 +699,51 @@ def get_pivot_role_template(context: Context, source, organizationUri=None):
...
+        if not cdk_exec_policy_bucket or not cdk_exec_policy_bucket_key:
+            raise exceptions.AWSResourceNotFound(
+                action='GET_CDK_EXEC_POLICY_TEMPLATE',
+                message='CDK Exec Yaml template file could not be found on Amazon S3 bucket',
+            )
```
dlpzx, line 723: I think this exception should be added in line 742, as it is the moment in which we actually check if the file exists and we can download it

backend/dataall/api/Objects/Environment/resolvers.py
```diff
@@ -699,7 +699,51 @@ def get_pivot_role_template(context: Context, source, organizationUri=None):
...
+        try:
+            s3_client = boto3.client(
+                's3',
+                region_name=os.getenv('AWS_REGION', 'eu-central-1'),
```
dlpzx, line 727: This is a very odd default region, maybe we can think of a configurable default

backend/dataall/api/Objects/Environment/resolvers.py
dlpzx: I just realized that you were following the example of the pivotRole template. So most of my comments here are not applicable for this PR, we will open a separate issue to improve how we get both the pivotRole template and the cdkExec role template in a separate PR.

backend/dataall/api/Objects/Environment/resolvers.py
dlpzx: Not applicable, separate PR

backend/dataall/api/Objects/Environment/resolvers.py
dlpzx: Not applicable, separate PR

backend/dataall/api/Objects/Environment/resolvers.py
dlpzx: Not applicable, separate PR

deploy/cdk_exec_policy/cdkExecPolicy.yaml
```diff
@@ -0,0 +1,180 @@
...
+              - 'iam:CreatePolicy'
+              - 'iam:GetPolicy'
+            Effect: Allow
+            Resource:
+              - !Sub 'arn:aws:iam::${AWS::AccountId}:policy/*'
+          - Sid: IAMRole
+            Action:
+              - 'iam:AttachRolePolicy'
+              - 'iam:CreateRole'
+              - 'iam:CreateServiceLinkedRole'
```
dlpzx, line 54: What about role deletion? What happens when we delete an environment or dataset stack?

deploy/cdk_exec_policy/cdkExecPolicy.yaml
```diff
@@ -0,0 +1,180 @@
...
+              - 's3:PutBucketAcl'
+              - 's3:PutBucketNotification'
+            Effect: Allow
+            Resource:
+              - !Sub 'arn:aws:s3:::${EnvironmentResourcePrefix}-logging-*'
+          - Sid: ReadBuckets
+            Action:
+              - 'kms:CreateAlias'
+              - 'kms:CreateGrant'
+              - 'kms:Decrypt'
```
dlpzx, line 77: Why do we need Encrypt/Decrypt permissions for this role. It is a role to create resources right?

deploy/cdk_exec_policy/cdkExecPolicy.yaml
```diff
@@ -0,0 +1,180 @@
...
+Description: Custom least privilege IAM policy for linking environments to dataall
+Parameters:
+  AwsAccountId:
+    Description: AWS AccountId of the dataall INFRASTRUCTURE account that we wish to link this environment with.
+    Type: String
+  PolicyName:
+    Description: IAM policy name (copy/paste from the UI)
+    Type: String
+    Default: 'DataAllCustomCDKPolicy'
+  EnvironmentResourcePrefix:
```
dlpzx, line 11: We are barely using this prefix in our policies and we could really restrict access to prefixed resources only (e.g. all S3 buckets created will be repfixed with this resource)

deploy/cdk_exec_policy/cdkExecPolicy.yaml
```diff
@@ -0,0 +1,180 @@
...
+              - 'iam:PutRolePolicy'
+            Effect: Allow
+            Resource: '*'
+          - Sid: KMSCreateAlias
+            Action:
+              - 'kms:CreateAlias'
+            Effect: Allow
+            Resource:
+              - !Sub 'arn:aws:kms:${AWS::Region}:${AWS::AccountId}:alias/*'
+          - Sid: KMSKey
```
dlpzx, line 66: I don't understand this statement, what is it for?

dlpzx (changes requested):
It needs some changes. Plus I see that it is based out of main so it introduces conflicts in v1.6.0 branch. We either rebase to v1m6m0 or we wait to do the PR to main. What do you think?

frontend/src/views/Environments/EnvironmentCreateForm.js
```diff
@@ -285,6 +296,42 @@ const EnvironmentCreateForm = (props) => {
                   {`cdk bootstrap --trust ${trustedAccount} -c @aws-cdk/core:newStyleStackSynthesis=true --cloudformation-execution-policies arn:aws:iam::aws:policy/AdministratorAccess aws://ACCOUNT_ID/REGION`}
```
dlpzx, line 296: Maybe we can remove this command from here as we are specifying another command to bootstrap in lines 331-333

deploy/cdk_exec_policy/cdkExecPolicy.yaml
```diff
@@ -0,0 +1,180 @@
+AWSTemplateFormatVersion: 2010-09-09
```
dlpzx, line 1: I do not see any permissions related to tags. When you created datasets and environments using this execution role, did the resources get tagged?

noah-paige:
I have tested the following with the restricted CDK Execution Role:

- [x] Create Environment
- [x] Update Environment (All Play Features Enabled)
- [x] Delete Environment

- [x] Create Notebook
- [x] Delete Notebook

- [x] Create ML Studio User Profile
- [x] Delete ML Studio User Profile

- [x] Create Dataset
- [x] Update Dataset
- [x] Delete Dataset

- [x] Create All Pipeline Types w/ CDK Exec Policy Passed on DDK Bootstrap
- [x] Delete All Pipeline Types  w/ CDK Exec Policy Passed on DDK Bootstrap

**Comments:**
For the `CloudfromationTwo` Statement - if my understanding is correct it appears that this statement giving the execution role access to ALL actions as long as the role is being used by CloudFormation. If so is this not very similar to adding full access to the CDK Execution role? Can we possible restrict this further?

Additionally, in my testing I had to add the following permissions to the CDK Execution Role to get all create/delete actions to work as expected:

_Create Dataset_
iam:TagRole
Iam:UnTagRole

_Delete Dataset_
Iam:DeleteRole
Iam:DetachRolePolicy
iam:DeleteRolePolicy
lambda:RemovePermission
S3:DeleteBucketPolicy

_Create CDK Pipeline_
codepipeline:TagResource
codepipeline:UntagResource
codepipeline:CreatePipeline
codepipeline:UpdatePipeline
codepipeline:StartPipelineExecution
codepipeline:GetPipeline
iam:PassRole
iam:UpdateAssumeRolePolicy
codepipeline:GetPipelineState
events:PutRule
events:DescribeRule
eventsPutTargets

_Delete CDK Pipeline_
events:RemoveTargets
events.DeleteRule
codepipeline:DeletePipeline
s3:DeleteBucket

Delete Environment
iam:ListPolicyVersions
iam:DeletePolicy

NOTE: For Data All Pipelines this is a little tricky since the DDK execution role permissions will be used to both set up the foundational CICD Pipeline (this is managed by data all and known) AND deploy the infra to the respective development environments (this is customized by the customer and can be extended however they see fit per their use case). 

Because we will not know what a person would want to create as pipeline of their stack deployed in the pipeline should we opt to leave the default `ddk bootstrap` command with broader permissions? Curious on thoughts here

frontend/src/views/Environments/EnvironmentCreateForm.js
dlpzx: Okeeey, I see you want to leave options!

mourya-33:
Create Environment

Update Environment (All Play Features Enabled)

Delete Environment

Create Notebook
Updated the stack to add additional required permissions and tested the following.

Delete Notebook
Create ML Studio User Profile
Delete ML Studio User Profile
Create Dataset
Update Dataset
Delete Dataset
Dashboards

deploy/cdk_exec_policy/cdkExecPolicy.yaml
dlpzx: For future enhancements

deploy/cdk_exec_policy/cdkExecPolicy.yaml
```diff
@@ -0,0 +1,280 @@
...
+            Resource: '*'
+          - Sid: ec2
+            Action:
+              - 'ec2:DescribeSecurityGroups'
+              - 'ec2:CreateSecurityGroup'
+              - 'ec2:createTags'
+              - 'ec2:DeleteSecurityGroup'
+              - 'ec2:RevokeSecurityGroupEgress'
+              - 'ec2:AuthorizeSecurityGroupIngress'
+              - 'ec2:AuthorizeSecurityGroupEgress'
```
noah-paige, line 35: When testing creating a new Environment with no default VPC and ML Studio Enabled, required some additional permissions for stack creation:
```
                "ec2:RevokeSecurityGroupIngress",
                "ec2:AttachInternetGateway",
                "ec2:DetachInternetGateway",
                "ec2:DescribeInternetGateways",
                "ec2:DescribeRouteTables",
                "ec2:AssociateRouteTable",
                "ec2:DisassociateRouteTable",
                "ec2:CreateRoute",
                "ec2:DeleteRoute"
```


NOTE: Adding the above additional permissions may cause a character max length exceeded in the policy. I will suggest 1 area where I think we can consolidate permissions in the YAML File 

In the long-term I imagine we will have to split this out into multiple policies and pass a list of execution policies to the CDK bootstrap command as we refine this permission set

deploy/cdk_exec_policy/cdkExecPolicy.yaml
```diff
@@ -0,0 +1,280 @@
...
+              - 'iam:ListPolicies'
+              - 'iam:ListAttachedRolePolicies'
+              - 'iam:ListGenerateServiceLastAccessedDetails'
+              - 'iam:ListEntitiesForPolicy'
+              - 'iam:ListPoliciesGrantingServiceAccess'
+              - 'iam:ListRoles'
+              - 'iam:GetServiceLastAccessedDetails'
+              - 'iam:ListAccountAliases'
+              - 'iam:ListRolePolicies'
+            Effect: Allow
```
noah-paige, line 76: Here I think we can add  `iam:List*` and remove each call out of `iam:List...` to consolidate this statement

noah-paige (changes requested):

deploy/cdk_exec_policy/cdkExecPolicy.yaml
mourya-33: I did both the above changes. Yes we should refine this further to split into multiple policies for the long-term.

noah-paige (approved):
Re-tested creating and deleting of all data all stacks - good to merge from me

dlpzx (approved):
Looks good. It can serve as starting point but there are a couple of possible enhacements

dlpzx:
Found issues during deployment of branch V1.6.0 introduced by this PR. I am opening a separate PR for this.

Issue #531, opened 2023-06-22
data.all should not be able to delete all Glue databases and tables in Environment account


jeffshep:
### Describe the bug

A fundamental issue with data.all

Deleting a dataset, that was imported from existing resources, results in the Glue database being deleted in the AWS environment. The Glue resources are not owned by data.all, they are owned by a data pipeline in the account already, and should exist independently of whether the data is shared with data.all or not.

### How to Reproduce

In an AWS account, create a Glue database and table, pointing to data in S3.
In data.all, Import a new dataset with these resources.
Wait for data.all to create the CloudFormation stack to import the dataset successfully.
Delete the dataset in data.all.
Observe the dataall-gluedb-lf-handler-UUID resource delete the Glue database and table

### Expected behavior

In an AWS account, create a Glue database and table, pointing to data in S3.
In data.all, Import a new dataset with these resources.
Wait for data.all to create the CloudFormation stack to import the dataset successfully.
Delete the dataset in data.all.
The database and table should remain in the AWS account, data.all permissions should be removed.

### OS

n/a

### Python version

n/a

### AWS data.all version

1.5.4

dlpzx:
Hi @jeffshep, thanks for raising the issue. Yes we are aware of this and were discussing internally how to tackle it best. We have 2 alternatives:
- Do not delete databases in any case - do not delete imported databases or data.all created databases. The reason being that Glue databases should be considered differently and only be deleted manually (as it happens with s3 buckets for example)
- Do not delete imported databases. We can clean-up data.all created ones but the imported ones should always be kept and as you said above, data.all should respect existing resources

I would like to know your opinion on it so that we can start working on it and include it in V1.6 release (mid July)

jeffshep:
@dlpzx Thanks for the quick response, I would strongly advocate option 2. If I was using data.all to create resources I would expect it to delete them when I delete a dataset.

dlpzx:
Thank you @jeffshep, we will try to include is as optional for v1.6 and at that state it will keep imported databases only.

dlpzx:
It was implemented in #524, it is now under validation

PR #498, opened 2023-06-06
feat: add share reason in share requests

noah-paige:
### Feature or Bugfix
- Feature

### Detail
- Expose the Share Request Comments as Part of the Share View for the Approver to See 
- Allow Approver to Provide Comments when Rejecting a Share for the Requester to See

### Relates
- [#422 ]

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #565, opened 2023-07-11
V1.6.0 features

dlpzx:
### Feature or Bugfix
Release PR with the following list of features. Refer to each PR for the details

### Detail
- #498 
- #482 
- #543
- #524 (which also solves #531)
- #532 
- #535 
- #497 
- #515
- #529 
- #562 
- #455 
- #572 
- #567 
- #573 
- #579 
- #578 
- #582 

### Breaking changes - release notes
- ⚠️ IMPORTANT: upgrade to a version >V1.5.0 before upgrading to V1.6 to avoid deletion of resources in custom resource deletion
- ⚠️ IMPORTANT: requires an update of environments and then datasets after upgrading. Either using cdk.json parameter `enable_update_dataall_stacks_in_cicd_pipeline`, waiting for overnight update stack task, or manually updating first environments and then datasets
- CloudFront distribution replace for #529 
- Additional EC2 permissions in CDK Synth CodeBuild stage for #543 --> this can be avoided by upgrading to v1.5.6 before upgrading to v1.6.0
- local development affected by more restrictive pivotRole trust policy

### Relates
V1.6.0 release


dlpzx:
During validation 2 issues where found:
- Imported Dataset stack creates non-prefixed resources. As a result the pivot role with restricted permissions cannot access the resources and features such as "StartCrawler" fail.
- Support for SSE-S3 default encryption buckets needed. Enforcing the KMS key as input removes the option for customers to use the default S3 encryption settings.

I will work on this items today @noah-paige. We can meet later to discuss on them

noah-paige (approved):
Deployed and tested v1.6 on AWS for all data.all components

The changes look good to me! Approving this PR

PR #569, opened 2023-07-12
Migrate to a new permission checker

nikpodsh:
There is a new permission checker API in the modularization branch. It allows us get away from passing unused variables into the methods. 
The previous permission checker used a certain signature that expected special parameters  passed in a certain order. Though, most of the time those passed parameters were not used in the method, but only were needed for the permission checker decorator.
The other issue with the previous permission checker, that it had forced us to pass a request data as a dictionary e.g. in the resolver we had had some parameters and had to put them into dictionary and unpack them back in the method. It was not memory inefficient
All in all, a new permission checker doesn't require to follow previous agreements which will us to write  shorter and more concise API, we can utilize type checking and avoid unnecessary memory allocation during the object creation. 

The PR is waiting other PR to merged.


backend/dataall/db/api/environment.py
```diff
@@ -21,29 +19,32 @@
...
 
 log = logging.getLogger(__name__)
 
 
 class Environment:
     @staticmethod
-    @has_tenant_perm(permissions.MANAGE_ENVIRONMENTS)
-    @has_resource_perm(permissions.LINK_ENVIRONMENT)
-    def create_environment(session, username, groups, uri, data=None, check_perm=None):
+    @has_resource_permission(permissions.MANAGE_ENVIRONMENTS)
```
dbalintx, line 37: I think decorators are mixed up: 
tenant -> manage env
resource -> link env

backend/dataall/db/api/environment.py
```diff
@@ -392,20 +382,19 @@ def update_group_permissions(
...
-        session, username, groups, uri, data=None, check_perm=None
-    ):
-        if not data:
-            raise exceptions.RequiredParameter('data')
-        if not data.get('groupUri'):
-            raise exceptions.RequiredParameter('groupUri')
+    @has_resource_permission(permissions.LIST_ENVIRONMENT_GROUP_PERMISSIONS)
+    def list_group_permissions(session, uri, group_uri):
+        # the permission checked
+        return Environment.list_group_permissions(session, uri, group_uri)
```
dbalintx, line 388:
```suggestion
        return Environment.list_group_permissions_internal(session, uri, group_uri)
```

backend/dataall/db/api/environment.py
```diff
@@ -622,29 +595,20 @@ def query_environment_invited_groups(
...
 
     @staticmethod
-    @has_resource_perm(permissions.LIST_ENVIRONMENT_GROUPS)
-    def list_environment_invited_groups(
-        session, username, groups, uri, data=None, check_perm=None
-    ) -> dict:
-        return Environment.query_environment_invited_groups(
-            session, username, groups, uri, data
-        ).all()
+    def list_environment_invited_groups(session, uri):
```
dbalintx, line 607: permission decorator is missing from here

backend/dataall/db/api/environment.py
nikpodsh: Yes, my bad! :)

backend/dataall/db/api/environment.py
nikpodsh: This method is used while deploying env stack. We can't put a decorator, because we are out of user context

dbalintx (approved):

PR #573, opened 2023-07-13
Networking Configuration Support for Multi-Region Data All Resources

noah-paige:
### Feature or Bugfix
- Feature
- Bugfix

### Detail
- Allow Traffic to other AWS Regions to support Data All resources in other regions (e.g. Deployment in `us-east-1` and Link Environment in `us-west-2`)
- Based on Route Priority in AWS [Docs](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Route_Tables.html#route-tables-priority) routes will prioritize the most specific CIDR Blocks first as preferred method of traffic
  - Same region resources as the deployment account region should defer to private communication while deployment of resources to regions different from the deployment region will be able to use NAT Gateway from the private subnet to deploy the stacks
  - Alternative to this solution is to specify VPC Endpoints for every Region where data all related resources could/would be deployed to

### DIFF
--- a/deploy/stacks/container.py
+++ b/deploy/stacks/container.py
@@ -393,11 +393,11 @@ class ContainerStack(pyNestedClass):
                         'Allow Lambda to ECS Connection'
                     )
 
-            # Add NAT Gateway Access for RAM API Access
-            sg_connection.add_egress_rule(
+            # Add NAT Gateway Access for Cross-region requests in same region the more specific rules apply
+            sg_connection.allow_to(
                 ec2.Peer.any_ipv4(),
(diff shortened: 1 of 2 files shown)


dlpzx:
I agree on using NAT GW for cross-region requests. VPC endpoints in each region + peering with each region + more routing seems a bit overkill. I will add it to the tests I am doing for backwards compatibility with a deployment in eu-west-1 that has environments in eu-west-1 and eu-west-2

dlpzx:
@noah-paige I fixed a small bug and I am retesting again

dlpzx (approved):

PR #581, opened 2023-07-17
Frontend Modularization: Catalog

itsmo-amzn:
### Feature or Bugfix

- Refactoring

### Detail
- Move catalog views into catalog module
- Move services that were only used by catalog into catalog module
- Move catalogs components into catalog module

### Relates
- [Frontend modularization](https://github.com/awslabs/aws-dataall/issues/398)


frontend/src/modules/Catalog/components/GlossarySearch.js
```diff
@@ -319,4 +319,3 @@ GlossarySearch.propTypes = {
   setQuery: PropTypes.func.isRequired,
   matches: PropTypes.array.isRequired
 };
-export default GlossarySearch;
```
jorgeig-space, line 322: For 2.0.1: we should rename `GlossarySearch` and `GlossarySearchComponent` to be more consistent, i.e., when we read the name, we should have an idea of what each entity is: what is `GlossarySearch` vs `GlossarySearchComponent`? Why does one have a qualifier and the other doesn't?

frontend/src/modules/Catalog/components/Hit.js
```diff
@@ -45,7 +45,7 @@ HitICon.propTypes = {
   hit: PropTypes.object.isRequired
 };
 
-const Hit = ({ hit }) => {
+export const Hit = ({ hit }) => {
```
jorgeig-space, line 48: For 2.0.1: We should rename `Hit` to be more descriptive: `SearchHit`? `GlossarySearchResultHit`?

jorgeig-space:
LGTM w.r.t. modularization. Left some comments for 2.0.1.

jorgeig-space:
@itsmo-amzn is it worth it merging #581 first and then using relative imports in this PR already?

frontend/src/modules/Catalog/components/Hit.js
itsmo-amzn: I believe `GlossarySearchResultItem` is much more descriptive.

frontend/src/modules/Catalog/components/GlossarySearch.js
itsmo-amzn: renamed to `GlossarySearchWrapper` and `GlossarySearchUI`

itsmo-amzn:
> @itsmo-amzn is it worth it merging #587 first and then using relative imports in this PR already?

Yes. I rebased to fetch the new changes, and updated all relative imports in the Catalog module to reflect this

frontend/src/services/graphql/index.js
```diff
@@ -1,7 +1,6 @@
 export * from './Activity';
 export * from './AirflowCluster';
 export * from './ApiKeys';
-export * from './Catalog';
```
nikpodsh, line 4: I wonder if we want to disable catalog in the routes when Dashboards and Datasets are inactive? 
Otherwise, it will be just empty

itsmo-amzn: @jorgeig-space can you weigh-in on this?

jorgeig-space: @dlpzx wdyt? I think ideally we should disable the Catalog if its functionality does not make sense when the other two modules that use it are inactive.

dlpzx: For me `Datasets` and `Dashboards` should have a depends on on the `Catalog` module like it happens with `Datasets` and `Dataset_sharing`

dlpzx (commented):
looks good. I like the renaming and the separation of components and views. Plus the improvements in the imports

frontend/src/services/graphql/index.js
itsmo-amzn: @dlpzx can you please elaborate?

dlpzx: Yes, sorry I was a bit cryptic. In an ideal world, the Catalog module should be enabled if either Datasets or Dashboards are enabled. Otherwise it should be disabled. But currently it is considered `core` in the backend modularization. @nikpodsh wdyt?
option 1: keep Catalog in core (backend) or always enabled (frontend)
option 2: modularize Catalog with Dataset and Dashboards depends on Catalog (backend) and make Catalog dependent on Datasets and Dashboards (frontend)
option 3: keep Catalog in core (backend) and make Catalog dependent on Datasets and Dashboards (frontend). We will modularize Catalog (and its infrastructure) after V2.0.0 release

frontend/src/routes.js
```diff
@@ -219,10 +217,11 @@ const routes = [
           }
         ]
       },
-      {
-        path: 'catalog',
-        element: <Catalog />
-      },
+      config.modules.datasets.active &&
```
nikpodsh, line 220: `&&` is wrong. The logic here: `if at least one of (dashboards, dashboards) is not present, catalog is not displayed` instead it should be `if all of them are not present, catalog is not displayed`

nikpodsh (approved):

Issue #398, opened 2023-03-31
Frontend Modularization Planning


AmrSaber:
# Frontend Modularization Plan

## Current Frontend Structure

The current structure of the frontend application is as follows

```
src
├── api -- contains integration API files
├── components -- All components used in the project
├── contexts -- Used contexts (provide some info to app)
├── hooks -- All hooks used in the project
├── icons -- All icons used in the project
├── store -- Common redux store files
├── theme -- Themes used in the app
├── utils -- Utilities used in different parts of the application
├── views -- All the screens in the application
├── App.js -- React root
├── index.js -- Project root
├── routes.js -- Project routes
└── constants.js -- All Constants used in the application
```

As you see, the files are structure based on their logical properties, and not based on their module. But a lot of the directories already contains common code used by different parts of the application that we could consider modules, we will make use of that fact later on.

### The Problem With The Current Structure

The project is currently structured in a manner that describes the "What" of each component of the code. As the scope and the size of the project get bigger and bigger, this structure gets in the way of contributing to the project because of 2 problems:

1. To update an existing module, you will need to go through all the directories and figure out which of them is related to the module you want to update
1. To add a new module, the process is not well defined at all, you will need to go around the project creating files everywhere which also increases the difficulty of the first problem

Because of this, we agreed that we will split the project into meaningful modules, so that contributing to an existing module is almost trivial, you will know exactly what files are related to that module. And also adding a new module will be a well defined process so that the project is kept organized.

## Modular Structure

```
src
├── authentication
│   ├── components
│   ├── contexts
│   │   ├── AuthContext.js
│   │   ├── CognitoAuthContext.js // Previously AmplitudeContext.js
│   │   ├── LocalAuthContext.js // Previously LocalContext.js
│   │   └── index.js
│   ├── hooks
│   └── index.js
├── design
│   ├── components
│   ├── contexts
│   │   ├── UXSettingsContext.js
│   │   └── index.js
│   ├── hooks
│   ├── icons
│   ├── themes
│   └── index.js
├── globalErrors // Previously store
├── modules
│   ├── Misc
│   ├── MLStudio
│   ├── ...
│   └── Notebooks
│       ├── components
│       ├── hooks
│       ├── services
│       ├── views
│       └── constants.js
├── services // Previously api
│   ├── graphql
│   └── hooks
├── utils
├── App.js
├── index.js
└── routes.js
```

The description of each of the new directory is as follows:

[...]

PR #590, opened 2023-07-20
Frontend Modularization: Datasets

itsmo-amzn:
### Feature or Bugfix

- Refactoring

### Detail
- Move datasets views into dataset module
- Move services that were only used by datasets into datasets module
- Move datasets components into datasets module

### Relates
- #398 


frontend/src/modules/Datasets/views/DatasetView.js
```diff
@@ -31,25 +31,21 @@ import {
   PencilAltIcon,
   UpVoteButton,
   useSettings
-} from '../../../../design';
-import { SET_ERROR, useDispatch } from '../../../../globalErrors';
+} from '../../../design';
```
nikpodsh, line 34: routes.js should be fixed to match new paths for all view

frontend/src/modules/Datasets/views/DatasetCreateForm.js
```diff
@@ -29,15 +29,15 @@ import {
   ChipInput,
   Defaults,
   useSettings
-} from '../../../../design';
-import { SET_ERROR, useDispatch } from '../../../../globalErrors';
+} from '../../../design';
```
nikpodsh, line 32: Datasets (and everything related to it as tables, locations/folders, shares) is one of the optional modules. The views should not be shown if the module is disabled. We should do the same as it has been done for notebooks

itsmo-amzn: can you please elaborate?

nikpodsh: We shouldn't show dataset's forms if the dataset module is inactive. It was one of the requirements of the modularization: s customer can specify what modules are active/inactive in the config.json. 
The other example of the module that we can make inactive is `notebook`. If we set `modules.notebook.active = false` in the config.json, we won't see anything related to notebooks. For the reference, please take a look at #415. 
The list of the optional [modules](https://github.com/awslabs/aws-dataall/blob/modularization-main/config.json)

itsmo-amzn:
@jorgeig-space, and @nikpodsh 
I moved all shared components into the Shared module, and shared services into services module.
The PR is quite big as it affects many other modules that are pending modularization.

I also updated the datasets module to be optional in case the user disables it in the configs.json file

frontend/src/services/graphql/Datasets/updateDatasetStorageLocation.js
nikpodsh, line 1: I wonder how we are going to deal with imports for all dataset related enitities like Folders and Tables ? e.g. `FolderEditForm` needs this `updateDatasetStorageLocation.js`, but `FolderEditForm`  will end up in `Folder` module

I think we need to discuss:
1) if we are going to allow importing from one module to another e.g. `FolderCreationForm` will have `import {addDatasetStorageLocation} from ../../../Datasets/services`. 
2) if we are going to put **shared for Dataset & co only**  services to the Shared folder
3) Come up with a bit different layout for datasets

frontend/src/services/graphql/Datasets/index.js
```diff
@@ -0,0 +1,8 @@
+export * from './addDatasetStorageLocation';
```
nikpodsh, line 1: All imports for Folder's view should be fixed according to a new location of services

frontend/src/modules/Shared/index.js
```diff
@@ -0,0 +1,9 @@
+export * from './FeedComments';
```
nikpodsh, line 1: Is it a comprehensive list of all `Shared` components or is it just what has been found so far? 
I am asking because if the list keeps growing (in the scope of modularization) we might need to think about reorganizing this into something like 
```
    Shared
           Comments
                  FeedCommentAdd
                  FeedComments
           Stack
                  Stack Stuff
           KeyValueTags
                  Key Value Stuff etc
```
And import it as `import { FeedComments } from 'modules/Shared/Comments';`
But it don't expect anything to pop up we can proceed

frontend/src/modules/Shared/index.js
```diff
@@ -0,0 +1,9 @@
...
+export * from './FeedCommentAdd';
+export * from './KeyValueTagList';
+export * from './KeyValueTagUpdateForm';
+export * from './ShareInboxList';
```
nikpodsh, line 5: `ShareInboxList` and `ShareInboxListItem` are pure dataset functionality. If datasets are off `listDatasetShareObjects` and `getShareRequestsToMe` will fail

frontend/src/modules/Datasets/services/index.js
```diff
@@ -1,19 +1,11 @@
-export * from './addDatasetStorageLocation';
 export * from './createDataset';
 export * from './deleteDataset';
 export * from './generateDatasetAccessToken';
-export * from './getDataset';
-export * from './getDatasetAssumeRoleUrl';
 export * from './getDatasetPresignedUrl';
 export * from './getDatasetSchema';
```
nikpodsh, line 5: `getDatasetSchema` is no longer used and can be deleted as far as I see

frontend/src/modules/Datasets/services/index.js
```diff
@@ -1,19 +1,11 @@
-export * from './addDatasetStorageLocation';
 export * from './createDataset';
```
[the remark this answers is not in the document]
itsmo-amzn: Sure, will bring this up during the next standup.

frontend/src/modules/Shared/index.js
nikpodsh: It works for me. Let's leave it there for a while. But we should not forget to return back to it when we modularize dataset shares.

nikpodsh (approved):

PR #591, opened 2023-07-20
Fix wrong update of externalId for pivotRole

### LINKED ISSUE #589: Replaced externalId in V1.6.0
### Describe the bug

When upgrading to V1.6.0 we noticed that our environments go errors such as "roleXYZ cannot assume pivotRole". The issue that we noticed is that the externalId was replaced by a new value, thus the manually created pivotRoles will stop being assumable by the central account.

The root cause is that the `_get_external_id_value` function that runs to create the parameter stack in the backend infra is synthesized in the tooling account. Instead of verifying this parameter in the infra account it is wrongly searching for it in the tooling account.

### How to Reproduce

Upgrade to v1.6. with manual pivot roles

### Expected behavior

_No response_

### Your project

_No response_

### Screenshots

_No response_

### OS

n/a

### Python version

n/a

### AWS data.all version

v1.6.0

### Additional context

_No response_

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
Fixes #589 by: 
- using CDK constructs to check the existence of an externalID in Secrets Manager
- using boto3 calls using the CDK look up role in the deployment accounts to find an externalID in the Systems Manager Parameter Store

## Manual operations needed ONLY if upgrading. Fresh deployments are unaffected

In the first run the CodePipeline will fail in the `CDK Synth` stage if no additional changes are done:
```
botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the AssumeRole operation: User: arn:aws:sts::111111111111:assumed-role/SOME ROLE/... is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::222222222222:role/cdk-hnb659fds-lookup-role-22222222222-eu-west-1
```
CodeBuild needs additional permissions to assume the IAM role in the CDK Synth stage. Since we cannot update this CodeBuild stage without running it, the permissions need to be added manually.

### Upgrading from V1.6.0 to v1.6.1
The role that we need to update is a role named `<PREFIX>-<GITBRANCH>-codebuild-baseline-role`. It will say it in the error message in the CodeBuild logs

 1. Go to the IAM role (`<PREFIX>-<GITBRANCH>-codebuild-baseline-role`) and click on `Add permissions` > `Create inline policy`
image
2. Update the policy, use the JSON and copy the policy below:
image

The policy of the Codebuild execution role need to include the following:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": "sts:AssumeRole",
            "Resource": "arn:aws:iam::*:role/cdk-hnb659fds-lookup-role*"
        }
    ]
}
```
After the pipeline has successfully run, go back to the IAM role and remove the manually added policy. The policy is now added as part of infrastructure as code.
image

### Upgrading from <V1.6.0 to v1.6.1
The error points at a different role some. A role created by CDK that looks like the following in the CodeBuild logs:
```
botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the AssumeRole operation: User: arn:aws:sts:::111111111111:assumed-role/dataall-sbx8-cicd-stack-dataallsbx8cdkpipelinePipe-HMXY7D9OX4FM/AWSCodeBuild-30c50765-4529-4d20-99ce-88f82139a82c is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::22222222222:role/cdk-hnb659fds-lookup-role-22222222222-eu-west-1
```
We find the role and update it as we explained in the "Upgrading from V1.6.0 to v1.6.1" section. 
image

Once that is done, retry the CodeBuild Synth stage. In this case you do NOT need to cleanup the manually added policies as this role will be deleted.

## Tests
[X] Tested by merging this branch into a deployment initially in V1.6.0 -> SSM parameter value unmodified
[X] Tested by merging this branch into a deployment initially in V1.5.6 -> Secret value copied to SSM parameter unmodified
[X] Tested by deploying this branch on a fresh account

### Relates
- #589

### DIFF
--- a/deploy/stacks/param_store_stack.py
+++ b/deploy/stacks/param_store_stack.py
@@ -147,7 +147,11 @@ def _get_external_id_value(envname, account_id, region):
         return parameter_value
     except:
         try:
-            secret_value = SecretValue.secrets_manager(secret_id).unsafe_unwrap()
+            secrets_client = session.client('secretsmanager', region_name=region)
+            if secrets_client.describe_secret(SecretId=secret_id):
+                secret_value = SecretValue.secrets_manager(secret_id).unsafe_unwrap()
+            else:
+                raise Exception
             return secret_value
         except:
             return False
--- a/deploy/stacks/param_store_stack.py
+++ b/deploy/stacks/param_store_stack.py
@@ -1,6 +1,5 @@
 import random
 import string
-import os
 
 import boto3
 from aws_cdk import (
@@ -105,9 +104,6 @@ class ParamStoreStack(pyNestedClass):
             string_value=str(pivot_role_name),
             description=f"Stores dataall pivot role name for environment {envname}",
         )
-
-        # account_id = os.environ.get("CDK_DEPLOY_ACCOUNT", os.environ["CDK_DEFAULT_ACCOUNT"])
-        # region = os.environ.get("CDK_DEPLOY_REGION", os.environ["CDK_DEFAULT_REGION"])
 
         existing_external_id = _get_external_id_value(envname=envname, account_id=self.account, region=self.region)
         external_id_value = existing_external_id if existing_external_id else _generate_external_id()
--- a/deploy/stacks/param_store_stack.py
+++ b/deploy/stacks/param_store_stack.py
@@ -104,7 +104,7 @@ class ParamStoreStack(pyNestedClass):
         )
         try:
             parameter_path = f"/dataall/{envname}/pivotRole/externalId"
-            external_id_value = aws_ssm.StringParameter.value_for_string_paramete(self, parameter_path)
+            external_id_value = aws_ssm.StringParameter.value_for_string_parameter(self, parameter_path)
         except:
             secret_id = f"dataall-externalId-{envname}"
             try:
--- a/deploy/stacks/param_store_stack.py
+++ b/deploy/stacks/param_store_stack.py
@@ -1,5 +1,6 @@
 import random
 import string
+import os
 
 import boto3
 from aws_cdk import (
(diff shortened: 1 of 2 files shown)


dlpzx:
We have 2 approaches, I have tried them both.
## 1. use boto3 calls to the infra accounts from the CICD pipeline 
It is the current code in the PR. It assumes the CDK look-up-role in the infra accounts. 
- CONS: CodeBuild needs additional permissions to assume the IAM role.
- PROS: It works well

## 2. use CDK look-up resources from the infra account while the stack is deployed
We tried several CDK constructs for this. 
- CONS: Either they did not work or they needed CDK synth with infra account credentials. --> which I am not sure how is going to work with a fresh deployment.
- CONS: bad handling of SSM and Secrets together - exceptions in CDK are not as usable as in normal python
- PROS: no additional permissions granted


In conclusion, because of how unstable the solution with CDK constructs is (I am specially worried about the user experience for fresh deployments) I am more inclined to use boto3. What do you think? @noah-paige

noah-paige:
I agree with opting for boto3 at this point since the CDK Construct methods seem to not work as we would like

The main issue I see with Option 1 is in my testing the default cdk look up role does not have the required permissions to get the secret either

dlpzx:
I checked the permission on the role and it has ` "secretsmanager:Describe*",` damm. Let me review that again

noah-paige:
Yeah but it has `kms:Decrypt` Deny and does not have `secretsmanager:GetSecretValue`

noah-paige:
What if we use boto3 for SSM and CDK Constructs to get secret... I am testing this approach and I think it works as expected 

Will still have to add permissions to assume role in `Synth` step

deploy/stacks/param_store_stack.py
```diff
@@ -128,7 +146,10 @@ def _get_external_id_value(envname, region):
     except:
         try:
             secrets_client = session.client('secretsmanager', region_name=region)
-            secret_value = secrets_client.get_secret_value(SecretId=secret_id)['SecretString']
+            if secrets_client.describe_secret(SecretId=secret_id):
+                secret_value = SecretValue.secrets_manager(secret_id).unsafe_unwrap()
```
dbalintx, line 150: minor: `secret_value` could be returned directly here

dlpzx: I think it was already resolved

dbalintx (approved):

PR #596, opened 2023-07-24
[Proposal] A new GraphQL API

nikpodsh:
I'd like to propose a new GraphQL API for data.all
We have the API for GraphQL in the data.all that defines the schema. This API is cumbersome, have a lot of boilerplate and hard to use in my very humble opinion. For example, we define the input type this way:
```
UpdateWorksheetInput = gql.InputType(
    name='UpdateWorksheetInput',
    arguments=[
        gql.Argument(name='label', type=gql.String),
        gql.Argument(name='description', type=gql.String),
        gql.Argument(name='tags', type=gql.ArrayType(gql.String)),
        gql.Argument(name='sqlBody', type=gql.String),
        gql.Argument(name='chartConfig', type=gql.Ref('WorksheetChartConfigInput')),
    ],
)
```
1) So we have chartConfig that has type string, but in the code we don't utilize it:
```
def update_worksheet(
    context: Context, source, worksheetUri: str = None, input: dict = None
):
    here code
```
In the method above, we use input as a json dict that we pass across whole hierarchy of methods. We don't utilize the types and do the double work to get the params as a key of dictionary.
2) If we want to change the param name we should not forget to change every key for this param when we retrieve the data. This makes this approach very error-prone.
3) We have to duplicate the code. For example, we have duplication for every Enum in api (GraphQLEnumMapper) and DB. 
4) No way to define default variables. We do it in the code like this `description=data.get('description', 'No description provided')`
5) Hard to validate. We should extract every value and check it in the params and if we don't have the optional value we should put the default. It's one of the reason that validation happens not once the request has come but when we are going to use the parameters (after establish db session and other)
6) The schema is not reusable. For example, the filter is duplicated everywhere, but it can be reused:
```
WorksheetFilter = gql.InputType(
    name='WorksheetFilter',
    arguments=[
        gql.Argument(name='term', type=gql.String),
        gql.Argument(name='page', type=gql.Integer),
        gql.Argument(name='pageSize', type=gql.Integer),
    ],
)
```
\+ You can't define a class hierarchy like `Resource -> Notebook -> SagemakerNotebook`
7) The references to other objects are strings: like `gql.Ref('Environment')`. To delete or check that a object is used, we have to full search the whole project. Since the same names are used in many places it's hard to track the real usage.
8) Allocating resources and requesting the same stuff over and over again. For instance, the following object will create a db session 5 times and request environment a few times, though it could do it only once.
```
Dashboard = gql.ObjectType(
    name='Dashboard',
    fields=[
        gql.Field('dashboardUri', type=gql.ID),
        gql.Field('name', type=gql.String),
        gql.Field('label', type=gql.String),
        gql.Field('description', type=gql.String),
        gql.Field('DashboardId', type=gql.String),
        gql.Field('tags', type=gql.ArrayType(gql.String)),
        gql.Field('created', type=gql.String),
        gql.Field('updated', type=gql.String),
        gql.Field('owner', type=gql.String),
        gql.Field('SamlGroupName', type=gql.String),
        gql.Field(
            'organization',
            type=gql.Ref('Organization'),
            resolver=get_dashboard_organization,
        ),
        gql.Field(
            'environment',
            type=gql.Ref('Environment'),
            resolver=resolve_environment,
        ),
        gql.Field(
            'userRoleForDashboard',
            type=DashboardRole.toGraphQLEnum(),
            resolver=resolve_user_role,
        ),
        gql.Field(
            name='terms',
            type=gql.Ref('TermSearchResult'),
            resolver=resolve_glossary_terms,
        ),
        gql.Field(
            'upvotes',
            type=gql.Integer,
            resolver=resolve_upvotes,
        ),
    ],
)
```

&nbsp;
&nbsp;
&nbsp;

### A new way
There is a new way to create schema. The new schema resolves all the mentioned problems. 
To create a mutation:
```
@api_mutation(name="requestDashboardShare")
def request_dashboard_share(principalId: str, dashboardUri: str) -> DashboardShareDto:
    return DashboardShareService.request_dashboard_share(uri=dashboardUri, principal_id=principalId)
```
To create a query:
```
@api_query(name="listDashboardShares")
def list_dashboard_shares(dashboardUri: str, filter: DashboardShareFilter) -> DashboardShareSearchResults:
    return DashboardShareService.list_dashboard_shares(uri=dashboardUri, data=filter)
```
To define the input type:
```
@api_input("ImportDashboardInput")
@dataclass
class ImportDashboardInput(gql.InputType):
    label: str
    environmentUri: str
    dashboardId: str
    SamlGroupName: str
    description = "No description provided"
    tags: [str] = field(default_factory=list)
    terms: [str] = field(default_factory=list)
```
To define object:
```
@api_object("DashboardShare")
@dataclass
class DashboardShareDto(gql.ObjectType):
    shareUri: str
    dashboardUri: str
    name: str
    label: str
    SamlGroupName: str
    status: str
    owner: str
    tags: str
    created: str
    updated: str
```
All those are the same as previous:
```
listDashboardShares = gql.QueryField(
    name='listDashboardShares',
    args=[
        gql.Argument(name='dashboardUri', type=gql.NonNullableType(gql.String)),
        gql.Argument(name='filter', type=gql.Ref('DashboardShareFilter')),
    ],
    resolver=list_dashboard_shares,
    type=gql.Ref('DashboardShareSearchResults'),
)

requestDashboardShare = gql.MutationField(
    name='requestDashboardShare',
    type=gql.Ref('DashboardShare'),
    args=[
        gql.Argument(name='principalId', type=gql.NonNullableType(gql.String)),
        gql.Argument(name='dashboardUri', type=gql.NonNullableType(gql.String)),
    ],
    resolver=request_dashboard_share,
)


ImportDashboardInput = gql.InputType(
    name='ImportDashboardInput',
    arguments=[
        gql.Argument(name='label', type=gql.NonNullableType(gql.String)),
        gql.Argument(name='environmentUri', type=gql.NonNullableType(gql.String)),
        gql.Argument(name='description', type=gql.String),
        gql.Argument(name='SamlGroupName', type=gql.String),
        gql.Argument(name='tags', type=gql.ArrayType(gql.String)),
        gql.Argument(name='dashboardId', type=gql.NonNullableType(gql.String)),
        gql.Argument(name='terms', type=gql.ArrayType(gql.String)),
    ],
)

DashboardShare = gql.ObjectType(
    name='DashboardShare',
    fields=[
        gql.Field('shareUri', type=gql.ID),
        gql.Field('dashboardUri', type=gql.ID),
        gql.Field('name', type=gql.String),
        gql.Field('label', type=gql.String),
        gql.Field('SamlGroupName', type=gql.String),
        gql.Field('status', type=gql.String),
        gql.Field('owner', type=gql.String),
        gql.Field('tags', type=gql.ArrayType(gql.String)),
        gql.Field('created', type=gql.String),
        gql.Field('updated', type=gql.String),
    ],
)
```

&nbsp;
&nbsp;
&nbsp;

It's a proposal so you don't need to review the entire PR, just a schema of dashboards, how it has changed. The rest can be skipped for now. 
```
backend/dataall/modules/dashboards/api/*
```

This PR is not final, but the changes are already working and define the schema the same way as previously. However, there are a few edge cases that we need to deal with and test to write. I haven't done with them because they require time and it's just a proposal for now. 

If you have any question how new schema resolves the aforementioned problem don't hesitate to ask :) 

There are a few very small nuances for migration from old API to the one that we should keep in mind, mostly related to default values. I can create a doc how to migrate to a new schema if we want to accept this change

**This PR is out of the scope of the modularization.**


backend/dataall/base/api/graphql_api.py
```diff
@@ -0,0 +1,183 @@
...
+    def decorator(f):
+        fn, fn_decorator = process_func(f)
+
+        hints = get_type_hints(fn)
+        signature = inspect.signature(fn)
+        num_args = len(signature.parameters)
+
+        if "return" not in hints:
+            raise ValueError(f"Function {fn.__name__} doesn't specify the return type")
+
```
dlpzx, line 84: Ahh I see what you are doing. You are using hints to get the parameters that include hints and then we compare this with the signature parameters to check if everything contains a type hint! + check that return also includes hints

backend/dataall/base/api/graphql_api.py
```diff
@@ -0,0 +1,183 @@
...
+        field(
+            name=name,
+            args=gql_args if gql_args else None,
+            type=return_type,
+            resolver=fn_decorator(decorated),
+            api_version=2,
+        )
+
+        return fn_decorator(decorated)
```
dlpzx, line 129: Why do we return the resolver instead of returning the whole Query/Mutation Field?

backend/dataall/base/api/graphql_api.py
```diff
@@ -0,0 +1,183 @@
...
+    if name in _object_names:
+        raise ValueError(f"The object {name} already exists in the schema")
+    _object_names.add(name)
+
+    def class_decorator(cls):
+        _objects[cls] = name
+
+        hints = get_type_hints(cls)
+        gql_args = []
+
```
dlpzx, line 143: Since input and objects are dataclasses we can assume that their fields have type annotations, but we still need to check the types

backend/dataall/modules/dashboards/api/dashboard_schema.py
```diff
@@ -0,0 +1,109 @@
...
+    label: str
+    description: str
+    DashboardId: str
+    tags: [str]
+    created: str
+    updated: str
+    owner: str
+    SamlGroupName: str
+    organization: gql.Ref('Organization') = None
+    environment: gql.Ref('Environment') = None
```
dlpzx, line 25: I don't understand how this is replacing the resolve_environment function

dlpzx:
Hi @nikpodsh I like this proposal. For me it is way easier to understand it in comparison with the previous, plus the benefits that you listed. I left 2 doubts on 2 things that I did not get completely

backend/dataall/base/api/graphql_api.py
nikpodsh: Yes. Since GraphQL schema has to have type and I wanted to utilize this fact in the code. 
All functions that defines the schema should be typed and types must be scalars (bools, strs, ints, floats) or custom object.

backend/dataall/base/api/graphql_api.py
nikpodsh: It's needed for python. The query/mutation field will be once created inside the decorator, but we need the `fn_decorator(decorated)` to call the api function properly

backend/dataall/base/api/graphql_api.py
nikpodsh: Yes. The input type and the object fields should also have types. If it doesn't have a type/hint we don't know how to map it to the gql type

backend/dataall/modules/dashboards/api/dashboard_schema.py
nikpodsh: This is one of the tricky stuff I have mentioned above
It's not a replacement for `resolve_environment`. It just says that `DashboardDto` (change the name for it to avoid collision with `models.Dashboard`) has type with the reference to environment. **= None** means that the value is optional and isn't supplied all the time (can vary from method to method)

Why so? `DashboardDto` returned from a couple of functions: `import_dashboard`, `update_dashboard`, and `get_dashboard`. Only `get_dashboard` returns all the values. 
If you take a look at the `get_dashboard` you will see that `DashboardDto` is created with all information. It has dashboard, upvotes, terms, organization, env, and role.
organization and environment use the same type as before: `gql.Ref('Organization')` and `gql.Ref('Environment')` accordingly. We are resolving environment in the `get_dashboard` function
If we remove the default value for environment: `gql.Ref('Environment')` the functions `import_dashboard` and `update_dashboard` fails since they will try to return incomplete object.


The PR just mimics the previous schema, but it can be improved further since the objects and input types can be extended. I will demonstrate it here and update the PR to reflect this:
```
@api_object("Dashboard")
@dataclass
class DashboardView(gql.ObjectType): (renamed from DashboardDto)
    dashboardUri: gql.ID
    name: str
    label: str
    description: str
    DashboardId: str
    tags: [str]
    created: str
    updated: str
    owner: str


@api_object("DashboardOverview")
@dataclass
class DashboardOverview(DashboardView):  <------ Here we are extending the schema of DashboardView obj
    organization: gql.Ref('Organization')
    environment: gql.Ref('Environment') 
    userRoleForDashboard: DashboardRole
    terms: [str] 
    upvotes: int

    def __init__(
        self,
        dashboard: Dashboard,
        organization: Organization,
        role: DashboardRole,
        environment: Environment,
        terms,
        upvotes: int
    ):
        for attr in Dashboard.__dict__:
            if not attr.startswith('_'):
                setattr(self, attr, getattr(dashboard, attr))

        self.environment = environment
        self.organization = organization
        self.role = role
        self.upvotes = upvotes
        self.terms = terms
        self.userRoleForDashboard = role

```
`DashboardView` can be returned from `import_dashboard` and `update_dashboard` while `DashboardOverview` from `get_dashboard`


If all objects and input types are migrated we can remove `gql.Ref('Organization')` and replace it with `OrganizationView` or something like that

dlpzx:
@jorgeig-space should we consider this changes for V2.0.0?

dlpzx:
@nikpodsh can we close this? @noah-paige

dlpzx:
Closed in favor of

PR #608, opened 2023-07-30
Replace PostgreSQL 10 by YugabyteDB

FranckPachot:
In case you are interested by it, I tested with YugabyteDB instead of PostgreSQL. It is compatible and has the advantage to scale out in a cloud environment (plus high availability, no need to vacuum...). Here is the configuration for it. It adds the build of docker image to be a drop-in replacement. If you want to add it but keep PostgreSQL as the default, simply do not change the [docker-compose.yaml](https://github.com/FranckPachot/aws-dataall/blob/8f60710c2cd65b7f3bcca62fb052879c7b9bfa11/docker-compose.yaml#L58)

> By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

YugabyteDB is Apache 2.0

### DIFF
--- a/compose/yugabytedb/Dockerfile
+++ b/compose/yugabytedb/Dockerfile
@@ -1,0 +1,9 @@
+FROM docker.io/yugabytedb/yugabyte:latest
+COPY entrypoint.sh /usr/local/bin/
+COPY healthcheck.sh /usr/local/bin/
+COPY config.flags /tmp/
+RUN chmod a+x /usr/local/bin/entrypoint.sh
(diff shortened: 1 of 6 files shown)


dlpzx:
Hi @FranckPachot sorry for the very late response. We have considered your PR and for the moment we will continue using Postgresql as it mimics what we deploy in RDS. If in the future we explore other databases we will come back to this PR. Thank you so much for the interest. Please do not hesitate to make any other contribution, we will not take as much to answer as this time.

Issue #610, opened 2023-07-31
Pivot role modularization


dlpzx:
### Problem statement
As part of v2.0.0, different functionalities of data.all can be configured, enabled or disabled as "modules". When creating new modules or extracting the existing ones we would want to add or remove certain IAM policies from the pivot role that because they are now needed or now longer needed.

### Solution
- I would like a solution where we can define the set of IAM policies that the pivot Role needs to perform SDK calls in AWS that are particular to a module. This set of policies will only be added to the pivotRole YAML and CDK files if the module is enabled.
- Since the policies refer to AWS calls, this implementation will be part of the `aws` package of each module + core modules
- Directly inside the module code we will configure if a particular policy is needed for a feature of the module.

dlpzx:
Completed!

PR #616, opened 2023-08-02
modularization: pivot role policies

dlpzx:
### Feature or Bugfix
- Feature and refactoring

### Detail

Implement a class to add IAM Policy statements to the pivotRole policies from the modules. The additional pivot role policies  needed for each module can be defined inside the module code as in the following picture:
image

Policies that are needed independently from the modules are defined in `dataall/core/environment/cdk/pivot_role_core_policies`.

I moved all permissions, including those for datasets and data_sharing inside their modules. 

In addition:
- Splitted MLStudio and Notebooks pivot role permissions
- Renamed policies added to the environment team roles and moved them to `core/environment/cdk`

### Questions
- Some modules share permissions so there is some code-duplication (around 10 lines). In my opinion it is minimal, and it is not worth mixing module permissions, but I wanted to bring it up.
- With this PR we can only update based on modules the `dataallPivotRole-cdk` that is created as part of the environment stack when auto-create is set to true. Should we implement something for manually created pivot roles? These are roles that are manually deployed in the environment account and cannot be handled by data.all.

### Relates
- #610 


dlpzx:
@nikpodsh I need to re-test the latest changes, but overall it is done

backend/dataall/modules/datapipelines/cdk/__init__.py
```diff
@@ -1,11 +1,13 @@
...
 __all__ = [
     'datapipelines_cdk_cli_wrapper_extension',
     'datapipelines_cdk_pipeline',
-    'datapipelines_stepfunctions_policy',
-    'datapipelines_lambda_policy',
-    'datapipelines_codestar_policy',
-    'datapipelines_pipeline'
+    'env_role_datapipelines_stepfunctions_policy',
+    'env_role_datapipelines_lambda_policy',
+    'env_role_datapipelines_policy',
```
nikpodsh, line 10: There is no `env_role_datapipelines_policy` in the directory. The file should be adjusted

backend/dataall/core/environment/cdk/pivot_role_core_policies/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    ram,
+    s3,
+    sns,
+    sqs,
+    ssm,
+    sts
+)
+
+__all__ = [
+    "athena",
```
nikpodsh, line 21: Many files were migrated to env_role_core_policies

nikpodsh:
We need to fix linting and tests :)

nikpodsh:
I like the approach. Good job, @dlpzx! 👍

dlpzx:
@nikpodsh I have tested the following in my local deployment:

- [X] creation of environment with auto-created pivot role - to verify correct deployment. I checked that it includes all permissions of all modules enabled.
- [ ] creation of dataset - to verify that the custom resources work correctly
- [ ] creation of notebook/start/stop/delete - because I splitted ML Studio and Notebooks permissions
- [ ] creation of ML Studio user/start session/delete - because I splitted ML Studio and Notebooks permissions

backend/dataall/core/environment/cdk/env_role_core_policies/athena.py
nikpodsh, line 1: Is athena related to env? I thought we used it only for worksheets

dlpzx: Kind of, it is used in Worksheets and in the Preview functionality of Datasets. In addition, as part of the environment stack we deploy an Athena workgroup per team. Because of this last reason, it makes sense that any team role can access the team-athena workgroup independently from Worksheets or Datasets.

An alternative would be to extract the workgroup creation as an EnvironmentExtension and move the athena permissions to the module. But then the question is: should we add it as part of the Datasets module or the Worksheets module?

dlpzx: I also noticed that the environment custom resources are part of the environment stack, while they are particular to the Dataset module. Do you think that we should open another PR for this?

nikpodsh: >  it is used in Worksheets and in the Preview functionality of Datasets

Got it! It's too versatile, let's keep it in env.
> I also noticed that the environment custom resources are part of the environment stack, while they are particular to the Dataset module.

I have also noticed it while extracting the glue profiler. With the glue profiler, I was sure that it was used only for datasets , but for the rest - no. So, I decided to leave it in env. 
Answering your questions: yes, we should if it's dataset-only functionality :)

nikpodsh (approved):

backend/dataall/core/environment/cdk/env_role_core_policies/athena.py
dlpzx: Then, should we merge this PR and I will start working on another one for dataset extensions only?

PR #621, opened 2023-08-02
Frontend Modularization: Tables

maryamkhidir:
### Feature or Bugfix
- Refactoring

### Detail
- Tables folder modularization into views, components and services
- Refactored relative imports used in the module to use absolute paths
- Removed unused component
- Updated routes to reflect new file paths

### Relates


frontend/src/modules/Tables/views/TableView.js
```diff
@@ -27,12 +27,15 @@ import {
   useSettings
 } from 'design';
 import { SET_ERROR, useDispatch } from 'globalErrors';
-import { deleteDatasetTable, getDatasetTable, useClient } from 'services';
+import { deleteDatasetTable, useClient } from 'services';
```
dlpzx, line 30: Why is the `deleteDatasetTable` part of the global services? Shouldn't it be part of the table/services

frontend/src/services/graphql/DatasetTable/index.js
```diff
@@ -1,11 +1,3 @@
 export * from './deleteDatasetTable';
-export * from './getDatasetTable';
-export * from './getDatasetTableProfilingRun';
 export * from './getSharedDatasetTables';
 export * from './listDatasetTableColumns';
```
dlpzx, line 3: This is a generic question. For some of the PRs I am seeing that whenever a service is used in 2 different modules it is placed under `services/graphql/moduleName` instead of in `modules/serviceName/services`. In my opinion is a bit weird that services that are very particular to a module and will be customized by customers are placed outside the module directory. Why do we do it like this? (I re-read Amr's planning PR and I think he did not consider this scenario) @nikpodsh wdyt?

nikpodsh: Yes, it might seem a bit awkward, but it's by design. If a service belongs to one module only, it will be placed into that module and imported using relative imports `from ../services`, but if it's shared among a few modules it will stay in global services. Fundamentally, those global services are just common code. 
The modules in the frontend are more granular e.g. Folders, Tables, Datasets are separate modules meanwhile it's only one module in the backend. It allows to have less code per module, but more code in common (luckily only in services). We could consider to create a deeper structure (like we have in BE) that would address most of the imports (yet no all of them 😅). However, it might become a bit more complicated and we would get away from the planned structure.

dlpzx: Thanks for the explanation, then let's stick with the plan. It is not bad that we highlight that a component is used in multiple modules

dlpzx (approved):
lgtm

Issue #618, opened 2023-08-02
cross-region deployment failure in DBMigration stage - prefix list id does not exist


dlpzx:
### Describe the bug

I am trying to deploy the solution in eu-central-1 having the tooling in us-east-1. The pipeline progresses as expected until it reaches the backend deployment stage, where it fails with the following:

<img width="1113" alt="image" src="https://github.com/awslabs/aws-dataall/assets/71252798/e4cb8599-1f4b-4af3-80a7-2ad2b54d6ff5">

### How to Reproduce

```
*P.S. Please do not attach files as it's considered a security risk. Add code snippets directly in the message body as much as possible.*
```

### OS

n/a

### Python version

n/a

### AWS data.all version

v1.6.1

dlpzx:
In other deployments that use the same region for both tooling and infra the issue does not appear. So I tested the output of `aws ec2 describe-prefix-list` in each of the regions:

For `us-east-1`:
```
{
    "PrefixLists": [
        {
            "Cidrs": [
                "16.182.0.0/16",
                "18.34.0.0/19",
                "52.216.0.0/15",
                "54.231.0.0/16",
                "3.5.0.0/19",
                "18.34.232.0/21"
            ],
            "PrefixListId": "pl-63a5400a",
            "PrefixListName": "com.amazonaws.us-east-1.s3"
        },
        {
            "Cidrs": [
                "35.71.70.0/24",
                "3.218.182.0/24",
                "3.218.180.0/23",
                "35.71.68.0/23",
                "52.94.0.0/22",
                "52.119.224.0/20"
            ],
            "PrefixListId": "pl-02cd2c6b",
            "PrefixListName": "com.amazonaws.us-east-1.dynamodb"
        }
    ]
}
```

For `eu-central-1`:
```
{
    "PrefixLists": [
        {
            "Cidrs": [
                "52.219.44.0/22",
                "52.219.208.0/22",
                "3.5.136.0/22",
                "52.219.218.0/24",
                "52.219.140.0/24",
                "16.12.32.0/22",
                "52.219.72.0/22",
                "3.5.134.0/23",
                "16.12.24.0/21",
                "52.219.168.0/22"
            ],
            "PrefixListId": "pl-6ea54007",
            "PrefixListName": "com.amazonaws.eu-central-1.s3"
        },
        {
            "Cidrs": [
                "35.71.105.128/26",
                "52.94.17.0/24",
                "35.71.105.0/25"
            ],
[...]

PR #624, opened 2023-08-03
get prefix list ids for dbmigration for infra region

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
- get the prefix id list for S3 from the infra region. We need the prefix id to connect the dbmigration stage with the S3 bucket containing the migration scripts (add it in the security groups)

### Relates
- #618 


dlpzx:
When testing the fix, the dbmigration nested stack gets created successfully, but another error arises:
image

dlpzx:
I will open a new GitHub issue to tackle the cross-region deployment issue as a whole. This PR will only fix the part of creating the CodeBuild migration stage, other PRs will be needed to allow the cross-region deployment

PR #635, opened 2023-08-07
Disabling UI features within modules

blitzmohit:
Feature or Bugfix

- Feature (follow-up to #538)

Detail

- Ability to turn disable features via a single configuration
- Define feature boundaries for
   - AWS token/console actions on both dataset & env
   - File uploads/actions

Relates

-   https://github.com/awslabs/aws-dataall/issues/473

### DIFF
--- a/frontend/src/views/Datasets/DatasetData.js
+++ b/frontend/src/views/Datasets/DatasetData.js
@@ -3,15 +3,18 @@ import React from 'react';
 import { Box } from '@mui/material';
 import DatasetTables from './DatasetTables';
 import DatasetFolders from './DatasetFolders';
+import config from '../../generated/config.json';
 
 const DatasetData = ({ dataset, isAdmin }) => (
   <Box>
     <Box>
       <DatasetTables dataset={dataset} isAdmin={isAdmin} />
     </Box>
-    <Box sx={{ mt: 3 }}>
-      <DatasetFolders dataset={dataset} isAdmin={isAdmin} />
-    </Box>
+    {config.modules.datasets.features.file_actions === true && (
+      <Box sx={{ mt: 3 }}>
+        <DatasetFolders dataset={dataset} isAdmin={isAdmin} />
+      </Box>
+    )}
   </Box>
 );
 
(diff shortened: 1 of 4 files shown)


frontend/src/views/Datasets/DatasetData.js
```diff
@@ -3,15 +3,18 @@ import React from 'react';
...
 
 const DatasetData = ({ dataset, isAdmin }) => (
   <Box>
     <Box>
       <DatasetTables dataset={dataset} isAdmin={isAdmin} />
     </Box>
-    <Box sx={{ mt: 3 }}>
-      <DatasetFolders dataset={dataset} isAdmin={isAdmin} />
-    </Box>
+    {config.modules.datasets.features.file_actions === true && (
```
itsmo-amzn, line 13: Can you add null check here?
It may cause errors if the configs file isn't defined properly

frontend/src/views/Datasets/DatasetView.js
```diff
@@ -80,11 +81,13 @@ const DatasetView = () => {
         value: 'shares',
         icon: <ShareOutlined fontSize="small" />
       });
-      tabs.push({
-        label: 'Upload',
-        value: 'upload',
-        icon: <Upload fontSize="small" />
-      });
+      if (config.modules.datasets.features.file_uploads === true) {
```
itsmo-amzn, line 84: another `null` check here

frontend/src/views/Datasets/DatasetData.js
itsmo-amzn: I believe we should create a utility function called `bool isModuleEnabled(moduleName){}` that does this logic, instead of repeating the almost the same code everywhere (DRY). With a switch statement for module name, and it's the only one which will import the `configs.json` file. What do you think?

frontend/src/views/Environments/EnvironmentTeams.js
maryamkhidir, line 1: I think you should also put the toggle disable wrapper here to readjust the UI, so that when `env_aws_actions` is disabled, the table head cell is also hidden.

> https://github.com/awslabs/aws-dataall/blob/1c6f1763984522d450f2eed89d53d13a8aafebcc/frontend/src/views/Environments/EnvironmentTeams.js#L448

frontend/src/views/Datasets/DatasetData.js
blitzmohit: that makes sense to me, any modification we want to make in the logic can then be restricted to the new utility function. Will update the PR.

blitzmohit:
Closing in favor of https://github.com/awslabs/aws-dataall/pull/817, which includes all the requested changes

PR #639, opened 2023-08-08
Minor issues in the modularization

nikpodsh:
There is a small PRs that contains changes:
1) Creates Workflow for `modularization-main`. There were many issues with linting and testing, so it should be automated
2) The ML Studio extension shouldn't be deployed if there ML Studios are not enabled for the env
3) Fix of wrong env and dataset in share manager


backend/dataall/modules/mlstudio/cdk/mlstudio_stack.py
```diff
@@ -38,6 +38,11 @@ class SageMakerDomainExtension(EnvironmentStackExtension):
     @staticmethod
     def extent(setup: EnvironmentSetup):
         _environment = setup.environment()
+        with setup.get_engine().scoped_session() as session:
```
dlpzx, line 41: Good catch! thank you!

dlpzx: Maybe as a future improvement idea: could we use Enums or other type for the env parameter values? Right now we need to look for the particular string in the code, both in frontend and backend

nikpodsh: Yes, we should do something about that)

tests/modules/mlstudio/cdk/test_sagemaker_studio_stack.py
```diff
@@ -21,6 +24,7 @@ def __init__(self, scope, id, env, **kwargs):
             **kwargs,
         )
         self._environment = env
+        self._db = db
```
dlpzx, line 27: This is needed for this right?
image

dlpzx (commented):
Looks good, I just have one question in the tests/cdkproxy/conftest. Otherwise it looks good from my side

tests/cdkproxy/conftest.py
```diff
@@ -43,6 +43,10 @@ def env(db, org: Organization) -> Environment:
         )
         session.add(env)
         session.commit()
+        session.add(
```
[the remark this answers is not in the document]
nikpodsh: No, we don't have to. We right way to do it was to define the fixture in `tests/modules/mlstudio/cdk/conftest.py` as you mentioned.
Currently, we have too many references just to take some fixtures from modules to the rest conftest files that it deserve a dedicated PR to restructure fixtures and reduce duplication.

tests/modules/mlstudio/cdk/test_sagemaker_studio_stack.py
nikpodsh: Yep!

dlpzx (approved):

PR #641, opened 2023-08-08
Frontend Modularization: Shares

maryamkhidir:
### Feature or Bugfix
- Refactoring

### Detail
- Modularized Shares folder into components, views and services
- Refactored relative imports used in the module to use absolute paths
- Removed unused component: `ShareInboxTable.js`
- Updated routes to reflect new file paths

### Relates


frontend/src/services/graphql/ShareObject/index.js
```diff
@@ -1,11 +1,2 @@
-export * from './addSharedItem';
-export * from './approveShareObject';
 export * from './createShareObject';
```
dlpzx, line 1: `createShareobject` is placed in the global `services` because it is used in the Catalog module right? (https://github.com/awslabs/aws-dataall/pull/581/files#diff-d9b39fcf99c580c2d427aefaf3a10d31389835997ae58d4380d5755e87c37b97) 
Wouldn't it be better to move the `RequestAccessModal` inside the `Shares` module? I know that it was placed with the Catalog before but if I am a developer working on Shares the first place that I am going to look at is the `Shares` module, what do you think?

frontend/src/services/graphql/ShareObject/index.js
```diff
@@ -1,11 +1,2 @@
...
-export * from './deleteShareObject';
-export * from './getShareObject';
-export * from './getShareRequestsFromMe';
 export * from './getShareRequestsToMe';
```
dlpzx, line 2: Just to clarify, we need to export the `getShareRequestsToMe` to the global services because it is used in the `Datasets` module right?

frontend/src/modules/Shares/views/ShareList.js
```diff
@@ -14,9 +14,9 @@ import { Helmet } from 'react-helmet-async';
 import { FiSend } from 'react-icons/fi';
 import { RiInboxArchiveLine } from 'react-icons/ri';
 import { Link as RouterLink } from 'react-router-dom';
-import { ChevronRightIcon, useSettings } from '../../../../design';
+import { ChevronRightIcon, useSettings } from 'design';
 import { ShareInboxList } from 'modules/Shared';
```
dlpzx, line 18: I understand why we have the Inbox in `modules/Shared` and the outbox in `Shares` but it is a bit confusing as a developer to manage one in one place and the other in another place. I am quite ignorant here, but it is possible or recommended to have them both in `Shares` and import the inbox from `Shares` in the `Dataset` module

dlpzx (approved):
Looks good, just some comments. I am very new to frontend, so I might be totally off

frontend/src/services/graphql/ShareObject/index.js
maryamkhidir: I believe `RequestAccessModal` is being used by Catalog to display search results using access, resourcetypes and some other properties, so it is directly related and being used only by Catalog, therefore it should be in the same module.

maryamkhidir: And going by the functionality of the `createShareObject` service, although it's currently used by the Catalog module, it can also be used by other modules to share an item. That is why it is placed in the global services folder

frontend/src/modules/Shares/views/ShareList.js
maryamkhidir: Following the migration steps [here](https://github.com/awslabs/aws-dataall/issues/398), components or scripts shared among different modules should be under a common directory in this case `src/Shared/Shares`, however, I agree with your point, if we are following logical relationships every related script should be under the same folder `src/Shares`. :)

maryamkhidir: @itsmo-amzn I think for the `Shares` module, we should leave `ShareInboxList` and `ShareInboxListItem` in the `src/modules/Shares/components`, I agree that it is confusing to find `ShareOutboxList` in components and `ShareInboxList` in the global shared folder, what do you think?

itsmo-amzn: Good point! But how would this pan out in the whole modularization scheme?
According to this proposed design, we should keep all shared components/services in their original module directory, and in order to access them from other modules, we would simply use absolute addressing `import { ShareInboxList } from 'modules/Shares'`. I believe we discussed the approach for "Shared" modules before. But maybe we can make an exception for this module?

itsmo-amzn (approved):
Looks good to me.

frontend/src/modules/Shares/views/ShareList.js
maryamkhidir: I agree, an exception can be made for this, another thing to note is that other components that we put in the Shared folder (`Stack`, `Comments`, `KeyValueTags`) doesn't have a view. So technically they cannot be a module, going by that the `Shares` module is different

maryamkhidir: @dlpzx @itsmo-amzn This has been updated with [1945d73](https://github.com/awslabs/aws-dataall/pull/641/commits/1945d739e402799edb175722b604296b020bb962)

PR #643, opened 2023-08-09
All modules can be disabled at a time

nikpodsh:
There was a problem if we wanted to disable all modules at once:
A GraphQL union type cannot be empty. We resolve all unions programmatically depending whether a module is present or not. Some unions consist of module related parts only and, when all modules are disabled, the application fails to start.
To fix the issue we could try not to create a unions only, but the rest schema depends on it. To solve the issue, the following modules have been introduced: `feeds`, `catalog` and `vote`. Those are not standalone modules, but just dependencies, hence if no module using them is present they won't be loaded


backend/dataall/modules/catalog/api/types.py
dlpzx, line 1: I think we are missing the import of `GlossaryRole` used in line 40

backend/dataall/core/__init__.py
dlpzx, line 1: I see that we are not importing the `Activity` package. At first look it seems that we are only creating a table with nothing on it. Should we clean it up?

backend/dataall/base/api/__init__.py
```diff
@@ -32,6 +32,9 @@ def bootstrap():
         enumclass.toGraphQLEnum()
 
     for cls in classes.keys():
+        if not cls.class_instances:  # if there are no instances of cls registered in the app
+            continue
+
```
dlpzx, line 37: This is for the case there is no `QueryType` or `UnionType` or any other gql type to register right?

backend/dataall/base/loader.py
```diff
@@ -185,8 +191,13 @@ def _initialize_modules(modes: Set[ImportMode]):
...
 
 
-def _initialize_module(module):
+def _get_module_name(module):
+    return module[len(_MODULE_PREFIX) + 1:].split(".")[0]  # gets only top level module name
+
+
+def _initialize_module(module: Type[ModuleInterface]):
     module()  # call a constructor for initialization
+    _LOADED_MODULES.add(module.name())
```
dlpzx, line 201: In line 219-230: Why do we have the double `while to_add` block? Why do we need to `to_add` and `new_to_add` variables?

backend/dataall/modules/worksheets/__init__.py
```diff
@@ -16,8 +18,12 @@ class WorksheetApiModuleInterface(ModuleInterface):
     def is_supported(modes):
         return ImportMode.API in modes
 
+    @staticmethod
```
dlpzx, line 21: Why do worksheets depend on Feed?

tests/modules/catalog/test_glossary.py
```diff
@@ -1,8 +1,7 @@
 from datetime import datetime
 
-from dataall.core.catalog.db.glossary_models import GlossaryNode
-from dataall.core.environment.db.models import Environment
-from dataall.core.organizations.db.organization_models import Organization
+from dataall.modules.catalog.db.glossary_models import GlossaryNode
+from tests.api.conftest import *
```
dlpzx, line 4: Don't we need the `Environment` and `Organization` classes in line 9 for example?

dlpzx (commented):
I left some comments. I am very glad Catalog is also a module :) :) It does not need to be immeadiate, but I think customers will like to be able to not-deploy OpenSearch if teh Catalog is inactive

backend/dataall/modules/catalog/api/types.py
nikpodsh: `GlossaryRole` was imported from resolvers by `from .resolvers import *`
One more of the reasons to avoid using `*`. I will fix it

backend/dataall/modules/catalog/__init__.py
```diff
@@ -0,0 +1,16 @@
+from typing import Set
+
+from dataall.base.loader import ModuleInterface, ImportMode
+from dataall.modules.catalog import tasks
+
+
+class CatalogApiModuleInterface(ModuleInterface):
+    """Implements ModuleInterface for dashboard GraphQl lambda"""
```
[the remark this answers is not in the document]
nikpodsh: Sure!

backend/dataall/core/__init__.py
nikpodsh: I agree. I also think that API for activities would be beneficial. 
I haven't changed it since Activity is not queried in the project (the data is saved, but not retrieved). I decided to keep it as it is since it can some sort of audits

backend/dataall/base/api/__init__.py
nikpodsh: Yes, if we don't have `Union` type we should not fail (it can be achieved easily by disabling all modules)

backend/dataall/base/loader.py
nikpodsh: I implemented level by level BFS. It is working this way:
1) On the first iteration of first `while to_add` all active modules in the config file will be loaded.
2) If a module has a dependencies we will add them to `new_to_add` to load later (if they have not been loaded)
3) We switch `to_add` to `new_to_add` and repeat until all submodules are loaded

backend/dataall/core/stacks/api/mutations.py
```diff
@@ -1,5 +1,5 @@
 from dataall.base.api import gql
-from .resolvers import *
```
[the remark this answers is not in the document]
nikpodsh: Yes, we should !

backend/dataall/modules/worksheets/__init__.py
nikpodsh: Honestly, I have scrupulously just modularize it from main as we could post a message for a worksheet. 

But I have just checked it in the frontend and there is no button to post a message for a worksheet (possible, this feature had gone away even before `WorksheetShare` ). Thought, I might have overlooked it
Do you want me to delete this ?

tests/conftest.py
```diff
@@ -17,19 +16,32 @@ def ignore_module_tests_if_not_active():
...
+    exclude_tests = all_module_files - active_module_tests
+
+    # here is a small hack to satisfy both glob and pytest. glob is using os.getcwd() which is root of the project
+    # while using "make test". pytest is using test directory. Here is why we add "tests" prefix for glob and
+    # remove it for pytest
+    prefix_to_remove = f"tests{os.sep}"
+
+    # migrate to remove prefix when runtime > 3.8
+    exclude_tests = [excluded[len(prefix_to_remove):] for excluded in exclude_tests]
+    collect_ignore_glob.extend(exclude_tests)
```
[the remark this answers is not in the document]
nikpodsh: If I understood your question:
It's a pytest feature. When there is a variable `collect_ignore_glob`, pytest doesn't collect the test with accordance with the variable

tests/modules/catalog/test_glossary.py
nikpodsh: They are being taken from `from tests.api.conftest import *` which is bad indeed. However, no worries:  #656 comes to rescue 😄

backend/dataall/modules/worksheets/__init__.py
dlpzx: I would remove it for the moment, just to make it clear that it is unused in worksheets

dlpzx (approved):
I remove `Feed` from the `Worksheets` dependencies

nikpodsh:
@dlpzx thank you for taking care of feeds for the worksheets :)

PR #647, opened 2023-08-09
Move custom resources to Dataset Environment extension

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail
- moved dataset related custom resources to `Dataset` module
- created EnvironmentExtension for the custom resources

### Relates


backend/dataall/modules/datasets/cdk/dataset_custom_resources_extension.py
```diff
@@ -0,0 +1,239 @@
...
+            environment=setup.environment(),
+            group_roles=setup.group_roles,
+            default_role=setup.default_role
+        )
+
+        # Lakeformation default settings custom resource
+        # Set PivotRole as Lake Formation data lake admin
+        entry_point = str(
+            pathlib.PosixPath(os.path.dirname(__file__),
+                              './assets/lakeformationdefaultsettings').resolve()
```
nikpodsh, line 38: In `DatasetGlueProfilerExtension`, the relative path is different. We need to check which one is correct

nikpodsh:
I have only one question. The rest changes look good to me!

backend/dataall/modules/datasets/cdk/dataset_custom_resources_extension.py
dlpzx: I tested the creation of the Dataset with the extensions and it worked fine with this paths. At first I had the same level for both but i think because of the libraries we use they are defined diferently

nikpodsh (approved):

PR #649, opened 2023-08-09
Resolve Dataset Profiling Glue Job

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
-  Specify `SPARK_VERSION` as an environment variable for `pydeequ` before import
- Add IAM Permissions to Dataset IAM Role to Allow for Glue Job logging in CloudWatch
- Add LF Permissions to resolve insufficient permissions error thrown when looking for `default` database

### Relates

### DIFF
--- a/backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
+++ b/backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
@@ -48,6 +48,13 @@ def on_create(event):
         exists = True
     except ClientError as e:
         pass
+    
+    default_db_exists = False
+    try:
+        glue_client.get_database(Name="default")
+        exists = True
+    except ClientError as e:
+        pass
 
     if not exists:
         try:
@@ -63,7 +70,7 @@ def on_create(event):
             raise Exception(f"Could not create Glue Database {props['DatabaseInput']['Name']} in aws://{AWS_ACCOUNT}/{AWS_REGION}, received {str(e)}")
 
     Entries = []
-    for i, role_arn in enumerate(props.get('DatabaseAdministrators')):
+    for i, role_arn in enumerate(props.get('DatabaseAdministrators', [])):
         Entries.append(
             {
                 'Id': str(uuid.uuid4()),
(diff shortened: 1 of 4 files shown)


noah-paige:
To note for this PR - Glue Jobs require the permissions to be able to verify the existence of the `default` database or create the `default` database if it does not exist [AWS Docs](https://docs.aws.amazon.com/glue/latest/dg/data-quality-trouble.html#data-quality-trouble-error-7)

This is needed for Glue Jobs that `Use Data Catalog as the Hive metastore` which is needed for our Glue Profiling Job

In order to accommodate this requirement, the PR adds code to grant `DESCRIBE` LF Permissions to the Dataset Principals (Dataset Role, EnvGroup Role, PivotRole) in the dataset Custom Resource Lambda

If the `default` database does not exist with the above changes in this PR the Glue Profiling Job will still fail as of now...

backend/dataall/cdkproxy/assets/gluedatabasecustomresource/index.py
```diff
@@ -49,6 +49,13 @@ def on_create(event):
     except ClientError as e:
         pass
 
+    default_db_exists = False
+    try:
```
dlpzx, line 53: What happens if the default does not exist?

dlpzx:
So now everytime we create a dataset, the dataset admin role will get permissions to the default database. Permissions to this database are needed to run the profiling job. Do you know why? is it generic for any Glue Job you need permissions to the default database?

dlpzx:
We also need to fix it in `modularization-main`

dlpzx (approved):
And thank you, this is a long-needed fix :)

nikpodsh:
@dlpzx, @noah-paige  I will gladly replicate these changes to mod-main :)

noah-paige:
> So now every time we create a dataset, the dataset admin role will get permissions to the default database. Permissions to this database are needed to run the profiling job. Do you know why? is it generic for any Glue Job you need permissions to the default database?

I know in documentation it calls out that Glue will also try to verify the default DB only if you enable `Use Data Catalog as the Hive metastore`

The permissions required are only `DESCRIBE` on the DB level since all Glue is doing is verifying the `default` DB exists by calling `GetDatabase` API

PR #657, opened 2023-08-15
Refactor module enablement

itsmo-amzn:
### Feature or Bugfix

- Refactoring

### Detail
- Refactor module enablement

### Relates
- #398 

### DIFF
--- a/frontend/src/design/components/layout/DefaultSidebar.js
+++ b/frontend/src/design/components/layout/DefaultSidebar.js
@@ -17,7 +17,7 @@ import { MdShowChart } from 'react-icons/md';
 import { SiJupyter } from 'react-icons/si';
 import { VscBook } from 'react-icons/vsc';
 import { useLocation } from 'react-router-dom';
-import config from '../../../generated/config.json';
+import { ModuleNames, isModuleEnabled } from 'utils';
 import { useSettings } from '../../hooks';
 import { NavSection } from '../NavSection';
 import { Scrollbar } from '../Scrollbar';
@@ -28,7 +28,7 @@ export const DefaultSidebar = ({ openDrawer, onOpenDrawerChange }) => {
       title: 'Catalog',
       path: '/console/catalog',
       icon: <VscBook size={15} />,
-      active: config.modules.datasets.active || config.modules.dashboards.active
+      active: isModuleEnabled(ModuleNames.CATALOG)
     };
 
     const datasetsSection = {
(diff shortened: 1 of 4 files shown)


frontend/src/design/components/layout/DefaultSidebar.js
```diff
@@ -28,7 +28,7 @@ export const DefaultSidebar = ({ openDrawer, onOpenDrawerChange }) => {
       title: 'Catalog',
       path: '/console/catalog',
       icon: <VscBook size={15} />,
-      active: config.modules.datasets.active || config.modules.dashboards.active
+      active: isModuleEnabled(ModuleNames.CATALOG)
     };
 
     const datasetsSection = {
```
noah-paige, line 34: Out of curiosity - why do we not pass isModuleEnabled(...) for Datasets, Shares, and Glossaries?

frontend/src/utils/helpers/moduleUtils.js
```diff
@@ -0,0 +1,38 @@
...
+  GLOSSARIES: 'glossaries',
+  WORKSHEETS: 'worksheets',
+  NOTEBOOKS: 'notebooks',
+  MLSTUDIO: 'mlstudio',
+  PIPELINES: 'datapipelines',
+  DASHBOARDS: 'dashboards'
+};
+
+function isModuleEnabled(module) {
+  if (module === ModuleNames.CATALOG) {
```
maryamkhidir, line 17: This means Catalog can only be enabled if Datasets or Dashboards is enabled, is this the only module that have this conditional dependency? I think, if this can occur for other modules, it will be better to just create a method that handles that generally. That way it is reusable.

frontend/src/utils/helpers/moduleUtils.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable no-restricted-properties */
+import config from '../../generated/config.json';
+
+const ModuleNames = {
```
maryamkhidir, line 4: I like this.

frontend/src/design/components/layout/DefaultSidebar.js
itsmo-amzn: We still need to decide on the behavior for module enablement for `Datasets`, because it would potentially disable other modules as you mention like `Shares`, `Glossaries`,etc. But as for this PR, we're just doing the refactoring (potential bug fixes also)

frontend/src/utils/helpers/moduleUtils.js
itsmo-amzn: This function acts as the single handler for checking module enablement. You can adjust the rules depending on your module's requirements. It acts as a `switch` case for the modules. In our case, only `catalog` has the a complex condition to be enabled, and the default behavior for other modules, is simply checking the `config.json` for `active` status. For future modules, you can adjust the condition by simply inserting an `if` case to check your module, and add your checks accordingly.

noah-paige: I think could make it more flexible allowing to pass a list of module names and do checks on each to see if module should be loaded ... but for now this PR looks good to me and can think on if something like that would help moving forward

maryamkhidir (approved):
LGTM

PR #675, opened 2023-08-18
Fix module dependency loading.

nikpodsh:
The corresponding module interface for a dependency should be loaded locally

- Bugfix

There is a issue with the loading of modules for cdkproxy task. The issue appears since the loading of dependency happens in the module scope, but should be local for each `ModuleInterface`

For instance, when there is `from dataall.modules.catalog import api` in `dataall.modules.datasets.__init__` the Catalog API will be loaded for each dataset module interfaces.

### DIFF
--- a/backend/dataall/modules/vote/__init__.py
+++ b/backend/dataall/modules/vote/__init__.py
@@ -1,7 +1,6 @@
 from typing import Set, List, Type
 
 from dataall.base.loader import ModuleInterface, ImportMode
-from dataall.modules.vote import api
 
 
 class VoteApiModuleInterface(ModuleInterface):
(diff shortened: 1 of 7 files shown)


backend/dataall/modules/vote/__init__.py
```diff
@@ -1,7 +1,6 @@
 from typing import Set, List, Type
 
 from dataall.base.loader import ModuleInterface, ImportMode
-from dataall.modules.catalog import CatalogApiModuleInterface
 from dataall.modules.vote import api
```
dbalintx, line 4: this import is unnecessary as its imported in the MI class

nikpodsh: Agreed!

dbalintx (approved):

PR #679, opened 2023-08-21
Enforce module enablement rules for remaining modules

itsmo-amzn:
### Feature or Bugfix

- Feature

### Detail
- module enablement for `datasets`, `catalog`, and `glossaries`

### Relates
- #398 

### DIFF
--- a/frontend/src/design/components/layout/DefaultSidebar.js
+++ b/frontend/src/design/components/layout/DefaultSidebar.js
@@ -34,19 +34,22 @@ export const DefaultSidebar = ({ openDrawer, onOpenDrawerChange }) => {
     const datasetsSection = {
       title: 'Datasets',
       path: '/console/datasets',
-      icon: <FiPackage size={15} />
+      icon: <FiPackage size={15} />,
+      active: isModuleEnabled(ModuleNames.DATASETS)
     };
 
     const sharesSection = {
       title: 'Shares',
       path: '/console/shares',
-      icon: <ShareOutlined size={15} />
+      icon: <ShareOutlined size={15} />,
+      active: isModuleEnabled(ModuleNames.SHARES)
     };
 
     const glossariesSection = {
       title: 'Glossaries',
       path: '/console/glossaries',
-      icon: <BsIcons.BsTag size={15} />
+      icon: <BsIcons.BsTag size={15} />,
+      active: isModuleEnabled(ModuleNames.GLOSSARIES)
     };
 
     const worksheetsSection = {
--- a/frontend/src/utils/helpers/moduleUtils.js
+++ b/frontend/src/utils/helpers/moduleUtils.js
@@ -14,11 +14,14 @@ const ModuleNames = {
 };
 
 function isModuleEnabled(module) {
-  if (module === ModuleNames.CATALOG) {
+  if (module === ModuleNames.CATALOG || module === ModuleNames.GLOSSARIES) {
     return (
       getModuleActiveStatus(ModuleNames.DATASETS) ||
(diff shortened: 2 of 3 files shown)


frontend/src/utils/helpers/moduleUtils.js
```diff
@@ -14,12 +14,15 @@ const ModuleNames = {
 };
 
 function isModuleEnabled(module) {
-  if (module === ModuleNames.CATALOG) {
```
dlpzx, line 17: This is good for V2.0.0 but maybe we can already propose abstractions to handle dependencies that are more generic, similar to the `depends_on` method in the backend. wdyt?

frontend/src/design/components/layout/DefaultSidebar.js
```diff
@@ -34,19 +34,22 @@ export const DefaultSidebar = ({ openDrawer, onOpenDrawerChange }) => {
...
+      icon: <FiPackage size={15} />,
+      active: isModuleEnabled(ModuleNames.DATASETS)
     };
 
     const sharesSection = {
       title: 'Shares',
       path: '/console/shares',
-      icon: <ShareOutlined size={15} />
+      icon: <ShareOutlined size={15} />,
+      active: isModuleEnabled(ModuleNames.SHARES)
```
dlpzx, line 45: Let me understand this a bit taking Shares as example. We use `isModuleEnabled` with the `Shares` module. But then, we do not check Shares active param in the config file, we directly check the Datasets active parameter. For me is a little misleading, I would expect that we verify the param of shares in config.json

dlpzx (approved):
Approved. I think the functionality works as expected but I would work on better module dependency management post V2.0.0 release

frontend/src/design/components/layout/DefaultSidebar.js
itsmo-amzn: I believe there's no entry for `shares` in the `configs.json` file, and as we agreed before, `shares` should be enabled if and only if `datasets` module is enabled. I understand it's a bit confusing, but this is by far the best approach we can take for module enablement dependencies management in the frontend, because if we define every entry in the `configs.json` file and keep it up to the user to define which module should be enabled, we might end up with conflicting active statuses. wdyt?

frontend/src/utils/helpers/moduleUtils.js
itsmo-amzn: could you please elaborate? I'm not quite familiar with backend.

dlpzx: In this `utils` we are defining that `Shares` depends on `Datasets` and that `Catalog` and `Glossaries` depend on `Datasets` and `Dashboards`. I find it strange that the dependencies of the module are defined in the utils and not in the module itself. For me `isModuleEnabled` should check the dependencies with something like getModuleDependencies(module) and then return getModuleActiveStatus(dependency). An abstraction of the dependencies instead of hardcoding Datasets, Dashboards....

dlpzx: In the backend the depedency is defined in the `__init__` of each module, I am not sure what would be the equivalent in frontend

frontend/src/design/components/layout/DefaultSidebar.js
dlpzx: `shares` and `catalog` can be defined in the `config.json`, glossaries is part of catalog in the backend

frontend/src/utils/helpers/moduleUtils.js
itsmo-amzn: Make perfect sense, I'll open an issue with the proposed suggestion.

PR #689, opened 2023-08-22
Fix auth issue happening on the frontend

nikpodsh:
Fix of auth issue in frontend. It needs additional investigation where exactly (on what line/s) the issue is happening not to delete all good stuff from the `package.json`, but it should be fine as a temporary workaround

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -12,8 +12,8 @@
     "prebuild": "yarn copy-config",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "lint": "eslint --ext js src",
-    "lint:fix": "eslint --fix --ext js src"
+    "lint": "node_modules/.bin/eslint --ext js src",
+    "lint-fix": "node_modules/.bin/eslint --fix --ext js src"
   },
   "dependencies": {
     "@apollo/client": "^3.3.19",
@@ -22,11 +22,11 @@
     "@emotion/styled": "^11.8.1",
     "@monaco-editor/react": "^4.3.1",
     "@mui/icons-material": "^5.5.1",
+    "@mui/x-date-pickers": "^5.0.0",
     "@mui/lab": "^5.0.0-alpha.74",
     "@mui/material": "^5.5.2",
     "@mui/styles": "^5.5.1",
     "@mui/x-data-grid": "^5.7.0",
-    "@mui/x-date-pickers": "^5.0.0",
     "@reduxjs/toolkit": "^1.8.0",
     "@testing-library/jest-dom": "^5.16.2",
     "@testing-library/react": "^12.1.4",
@@ -64,7 +64,7 @@
     "web-vitals": "^2.1.4",
     "yup": "^0.32.11"
   },
-  "overrides": {
+  "overrides" : {
     "@appbaseio/reactivesearch": {
       "react-redux": "^7.2.6"
     }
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -74,84 +74,13 @@
...
-      "import"
-    ],
     "extends": [
       "react-app",
-      "react-app/jest",
-      "plugin:prettier/recommended",
-      "plugin:import/recommended"
-    ],
-    "settings": {
-      "import/resolver": {
```
itsmo-amzn, line 96: This might cause all absolute imports to fail

nikpodsh: It shouldn't. `settings` are declared for `eslintConfig`, so the only thing that can fail is linting. It worked fine in pipeline. I assume, that it's because we returned back to original linting in this PR

frontend/package.json
```diff
@@ -12,8 +12,8 @@
     "prebuild": "yarn copy-config",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "lint": "eslint --ext js src",
-    "lint:fix": "eslint --fix --ext js src"
+    "lint": "node_modules/.bin/eslint --ext js src",
```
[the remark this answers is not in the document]
nikpodsh: I can only guess, since I haven't tested it without the prefix: 
I assume that `eslint` is not in the PATH. It was added to path in mod-main by one of those:
```
    "eslint-config-prettier": "^8.8.0",
    "eslint-plugin-import": "^2.27.5",
    "eslint-import-resolver-alias": "^1.1.2",
    "eslint-plugin-prettier": "^4.2.1",
```
Though, it's just an assumption.

nikpodsh:
I have updated the workaround and now, it can become a regular PR.
The problem: something was wrong with Amplify. It might have that our yarn.lock was inconsistent with Amplify version, or the amplify version itself has a bug. 
Both, main and modularization had "aws-amplify": "^5.2.6", which basically meant: use any version higher than > 5.2.6 and < 6.0.0. BUT, main used version 5.3.8 in yarn.lock, while modularization used 5.2.6. 
Why so? My best bet: there was a merge conflict between the frontend mod branch and main and not to resolve thousands lines conflicts we regenerated `yarn.lock` which was generated correctly, but we lost the changes from the main and hence, the error. 
I have checked the issue in my account and it worked. If anybody confirm that the issue is working for him/her as well, I will convert this to a normal PR and merge it. If not, I'll drop the last commit

dbalintx:
I can confirm that the fix is working in a non-internet facing deployment

maryamkhidir:
> I have updated the workaround and now, it can become a regular PR. The problem: something was wrong with Amplify. It might have that our yarn.lock was inconsistent with Amplify version, or the amplify version itself has a bug. Both, main and modularization had "aws-amplify": "^5.2.6", which basically meant: use any version higher than > 5.2.6 and < 6.0.0. BUT, main used version 5.3.8 in yarn.lock, while modularization used 5.2.6. Why so? My best bet: there was a merge conflict between the frontend mod branch and main and not to resolve thousands lines conflicts we regenerated `yarn.lock` which was generated correctly, but we lost the changes from the main and hence, the error. I have checked the issue in my account and it worked. If anybody confirm that the issue is working for him/her as well, I will convert this to a normal PR and merge it. If not, I'll drop the last commit

Yeah, I can also confirm this worked for the backward compatibility deployment

noah-paige (approved):
looks good to me

PR #701, opened 2023-08-23
DA v2: Dataset IAM Role Default DB Glue Permissions

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- In the case where the `default` Glue Database does not already exist
  - Glue Script is failing with "DATASET IAM ROLE  is not authorized to perform: glue:CreateDatabase on resource XXX"

- Root Casue
  - The Glue Job needs to verify that the default database exists and if it does not it needs to create the default DB
  - Adding the glue:CreateDatabase permissions on the catalog and only for the default DB resolves the issue

- POTENTIAL ISSUE:
  - In the Custom Resource of Dataset Stack we add permission for the dataset IAM Role to the default DB if it exists

  - In the scenario where default DB does not already exist and we create Dataset1 and Dataset2
  - Then Dataset1 Role creates default DB when running profiling job and the Dataset2 IAM Role never gets proper LF Permissions 
  - On testing - I tried a Dataset Stack update but it determines "No Change Set" on the Cloudformation Stack so the Custom Resource does not re-run to add the required permissions
  - A potential solution is to pass a randomly generated uuid as part of the Custom Resource properties to trigger a run of the Lambda each time? 

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/datasets/cdk/assets/gluedatabasecustomresource/index.py
+++ b/backend/dataall/modules/datasets/cdk/assets/gluedatabasecustomresource/index.py
@@ -49,13 +49,6 @@ def on_create(event):
     except ClientError as e:
         pass
 
-    default_db_exists = False
-    try:
-        glue_client.get_database(Name="default")
-        default_db_exists = True
-    except ClientError as e:
-        pass
-
     if not exists:
         try:
             db_input = props.get('DatabaseInput').copy()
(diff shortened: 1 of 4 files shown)


dlpzx:
Hi @noah-paige this issue keeps coming back ahhahaha the potential issue that you pointed out led me to the conclusion that we should treat the default Glue database as a default setting of the environment account. So what I propose is to move the creation of the default database to the `backend/dataall/modules/datasets/cdk/assets/lakeformationdefaultsettings/index.py` this way we ensure that at dataset creation the default Glue database exists and we can grant Lake Formation permissions in the `gluedatabasecustomresource`.  Plus, I would like to restrict database creation action to the dataset role, it is just a role to read and write to the dataset resources. Managing resources outside of its own dataset should be denied.

I am testing this changes in my AWS deployment at the moment, I will update the PR branch if you agree and if the tests succeed.

noah-paige:
I agree with the proposed changes - will try out the code changes in my data.all deployment!

noah-paige:
Tested out the above and all the Glue Profiling Jobs work now! PR looks good to me

nikpodsh:
Tested it in my account. The fix is working

PR #703, opened 2023-08-24
DA v2: Worksheets View need default export

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Worksheet view is not loading with error:
```
Element type is invalid. Received a promise that resolves to: [missing argument]. Lazy element type must resolve to a class or function.[missing argument]
```

- From React [Docs](https://legacy.reactjs.org/docs/code-splitting.html#reactlazy):
```
React.lazy takes a function that must call a dynamic import(). This must return a Promise which resolves to a module with a default export containing a React component.
```

### Security
NA 
```

  - Is the input sanitized?
```

### DIFF
--- a/frontend/src/modules/Worksheets/views/WorksheetView.js
+++ b/frontend/src/modules/Worksheets/views/WorksheetView.js
@@ -50,7 +50,7 @@ import {
   WorksheetResult
 } from '../components';
 
-export const WorksheetView = () => {
+const WorksheetView = () => {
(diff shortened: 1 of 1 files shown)


noah-paige:
Tested in AWS Deployment - Worksheet View loads and able to run queries against data 

Opening PR for review

nikpodsh (approved):
Worked for me as well!

PR #705, opened 2023-08-25
DA v2: fix issues with modularization pipelines

dlpzx:
### Feature or Bugfix
- Refactoring
- datapipelines module is not following design guidelines. `api` =validation of input, `services`=check permissions and business logic
- there is a lot of unused code

### Detail
- move business logic to `services`
- implement validation of parameters in `api`
- clean up unused code in `datapipelines`: cat, ls and the `aws` clients

### Security

N/A

  - Is the input sanitized?


backend/dataall/modules/datapipelines/db/datapipelines_repositories.py
```diff
@@ -25,11 +32,52 @@ def count_resources(self, session, environment, group_uri) -> int:
...
+    def create_pipeline(
+            session,
+            username: str,
+            admin_group: str,
+            uri: str,
+            data: dict = None,
+    ) -> DataPipeline:
+        environment = EnvironmentService.get_environment_by_uri(session, uri)
+
+        pipeline: DataPipeline = DataPipeline(
```
dbalintx, line 44: I think instantiation of the object should happen in the service

dlpzx: I was also doubtful on this one because I found examples of us creating the RDS object in the services and in db. The logic I followed was: if I am a developer and I want to see how a DataPipeline was created in the database I will go to the db package. But I am not completely convinced by one or other approach. Why do you think it should be in `services`?

dlpzx:
Testing is still needed:
[ ] testing locally
[ ] testing in AWS

backend/dataall/modules/datapipelines/db/datapipelines_repositories.py
dbalintx: I think the mixed creation of objects in services and db happens due to differences around the object creation (eg. in some cases multiple things are done besides just creating the object). In this case, besides creating the object nothing else really happens, so I agree that its nicer to have it in db - I just remembered the pattern that object creation was done in services in many places.
Best would be to have all the object creations in either db or services, but its fine from my side to do this in a followup, also after the release.

dlpzx:
## Fix local development with pipelines
I just made some edits on the dev dockerfile:
image

dbalintx:
Copying of the dataall directory was already happening in the `prod/ecs/Dockerfile` before modularisation. This cookiecutter config was I believe our workaround for the ReadOnlyRootFileSystem change which made the datapipelines stack creation fail.
Seems that copying was not required so far for the local deployment in the `dev/Dockerfile`, but now that this config being added to the env vars for the datapipelines, it is missing it.
My guess is that its enough to COPY this yaml config file, but I'm fine if we copy the data.all directory like we do for `prod/ecs/Dockerfile`

dlpzx:
### Fix cdkpipelines unregistered stack issue

Balint and I found the root cause of the error. We needed to add the correct path to point to the `app.py` created in the `ddk init`. That is implemented in this commit

When trying the fix locally I noticed that we might face additional issues with the paths. Locally, I can see that the ddk repo was created in the following path: `/code/dataall/modules/datapipelines/cdk/dataall-cdk-hoy-2-grsiynkk/.git/` But then, it failed in the deployment:
`Failed to deploy stack 5yxtjojc due to [Errno 2] No such file or directory: '/code/dataall/base/cdkproxy/./cdkpipeline/dataall-cdk-hoy-2-grsiynkk/'`

I changed the path creation a bit to point to the correct folder. I will now start the testing again:

[ ] - local CodePipelines gitflow
[ ] - local CodePipelines trunk
[ ] - local CDK trunk
[ ] - AWS CodePipelines gitflow
[ ] - AWS CodePipelines trunk
[ ] - AWS CDK trunk

dlpzx:
We considered 3 alternatives to solve the import issue:
1. Place the load_modules for the CDK import mode somewhere where it would be imported in the deployment of stacks and in the instanciate stacks (app.py where is was originally placed). Advantage = simple, Disadvantage = unexpected behavior, we tried and there was no place where the stacks decorated where correctly instanciated.
2. Add a new import mode for the CLI extensions, keep the loading of CDK stacks in the app.py and add the load_modules for the new import mode in the deploy stacks function. Advantage = clear separattion of import modes. Import only what is needed. Disadvantage = more complexity by adding a new import mode
3. repeat the load_modules loading of the CDK import mode in the deploy function. Advantage = simpler Disadvantage = loading more things than necessary in the app.py and in the deploy function


We opted for alternative 2, also considering if we need to remove this functionality or test it to have it more isolated. 
- Added import mode: `ImportMode.CDK_CLI_EXTENSION` that imports the extensions only
- Added loading of this import mode in `dataall.base.cdkproxy.cdk_cli_wrapper.deploy_cdk_stack`

However, it uncovered one issue with the loader checks. 
- First, that the `dataset_base` module had a `True` for absolutely all import modes. In this PR we explicitly check from a list of supported modes
- Second, the `dataall.base.loader._check_loading_correct` check number 4 was checking all packages imported no matter the mode selected. Instead, it should check all packages imported that support a module and then compare it with the expected loaded packages from the config file.

dlpzx:
Left TODO items:
- The last [commit](https://github.com/awslabs/aws-dataall/pull/705/commits/264fce733b0358aebe71cbec62b00d5f6143f3ab) will break the local development of cdk-trunk pipelines. It is low priority and can be fixed in the dockerfile or after the release
- The loader check#4 is useless at the moment and needs to be reviewed. To keep the PRs clean I suggest to handle that in a separate PR

noah-paige:
Tested the latest code in this PR and ran into similar issues as before:

- [x] AWS CodePipelines gitflow
- [x]  AWS CodePipelines trunk
- [ ] AWS CDK trunk --> again a path error `/usr/bin/python3.8: can't open file '../../base/cdkproxy/app.py': [Errno 2] No such file or directory`

backend/dataall/base/loader.py
```diff
@@ -237,15 +238,16 @@ def _check_loading_correct(in_config: Set[str], modes: Set[ImportMode]):
...
     checked_module_names |= in_config
-    for module in sys.modules.keys():
-        if module.startswith(_MODULE_PREFIX) and module != __name__:  # skip loader
-            name = _get_module_name(module)
-            if name and name not in checked_module_names:
-                raise ImportError(f"The package {module} has been imported, but it doesn't contain ModuleInterface")
+    dataall_packages_names = [_get_module_name(package) for package in sys.modules.keys() if package.startswith(_MODULE_PREFIX) and package != __name__]
+    for module in _all_modules():
+        if module.name() in dataall_packages_names and module.is_supported(modes):
+            if module.name() and module.name() not in checked_module_names:
```
nikpodsh, line 247: The loader works correctly in this scenario: It's complaining that we are trying to import `database_base` which shouldn't be loaded. Why is it happening? Before new `CDK_CLI_EXTENSION` mode, we had imported `database_base` directly in `datasets` in `__init__.py`.  Since `dataset` is part of the config, the loader doesn't raise an error while importing init file.  `dataset_base` is not standalone module and is no longer support all modes as it used to be, thus `__init__.py` shouldn't be initialized for `database_base`. 
The way to solve it: move global `dataset_base` imports to corresponding `ModuleInterface`s contructors

backend/dataall/base/cdkproxy/cdk_cli_wrapper.py
```diff
@@ -70,6 +70,9 @@ def update_stack_output(session, stack):
 
 
 def deploy_cdk_stack(engine: Engine, stackid: str, app_path: str = None, path: str = None):
+    from dataall.base.loader import load_modules, ImportMode
+    load_modules(modes={ImportMode.CDK_CLI_EXTENSION})
```
nikpodsh, line 74: We should move it to `cdkproxy`

backend/dataall/modules/datasets_base/__init__.py
```diff
@@ -5,4 +5,12 @@
...
     def is_supported(modes: Set[ImportMode]) -> bool:
-        return True
+        supported_modes = [
+            ImportMode.API,
+            ImportMode.CDK,
+            ImportMode.HANDLERS,
+            ImportMode.STACK_UPDATER_TASK,
+            ImportMode.CATALOG_INDEXER_TASK
+        ]
+        supported = [x in supported_modes for x in modes]
```
nikpodsh, line 15: We can simplify it by 
```
supported_modes = {
    ImportMode.API,
    ImportMode.CDK,
    ImportMode.HANDLERS,
    ImportMode.STACK_UPDATER_TASK,
    ImportMode.CATALOG_INDEXER_TASK
}
return modes & supported_modes
```

backend/dataall/modules/datapipelines/cdk/__init__.py
```diff
@@ -1,13 +0,0 @@
-from dataall.modules.datapipelines.cdk import datapipelines_cdk_cli_wrapper_extension, datapipelines_cdk_pipeline, \
-    env_role_datapipelines_stepfunctions_policy, env_role_datapipelines_lambda_policy, env_role_datapipelines_cicd_policy, datapipelines_pipeline, \
-    pivot_role_datapipelines_policy
-
-__all__ = [
```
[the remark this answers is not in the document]
dbalintx: Adriana moved it to the datapipelines CDK ModuleInterface init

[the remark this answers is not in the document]
nikpodsh: Then a minor comment: I'd import them by the name of the class, not by file. Otherwise, It's harder to track usages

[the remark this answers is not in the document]
dbalintx: sure!

dbalintx:
With the current state I could deploy all 3 pipelines types to AWS.

backend/dataall/core/stacks/tasks/cdkproxy.py
```diff
@@ -11,6 +12,10 @@
     root.addHandler(logging.StreamHandler(sys.stdout))
 logger = logging.getLogger(__name__)
 
+load_modules(modes={ImportMode.CDK_CLI_EXTENSION})
+logger.warning(f'Loading  _CDK_CLI_WRAPPER_EXTENSIONS {_CDK_CLI_WRAPPER_EXTENSIONS}')
```
nikpodsh, line 16: Warning is too much, isn't it? I'd go with debug

nikpodsh (approved):

PR #706, opened 2023-08-25
Add env URI to role share query

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- If I have Group_1 that is invited to Env1 and Env2 and has existing shares for that group to Env1 - I still cannot remove the EnvGroup from Env2 (where no shares exists)
  - When we count resources to validate before removing group - for share objects in the `count_principal_resources()` we only filter by principalId and principalType but we also need to filter by environmentUri

- This PR fixes query to add environmentURI filter when counting envGroup resources before removing groups

### Security
NA 
```

  - Is the input sanitized?
```

### DIFF
--- a/backend/dataall/core/environment/services/environment_service.py
+++ b/backend/dataall/core/environment/services/environment_service.py
@@ -332,6 +332,15 @@ class EnvironmentService:
                 message=f'Team: {group} has created {group_env_objects_count} resources on this environment.',
             )
 
+        group_env_consumption_roles = EnvironmentService.query_user_environment_consumption_roles(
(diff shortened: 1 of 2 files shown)


dlpzx (approved):

dbalintx (approved):

dlpzx:
Looks good! @noah-paige  I wondered if we also count the consumption roles when we remove a team from an environment. I just checked and I don't think we do. Should we define ConsumptionRoles as `EnvironmentResource`?  wdyt?

dbalintx (approved):
lgtm!

PR #718, opened 2023-08-31
Fix canary user pw creation

dbalintx:
In rare cases the current way of generating a Canary user password in Cognito can result in a string containing no numerical values, hence following error is thrown during deployment ( requires **enable_cw_canaries** config parameter set to True in cdk.json):
`botocore.errorfactory.InvalidPasswordException: An error occurred (InvalidPasswordException) when calling the AdminCreateUser operation: Password did not conform with password policy: Password must have numeric characters`

Changing the password creation to contain at least 1 uppercase and 1 numerical character.

### Feature or Bugfix
- Bugfix

### Detail
- <feature1 or bug1>
- <feature2 or bug2>

### Relates
- <URL or Ticket>

### Security

n/a

  - Is the input sanitized?

### DIFF
--- a/deploy/configs/cognito_urls_config.py
+++ b/deploy/configs/cognito_urls_config.py
@@ -7,6 +7,11 @@ import boto3
 from botocore.exceptions import ClientError
 
 
+def shuffle_password(pwd):
+    chars = list(pwd)
+    random.shuffle(chars)
+    return ''.join(chars)
+
 def setup_cognito(
     region,
     resource_prefix,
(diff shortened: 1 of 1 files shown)


deploy/configs/cognito_urls_config.py
```diff
@@ -110,11 +110,13 @@ def setup_cognito(
                         {'Name': 'email', 'Value': f'{username}@amazonaws.com'}
                     ],
                     TemporaryPassword='da@'
+                    + random.SystemRandom().choice(string.ascii_uppercase)
```
nikpodsh, line 113: Let's also shuffle the characters in the string to make it less predictable:
```
def shuffle_pwd(pwd):
    chars = list(pwd)
    random.shuffle(chars)
    return ''.join(chars)
```

noah-paige:
Minor: Is there any point to add `da@` if we are then shuffling the password? I would opt to keep `da@` prefix and shuffle the trailing characters

Otherwise looks good to me

noah-paige (approved):
Looks good to me

InasK6:
> Minor: Is there any point to add `da@` if we are then shuffling the password? I would opt to keep `da@` prefix and shuffle the trailing characters
> 
> Otherwise looks good to me

the `da@` seems useful to me to have the special character `@` which is required for the password definition , but I am not sure if the `da`  is really necessary

PR #725, opened 2023-09-04
Add conditions on catalog indexer tasks

nikpodsh:
The catalog indexer fails if all modules are disabled. Here is a small improvement to fix it 

Checked against:
1)  all disabled
2)  all enabled
3) datasets enabled only
4) dashboards enabled only

### Security
`N/A` . it's a small improvement for disable/enabling tasks based on modules. Based on

  - Is the input sanitized?

### DIFF
--- a/deploy/stacks/run_if.py
+++ b/deploy/stacks/run_if.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from .deploy_config import deploy_config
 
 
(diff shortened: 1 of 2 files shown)


deploy/stacks/run_if.py
```diff
@@ -17,16 +19,20 @@ def no_decorated(f):
     return fn, staticmethod if static_func else no_decorated
 
 
-def run_if(active_property: str):
+def run_if(active_properties: List[str]):
```
dlpzx, line 22: Also in the original one, I got a question shouldn't `active_properties` be of type List[bool]?

dlpzx (approved):
lgtm. I will test in my deployment as well

deploy/stacks/run_if.py
nikpodsh: It is possible to achieve. But it would duplicate the code a little in all files since we would need to get the value of properties (`deploy_config.get_property`) before calling the decorator. I don't see that we would benefit much by passing List[bool] at this point of time 😄  (Of course, it can easily change later with a new reqs)

PR #727, opened 2023-09-04
Fix frontend issues

itsmo-amzn:
### Feature or Bugfix

- Bugfix

### Detail
- disabled `datasets` tab in `environment` view when `datasets` module is disabled
- disabled module list item in environment features when module is disabled
- fix issue with landing route when `catalog` module is disabled
- `worksheets` tab enablement rules

### DIFF
--- a/frontend/src/modules/Environments/components/EnvironmentFeatures.js
+++ b/frontend/src/modules/Environments/components/EnvironmentFeatures.js
@@ -1,3 +1,5 @@
+import React from 'react';
+import PropTypes from 'prop-types';
 import {
   Card,
   CardContent,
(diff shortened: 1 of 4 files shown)


frontend/src/modules/Environments/components/EnvironmentFeatures.js
```diff
@@ -7,119 +9,73 @@ import {
...
+import { ModuleNames, isModuleEnabled } from 'utils';
 
 export const EnvironmentFeatures = (props) => {
   const { environment, ...other } = props;
 
+  // Define your list of features/modules with additional attributes
+  const features = [
+    {
+      title: 'Dashboards',
+      env: 'dashboardsEnabled',
```
dlpzx, line 22: Very minor: can we rename `env` as something that indicates that it is the env_parameter?

frontend/src/routes.js
```diff
@@ -409,7 +409,11 @@ const routes = [
     children: [
       {
         path: '',
-        element: <Catalog />
+        element: isModuleEnabled(ModuleNames.CATALOG) ? (
```
dlpzx, line 412: Instead of making the Catalog the default and then Organization. Could we define the `Environments` view as default always? 
- Environments better than Organizations because they are more used, they are the AWS accounts-link
- better to have some core as default and then forget about the modules in the defaults
- better to have environment as default because the Catalog performs queries to OpenSearch which at enterprise scale can be costly => let's save soooome money

frontend/src/utils/helpers/moduleUtils.js
```diff
@@ -23,6 +23,12 @@ function isModuleEnabled(module) {
   if (module === ModuleNames.SHARES) {
     return getModuleActiveStatus(ModuleNames.DATASETS);
   }
+  if (module === ModuleNames.WORKSHEETS) {
+    return (
+      getModuleActiveStatus(ModuleNames.DATASETS) &&
+      getModuleActiveStatus(ModuleNames.WORKSHEETS)
+    );
+  }
```
dlpzx, line 31: I think is is out of scope for V2.0.0 release, but we need a better way of managing dependencies in the frontend. Wdyt?

dlpzx (commented):
just a suggestion on the default routes that would save some money for customers

frontend/src/utils/helpers/moduleUtils.js
itsmo-amzn: Yes, I believe we discussed this earlier in the module enablement PR.

frontend/src/modules/Environments/views/EnvironmentView.js
```diff
@@ -52,7 +53,7 @@ const tabs = [
     value: 'teams',
     icon: <SupervisedUserCircleRounded fontSize="small" />
   },
-  {
+  isModuleEnabled(ModuleNames.DATASETS) && {
```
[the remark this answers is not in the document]
itsmo-amzn: good catch!

dlpzx (approved):
Looks good!

Issue #723, opened 2023-09-01
Launching Chat from Datasets leads to 400 error due to missing types, if features are disabled


rbernotas:
### Describe the bug

Datasets -> choose a dataset -> Chat

This leads to a 400 error if you have Dashboards/Worksheets/DataPipeline disabled.  The types are not able to be found and graphql api returns a 400 error.

Source is based on tag v2.0.0-beta1.

`
{
    "errors": [
        {
            "message": "Unknown type 'Dashboard'.",
            "locations": [
                {
                    "line": 17,
                    "column": 14
                }
            ]
        },
        {
            "message": "Unknown type 'Worksheet'.",
            "locations": [
                {
                    "line": 21,
                    "column": 14
                }
            ]
        },
        {
            "message": "Unknown type 'DataPipeline'. Did you mean 'DatasetLink'?",
            "locations": [
                {
                    "line": 25,
                    "column": 14
                }
            ]
        }
    ]
}
`

### How to Reproduce

Source is based on tag v2.0.0-beta1.

Disable Dashboards, Worksheets, and/or Data Pipeline features.

Datasets -> choose a dataset -> Chat

network error / 400 error should appear and the graphql errors will be present in the response.

### OS

n/a

### Python version

n/a

### AWS data.all version

v2.0.0-beta1

dlpzx:
Hi @rbernotas, I just tested the `main-v2` branch with Dashboards disabled and I was able to confirm this issue also in the V2.0 release (not only in beta) I will add it to our backlog of issues for the release. Thanks for being a release tester :)

noah-paige:
Hi @rbernotas  - thank you for opening up this issue. This linked PR above #729 should include the fix for the issue found on disabling modules and the Chat feature

As of now the code changes are reflected in the `main-v2` branch

PR #729, opened 2023-09-05
DA v2: Fix Get Query on Chat/Feed Feature 

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- If 1 of the dataset, dashboard, or datapipeline modules was disabled, the query to get Feed (i.e. the Chat Feature) would fail with an error similar to error:
  - `[GraphQL error]: Message: Unknown type 'DataPipeline'. Did you mean 'DatasetLink'?, Location: [object Object], Path: undefined`
  - This is due to the fixed inline fragments on the `getFeed(...)` query that do not exists such as `... on DatasetTable {...}`
   
- In this PR we remove the fragments as the `label` return value is not used in the `FeedComments` View

### Relates
- [#723](https://github.com/awslabs/aws-dataall/issues/723)

### Security

  - Is the input sanitized?

### DIFF
--- a/frontend/src/services/graphql/Feed/listFeedMessages.js
+++ b/frontend/src/services/graphql/Feed/listFeedMessages.js
@@ -13,24 +13,6 @@ export const listFeedMessages = ({ targetUri, targetType, filter }) => ({
       $filter: FeedMessageFilter
     ) {
       getFeed(targetUri: $targetUri, targetType: $targetType) {
-        target {
-          __typename
-          ... on DatasetTable {
-            label
-          }
-          ... on Dataset {
-            label
-          }
-          ... on DatasetStorageLocation {
-            label
-          }
-          ... on Dashboard {
-            label
-          }
-          ... on DataPipeline {
-            label
-          }
-        }
         messages(filter: $filter) {
           count
           hasNext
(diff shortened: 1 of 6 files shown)


noah-paige:
With the above changes - the `FeedRegistry` class is no longer used ... I can clean up the code to remove any reference to this unused class but have left it in for now as it could be used in future enhancements. 

What do you think @dlpzx ?

frontend/src/services/graphql/Feed/listFeedMessages.js
```diff
@@ -13,24 +13,6 @@ export const listFeedMessages = ({ targetUri, targetType, filter }) => ({
       $filter: FeedMessageFilter
     ) {
       getFeed(targetUri: $targetUri, targetType: $targetType) {
-        target {
```
nikpodsh, line 16: I believe we will have the same issue for `listGlossaryAssociations`

noah-paige: I would imagine so when one of dashboard or dataset are disabled - is this something that will be resolved as part of PR #731? Otherwise I can open a separate PR once I validate that error as well

dlpzx: @noah-paige I was not going to include it in the PR as it is already quite packed. Can we wait for #731 to be merged and then fix the feed?

dlpzx (approved):

frontend/src/services/graphql/Feed/listFeedMessages.js
noah-paige: sounds good - let's do that and I will merge this PR

PR #730, opened 2023-09-05
Fix/cdkpipeline local deploy

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Fix bug with local deployments creation of CDK Pipelines. Error Caused By:
  - Not registering `ImportMode.CDK_CLI_EXTENSION` since `load_modules()` for this import mode type was only being called for AWS Deployments
  - Incorrect path specified for the location of the cdk/ddk app created for CDK Pipelines in local deployments

- 

### Security

NA

```
  - Is the input sanitized?
```

### DIFF
--- a/backend/cdkproxymain.py
+++ b/backend/cdkproxymain.py
@@ -9,8 +9,6 @@ from fastapi import FastAPI, BackgroundTasks, status, Response
 
 import dataall.base.cdkproxy.cdk_cli_wrapper as wrapper
 from dataall.base import db
-from dataall.base.loader import load_modules, ImportMode
-from dataall.base.cdkproxy.stacks import StackManager
 from dataall.core.organizations.db.organization_models import Organization
 from dataall.core.stacks.db.stack_models import Stack
 
@@ -24,8 +22,6 @@ logger.warning(
     f"Application started for envname= `{ENVNAME}` DH_DOCKER_VERSION:{os.environ.get('DH_DOCKER_VERSION')}"
 )
 
-load_modules(modes={ImportMode.CDK})
-StackManager.registered_stacks()
 
 
 def connect():
--- a/backend/dataall/base/cdkproxy/cdk_cli_wrapper.py
+++ b/backend/dataall/base/cdkproxy/cdk_cli_wrapper.py
@@ -70,6 +70,8 @@ def update_stack_output(session, stack):
 
 
 def deploy_cdk_stack(engine: Engine, stackid: str, app_path: str = None, path: str = None):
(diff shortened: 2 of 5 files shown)


noah-paige:
Tested the following successfully:

Locally
- [ ] Creation of CDK Pipeline
- [ ] Creation of Trunk-based CodePipeline
- [ ] Creation of Gitflow CodePipeline

AWS Deployment
- [ ] Creation of CDK Pipeline
- [ ] Creation of Trunk-based CodePipeline
- [ ] Creation of Gitflow CodePipeline

backend/cdkproxymain.py
```diff
@@ -24,9 +22,6 @@
     f"Application started for envname= `{ENVNAME}` DH_DOCKER_VERSION:{os.environ.get('DH_DOCKER_VERSION')}"
 )
 
-load_modules(modes={ImportMode.CDK})
```
dbalintx, line 27: was this unnecessary because the loading also happens [in app.py ](https://github.com/awslabs/aws-dataall/blob/main-v2/backend/dataall/base/cdkproxy/app.py#L16)?

noah-paige: Yup exactly

dbalintx: My understanding is, that in case the pipeline type is CDK Pipeline (the one managed with DDK), this app.py is not going to be used, as another one is created with the DDK commands. See [here](https://github.com/awslabs/aws-dataall/blob/main-v2/backend/dataall/base/cdkproxy/cdk_cli_wrapper.py#L116) `app_path` is overwritten and [here](https://github.com/awslabs/aws-dataall/blob/main-v2/backend/dataall/modules/datapipelines/cdk/datapipelines_cdk_pipeline.py#L142) the new app file is created.

In this case, aren't we going to missed the CDK modules, as this loading is not going to happen in the other app.py file?

noah-paige: When the pipeline type is `cdkpipeline` we load the `DatapipelinesCdkCLIExtensionModuleInterface` which loads the `CDKPipeline()` class for us

The `CDKPipeline()` class initializes a new DDK App and CodeCommit Repo and calls `cdkdeploy` of that newly created `app.py`

For that DDK App to deploy there is no need to reference any additional data.all CDK module code - there just requires cdk and ddk pip libraries to be installed (as is the case in `requirements.txt`)

I do not think there should be any issue here with not having `load_modules({ImportMode.CDK})` anywhere but the `app.py`

dbalintx (approved):

Issue #540, opened 2023-06-29
OpenSearch stack failed during backend deploy due to length of policy name


rbernotas:
### Describe the bug

Our "envname" as described in cdk.json is "staging".  Using this envname, the dataall CICD backend stage CF stack is failing because of error in the OpenSearchCollectionEncryptionPolicy:

Properties validation failed for resource OpenSearchCollectionEncryptionPolicystaging with message: #Name: expected maxLength: 32, actual: 33 #Name: failed validation constraint for keyword [pattern]

The name of the policy is too long due to the envname "staging".

In the file aws-dataall/deploy/stacks/opensearch_serverless.py (L49-58):

`        cfn_encryption_policy = opensearchserverless.CfnSecurityPolicy(
            self,
            f'OpenSearchCollectionEncryptionPolicy{envname}',
            name=f'{resource_prefix}-{envname}-encryption-policy',
            type='encryption',
            policy=self._get_encryption_policy(
                collection_name=self.cfn_collection.name,
                kms_key_arn=key.key_arn,
            ),
        )`

The name here would be dataall-staging-encryption-policy (33 chars) when the limit must be 32 chars.

In the meantime a workaround fix would likely be to change the name to {resource_prefix}-{envname}-encrypt-policy.

In the long term there should be a fix to have a truncate mechanism to prevent the error in the Policy name length, or warn the user about the length of their envname in the cdk.json.

### How to Reproduce

Run a deploy with the envname in cdk.json of "staging".  OpenSearch deploy in the backend stack deploy will fail on the length of Policy names due to the length of the envname.

### OS

N/A

### Python version

N/A

### AWS data.all version

1.5

dlpzx:
Merged and released with v2.1.0 🚀

PR #750, opened 2023-09-12
Build Compliant Names for Opensearch Resources

noah-paige:
### Feature or Bugfix

- Enhancement / Bugfix

### Detail
- Ensure the names passed for OpenSearch Domain and OpenSearch Serverless Collection, Access Policy, Security Policy, and VPC Endpoint all follow naming conventions required by the service, meaning

    - The name must start with a lowercase letter
    - Must be between 3 and 28 characters
    - Valid characters are a-z (lowercase only), 0-9, and - (hyphen).

### Relates
- #540 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/base/utils/naming_convention.py
+++ b/backend/dataall/base/utils/naming_convention.py
@@ -13,6 +13,7 @@ class NamingConventionPattern(Enum):
     GLUE_ETL = {'regex': '[^a-zA-Z0-9-]', 'separator': '-', 'max_length': 52}
     NOTEBOOK = {'regex': '[^a-zA-Z0-9-]', 'separator': '-', 'max_length': 63}
     DEFAULT = {'regex': '[^a-zA-Z0-9-_]', 'separator': '-', 'max_length': 63}
+    OPENSEARCH = {'regex': '[^a-z0-9-]', 'separator': '-', 'max_length': 27}
 
 
 class NamingConventionService:
@@ -47,6 +48,9 @@ class NamingConventionService:
         elif self.service == NamingConventionPattern.NOTEBOOK:
             regex = self.service.NOTEBOOK.value['regex']
             return self.build_notebook_compliant_name(regex)
+        elif self.service == NamingConventionPattern.OPENSEARCH:
+            regex = self.service.OPENSEARCH.value['regex']
+            return self.build_opensearch_compliant_name(regex)
         else:
             regex = self.service.DEFAULT.value['regex']
             return self.build_default_compliant_name(regex)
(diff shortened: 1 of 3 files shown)


noah-paige:
**Note**: In the above PR I added a path to the `backend/dataall/utils` in the sys path in order to be able to import the `naming_convention.py` module in the `deploy/stacks/opensearch*` files

As one alternative we could remove the hardcoded `BACKEND_UTILS_PATH` and edit the import statement to be from `backend.dataall.base.utils.naming_convention` but that approach required to remove the imports listed in `backend/dataall/__init__.py` which later caused issues with Integration Tests.

Another alternative that could work is to export the `PYTHON_PATH` env variable as part of the CodeBuild build spec but I have not tested and would run into issues since the UpdatePipeline Step comes after the Build Step in the CodePipeline so the build spec would not be updated in time

So far with testing everything is still functional but please let me know if there is any perceived risk in the above change? @dlpzx 

Testing:

AWS w/ Opensearch Serverless

- [ ] CICD CodePipeline Succeeds (No Changes to the OpenSearch Serverless Collection)
- [ ] Frontend Views and Routes for Catalog all reachable
- [ ] Also tested with a longer encryption policy name (correctly built compliant name and recreated encryption policy only) 

AWS w/ Opensearch
- [x] CICD CodePipeline Succeeds (No Changes to the OpenSearch Domain)
- [x] Frontend Views and Routes for Catalog all reachable

noah-paige:
I also had to change the construct IDs of some of the existing OpenSearch Resources since if the `_set_os_compliant_name()` function returns a different name than the previous existed one than it will try to recreate the resource with the same ID which throws an error in CloudFormation

I would think this should not be a concern since any pre-existing OpenSearch resource names must already comply with the naming constraints and thus there should be no changeset or update on CloudFormation for the OpenSearch domain/collection and corresponding resources (I only ran into this issue when testing because I was manually editing the resource names to ensure the `_set_os_compliant_name()` function works correctly)

Still to be safe I only change construct IDs for:
- OpenSearch Serverless Network Policy
- OpenSearch Serverless Access Policy
- OpenSearch Serverless Encryption Policy
- OpenSearch Serverless VPC Endpoint

For the OpenSearch domain or serverless collection I kept the construct IDs the same so if for some reason the name does change and the resource needs to be recreate it will throw an error rather than trying to delete the OpenSearch domain/collection and cause additional problems

dlpzx:
Hi @noah-paige I like the regex and the usage of the utilities, but I am thinking that the creation of the OpenSearch cluster/domain... is out of the backend folder. What do you think if we create naming conventions in `deploy`? I am just brainstorming here, but maybe we can have a naming conventions utility for the infra resources and another one for the environment-account stacks

noah-paige:
@dlpzx - does it make sense to have 2 separate naming convention utilities or would it make sense to have 1 utility that is used for both `deploy/` and `backend/` . I imagine there may be some overlap so could avoid duplicate code but could add some complexity moving naming convention utility to parent directory and copying in to `backend/` similar to how we do for `config.json` in dockerfile

noah-paige:
Added a separate naming convention pattern for opensearch serverless (allows for a longer max length resource name)

noah-paige:
I had to make one small edit to the naming convention util to handle the case when `target_uri` is `None` - for OpenSearch naming cases

Successfully then deployed the latest to AWS and tested all of the different Naming Convention Patterns locally 

@dlpzx - I think the above is good to merge to `v2m1m0` branch but will wait for confirmation

dlpzx:
It's good from my side. I tested the deployment and locally and we will do additional validation before releasing v2.1 so you can go ahead and merge :)

Issue #738, opened 2023-09-07
Replace shell=true commands in subprocess function


dlpzx:
**Is your idea related to a problem? Please describe.**
From several scanning tools we get the following error regarding shell=true. Description: Found 'subprocess' function 'run' with 'shell=True'. This is dangerous because this call will spawn the command using a shell process. Doing so propagates current shell settings and variables, which makes it much easier for a malicious actor to execute commands. Use 'shell=False' instead. Details: https://sg.run/J92w  

This happens in:

* backend/dataall/base/cdkproxy/cdk_cli_wrapper.py (2 findings)
* backend/dataall/modules/datapipelines/cdk/datapipelines_cdk_pipeline.py (4 findings)
* backend/dataall/modules/datapipelines/cdk/datapipelines_pipeline.py (2 findings)
* deploy/stacks/solution_bundling.py (2 findings)

**Describe the solution you'd like**
We have evaluated that there is no risk as there is no input introduced by users in this command. Nevertheless, we can fix the issue and avoid shell=true, so we should do it.

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

dlpzx:
Merged and released with v2.1.0 🚀

PR #760, opened 2023-09-14
Fix: shell true semgrep

dlpzx:
### Feature or Bugfix
- Feature
- Bugfix

### Detail
As explained in the [semgrep docs](https://semgrep.dev/docs/cheat-sheets/python-command-injection/#1b-shelltrue): "Functions from the subprocess module have the shell argument for specifying if the command should be executed through the shell. Using shell=True is dangerous because it propagates current shell settings and variables. This means that variables, glob patterns, and other special shell features in the command string are processed before the command is run, making it much easier for a malicious actor to execute commands. The subprocess module allows you to start new processes, connect to their input/output/error pipes, and obtain their return codes. Methods such as Popen, run, call, check_call, check_output are intended for running commands provided as an argument ('args'). Allowing user input in a command that is passed as an argument to one of these methods can create an opportunity for a command injection vulnerability."

In our case the risk is not exposed as no user input is directly taken into the subprocess commands. Nevertheless we should strive for the highest standards on security and this PR works on replacing all the `shell=True` executions in the data.all 
code.

In this PR:
- when possible we have set `shell=False`
- in cases where the command was too complex a `CommandSanitizer` ensures that the input arguments are strings following the regex=`[a-zA-Z0-9-_]`

Testing: 
- [X] local testing - deployment of any stack (`backend/dataall/base/cdkproxy/cdk_cli_wrapper.py`)
- [X] local testing - deployment of cdk pipeline stack (`backend/dataall/modules/datapipelines/cdk/datapipelines_cdk_pipeline.py`)
- [X] local testing - deployment of codepipeline pipeline stack (`backend/dataall/modules/datapipelines/cdk/datapipelines_pipeline.py`)
- [ ] AWS testing - deployment of data.all
- [ ] AWS testing - deployment of any stack (`backend/dataall/base/cdkproxy/cdk_cli_wrapper.py`)
- [ ] AWS testing - deployment of cdk pipeline stack (`backend/dataall/modules/datapipelines/cdk/datapipelines_cdk_pipeline.py`)
- [ ] AWS testing - deployment of codepipeline pipeline stack (`backend/dataall/modules/datapipelines/cdk/datapipelines_pipeline.py`)

### Relates
- #738 

### Security

  - Is the input sanitized? ---> 🆗 This is exactly what this PR is trying to do


deploy/stacks/solution_bundling.py
```diff
@@ -14,18 +14,21 @@ def __init__(self, source_path=None):
...
-            [f'ls -ll {output_dir}'],  # nosemgrep
-            stderr=subprocess.STDOUT,  # nosemgrep
-            shell=True,  # nosec  # nosemgrep
+        subprocess.check_output(
+            ['pip', 'install', '-r', requirements_path, '-t', output_dir],
+            stderr=subprocess.STDOUT,
+            shell=False,
+        )
+
+        ls_output = subprocess.check_output(
```
noah-paige, line 29: do we need to store output in variable - does not seem like we are using `ls_output`

noah-paige (commented):
Found some errors (mentioned above) when testing AWS Deployment - may need to look into other workarounds when setting `shell=False` for cdkproxy

backend/dataall/base/cdkproxy/cdk_cli_wrapper.py
```diff
@@ -46,10 +47,15 @@ def aws_configure(profile_name='default'):
     print('..............................................')
     print(f"AWS_CONTAINER_CREDENTIALS_RELATIVE_URI: {os.getenv('AWS_CONTAINER_CREDENTIALS_RELATIVE_URI')}")
     cmd = ['curl', '169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI']
-    process = subprocess.run(' '.join(cmd), text=True, shell=True, encoding='utf-8', capture_output=True)  # nosec  # nosemgrep
+    process = subprocess.run(cmd, text=True, shell=False, encoding='utf-8', capture_output=True)
```
dlpzx, line 50: I am working on this item, after plotting the error logs I think the issue is on the url to get the ECS-EC2 credentials. I will use f-string to create the curl command since we can obtain it as environment variable. I am now testing it in AWS

backend/dataall/modules/datapipelines/cdk/datapipelines_cdk_pipeline.py
```diff
@@ -215,23 +228,17 @@ def git_push_repo(self):
...
     def clean_up_repo(path):
         if path:
-            precmd = [
-                'deactivate;',
-                'rm',
-                '-rf',
-                f"{path}"
-            ]
-
+            cmd = ['rm', '-rf', f"{path}"]
```
dlpzx, line 231: I removed the `deactivate` command. I think we do not need it. I am testing if there are any implications when deploying to AWS

dlpzx: Tested in AWS and working fine from my end

dlpzx:
We should also consider upgrading to DDK to the newest major release. They do not make use of the core library anymore and the bootstraping of accounts is replaced by CDK bootstraping, which simplified the usage of pipelines

Issue #766, opened 2023-09-15
CDK-nag actions from external forks do not have write permissions


dlpzx:
### Describe the bug

As happens in this PR, the cdk-nag action requires write permissions, but the maximum permissions allowed for a fork in a public repo are `read` permissions.
![image](https://github.com/awslabs/aws-dataall/assets/71252798/c276830e-6068-4c4e-a9da-87f8cc522bc3)

**alternatives**
1. use [ok-to-test](https://github.com/imjohnbo/ok-to-test) --> requires authentication, preferred is GitHub App (not possible for data.all),  other options: personal access token or OAuth app token
2. run cdk-nag action only on minor release branches (v2m2m0 to main) branch
3. run cdk-nag on `pull_request_target` after the PR is merged
4. run cdk-nag on schedule
5. Use other than OIDC
6. running without credentials --> maybe mocking the cdk.context.json, not adding fake values but mocking it

Advice:
6 woudl solve the issue, so preferred if possible. 2 and 3 can also be added.

### How to Reproduce
Open a PR from a fork.

### OS

n/a

### Python version

n/a

### AWS data.all version

7.0.0

anmolsgandhi:
Hi @dlpzx This is marked as high priority and in progress, if you are working on it id like to add it to v2.1.0 milestone to track it.

dlpzx:
Completed in #767

PR #767, opened 2023-09-15
Cron job for cdk-nag Tuesdays at 1AM

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
See issue #766 

### Relates
#766 

### Security

  - Is the input sanitized?

### DIFF
--- a/.github/workflows/cdk-nag.yml
+++ b/.github/workflows/cdk-nag.yml
@@ -2,16 +2,12 @@ name: CDK Nag
 
 on:
   workflow_dispatch:
-  push:
     paths:
-    - "deploy/**"
+      - "deploy/**"
     branches:
-    - main
-  pull_request:
-    paths:
-    - "deploy/**"
-    branches:
-    - main
+      - main
+  schedule:
+    - cron: '0 1 * * 2'
 
 permissions:
   id-token: write
--- a/.github/workflows/semgrep-schedule.yml
+++ b/.github/workflows/semgrep-schedule.yml
@@ -4,7 +4,6 @@ on:
   workflow_dispatch:
     branches:
       - main
-      - main-v2
   schedule:
     - cron:  '0 1 * * 2'
 


dlpzx:
The test works but it still requires some testing in a real AWS deployment.

deploy/stacks/backend_stack.py
```diff
@@ -362,13 +362,16 @@ def create_opensearch_serverless_stack(self):
...
+            response = ec2_client.describe_prefix_lists(
+                Filters=[
+                    {
+                        'Name': 'prefix-list-name',
+                        'Values': [f'com.amazonaws.{self.region}.s3']
+                    },
+                ]
+            )
+        except:
+            return ""
```
dlpzx, line 376: I am not particularly happy with this piece. I need to review other alternatives

dlpzx: We could read from environment variables the GITHUB_ACTIONS env var. The issue is that this stack is alter executed in the CICD pipelines. I think we would need to add the env var to the CodeBuild stage, a bit complex for just this.

The main issue is that the except is returning a valid output instead of trowing an exception. In a real deployment, it could happen that CodeBuild has no permissions to list S3 prefixes, then the prefix list id is set to "". Later when we set the security groups for ECS and CodeBuild we add outbound rules to that list id. It will be giving permissions to a wrong id. My guess is that the CloudFormation will fail as there are no prefix ids as empty strings. 

```
# Add S3 Prefix List Connection
            if s3_prefix_list:
                sg_connection.allow_to(
                    ec2.Connections(peer=ec2.Peer.prefix_list(s3_prefix_list)),
                    ec2.Port.tcp(443),
                    'Allow ECS Task to S3 Prefix List'
                )
```

noah-paige (approved):
Tested the code changes to ensure no changes to backend stack for an AWS Deployment 

Looks good to me

PR #769, opened 2023-09-15
Fix Check other share exists before clean up

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Fix method to detect if other share objects exist on the environment before cleaning up environment-level shared resources (i.e. RAM invitation and PivotRole permissions)
  - Originally, if EnvA had 2 shares approved and succeeded on DatasetB and EnvA rejects 1 of the pre-existing shares, the method `other_approved_share_object_exists` was returning `False`and deleting necessary permissions for the other existing Share 
  - Also disables the other existing shares ability to Revoke since pivotRole no longer has permissions

### Security
NA
```

  - Is the input sanitized?
```

### DIFF
--- a/backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
+++ b/backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
@@ -775,13 +775,18 @@ class ShareObjectRepository:
 
     @staticmethod
     def other_approved_share_object_exists(session, environment_uri, dataset_uri):
+        share_item_shared_states = ShareItemSM.get_share_item_shared_states()
         return (
             session.query(ShareObject)
+            .join(
+                ShareObjectItem,
+                ShareObject.shareUri == ShareObjectItem.shareUri,
+            )
             .filter(
                 and_(
-                    Environment.environmentUri == environment_uri,
-                    ShareObject.status == ShareObjectStatus.Approved.value,
+                    ShareObject.environmentUri == environment_uri,
                     ShareObject.datasetUri == dataset_uri,
+                    ShareObjectItem.status.in_(share_item_shared_states)
                 )
             )
             .all()


noah-paige:
Tested in AWS Deployment:

Pre-requisites
- Dataset X owned by Team X in Env X
- Team A and Team B in Env Y

Testing Scenarios:

- [x] Team A Requests Access to Dataset X Table and Folder in Env Y
  - Share Succeeds
- [x] Team B Requests Access to Dataset X Table and Folder in Env Y
  - Share Succeeds

- [x] Team B Revokes Share in Env Y
  - Revoke Succeeds
  - Team B loses access to data
  - Team A still has access to table and folder
- [ ] Team A Revokes Share in Env Y
  - Revoke Succeeds
  - Access Revoked to External Account (Env Y) since no other shares exist
  - **Works only when Requestor Enviornment has Dashboards Enabled**


Upon testing, another issue seems to come up when trying to revoke the QS dataall Group for the second time (for the second revoke in the above testing scenario). It fails with the following error message:

`An error occurred (InvalidInputException) when calling the RevokePermissions operation: No permissions revoked. Grantee has no permissions and no grantable permissions on resource.`


With the current implementation of QuickSight using 1 QS Group for all EnvGroups in the environment - I believe we will have to avoid revoking any permissions on QS Group unless there are no other existing shares in the environment for that dataset - @dlpzx let me know your thoughts on this

backend/dataall/modules/dataset_sharing/aws/lakeformation_client.py
```diff
@@ -254,46 +254,48 @@ def revoke_source_table_access(**data):
         source_table = data['source_table']
         source_accountid = data['source_accountid']
         for target_principal in target_principals:
-            try:
```
noah-paige, line 257: I am not a huge fan of handling the QS Principals - want to think on other ways to do this 

In theory I would think the permissions for the target QS Group on the source database/table would only get removed when the external access to the target Aws Account is removed (i.e. when `revoke_external_account_access_on_source_account()` is called)

Will have to test though...

noah-paige:
To test this fix again for the following:

- [ ] 1 Table Shared 1 Time Cross Account to EnvA (QS enabled)
  - Share Succeeds with Proper LF Permissions
  - Share is Revoked Successfully (no more permissions for Team Role, QS Group, + Pivot Role in EnvA)

- [ ] 1 Table Shared 2 Times Cross Account to EnvA (QS enabled)
  - Both Shares Succeed
  - When 1 Share Revoke - no permissions for Team1 but QS Group and Team2 still can access Table in EnvA
  - Second Share Revoked Successfully

- [ ] 2 Tables in Same Dataset Shared 1 Time Each to Cross Account to EnvA (QS enabled)
  - Both Shares Succeed
  - When 1 Share Revoke - no more permissions for Team Role, QS Group, + Pivot Role in EnvA for Table1 (still permissions for other table)
  - Second Share Revoked Successfully

- [ ] 1 Table Shared 1 Time Same Account (QS enabled)
  - Share Succeeds with Proper LF Permissions
  - Share is Revoked Successfully (no more permissions for Team Role, QS Group, + Pivot Role)

noah-paige:
There is one remaining thing to fix - we currently only revoke external account access for cross account shares when there are no other share objects with the same target env and source dataset URIs

If that is the case we revoke external account access to the target env for all of the `self.revoked_tables` in the revoke request

The above does not work when there is 2 table shared from the same dataset to the same env (either to the same or different teams). Ultimately this leaves remaining permissions in LakeFormation when revoking a table

Instead, we can revoke external access on a tabular level and check for other existing share items with the same tableUri that are shared to the same envUri. Then if there are no other existing shares on that table we can easily revoke external access to the env for that table

noah-paige:
Final Testing - all works below, PR ready for review

- [x] 1 Table Shared 1 Time Cross Account to EnvA (QS enabled)
    - Share Succeeds with Proper LF Permissions
    - Share is Revoked Successfully (no more permissions for Team Role, QS Group, + Pivot Role in EnvA)

- [x] 1 Table Shared 2 Times Cross Account to EnvA (QS enabled)
    - Both Shares Succeed
    - When 1 Share Revoke - no permissions for Team1 but QS Group and Team2 still can access Table in EnvA
    - Second Share Revoked Successfully

- [x] 2 Tables in Same Dataset Shared 1 Time Each to Cross Account to EnvA (QS enabled)
  - Both Shares Succeed
  - When 1 Share Revoke - no more permissions for Team Role, QS Group, + Pivot Role in EnvA for Table1 (still permissions for other table)
  - Second Share Revoked Successfully

backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
```diff
@@ -774,14 +774,20 @@ def find_all_share_items(session, share_uri, share_type):
...
-                    Environment.environmentUri == environment_uri,
-                    ShareObject.status == ShareObjectStatus.Approved.value,
-                    ShareObject.datasetUri == dataset_uri,
+                    ShareObject.environmentUri == environment_uri,
+                    ShareObjectItem.itemUri == item_uri,
+                    ShareObjectItem.shareItemUri != share_item_uri,
+                    ShareObjectItem.status.in_(share_item_shared_states),
                 )
             )
             .all()
```
dlpzx, line 793: A very minor improvement - Instead of returning all, we could use .first() and return the first appearance only for the case we have a lot of shares and share object items

backend/dataall/modules/dataset_sharing/services/share_managers/lf_share_manager.py
```diff
@@ -519,3 +514,13 @@ def glue_client(self):
...
+
+    def clean_up_share(self) -> bool:
+        """"
+        1) deletes deprecated shared db in target account
+        Returns
+        -------
+        True if clean-up succeeds
+        """
+        self.delete_shared_database()
+        return True
```
dlpzx, line 526: Maybe we can skip this function completely and directly call delete_shared_database in `dataall/modules/dataset_sharing/services/data_sharing_service.py:210`

backend/dataall/modules/dataset_sharing/services/share_processors/lf_process_cross_account_share.py
```diff
@@ -157,10 +157,23 @@ def process_revoked_shares(self) -> bool:
...
 
+                other_table_shares_in_env = False
+                if ShareObjectRepository.other_approved_share_item_table_exists(
+                    self.session,
+                    self.target_environment.environmentUri,
+                    share_item.itemUri,
+                    share_item.shareItemUri
+                ):
+                    other_table_shares_in_env = True
+                    principals = [p for p in principals if "arn:aws:quicksight" not in p]
```
dlpzx, line 168: I see now the issue that you commented about the Quicksight groups. You are right, there is one single QS group for the whole environment. Which means that removing permissions for the QS group should happen when we efectuate the "clean_up". The clean-up consists of deleting the sahred database, and I think by deleting the shared database the existing Lakeformation permissions also get deleted right? In that case we already got what we want.

backend/dataall/modules/dataset_sharing/services/share_managers/lf_share_manager.py
```diff
@@ -435,29 +431,28 @@ def revoke_external_account_access_on_source_account(self) -> [dict]:
...
+                    }
+                },
+                'Permissions': ['DESCRIBE', 'SELECT'],
+                'PermissionsWithGrantOption': ['DESCRIBE', 'SELECT'],
+            }
+        ]
+
+        LakeFormationClient.batch_revoke_permissions(
+            client, self.source_environment.AwsAccountId, revoke_entries
+        )
```
dlpzx, line 455: With one is enough? it makes sense as there is only one RAM invitation

dlpzx (commented):
Looks good, just a couple of minor comments

backend/dataall/modules/dataset_sharing/services/share_managers/lf_share_manager.py
noah-paige: This method is now called once for each table if that particular table is not being shared anywhere on the target account anymore

Before we were checking if the dataset is shared anywhere anymore and if not revoking each table's external target account access which could be problematic (doing dataset-level share checks and revoking on the table-level... explained more in later comment)

backend/dataall/modules/dataset_sharing/services/share_processors/lf_process_cross_account_share.py
noah-paige: For Each Revoked Table:
- Revoke Table Resource Link Access (ALL Principals)
- Revoke Source Table Access (ALL Principals)
- Delete Resource Link Table
- If no other existing shared Table Items in the share
    - Delete Shared DB
    - If no other approved share objects for dataset (i.e. `other_approved_share_object_exists`)
        - Revoke External Account Access

The Problems with the above:
- The method `other_approved_share_object_exists()` was filtering on Share Object Status of Approved (not getting all shared states so filtering incorrectly)
    - Meaning we were revoking target account access when there were still other shares for that dataset

- Could be prematurely removing QS Permissions (since 1 QS Group used for all teams in the env)
    - Could have QS Permissions revoked to tableA because 1 share revoked but another for a different team in same environment still exists and needs those QS Permissions

- If there was a share of 2 tables in 1 dataset to a team that were revoked at separate times
    - Since we iterate on `self.revoked_tables` to revoke external target account access to each table if no other shares for that dataset exist we would only catch the last table that is included in the self.revoked_table and may miss on some permissions to clean up
    - We are doing dataset-level share object exist checks but revoking external access on table-level


Now:
- Revoke Table Resource Link Access (ALL Principals)
- Check if the same table is currently shared to other teams in environment
    - If true revoke source table access only to IAM role (NOT QS if present)
    - If false revoke ALL principals (QS included) and revoke account access on the table-level
- Delete Resource Link Table
- If no other existing shared Table Items
    - Delete Shared DB

Fixes
- Fixes `other_approved_share_object_exists()` and makes check on table level to properly check if other share for that table item in that dataset exist on the target environment

- Does not remove QS Permissions on LakeFormation if there are other shares for that table in that environment

- Revokes external target account access on a table-level if there are no other shares on the table item in the target account
    - Table-level checks and table-level revokes on permissions

backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
noah-paige: good call - resolved

backend/dataall/modules/dataset_sharing/services/share_managers/lf_share_manager.py
noah-paige: Good idea - changed to directly call  `self.delete_shared_database()`

dlpzx:
@noah-paige needs to include #816 to avoid npm-audit issues

dlpzx (approved):

Issue #580, opened 2023-07-17
Limit Pivot Role S3 permissions


dlpzx:
🆕  [UPDATED WITH THE FEEDBACK FROM COMMENTS]
**Is your idea related to a problem? Please describe.**
Currently the data.all pivotRole requires permission to all S3 Buckets and KMS keys in the AWS account

**Describe the solution you'd like**
A solution in which access to the data on S3 buckets is restricted to specific roles only. We would like to prevent any data access from other accounts, especially if the pivot role gets compromised. 

---------------------
### In data.all central account
| role | used in | 
| ---- | ------- |
| graphql-role | Graphql Lambda function |
| worker-role | Worker Lambda function |
| esproxy-role | OpenSearch proxy Lambda function |
| ecs-role | ECS tasks|

### In the environment accounts
In addition there is also the cdk execution role used for CDK deployments, but it is not relevant for this issue.

| role | creation | permissions | trust policies (assumable by) |
| ---- | -------- | ------------ | -------------- |
| Pivot role | If AUTO-CREATE = True it is created as part of the environment stack, otherwise it is manually created | Read and write permissions to the TEAM-OWNED dataset buckets and to the Glue databases and tables + permissions to work with the enabled modules, but only to the TEAM-OWNED resources based on resource policies and | by `graphql-role`, `worker-role` and `ecs-role` from the central account |
| Environment teams' Roles | As part of the environment stack, one for each team that is added to the environment | Read and write permissions to the TEAM-OWNED dataset buckets and to the Glue databases and tables + permissions to work with the enabled modules, but only to the TEAM-OWNED resources based on resource policies and tags. |  by several module services + by the `pivot-Role` in the account |
| Dataset Role | As part of the dataset stack, for both imported and created datasets | Read and write permissions to the dataset bucket and to the Glue database and tables of the dataset + it can write profiling results in its corresponding folder in the environment bucket | by Glue + by the `pivot-Role` in the account |

In this diagram we can see all roles and the SDK calls that they need to perform. It includes the changes that we want to implement.

![IAMRoles drawio](https://github.com/awslabs/aws-dataall/assets/71252798/75a72591-1381-48b1-b8e7-b9f64c61828d)

---------------------

## Implementation
To be able to remove the S3 and KMS permissions from the pivotRole policies, we need to **remove the need** of these permissions and then modify the pivot role permissions. At the moment the pivot role is used to access data in the S3 Buckets in the following features:
[...]

PR #780, opened 2023-10-05
Feat: limit pivot role S3 permissions

dlpzx:
### Feature or Bugfix
- Feature

### Detail
The guiding principle is that:
1. dataset IAM role is the role accessing data
2. pivot role is the role used by the central account to perform SDK calls in the environment account

In this PR we
- Replace pivot role by dataset role in dataset Lake Formation registration
- Use pivot role to trigger upload files feature and create folder feature, but use the dataset IAM role to perform the putObject operations-> removes the need for read and `putObject` permissions. for the pivot role
- Redefine pivot role CDK stack to manage S3 buckets (bucket policies) for only the datasets S3 buckets that have been created or imported in the environment.
- implement IAM policy utils to handle the new dynamic policies. We need to verify that the created policy statements do not exceed the maximum policy size. In addition we replace the previous "divide in chunks of 10 statements" by a function that divides in chunks based on the size of the policy statements. This way we optimize the policy size, which helps us in reducing the number of managed policies attached to the pivot role. --> it can be re-used in other "chunkenization" of policies
- We did not implement force update of environments (pivot role nested stack) with new datasets added because it is already forced in `backend/dataall/modules/datasets/services/dataset_service.py`

### Backwards compatibility Testing

Pre-update setup:
- 1 environment (auto-created pivot role)
- 2 datasets in environment, 1 created, 1 imported: with tables and folders
- Run profiling jobs in tables

Update with the branch changes:
- [X] CICD pipeline runs successfully
- [X] Click update environment on environment -> successfully updated policy of pivot role with imported datasets in policy. Reduction of policies
- [X] Click update datasets --> registration in Lake formation updated to dataset role
- [X] Update files works
- [X] Create folder works
- [X] Crawler and profiling jobs work
 

### Relates
- #580 

### Security

  - Have you used the least-privilege principle? How? `In this PR we restrict the permissions of the pivot role, a super role that handles SDK calls in the environment accounts. Instead of granting permissions to all S3 buckets, we restrict it to data.all handled S3 buckets only`

### DIFF
--- a/backend/dataall/base/utils/iam_policy_utils.py
+++ b/backend/dataall/base/utils/iam_policy_utils.py
@@ -1,58 +1,95 @@
 from typing import List
+import logging
+from aws_cdk import aws_iam as iam
+
+logger = logging.getLogger(__name__)
 
 POLICY_LIMIT = 6144
+POLICY_HEADERS_BUFFER = 144
+MAXIMUM_NUMBER_MANAGED_POLICIES = 20  # Soft limit 10, hard limit 20
+
 
 def split_policy_statements_in_chunks(statements: List):
     """
-    Splits a list of IAM policy statements into a list of lists (chunks)
-    For each chunk, the size should remain below the POLICY LIMIT
+    Splitter used for IAM policies with an undefined number of statements
+    - Ensures that the size of the IAM policy remains below the POLICY LIMIT
+    - If it exceeds the POLICY LIMIT, it breaks the policy into multiple policies (chunks)
+    - Note the POLICY_HEADERS_BUFFER to account for the headers of the policy which usually take around ~60chars
     """
     chunks = []
     index = 0
-    print("Initial loop")
-    # Index = number of statements
-    print(f"statements = {len(statements)}")
+    statements_list_of_strings = [str(s.to_json()) for s in statements]
+    total_length = len(', '.join(statements_list_of_strings))
+    logger.info(f"Number of statements = {len(statements)}")
+    logger.info(f"Total length of statements = {total_length}")
+    max_length = max(statements_list_of_strings, key=len)
+    if len(max_length) > POLICY_LIMIT - POLICY_HEADERS_BUFFER:
+        raise Exception(f"Policy statement {max_length} exceeds maximum policy size")
     while index < len(statements):
         chunk = []
-        chunk_size = len(statements[index].to_string())
-        print(f"-----------------")
-        print(f"chunk = {chunk}")
-        print(f"chunk_size = {chunk_size}")
-        while chunk_size + len(statements[index].to_string()) < POLICY_LIMIT:
-            for statement in statements[index:]:
-                print(statement.to_string())
-                chunk.append(statement)
-                print(f"statement size= {len(statement.to_string())}")
-                chunk_size += len(statement.to_string())
-                index += 1
-                print(f"################")
-                #print(f"chunk = {chunk}")
-                print(f"chunk_size = {chunk_size}")
-                print(f"index={index}")
+        chunk_size = 0
+        while index < len(statements) and chunk_size + len(str(statements[index].to_json())) < POLICY_LIMIT - POLICY_HEADERS_BUFFER:
+            chunk.append(statements[index])
+            chunk_size += len(str(statements[index].to_json()))
+            index += 1
         chunks.append(chunk)
-        print(chunks)
-
+    logger.info(f"Total number of managed policies = {len(chunks)}")
+    if len(chunks) > MAXIMUM_NUMBER_MANAGED_POLICIES:
+        raise Exception("The number of policies calculated exceeds the allowed maximum number of managed policies")
     return chunks
 
 
-def split_policy_with_resources_in_statements(statement_without_resources, resources):
+def split_policy_with_resources_in_statements(base_sid, effect, actions, resources):
     """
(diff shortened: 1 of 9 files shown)


backend/dataall/base/utils/iam_policy_utils.py
```diff
@@ -0,0 +1,95 @@
...
+        return [resulting_statement]
+    else:
+        logger.info("Exceeding policy limit, splitting statement ...")
+        index = 0
+        split = 0
+        resulting_statements = []
+        while index < len(resources):
+            size = 0
+            res = []
+            while index < len(resources) and (size + len(resources[index]) + 5) < POLICY_LIMIT - POLICY_HEADERS_BUFFER - base_length:
```
noah-paige, line 82: what is the `+5` here? An additional buffer for resource list?

backend/dataall/modules/datasets/cdk/dataset_stack.py
```diff
@@ -242,7 +242,8 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
                     sid="ListDatasetBucket",
                     actions=[
                         "s3:ListBucket",
-                        "s3:GetBucketLocation"
+                        "s3:GetBucketLocation",
+                        "s3:GetBucketAcl"
```
noah-paige, line 246: Is this extra `s3:GetBucketAcl` permissions needed because Dataset Role now handling file uploads to S3 bucket using presigned url?

backend/dataall/modules/datasets_base/db/dataset_repositories.py
```diff
@@ -290,6 +290,27 @@ def query_environment_datasets(session, uri, filter) -> Query:
             )
         return query
 
+    @staticmethod
+    def query_environment_imported_datasets(session, uri, filter) -> Query:
```
noah-paige, line 294: Is there a reason we limit the restriction of S3 permissions to be only for imported datasets and for data.all created datasets we still have more general s3 permissions on a wildcard `resource_prefix` S3 ARN?

Have you looked into limiting s3 permissions to only the explicit set of s3 bucket ARNs that are mapped to data.all "Datasets" (either imported or created) or is there a risk of quickly reaching the policy size limit with that approach?

backend/dataall/base/utils/iam_policy_utils.py
```diff
@@ -0,0 +1,95 @@
+from typing import List
```
noah-paige, line 1: Can we maybe add a couple more comments on the double `while` loops to make easier to follow on what is going on / how we a determining the final policy statement set for each of the methods

noah-paige (commented):
Overall - I think the PR looks good. Left a few high level comments as per my understanding and will find some time likely tomorrow to do a deploy in my AWS data.all deployment with these code changes

Also - have we tested out `iam_policy_utils.py` for generating a policy that gets new or exceeds the 20 policy limit? Or have we tested with a policy with a resource list that needs to be split into 2 separate managed policies?

backend/dataall/base/utils/iam_policy_utils.py
dlpzx: It accounts for the additional 4 chars around a resource string. Having a list of resources ["res1", "res2", "res3"]. If we only account the chars of each of the items then we would be counting only res1res2res3. The `", "` between resources are 4 chars. Then at the beginning and end we have the extra " and [], that's why I added an extra +1 and rounded up to 5. I added a comment for this

backend/dataall/modules/datasets_base/db/dataset_repositories.py
dlpzx: I was worried about the policy size limitations, so to be optimal I only add the imported datasets to that policy. For the data.all created datasets there is already a policy that grants all those permissions. In `backend/dataall/core/environment/cdk/pivot_role_core_policies/s3.py` we are granting permissions to all data.all managed buckets. Besides optimizing the size of the policies and the amount of policies, I wanted to reduce the amount of times the pivot role is updated. It should not give any issues, but less CloudFormation operations are less opportunities for errors

```
            # S3 Managed Buckets
            iam.PolicyStatement(
                sid='ManagedBuckets',
                effect=iam.Effect.ALLOW,
                actions=[
                    's3:List*',
                    's3:Delete*',
                    's3:Get*',
                    's3:Put*'
                ],
                resources=[f'arn:aws:s3:::{self.env_resource_prefix}*'],
            ),
```

dlpzx:
@noah-paige I tested locally importing 20, 200, 400 and 500 datasets to stress out the policy utilities. 
Here is the result for the 500 imported datasets test:
image
image

When I tested in AWS I used 2 datasets only, do you think I should do more load testing?

backend/dataall/base/utils/iam_policy_utils.py
noah-paige: Makes sense - I think I was seeing `resources_str` include those additional characters for the `total_length` of the policy and makes sense then  checking for each resource to add to the statement

noah-paige:
Testing for 500 datasets locally is definitely sufficient in my view - wanted to make sure we had tested when reaching up to these edge cases that all works as expected

My understanding is that the functionality should be the exact same when forming the pivot role permission statements for either local or aws deployment so I think it is good!

noah-paige (approved):
Approving to merge to `v2m1m0` - lgtm!

Issue #655, opened 2023-08-14
Update Lambda runtime - before Nov 27, 2023


dlpzx:
### Describe the bug

Node [will deprecate some versions](https://github.com/nodejs/release#release-schedule) and after that AWS Lambda ([runtime deprecation plan](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html)) that are used as runtime for our Lambdas.

 **In Nov 27, 2023 the Lambda runtime node14 and Python3.7 will be deprecated!**

For this reason, we need to make sure that:
- we update the runtime versions in Lambdas created by data.all explicitly.
- we update the CDK version that we use regularly for Lambdas created by built-in features of CDK Constructs. 

### How to Reproduce

```
*P.S. Please do not attach files as it's considered a security risk. Add code snippets directly in the message body as much as possible.*
```

### OS

n/a

### Python version

n/a

### AWS data.all version

V2.0.0

dlpzx:
Merged and released with v2.1.0 🚀

PR #782, opened 2023-10-09
Update Lambda runtime

nikpodsh:
### Feature or Bugfix
Update

### Relates
See #655: 
> In Nov 27, 2023 the Lambda runtime node14 and Python3.7 will be deprecated!

Checked all lambdas that explicitly set the runtime engine: only cognito httpheader redirection lambda used node14.
All lambdas use python3.8 and node16 or node18.

For cdk dependencies: upgraded to a newest `aws-cdk-lib` `v2.99.0` just in case if python3.7 is hardcoded somewhere inside of 2.78.0 (shouldn't be)

### 
Testing: 
- [x] uploaded the changes to my isengard account
- [x] deployment is green 
- [x] could access app page, userguide page, and userguide from the app page.

### Security
`N/A` - upgraded to a newer version of node js

### DIFF
--- a/deploy/requirements.txt
+++ b/deploy/requirements.txt
@@ -1,4 +1,4 @@
-aws-cdk-lib==2.78.0
+aws-cdk-lib==2.99.0
 boto3-stubs==1.20.20
 boto3==1.24.85
 botocore==1.27.85
--- a/backend/dataall/base/cdkproxy/requirements.txt
+++ b/backend/dataall/base/cdkproxy/requirements.txt
@@ -1,4 +1,4 @@
-aws-cdk-lib==2.83.1
+aws-cdk-lib==2.99.0
 boto3==1.24.85
 boto3-stubs==1.24.85
 botocore==1.27.85
--- a/deploy/setup.py
+++ b/deploy/setup.py
@@ -11,5 +11,5 @@ setuptools.setup(
     author=__author__,
     package_dir={'': 'stacks'},
     packages=setuptools.find_packages(where='stacks'),
-    python_requires='>=3.7',
+    python_requires='>=3.8',
 )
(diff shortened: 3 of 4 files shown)


deploy/requirements.txt
```diff
@@ -1,4 +1,4 @@
-aws-cdk-lib==2.78.0
+aws-cdk-lib==2.99.0
```
noah-paige, line 1: Could we also update the `backend/dataall/base/cdkproxy/requirements.txt` to have the same dependency for `aws-cdk-lib==2.99.0` 

I was able to deploy the code changes listed here as well as the change to `aws-cdk-lib==2.99.0`  in the other `requirements.txt` as well successfully but do you foresee any issues doing so @dlpzx @nikpodsh ?

I will test accessing the frontend and deploying a stack to ensure all still works as well - I think these code changes look good but could we point them to `v2m1m0` branch?

noah-paige: I also tested creating/deploying the dataset stack - I would think be best to test create/delete of all the stack resources (i.e. env, notebook, ml studio, pipelines) if we do update the `aws-cdk-lib` requirement on cdkproxy as well 

This is something we would do holistically for validation testing on `v2m1m0` branch when ready for release if we do point this PR to that branch and add that additional code change but let me know your all thoughts when you get a chance

noah-paige: Lastly confirmed no Lambda Runtimes using Node14 of Python3.7 after the PR code changes implemented

dlpzx: I agree with updating `backend/dataall/base/cdkproxy/requirements.txt`. For testing all functionalities I would go for the final review in V2.1 validation. Otherwise it looks good to me

nikpodsh: Agreed, I missed that cdkproxy is also using it. Updated!

noah-paige (approved):
Looks good to me to merge to `v2m1m0` - thanks @nikpodsh !!

Issue #785, opened 2023-10-11
Refactor and migrate notifications from core to modules


dlpzx:
**Is your idea related to a problem? Please describe.**
The notification module in `core` is very specific to datasets, which goes against the design principles of keeping in modules the module specific functionalities and in core the bare minimum application modules. For this reason we would like to refactor the notifications code and turn it into a module the same way it is done for `feeds` or `votes`

**Describe the solution you'd like**
The list of tasks that need to be complete are:
- Add a registry in notifications and move it from core to modules
- Add the dependencies in the data_sharing or in the dataset module init file
- In the config.json, we need to declare the module data_sharing and add the feature notifications to it. Instead of doing it from core . Alternatively we could do it from the datasets module, whatever works best.
- Code for SES notifications developed in #734 would be divided into the notifications module and the way we register the notifications in the dataset/data_sharing module

dlpzx:
Merged and released with v2.1.0 🚀

Issue #734, opened 2023-09-06
Email Notification for Actions on Share Requests


TejasRGitHub:
**Is your idea related to a problem? Please describe.**
Whenever a share request is created and transitions from states ( approved, revoked, etc ) a notification is created. This notification is displayed on the bell icon on the UI . 

We want such a similar notification to be sent to the dataset owner, requester, etc via email 

**Describe the solution you'd like**
An email is sent on creation and on any transition of the share request to the dataset owner, requester, etc

A solution consisting of using SNS topic - for which dataset owner, requester of the dataset are subscribed - where any notification which is sent to the topic gets delivered to the email addresses could be one potential way to implement this.

Further, a way to filter recipient email list where email about the share request are sent to only specific email ids ( from the environment teams ) would be nice to have.

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

dlpzx:
Hi @TejasRGitHub , thanks for opening an issue this is a great feature idea. 

First, we need to define the actions that are going to be notified:
- share workflow actions - what you describe on approved, revoked...
- dataset updates - in addition, do we want requesters to get notified when a dataset gets new tables (for example)

Secondly, and this is more complex, we need to decide on the subscription architecture. Presumably we'll be using AWS SNS to get the notifications from the ECS share task + GraphQL API calls. I can brainstorm different alternatives, let me know if this is aligned with what you have in mind:
- One SNS topics in data.all infrastructure account
- One SNS topic per environment (in data.all environment account)
- One SNS topics per team (in data.all environment account)

### One SNS topics in data.all infrastructure account
PROS: 
- easy to send notifications, all go to the same topic

CONS: 
- end-users have not access to the topic, cannot add subscribers. 
- [Service quota limitations](https://docs.aws.amazon.com/general/latest/gr/sns.html#limits_sns), at first sight it does not look like we are going to hit them in the near future, but with a single-thread architecture we should always review just in case
- Needs implementation of [SNS filter policies](https://docs.aws.amazon.com/sns/latest/dg/sns-subscription-filter-policies.html) to segregate by all teams

### One SNS topic per environment
PROS:
- users have access to the topic
- more scalable
- Already part of environment stack --> subscriptions enabled!

CONS:
- Needs some implementation of [SNS filter policies](https://docs.aws.amazon.com/sns/latest/dg/sns-subscription-filter-policies.html) to segregate by teams in the environment
[...]

PR #790, opened 2023-10-13
Feat: Move notifications from core to modules to facilitate customization

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail
As a rule of thumb, we encourage customization of `modules` while changes in `core` should be avoided when possible. `notifications` is a component initially in core which is only used by `dataset_sharing`. To facilitate customization of the `notifications` module and also to clearly see its dependencies we have:

- Moved `notifications` code from core to modules as it is a reusable component that is not needed by any core component.
- Moved dataset_sharing references inside dataset_sharing module and left `notifications` independent from any other module
- Added depends_on in the dataset_sharing module to load notifications if the data_sharing module is imported.

# TODO:
- frontend changes in case notifications are not loaded
- not for this PR, but as a general note, we should clean up deprecated ECS tasks

### Relates
- #785 
- #734 

### Security

`N/A` just refactoring


dlpzx:
@noah-paige and I discussed internally that this PR is just for reference and will stay in Draft. First the PR for #734 needs to be merged and then we will do the necessary changes which developed in this Draft PR.

PR #794, opened 2023-10-15
Fix: condition when there are no public subnets

### LINKED ISSUE #792: Check public subnets before output in CloudFormation
**Is your idea related to a problem? Please describe.**
data.all assumes presence of public subnets for all VPCs, also those that are provided by the customer as input in the `cdk.json` file. In some cases customers might not have public subnets and have other ways of routing traffic to the internet. In those cases, 
deploy/stacks/vpc.py fails on the CfnOutput block when the VPC doesn’t have public subnets.

**Describe the solution you'd like**
Check if the VPC has public subnets and create the CFN output only if those exist.

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

lorchda:
### Feature or Bugfix

Fix #792.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,10 @@ documentation, we greatly value feedback and contributions from our community.
 Please read through this document before submitting any issues or pull requests to ensure we have all the necessary
 information to effectively respond to your bug report or contribution.
 
+## First Things First
+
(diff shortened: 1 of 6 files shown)


noah-paige (approved):
Deployed these changes to an AWS Deployment and CICD Completed successfully.

No changeset for the Cloudformation VPC as expected 

Looks good to me - will approve

@dlpzx would we want to point these changes to `v2m1m0` release branch instead of directly to `main`

dlpzx:
Yesterday we discussed internally and we will merge it to the release branch as it is inminent

PR #811, opened 2023-10-16
feat: Handle Pre-filtering of tables

anushka-singh:
### Feature or Bugfix

- Feature

### Detail
- For a dataset to make sense all the tables within a dataset should have their location pointing to the same place as the dataset S3 bucket. However it is possible that a database can have tables which do not point to the same bucket which is perfectly legal in LakeFormation. Therefore we propose that data.all automatically only lists tables that have the same S3 bucket location as the dataset. This will solve a problem for Yahoo where we want to import a database that contains many tables with different buckets. Additionally Catalog UI should also only list prefiltered tables.

### Testing
- Tested this in local env. I was able to create and share datasets even after pre-filtering process takes place.
- Will send separate PR for unit testing. 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/datasets/api/table/mutations.py
+++ b/backend/dataall/modules/datasets/api/table/mutations.py
@@ -23,7 +23,9 @@ deleteDatasetTable = gql.MutationField(
 
 syncTables = gql.MutationField(
     name='syncTables',
-    args=[gql.Argument(name='datasetUri', type=gql.NonNullableType(gql.String))],
+    args=[
+        gql.Argument(name='datasetUri', type=gql.NonNullableType(gql.String))
+    ],
     type=gql.Ref('DatasetTableSearchResult'),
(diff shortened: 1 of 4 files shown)


noah-paige (approved):
Tested the following in AWS Deployment

Pre-existing dataset with 2 tables:
- [x] Dataset Synchronize successfully and loads 2 tables

Pre-existing dataset with 0 table:
- [x] Dataset Synchronize successfully and loads 0 tables

Import new dataset (2 tables in 2 different buckets)
- [x] Synchronize only loads 1 table
- [x] Share Request for dataset with tables filtered and can only add the 1 table
- [x] Create worksheet and can only query the 1 table 

ECS Table Sync Task
- [x] Successfully Runs and finds appropriate number of tables for each dataset

I think the code changes look good to me and UI working as expected! - approving to merge to `v2m1m0`

Issue #797, opened 2023-10-16
Make DNS alias record creation toggleable / optional


dlpzx:
**Is your idea related to a problem? Please describe.**
data.all assumes that we can create Route 53 records in customer environment. But this a restricted action that is forbidden with SCPs:
 ```
API: route53:GetHostedZone User: aws:sts::111111111111:assumed-role/cdk-hnb659fds-cfn-exec-role-111111111111-eu-central-1/AWSCloudFormation is not authorized to perform: route53:GetHostedZone on resource: arn:aws:route53mhostedzone/XXXXX with an explicit deny in a service control policy. 
```

**Describe the solution you'd like**

I commented out hosted zone requirement in `deploy/stacks/albfront_stack.py` but we need to evaluate the impact of this action. So it first needs investigation, and if it can be avoided, we would like a solution that makes the creation of DNS alias optional.

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

lorchda:
Make `hosted_zone_name` and `hosted_zone_id` configuration parameters optional.

The `ecs_patterns.ApplicationLoadBalancedFargateService` that you are using in https://github.com/awslabs/aws-dataall/blob/main/deploy/stacks/albfront_stack.py does not require the `domain_zone` parameter (optional), so it can be left out.

lorchda:
See description in #812 for more details.

lorchda:
+ documentation update

lorchda:
Documentation update.

dlpzx:
Merged and released with v2.1.0 🚀

PR #812, opened 2023-10-17
Make hosted_zone_id optional, code update

lorchda:
### Feature or Bugfix
- Bugfix

### Detail
- Make `hosted_zone_id` optional, code update

### Relates
- #797 

### Security

fetching data from storage outside the application (e.g. a database, an S3 bucket)?  N/A
  - Is the input sanitized? N/A

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. YES

### Description

Make `hosted_zone_id` optional and provide `HostedZoneId` and `DNSName` in CloudFormation Stack Output, so users can create their own [Route53 AliasTarget](https://docs.aws.amazon.com/Route53/latest/APIReference/API_AliasTarget.html).

Following validation checks in `ecs_patterns.ApplicationLoadBalancedFargateService` were considered:
* `frontend_alternate_domain` and `userguide_alternate_domain` have to be `None` when the `hosted_zone` is `None`, see checks in [multiple-target-groups-service-base.ts#L463](https://github.com/aws/aws-cdk/blob/c445b8cc6e20d17e4a536f17262646b291a0fe36/packages/aws-cdk-lib/aws-ecs-patterns/lib/base/network-multiple-target-groups-service-base.ts#L463), or else a `A Route53 hosted domain zone name is required to configure the specified domain name` error is raised
* for a HTTPS ALB listener, only the `certificate` is ultimately required, and not the `domainName` or `domainZone`, as per evaluation logic in [application-load-balanced-service-base.ts#L509](https://github.com/aws/aws-cdk/blob/c445b8cc6e20d17e4a536f17262646b291a0fe36/packages/aws-cdk-lib/aws-ecs-patterns/lib/base/application-load-balanced-service-base.ts#L509)

### DIFF
--- a/deploy/stacks/albfront_stack.py
+++ b/deploy/stacks/albfront_stack.py
@@ -149,7 +149,7 @@ class AlbFrontStack(Stack):
         if custom_domain and custom_domain.get('certificate_arn'):
             certificate = acm.Certificate.from_certificate_arn(self, "CustomDomainCertificate",
                                                                custom_domain.get('certificate_arn'))
-        else:
+        elif custom_domain and custom_domain.get('hosted_zone_name'):
             certificate = acm.Certificate(
                 self,
                 'CustomDomainCertificate',
@@ -157,6 +157,8 @@ class AlbFrontStack(Stack):
                 subject_alternative_names=[f'*.{custom_domain["hosted_zone_name"]}'],
                 validation=acm.CertificateValidation.from_dns(hosted_zone=hosted_zone),
             )
+        else:
+            raise ValueError("Configuration parameter custom_domain['hosted_zone_name'] in cdk.json is REQUIRED if internet_facing=false")
 
         frontend_sg = ec2.SecurityGroup(
             self,
--- a/deploy/stacks/albfront_stack.py
+++ b/deploy/stacks/albfront_stack.py
@@ -10,6 +10,7 @@ from aws_cdk import (
     aws_elasticloadbalancing as elb,
     aws_s3 as s3,
     Stack,
+    CfnOutput,
     Duration,
     RemovalPolicy,
     Fn,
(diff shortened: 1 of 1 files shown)


lorchda:
This is what the [documentation](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_ecs_patterns/ApplicationLoadBalancedFargateService.html) says for the parameters to `ecs_patterns.ApplicationLoadBalancedFargateService`:

* domain_name (Optional[str]) – The domain name for the service, e.g. “api.example.com.”.
  Default: - No domain name.
* domain_zone (Optional[IHostedZone]) – The Route53 hosted zone for the domain, e.g. “example.com.”.
  Default: - No Route53 hosted domain zone.

dlpzx:
Hi @lorchda, I am looking into this PR today. The albfront stack will only be deployed for VPC-facing frontend. In our documentation we specify that the custom_domain MUST be specified for this type of deployment. 

image

From the code I think there is going to be an issue in lines 149-159

lorchda:
@dlpzx yes noticed that as well. My thinking was that `custom_domain['hosted_zone_name']` would raise a not very descriptive `KeyError` along with a stack trace, but would otherwise abort as desired. This could be improved by providing a user-friendly error message that checks for the presence of `custom_domain['hosted_zone_name']`.

However, this is an unrelated issue that already existed *before* this PR. This PR is related only to `hosted_zoned_id`, and the semantics around `hosted_zone_name` (which *must* be provided) remains the same.

Example configuration:

```json
    "DeploymentEnvironments": [
      {
        "internet_facing": false,
        "custom_domain": {
          "hosted_zone_name": "dataall.example.internal",
          "certificate_arn": "arn:aws:acm:REGION:ACCOUNT_ID:certificate/UID"
        }
        ...
      }
    ]
```

lorchda:
`custom_domain['hosted_zone_name']`
raises a `KeyError` when the key does not exist

`custom_domain.get('hosted_zone_name')`
returns `None` if the key does not exist

dlpzx:
Hi @lorchda, thanks for the quick response. It is true that the certificate problem is more of a cosmetic issue, but based on our experience with customers we really need to provide an error message to users for them to understand the issue. This problem was not there before because `custom_domain['hosted_zone_name']` was mandatory (at least in the documentation) for an internet_facing=false.

It is still unclear to me what your final architecture looks like. The target is to not provide any route53 zone or certificate and have the ALB defined without those optional parameters right? In your cdk.json the parameter `"custom_domain"` does not exist right?

lorchda:
@dlpzx I added a fix for this (unrelated) issue to check for `custom_domain['hosted_zone_name']`

dlpzx:
Thank you @lorchda, the PR looks good. I will do a final test by deploying it to AWS and we can merge

dlpzx (approved):

PR #817, opened 2023-10-18
feat: adding frontend and backend feature flags

zsaltys:
### Feature or Bugfix
- Feature

### Detail
- Adding frontend support for all feature flags defined in config.json with a new util method isFeatureEnabled
- Adding a new flag **preview_data** in the datasets module to control whether previewing data is allowed
- Adding a new flag **glue_crawler** in the datasets module to control whether running glue crawler is allowed
- Updating environment features to be hidden or visible based on whether the module is active. Adding a new util isAnyFeatureModuleEnabled to check whether to render the entire feature box.

### Relates
N/A

### Security
Not relevant

### DIFF
--- a/backend/dataall/modules/datasets/api/dataset/resolvers.py
+++ b/backend/dataall/modules/datasets/api/dataset/resolvers.py
@@ -127,6 +127,7 @@ def get_dataset_statistics(context: Context, source: Dataset, **kwargs):
 def get_dataset_assume_role_url(context: Context, source, datasetUri: str = None):
     return DatasetService.get_dataset_assume_role_url(uri=datasetUri)
 
+
 @is_feature_enabled('modules.datasets.features.glue_crawler')
 def start_crawler(context: Context, source, datasetUri: str, input: dict = None):
     return DatasetService.start_crawler(uri=datasetUri, data=input)
--- a/backend/dataall/modules/datasets/api/dataset/resolvers.py
+++ b/backend/dataall/modules/datasets/api/dataset/resolvers.py
@@ -127,7 +127,7 @@ def get_dataset_statistics(context: Context, source: Dataset, **kwargs):
 def get_dataset_assume_role_url(context: Context, source, datasetUri: str = None):
     return DatasetService.get_dataset_assume_role_url(uri=datasetUri)
 
-
+@is_feature_enabled('modules.datasets.features.glue_crawler')
 def start_crawler(context: Context, source, datasetUri: str, input: dict = None):
     return DatasetService.start_crawler(uri=datasetUri, data=input)
 
--- a/backend/dataall/modules/datasets/api/table/resolvers.py
+++ b/backend/dataall/modules/datasets/api/table/resolvers.py
@@ -1,5 +1,6 @@
 import logging
 
+from dataall.core.feature_toggle_checker import is_feature_enabled
 from dataall.modules.catalog.db.glossary_repositories import GlossaryRepository
 from dataall.modules.datasets.api.dataset.resolvers import get_dataset
 from dataall.base.api.context import Context
@@ -22,7 +23,7 @@ def delete_table(context, source, tableUri: str = None):
         return False
     return DatasetTableService.delete_table(uri=tableUri)
 
-
+@is_feature_enabled('modules.datasets.features.preview_data')
 def preview(context, source, tableUri: str = None):
     if not tableUri:
         return None
--- a/backend/dataall/modules/datasets/api/table/resolvers.py
+++ b/backend/dataall/modules/datasets/api/table/resolvers.py
@@ -23,6 +23,7 @@ def delete_table(context, source, tableUri: str = None):
         return False
     return DatasetTableService.delete_table(uri=tableUri)
 
+
 @is_feature_enabled('modules.datasets.features.preview_data')
 def preview(context, source, tableUri: str = None):
     if not tableUri:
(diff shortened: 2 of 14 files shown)


frontend/src/utils/helpers/moduleUtils.js
```diff
@@ -33,6 +33,15 @@ function isModuleEnabled(module) {
   return getModuleActiveStatus(module);
 }
 
+function isAnyFeatureModuleEnabled() {
```
zsaltys, line 36: I'm a little unsure about this method.. We basically have no nice way to tell if a module is a feature module or not. If we want to avoid this hardcoded if.. We could add a flag into the config.json that a module is considered a feature module... 

Or perhaps make a new component like EnvironmentFeatures.js but with Toggles? And loop through the array? Let me know what you prefer.

frontend/src/modules/Environments/components/EnvironmentTeams.js
```diff
@@ -215,9 +193,34 @@ function TeamRow({ team, environment, fetchItems }) {
...
+              <CopyAllOutlined
+                sx={{
+                  color:
+                    theme.palette.mode === 'dark'
+                      ? theme.palette.primary.contrastText
+                      : theme.palette.primary.main
+                }}
+              />
+            </LoadingButton>
+            {team.groupUri !== environment.SamlGroupName && (
```
noah-paige, line 209: Would this also remove the ability the `removeGroup` from the environment? Should this be loaded regardless of `isFeatureEnabled('core', 'env_aws_actions')`?

backend/dataall/modules/datasets/api/dataset/resolvers.py
```diff
@@ -127,7 +127,7 @@ def get_dataset_statistics(context: Context, source: Dataset, **kwargs):
 def get_dataset_assume_role_url(context: Context, source, datasetUri: str = None):
     return DatasetService.get_dataset_assume_role_url(uri=datasetUri)
 
-
+@is_feature_enabled('modules.datasets.features.glue_crawler')
```
noah-paige, line 130: Linting Error - expected 2 blank lines, found 1

backend/dataall/modules/datasets/api/table/resolvers.py
```diff
@@ -22,7 +23,7 @@ def delete_table(context, source, tableUri: str = None):
         return False
     return DatasetTableService.delete_table(uri=tableUri)
 
-
+@is_feature_enabled('modules.datasets.features.preview_data')
```
noah-paige, line 26: expected 2 blank lines, found 1

noah-paige:
I tested in an AWS Deployment with the feature flags both on and off and is working as expected

One other thing I noted is that we use default `useState('preview')` in `TableView.js` to default to the Preview Tab. If that tab is disabled then when as a dataset owner navigating to the path `/console/datasets/table/${table.tableUri}` it will first have nothing but the tab options showing which I user will then have to select (see screenshot)

Screenshot 2023-10-23 at 4 54 26 PM

Alternatively, we can make overview the default Table tab opened (i.e. `useState('overview')`) - but trying to think through if this would cause any issues for a Requestor since when a requestor navigates to the catalog and clicks on a table to inspect from the catalog they also get navigated to `/console/datasets/table/${table.tableUri}`

backend/dataall/modules/datasets/api/dataset/resolvers.py
zsaltys: fixed

backend/dataall/modules/datasets/api/table/resolvers.py
zsaltys: fixed

frontend/src/modules/Environments/components/EnvironmentTeams.js
zsaltys: @noah-paige apologies! did not realize this was happening. Fixed!

zsaltys:
> useState('preview')

frontend/src/modules/Tables/views/TableView.js
```diff
@@ -157,7 +163,7 @@ const TableView = () => {
   const client = useClient();
   const navigate = useNavigate();
   const [table, setTable] = useState({});
-  const [currentTab, setCurrentTab] = useState('preview');
+  const [currentTab, setCurrentTab] = useState(tabs[0].value);
```
zsaltys, line 166: @noah-paige Actually this was a mistake on my side... When contributing back Mohit's changes I forgot to pick this up from our branch. This will now always default to the first tab on the UI.

noah-paige (approved):
Deployed these changes to AWS and tested with all the feature flags enabled and disabled - all working as expected

Also tested disabling a module and it is removed from the Env Feature Management card appropriately

noah-paige:
Thank you @zsaltys for this PR - I will merge this through to `v2m1m0` tomorrow barring any additional review comments found by @dlpzx

frontend/src/utils/helpers/moduleUtils.js
dlpzx: To this issue we also have the challenge of managing dependencies. At the moment they are hardcoded in the `isModuleEnabled` function. Maybe we can think of a generic way of defining some module properties in the frontend as we do in the `__init__` files for backend. I spoke about this with Mo as one of the future improvements for frontend. For the moment i think it is good

PR #818, opened 2023-10-18
Email Notification on Share Workflow - Issue - 734

TejasRGitHub:
### Feature or Bugfix
- Feature

### Detail

Whenever a share request is created and transitions from states ( approved, revoked, etc ) a notification is created. This notification is displayed on the bell icon on the UI .

We want such a similar notification to be sent to the dataset owner, requester, etc via email

Please take a look at Github Issue 734 For more details - https://github.com/awslabs/aws-dataall/issues/734

### Relates
- https://github.com/awslabs/aws-dataall/issues/734

### Security

  - Is the input sanitized?
  - Have you used the least-privilege principle? How? --> **Permission granted for SES:sendEmail to Lambda on resources - (Ses identity and configuration set ) , Also created KMS and SNS for SES setup to handle email bounces . Used least privleged and restricted access on both whenever required. **


backend/api_handler.py
```diff
@@ -115,6 +115,7 @@ def handler(event, context):
 
     if 'authorizer' in event['requestContext']:
         username = event['requestContext']['authorizer']['claims']['email']
+        email_id = event['requestContext']['authorizer']['claims']['email']
```
dlpzx, line 118: Why don't we use the username? It seems to be have the same value as the email. Is this to make it clear that we the variable contains the email?

backend/dataall/base/services/IdentityProvider.py
```diff
@@ -0,0 +1,8 @@
+import abc
```
dlpzx, line 1: Here we are introducing a new directory in the `base` package called `services`. I see where the name comes from, we are defining "non-aws generic services". But since we are already using `services` quite a lot in the repo, when we refer to business logic of a particular data.all object, I think the name might lead to confusion.

dlpzx: - Option 1: we rename the `services` directory
- Option 2: The package `aws` could be called something else like `cloud` and would account for all cloud-services. That way we could define IdentityProviders as part of the newly-named `aws` package. Since, right now IdentityProviders is only used for Cognito. I suggest that for the time being we move it to `aws`.
- Option 3: we move IdentityProviders to move to `utils`

backend/dataall/core/notifications/services/email_notification_provider.py
```diff
@@ -0,0 +1,77 @@
...
+        return email_list
+
+    @staticmethod
+    def get_email_provider_instance(datasetOwnerGroup, datasetStewards, requesterGroupName, requesterEmailId):
+        return EmailNotificationProvider(datasetOwnerGroup, datasetStewards, requesterGroupName, requesterEmailId, Ses.get_ses_client())
+
+    @staticmethod
+    def send_email_task(requesterEmailId, subject, message, data):
+        # Get instance of the email provider
+        email_provider = EmailNotificationProvider.get_email_provider_instance(
```
dlpzx, line 42: The data dict is very specific to share notifications. Can we make it more generic? Like data.get('senderGroups'). data.get('recipientGroup')

dlpzx: We can do it in this PR or in the one that I am working on to move notifications to modules as a standalone module that does not depend on datasets. What do you think?

backend/dataall/modules/dataset_sharing/services/share_item_service.py
```diff
@@ -63,7 +63,7 @@ def revoke_items_share_object(uri, revoked_uris):
                 resource_uri=dataset.datasetUri,
             )
 
-            ShareNotificationService.notify_share_object_rejection(session, context.username, dataset, share)
+            ShareNotificationService.notify_share_object_rejection(session, context.username, context.email_id, dataset, share)
```
dlpzx, line 66: Very similar to the inital definition of email_id in context. Is there a difference in other IdPs between the username and the email?

deploy/stacks/backend_stack.py
```diff
@@ -127,6 +132,12 @@ def __init__(
             **kwargs,
         )
 
+        # Create the SES Stack
+        ses_stack = None
```
dlpzx, line 136: Unused variable `ses_stack`

deploy/stacks/backend_stack.py
```diff
@@ -127,6 +132,12 @@ def __init__(
...
+        if custom_domain == None and email_notification_sender_email_id == None and deploy_config.get_property("core.share_notifications.email.active") == True :
+            raise Exception("Cannot Create SES Stack For email notification as Custom Domain and Sender Email Id are not present. Either Disable Email Notification Config or add Custom Domain")
+        ses_stack = self.create_ses_stack(custom_domain, envname, kwargs, resource_prefix)
```
dlpzx, line 139: In the if-clause in line 137 I think there is a problem. 
When deploy_config.get_property("core.share_notifications.email.active") == False, the if condition is not true, the exception is not raised and we end up creating the ses stack.

# Create the SES Stack
        ses_stack = None
        if deploy_config.get_property("core.share_notifications.email.active") == True :
            if custom_domain == None and email_notification_sender_email_id == None:
                raise Exception(".....")
            ses_stack = self.create_ses_stack(custom_domain, envname, kwargs, resource_prefix)

backend/dataall/modules/dataset_sharing/services/share_notification_service.py
```diff
@@ -96,3 +128,33 @@ def _get_dataset_stewards(dataset):
...
+
+                notification_task: Task = Task(
+                    action='notification.service',
+                    targetUri=share.shareUri,
+                    payload={
+                        'notificationType' : share_notification_config_type,
+                        'shareRequestUserEmail': email_id,
+                        'subject': subject,
+                        'message': msg,
+                        'requesterGroupName': requester_group_name,
```
dlpzx, line 150: Ideally the payload does not include any logic relative to the datasets and datasharing. Since we already have created the message `requesterGroupName`, `datasetOwnerGroup` and `datasetStewardsGroup` could be passed as a single field `recipients`. We will need to do the check `notification_configs['email']['features']['group_notifications'] == True:` in email_notification_provider line 56 in the share_notification service

dlpzx (changes requested):
Thanks for the PR! There are some comments related to the overall structure and some minor things on naming conventions.

backend/api_handler.py
TejasRGitHub: Yes, this was to make it clear that the variable contains email_id. As we are fetching the email attribute from the authorizer claim. I didn't want to refactor the username field and change it everywhere so I created a new field

backend/dataall/base/services/IdentityProvider.py
TejasRGitHub: I like option 3. I will place the IdentityProviders File in utils. I have updated this in next commit

backend/dataall/core/notifications/services/email_notification_provider.py
TejasRGitHub: As discussed, I have handled and refactored such that it becomes generic. Now there will will be two fields i.e. recipient_group_list and recipient_email_list ( in case group_notifications config is false this variable will contain the requester email id ) . These fields will be instead of the current fields like datasetOwnerGroup, datasetStewardsGroup, requesterGroupName

backend/dataall/core/notifications/services/email_notification_service.py
```diff
@@ -0,0 +1,8 @@
+import abc
+
+
+class EmailNotificationService:
+
+    @abc.abstractmethod
+    def send_email(self, to, message, subject):
+        raise NotImplementedError
```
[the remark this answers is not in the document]
TejasRGitHub: Updated in the next commit . Agreed

backend/dataall/modules/dataset_sharing/services/share_item_service.py
TejasRGitHub: I created this new context field so as to have clear usage of what is being passed. At that time I wanted to rename username to email_id but that would have added some more refactoring into entire data.all backend code where the `username` field is used.

backend/dataall/modules/dataset_sharing/services/share_notification_service.py
TejasRGitHub: As discussed made changes in the next commit

TejasRGitHub:
> Thanks for the PR! There are some comments related to the overall structure and some minor things on naming conventions.

Hi @dlpzx  . Thanks for the in-depth review. I have updated the code to reflect the changed asked in the comments. After the changes , I have carried out one round of sanity tests on my local dev environment.

backend/dataall/core/notifications/services/email_notification_provider.py
dlpzx: Awesome!!

dlpzx:
Hi @TejasRGitHub I have reviewed your changes, looking good! I will approve and test together with #790

dlpzx (approved):

deploy/stacks/ses_stack.py
```diff
@@ -0,0 +1,92 @@
...
+
+        self.KMS_SNS.add_to_resource_policy(
+            iam.PolicyStatement(
+                effect=iam.Effect.ALLOW,
+                actions=[
+                    "kms:Decrypt",
+                    "kms:GenerateDataKey*"
+                ],
+                principals=[
+                    iam.AnyPrincipal()
```
noah-paige, line 74: Does this require `AnyPrincipal()` - can we not restrict this to `AccountPrincipal()` at the minimum - seems very open

TejasRGitHub: Hi @noah-paige , we could actually add the ARN in the `conditions` to the principals and remove the `AnyPrincipal()`, 

@dlpzx , do you mind doing this small change when you are moving notifications from core to modules ?

dlpzx: Yes I will add this change in #822

deploy/stacks/ses_stack.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class SesStack(pyNestedClass):
+    def __init__(
+            self,
+            scope,
+            id,
+            envname='dev',
+            resource_prefix='dataall',
+            custom_domain=None,
```
dlpzx, line 22: @TejasRGitHub, why don't we pass the email id here? Aren't we going to use it in the EmailIdentity creation as identity parameter?

deploy/stacks/ses_stack.py
```diff
@@ -0,0 +1,92 @@
...
+            self,
+            f'{resource_prefix}-{envname}-SNS-Email-Bounce-Topic',
+            display_name="SNS-Email-Bounce-Topic",
+            topic_name=f'{resource_prefix}-{envname}-SNS-Email-Bounce-Topic',
+            master_key=self.KMS_SNS,
+        )
+
+        self.sns.apply_removal_policy(RemovalPolicy.DESTROY)
+
+        hosted_zone = route53.HostedZone.from_hosted_zone_attributes(
```
dlpzx, line 45: If we use email_id instead of custom_domain, and for some reason the custom domain is not there then this piece of code will make the stack fail. We need to add a conditional block and in case it is not there use the email_id in identity below

dlpzx: I fixed it in this commit: https://github.com/awslabs/aws-dataall/pull/822/commits/ee2cc20d3ea57a3af053a6d024d413e9826e6102 let me know if you agree with the changes

deploy/stacks/ses_stack.py
dlpzx: Done in this [commit](https://github.com/awslabs/aws-dataall/pull/822/commits/7a3e481e35ad5654c83e3585bda71195ed9ce7fd)

deploy/stacks/ses_stack.py
TejasRGitHub: Hi @dlpzx , the email id is not used to create an email identity. We are only using the email id as a sender email id. The way currently SES stack is working , is that it is verifying the custom_domain and not the email id. When the custom_domain is verified any email id on that domain can be used to send an email. Also, verifying an email id could be possible in case the user has access to the inbox of that email id. Apart from that if an email id is verified then the user also has to make entries for SPF and DKIM into their DNS server (https://docs.aws.amazon.com/ses/latest/dg/email-authentication-methods.html) for the recipient to receive the email properly ( without it being sent to spam)

deploy/stacks/ses_stack.py
TejasRGitHub: Currently, only the custom_domain is used to verify with SES. My intent was not to verify the email address as identity as, that would mean that the user has to verify the email address and also make couple of entires in their DNS server ( for DKIM and SPF ). 

For example, if the user wants to use an email ids like, data-all-notifications@some-non-aws-domain.com. Then in this case the user has to verify that the email id belongs to him by clicking on the verification email and also authenticate the email address with DKIM and SPF, for which the user needs to have access to the DNS server in someway

PR #822, opened 2023-10-20
Feat: Refactor notifications from core to modules

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail
As a rule of thumb, we encourage customization of `modules` while changes in `core` should be avoided when possible. `notifications` is a component initially in core which is only used by `dataset_sharing`. To facilitate customization of the `notifications` module and also to clearly see its dependencies we have:

- Moved `notifications` code from core to modules as it is a reusable component that is not needed by any core component.
- Moved dataset_sharing references inside dataset_sharing module and left `notifications` independent from any other module (done mostly in #734, so credits to @TejasRGitHub)
- Added depends_on in the dataset_sharing module to load notifications if the data_sharing module is imported.
- Modified frontend navigation bar to make it conditional of the notifications module
- Added migration script to modify the notification type column
- Fix tests from #734, some references on the payload of the notification tasks were wrong
- Small fixes to SES stack: added account in KMS policy and email_id as input

### [WIP] Testing
Local testing
- [ ] loading of notifications with datasets enabled
- [ ] ...

AWS testing
- [ ] CICD pipeline succeds

### Other remarks
Not for this PR, but as a general note, we should clean up deprecated ECS tasks

### Relates
- #785 
- #734 

### Security

`N/A` just refactoring


noah-paige:
Just a couple of notes that I meant to leave on the previous Email notifciation PR  - @dlpzx 

- Can we make `iam.AnyPrincipal()` drilled down to `iam.AccountPrincipal()` in the KMS Key Policy for SNS Topic in SES Stack

- Enhancement Idea (can be implemented later): On revokes it is for particular items but we just say revoked/rejected on dataset level 
    - Should we have different notifications for revoked share items and display the item names that are revoked verse rejected share objects?

dlpzx:
@noah-paige, @TejasRGitHub I got an enhancement idea. Currently who receives a notification is still a little messy:

-------------
SHARE_SUBMISSION
Notification in UI sent to:
    - dataset.owner
Email sent to:
    - share submitter user: user that clicked on "submit" share request OR share.groupUri (if group_notifications=true)
    - dataset.SamlAdminGroupName
    - dataset.stewards

-------------
SHARE_APPROVAL
Notification in UI sent to:
    - dataset.owner
    - dataset.SamlAdminGroupName
    - dataset.stewards
    - share.owner: user that created the share request
Email sent to:
    - share approver user: user that clicked on "approve" share request OR share.groupUri (if group_notifications=true)
    - dataset.SamlAdminGroupName
    - dataset.stewards

-------------
SHARE_REJECTED
Notification in UI sent to:
    - dataset.owner
    - dataset.SamlAdminGroupName
    - dataset.stewards
    - share.owner: user that created the share request
Email sent to:
    - share approver user: user that clicked on "reject" share request OR share.groupUri (if group_notifications=true)
    - dataset.SamlAdminGroupName
    - dataset.stewards

dlpzx:
After agreeing with @TejasRGitHub offline, the PR is ready for review with the latest changes to simplify the notifications recipients.

backend/dataall/modules/dataset_sharing/services/share_item_service.py
```diff
@@ -63,7 +63,11 @@ def revoke_items_share_object(uri, revoked_uris):
                 resource_uri=dataset.datasetUri,
             )
 
-            ShareNotificationService.notify_share_object_rejection(session, context.username, context.email_id, dataset, share)
+            ShareNotificationService(
+                session=session,
+                dataset=dataset,
+                share=share
+            ).notify_share_object_rejection(email_id=context.username)
```
TejasRGitHub, line 70: Do we need the email_id=context.username field if we are using the share.owner which is also the email id of the user creating the share ?

config.json
```diff
@@ -15,6 +15,14 @@
                 "file_uploads": true,
                 "file_actions": true,
                 "aws_actions": true
+            },
+            "share_notifications": {
+                "email": {
+                    "active": true,
```
noah-paige, line 21: Should we default this to `false` as it requires `custom_domain` or `email_sender_id` in `cdk.json` which is not mandatory parameters?

backend/dataall/modules/dataset_sharing/services/share_item_service.py
noah-paige: I think we do need to pass `context.username` to know who is performing the operation - i.e. for approvals/rejections/revokes what user rejected or approved the share? (this could be any member of the dataset SamlAdminGroup or stewards - or even requestorTeam revoking some items)

We use that information for the email message sent to the respective users/groups

noah-paige:
I tested this feature with `datasets.share_notifications.email.active = False` and all deployed successfully. One thing I noticed is that Notifications only get listed for the share.owner and not for the dataset.SamlAdminGroup or dataset.steward

I believe this is because in the `ShareNotificationService` for `self.notification_target_users` we add targeted users of  dataset.SamlAdminGroupName (Group) , dataset.steward (Group), and share.owner (User Email) and have 3 notifications created in table for each with each of the above targeted users as the `Notification.username`

Then in `NotificationPopover` when we list Notifications for the use with API `listNotifications()` the backend resolver filters by `context.username` (i.e. `models.Notification.username == username`) and does not add another check for if user is a part of the group which is specified in the username

Either we have to add 

1 - an additional conditional in the filter to say if user's username is the same email as in Notification.username or user's groups is one of the groups specified in Notification.username (and remove duplicates)

2 - get a list of user's emails that are within the dataset.SamlAdminGroup and dataset.stewards and add a new Notification object to the table for each distinct user email with Notification.username=user's email

backend/dataall/modules/dataset_sharing/services/share_notification_service.py
```diff
@@ -1,152 +1,140 @@
...
-            message=msg,
-        )]
-        session.add_all(notifications)
-
-        ShareNotificationService.create_notification_task(session, email_id, dataset, share, subject, msg)
+    """
+    Notifications sent to:
+        - dataset.SamlAdminGroupName
+        - dataset.stewards
+        - share.groupUri
```
noah-paige, line 29: Looking at the way we set `self.notification_target_users = self._get_share_object_targeted_users()` it looks like we send notifications to 
        - dataset.SamlAdminGroupName
        - dataset.stewards
        - share.owner (NOT share.groupUri)

noah-paige: Should we handle this similar to how we do for email (if group_notifications=true have it to share.groupUri and if not just to share.owner)?

backend/dataall/modules/dataset_sharing/services/share_notification_service.py
```diff
@@ -1,152 +1,140 @@
...
+            - share.owner (person that opened the request) OR share.groupUri (if group_notifications=true)
+        """
+        share_notification_config = config.get_property('modules.datasets.share_notifications')
+
+        notification_recipient_groups_list = [self.dataset.SamlAdminGroupName, self.dataset.stewards]
         notification_recipient_email_ids = []
 
         if share_notification_config['email']['features']['group_notifications'] == True:
-            notification_recipient_groups_list.append(share.groupUri)
+            notification_recipient_groups_list.append(self.share.groupUri)
```
noah-paige, line 128: Should we only check if the `group_notifications` piece of this feature is true/false after we determine the `email` feature is active at all?

Meaning if I just set `modules.datasets.share_notifications.email.active = False` and specify nothing for group_notifications I think this will fail

deploy/stacks/backend_stack.py
```diff
@@ -361,10 +361,10 @@ def __init__(
                 internet_facing=internet_facing,
             )
 
-    @run_if(["core.share_notifications.email.active"])
-    def create_ses_stack(self, custom_domain, envname, kwargs, resource_prefix, email_notification_sender_email_id):
-        if custom_domain == None and email_notification_sender_email_id == None and deploy_config.get_property("core.share_notifications.email.active") == True:
-            raise Exception("Cannot Create SES Stack For email notification as Custom Domain and Sender Email Id are not present. Either Disable Email Notification Config or add Custom Domain")
+    @run_if(["modules.datasets.share_notifications.email.active"])
+    def create_ses_stack(self, custom_domain, envname, kwargs, resource_prefix):
+        if custom_domain == None:
```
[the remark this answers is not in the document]
noah-paige: UPDATE - looking at the code and the way we create `ses.EmailIdentity()` we are using the custom domain hosted zone create

I think to keep things simple for the initial version of this feature maybe we also use the custom_domain hosted zone (which will be required for email notifications) and we can remove the `email_sender_id` parameter from `cdk.json` -  @dlpzx @TejasRGitHub let me know thoughts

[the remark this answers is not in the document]
TejasRGitHub: I think just checking custom_domain to None would do. Initially , I was checking the custom_domain, email_notification_sender_email_id and the config for email notification. 

1. The config for email notification is now checked in the `run_if` decorator 
2. If the `email_notification_sender_email_id` is not present and email notification config is true then we can still use noreply@custom_domain as a default email-id .
3. If the custom_domain is not present ( and even is the email notification config is true and also the `email_notification_sender_email_id` is presnet / absent) then only we don't want the SES stack to be formed.

backend/dataall/modules/dataset_sharing/services/share_item_service.py
TejasRGitHub: Yeah that's right!. I get it now. The message in the email body/ UI notification needs to have that information

dlpzx: It is passed as email_id, because I wanted to account for your suggestion @TejasRGitHub an make it clear that the username is actually the email. Just in case I checked the definition of the "notify-X" functions and they use the email_id for to create the message: `def notify_share_object_submission(self, email_id: str):`

config.json
dlpzx: Good point! Making that change now, I don't want customers to start having unexpected issues out of nowwhere

backend/dataall/modules/dataset_sharing/services/share_notification_service.py
dlpzx: I had the same dilemma! My first thought was, let's keep both the same for consistency. It is what makes sense right? But then, looking at it from a different angle: the reason why we have the group notifications is because there will be extra emails to everyone, it is a matter of people being bothered with many emails. A small notification in the UI does not harm anyone and even if you are not "poked with an email" to attend the share request it gives users some visibility on it. I can change it because I am very undecided between the 2 options. @TejasRGitHub what is your opinion?

deploy/stacks/backend_stack.py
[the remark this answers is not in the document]
dlpzx: @noah-paige, @TejasRGitHub 
- good point on checking the hosted zone of the `custom_domain` parameter, totally up for that. 
- When doing the initial reviews I think both of us got it wrong and the `email_sender_id` is not a required parameter, but an optional overwrite. If the parameter is going to be used (question for @TejasRGitHub) I say we should keep it. Now, where should we keep it? in cdk.json or config.json? For me the best options are either as 1. an additional subparameter in the `custom_domain` section of `cdk.json` or 2. as a feature of `share_notifications` in `config.json` or 3. as a feature of `core` in `config.json`

backend/dataall/modules/dataset_sharing/services/share_notification_service.py
dlpzx: Added a couple of extra checks! Thank you

dlpzx: And as for the problem with the notifications and who sees them as the output of listNotifications(). That is an issue that was there originally. Let me see if I can fix it as part of this PR @noah-paige

dlpzx: I have done the changes +  some clean-up of the notifications queries. I noticed that there are 2 queries that are currently unused in the frontend (`countDeletedNotifications`, `countReadNotifications`). Should we remove this unused code as well? For the moment I am leaving a comment

dlpzx:
@noah-paige, @TejasRGitHub 
I am testing the following scenario:
```
"share_notifications": {
                "email": {
                    "active": false
                }
            }
```

dlpzx:
@noah-paige @TejasRGitHub I still see one possible enhancement for this feature, but maybe it is out of the scope for this PR. I modified the notifications to take a group as `recipient`. That way notifications can be defined for a complete team. The issue is that a single user can delete the notification for the whole team, so only that user sees the notification although the whole team has received it.

Option 1: One of the options that you pointed out @noah-paige was to create a notification record for each user. But that means going into cognito and getting the emails of each of the users belonging to a group. The downside is that we depend on Cognito as underlying technology and that not all users in a group are data.all active users

Option2: have a window where all users can see all notifications. A proper frontend view for notifications. Even if a user marks a notification as read, all other members of the team can go and check the notifications so that the information is not lost. And we make notifications more user-friendly. At the moment I don't think they are that usable

Option 3: we store the users that have "marked as read" the notification in a different table --> not the ideal for a relational database to store


They are not exclusive. In my opinion option 2, having notifications views, is always a positive addition

noah-paige (approved):
Did another review of the latest commits and tested out the latest changes in an AWS Deployment with `share_notifications.email.active` set to `false`

Approving now to merge to `v2m1m0` - looks good to me

deploy/stacks/backend_stack.py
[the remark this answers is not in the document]
TejasRGitHub: Hi @dlpzx , yes you are right in which the `email_sender_id` is not a required parameter. I had added that thinking if any one might want to have some other sender email id. But I am open to not having it in the first place in the `cdk.json`. If we remove it the sender email id will always be noreply@`custom_domain`. 

If we are keeping it, having the `email_sender_id ` in the `custom_domain`, doesn't make sense as it is not a domain related thing. If we want to move it out of `cdk.json`, I think probably in the `share_notifications` in `config.json`.

backend/dataall/modules/dataset_sharing/services/share_notification_service.py
TejasRGitHub: Hi @noah-paige , @dlpzx , 

I agree we can have notifications based on the `group_notifications` settings for the UI as well.

backend/dataall/base/aws/cognito.py
```diff
@@ -15,8 +15,6 @@ def __init__(self):
     def get_user_emailids_from_group(self, groupName):
         try:
             envname = os.getenv('envname', 'local')
-            if envname == 'local' or envname == 'dkrcompose':
```
TejasRGitHub, line 18: @dlpzx , after removing, while working on local development environment the cognito calls do fail. In that case, do you maybe want to keep this if condition to check if its local dev environment and then raise an Exception ?

backend/dataall/modules/dataset_sharing/services/share_notification_service.py
TejasRGitHub: If we do this, i.e. have consistent logic for both UI and Email notifications then a single function _get_targeted_users... can be used for both the notification types

deploy/stacks/backend_stack.py
```diff
@@ -133,11 +132,15 @@ def __init__(
...
 
         # Create the SES Stack
-        ses_stack = self.create_ses_stack(custom_domain, envname, kwargs, resource_prefix, email_notification_sender_email_id)
+        ses_stack = self.create_ses_stack(custom_domain, envname, kwargs, resource_prefix)
 
         repo = ecr.Repository.from_repository_arn(
             self, 'ECRREPO', repository_arn=ecr_repository
         )
+        if None not in [custom_domain, ses_stack]:
+            email_sender = custom_domain.get('email_notification_sender_email_id', None) if custom_domain.get('email_notification_sender_email_id', None) != None else f'noreply@{custom_domain.get("hosted_zone_name")}'
```
TejasRGitHub, line 141: So here we are saying that we are keeping the `email_notification_sender_email_id` is going to be in `custom_domain` of `cdk.json`. If we are going forward with this, then I will make changes in the documentation on `gh-pages` github branch. I was not thinking of having it in the `custom_domain` but I get it and that it makes sense to club these

TejasRGitHub (commented):
Looks Good to me as well. @dlpzx , please let me know where are you planning to keep the `email_sender_id` parameter value and I will update the documentation and create PR on the `gh-pages`.

deploy/stacks/backend_stack.py
dlpzx: @TejasRGitHub, I think we are going to keep it as part of the custom_domain section. Because even if it is not a route53 parameter per se, it is not a standalone parameter. It can only exist if custom domain exists

deploy/stacks/backend_stack.py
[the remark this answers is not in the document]
dlpzx: It is commented below, but the decision made was to keep the `email_notification_sender_email_id` as part of the custom_domain block as it can only exist when the custom domain is provided
```
"custom_domain": {
          "hosted_zone_name": "string_ROUTE_53_EXISTING_DOMAIN_NAME|DEFAULT=None, REQUIRED if internet_facing=false",
          "hosted_zone_id": "string_ROUTE_53_EXISTING_HOSTED_ZONE_ID|DEFAULT=None, REQUIRED if internet_facing=false",
          "certificate_arn": "string_AWS_CERTIFICATE_MANAGER_ARN|DEFAULT=None, REQUIRED if internet_facing=false",
          "email_notification_sender_email_id": "string_EMAIL_NOTIFICATION_SENDER_EMAIL_ID|DEFAULT=None"
        },
```

backend/dataall/modules/dataset_sharing/services/share_notification_service.py
dlpzx: This will be handled in future enhancements to notifications

backend/dataall/modules/dataset_sharing/services/share_notification_service.py
dlpzx: I am going to leave them independent from each other. The final reason for this is that not all customers will use the `share_notifications` block and the `group_notifications` is a parameter specific to the email notifications

backend/dataall/base/aws/cognito.py
dlpzx: Good catch! Let me review because I have not deployed it locally in a while

dlpzx: I think we need to add it to both places but as a conditional. Let me explain: in the local development we might use the account with a Cognito user pool with local users and we might want to test SES email if we have the right set up in the AWS account. 
- For Cognito local setup, developers will need to create the SSM parameter for the dkrcompose environment and make sure it points at a Cognito user pool with users and groups used in the local deployment. The use case for this will be to test Cognito boto3 calls and the code that handles the response in data.all
- For SES we just need to perform the preparation as if it was a deployment to AWS. I see the point of testing but I also see the scenario where developers have that particular config.json and do not have SES in their development accounts. That's why I added a similar block

dlpzx: I tested it and committed it [here](https://github.com/awslabs/aws-dataall/pull/822/commits/4e9a05b21d91f2959b4e5f8e2ca25d57a56bb261)

noah-paige:
I think the latest changes to add exceptions for local deployment look good as well - think good to merge @dlpzx

TejasRGitHub (approved):
Looks good to me

PR #828, opened 2023-10-24
feat: custom env linking text

zsaltys:
### Feature or Bugfix
- Feature

### Detail
- Adding a way to override env linking prerequisites for those that do not want to use manual steps and have other ways to achieve the same thing. You can see in this picture how we use it:

Screenshot 2023-10-24 at 16 02 49

### Security

The only security concern is that there's an html value coming out of config.json which gets set as innerHTML. I don't think this would be an issue unless someone would find a way to override the config. value as an attack.


dlpzx:
Hi @zsaltys, thanks for opening a PR :) We will review with @maryamkhidir and @itsmo-amzn after v2.1.0 is released

zsaltys:
@dlpzx thanks!

frontend/src/modules/Environments/views/EnvironmentCreateForm.js
```diff
@@ -298,171 +299,192 @@ const EnvironmentCreateForm = (props) => {
...
                 <Box>
-                  <Typography color="textSecondary" variant="subtitle2">
-                    3. As part of the environment CloudFormation stack data.all
-                    will create an IAM role (Pivot Role) to manage AWS
-                    operations in the environment AWS Account.
-                  </Typography>
+                  <Typography
+                    color="textSecondary"
+                    variant="subtitle2"
+                    dangerouslySetInnerHTML={{
```
mkhidir-aws, line 307: We need to sanitize the text before adding to prevent XSS attack. A good way is to use [DOMPurify](https://github.com/cure53/DOMPurify).
Maybe we create a small component wrapper like this:
`<SanitizedText text={config.core.custom_env_linking_text} />`

zsaltys: @noah-paige @dlpzx are you happy with this? Do you want me to introduce a new component and pull in a new library DOMPurify?

zsaltys: @mkhidir-aws I don't really see the point of sanitizing it because it's coming out of a config? Why would we want to sanitize it?

maryamkhidir: There is any possibility that the config file can be modified by external sources and it contains content that could be changed, I think it's best to sanitize it.

frontend/src/modules/Environments/views/EnvironmentCreateForm.js
```diff
@@ -298,171 +299,192 @@ const EnvironmentCreateForm = (props) => {
...
-                          {`cdk bootstrap --trust ${trustedAccount} -c @aws-cdk/core:newStyleStackSynthesis=true --cloudformation-execution-policies arn:aws:iam::ACCOUNT_ID:policy/DataAllCustomCDKPolicy aws://ACCOUNT_ID/REGION`}
-                        </Typography>
-                      </CardContent>
-                    </Card>
-                  </Grid>
-                </Grid>
-              </Box>
-              {process.env.REACT_APP_ENABLE_PIVOT_ROLE_AUTO_CREATE ===
-              'True' ? (
+              {config.core.custom_env_linking_text !== undefined ? (
```
itsmo-amzn, line 302: I'm not quite sure I understand what this custom `html` does. and if there's a backend component required to support it.

Could you please provide an example of this html?

itsmo-amzn: I understand only customers have access to this `config.json` file to configure their deployment, but it's not a good idea to enable html injection into the frontend.

zsaltys:
@itsmo-amzn @maryamkhidir I will make a PR that uses DOMPurify and a new react component? Will that be acceptable to merge this in?

maryamkhidir:
@zsaltys Yes, that works

zsaltys:
@maryamkhidir @itsmo-amzn I have added support for DOMPurify. I have close to no experience with React or NPM so apologies if I've made some mistakes. Please let me know if there's anything else I should fix.

noah-paige:
Changed the target branch to `main` - will need to pull latest from main to resolve above conflicts before merging this PR

maryamkhidir:
@zsaltys Taking a look

frontend/src/design/components/SanitizedHTML.js
```diff
@@ -0,0 +1,12 @@
+import DOMPurify from 'dompurify';
+
+export const SanitizedHTML = ({ dirtyHTML }) => {
+  const defaultOptions = {
+    ALLOWED_TAGS: ['b', 'i', 'em', 'strong', 'a'],
+    ALLOWED_ATTR: ['href']
```
maryamkhidir, line 6: Are you expecting a link from the config? If not I don't think you should allow `href`. Otherwise, it looks good.

dlpzx:
Thanks for picking it up @zsaltys, the PR also needs to be updates with the latest changes in `main` to avoid the conflicts before merge

frontend/src/design/components/SanitizedHTML.js
zsaltys: @maryamkhidir yes we are using links. That's the main reason I needed HTML support.

zsaltys:
Re-raised this PR to the main branch: https://github.com/awslabs/aws-dataall/pull/934/files .. To avoid solving a huge merge conflict by changing the base on this branch.

Issue #284, opened 2023-01-31
S3 Bucket data sharing


dlpzx:
**Is your feature request related to a problem? Please describe.**
When I use tools in my ETL or data consumption applications, I often need direct access to the source data S3 Bucket. Lake Formation in some cases does not perform sufficiently fast and S3 access points do not integrate with some of the used tools. At the moment we solve this challenge by manually providing permission in the Bucket policy of the source Bucket.

**Describe the solution you'd like**
For this particular cases we would like an automated way to manage access to S3 Buckets. 
- We want a self-service workflow with sharing request and approval and at the same time 
- we want to use S3 Bucket policies under-the-hood.

**Describe alternatives you've considered**
We can re-use data.all sharing workflow and include a new type of sharing, S3 Bucket. To not over-complicate the resulting bucket policy we would not require a prefix-level grained access, granting access grants access to the whole S3 Bucket.

Changes needed:
(see comments below)

## Question
- **How should we manage imported Datasets?** When we import a Dataset, we are importing an existing S3 Bucket with its corresponding bucket policy which might be managed by other processes outside data.all. We need to define if data.all a) will ignore these buckets, b) can overwrite the bucket policy (I am not sure how) or c) if it will re-create it and manage it.
- **How do we define that a dataset is s3-bucket-shareable?** We need to clarify what is the user process to categorize a dataset as S3-bucket-shareable.
- Can a dataset be shared with LF-tables, Access points-folders and S3-Bucket mechanisms simultaneously? What are the limitations?
- How do we handle KMS keys for imported buckets? and in sharing how do we specify CMK keys?

## Design decisions and assumptions – We should write ADRs in the development fork
-	We will first implement the feature on data.all created datasets and then we will deal with imported datasets. 
-	Users can choose between 2 sharing models, granular-share and full-share. Dataset owners decide which model applies to their Dataset and shares will follow the chosen model. In the granular-share model, we share Tables (Glue tables) using Lake Formation and Folders (S3 prefixes) using S3 Access points, this is what is currently implemented. In the full-share model, the whole S3 Bucket is shared using Bucket policies and the whole Glue database is shared using Lake Formation. We will first implement the full S3 Bucket share and as optional the full database share.

dlpzx:
I have created a branch to include the S3 Bucket policy changes (step 4 in the changes needed). However it is only for visibility, this branch should be rebased as V1.4.0 includes significant changes in the sharing workflow implementation.

[...]

PR #830, opened 2023-10-25
Feat: pivot role limit kms

dlpzx:
### Feature or Bugfix
- Feature

### Detail
- read KMS keys with an alias prefixed by the environment resource prefix
- read KMS keys imported in imported datasets
- restrict pivot role policies to the KMS keys created by data.all and those imported in the imported datasets
- move kms client from data_sharing to base as it is used in environments and datasets

### Relates
- #580

### Security

This PR restricts the IAM policies of the pivot role, following the least privilege permissions principle

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/base/aws/kms.py
+++ b/backend/dataall/base/aws/kms.py
@@ -74,7 +74,9 @@ class KmsClient:
         try:
             paginator = self._client.get_paginator('list_aliases')
             for page in paginator.paginate():
-                prefix_aliases = [alias["AliasName"] for alias in page['Aliases'] if alias.startswith(key_alias_prefix)]
+                print(str(page['Aliases']))
+                prefix_aliases = [alias["AliasName"] for alias in page['Aliases'] if alias["AliasName"].startswith(f"alias/{key_alias_prefix}")]
+                print(prefix_aliases)
         except Exception as e:
             log.error(
                 f'Failed to list kms key aliases in account {self._account_id}: {e}'
--- a/backend/dataall/base/aws/kms.py
+++ b/backend/dataall/base/aws/kms.py
@@ -8,8 +8,8 @@ log = logging.getLogger(__name__)
 class KmsClient:
     _DEFAULT_POLICY_NAME = "default"
 
-    def __init__(self, account_id: str, region: str):
-        session = SessionHelper.remote_session(accountid=account_id)
+    def __init__(self, account_id: str, region: str, role=None):
+        session = SessionHelper.remote_session(accountid=account_id, role=role)
         self._client = session.client('kms', region_name=region)
         self._account_id = account_id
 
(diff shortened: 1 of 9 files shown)


dlpzx:
I have tested locally with a real imported datasets and some fake datasets to put the iam_policy util under stress.
For 500 imported datasets these are the resulting pivot role policies:

Screenshot 2023-10-26 at 14 09 57


Screenshot 2023-10-26 at 14 10 19

backend/dataall/base/aws/kms.py
```diff
@@ -69,3 +69,17 @@ def check_key_exists(self, key_alias: str):
             return None
         else:
             return key_exist
+
```
dlpzx, line 72: This function is not used at the moment. @noah-paige should we remove it and move the `kms` handler back to the data_sharing module?

dlpzx: I moved it back to avoid issues with the work done in #284

noah-paige:
Tested this Pr out in an AWS Deployment - all permissions added to PivotRole look correct but I am also facing the same error: 
`An error occurred (MalformedPolicyDocumentException) when calling the PutKeyPolicy operation: The new key policy will not allow you to update the key policy in the future.`

When doing folder sharing on imported buckets only

noah-paige:
> Tested this PR out in an AWS Deployment - all permissions added to PivotRole look correct but I am also facing the same error: `An error occurred (MalformedPolicyDocumentException) when calling the PutKeyPolicy operation: The new key policy will not allow you to update the key policy in the future.`
> 
> When doing folder sharing on imported buckets only

After a bit of testing - I found the only way I could resolve the `MalformedPolicyDocumentException` is to explicitly add in the imported KMS Key Policy the following permission set to the PivotRole:
```
                "kms:Decrypt",
                "kms:Encrypt",
                "kms:GenerateDataKey*",
                "kms:PutKeyPolicy",
                "kms:GetKeyPolicy",
                "kms:ReEncrypt*",
                "kms:TagResource",
                "kms:UntagResource"
```

Even if you enabled IAM privileges to the KMS key policy similar to:
```
{
  "Sid": "Enable IAM User Permissions",
  "Effect": "Allow",
  "Principal": {
    "AWS": "arn:aws:iam::111122223333:root"
   },
  "Action": "kms:*",
  "Resource": "*"
}
```

and then add the same kms permissions set to the Pivot Role's IAM Policies it would not work (needed to be explicitly in KMS)

This could be helpful to call out in documentation but implementation would be required by the data.all user importing the key to ensure that they are giving proper access to data.all's pivotRole for their pre-created infra

dlpzx:
@noah-paige thanks for the review and for looking into the `MalformedPolicyDocumentException` I think we can tackle that issue with documentation, I added a small comment in the import section

documentation/userguide/docs/datasets.md
```diff
@@ -101,6 +101,42 @@ the fields of a newly created dataset you have to specify the S3 bucket and opti
 is left empty, data.all will create a Glue database pointing at the S3 Bucket. As for the KMS key Alias, data.all assumes that if nothing is specified
 the S3 Bucket is encrypted with SSE-S3 encryption.
 
+!!! danger "Imported KMS key and S3 Bucket policies requirements"
+    Data.all pivot role will handle data sharing on the imported Bucket and KMS key (if imported). Make sure that
+    the resource policies allow the pivot role to manage them. For the KMS key policy, explicit permissions are needed. See an example below.
+
+
+### KMS key policy
+In the KMS key policy we need to grant explicit permission to the pivot role. Note that this block is needed even if
```
zsaltys, line 110: @dlpzx this is not correct. You do not need to do this if the key has root level policy.. then the permissions on pivotRole are enough

Issue #614, opened 2023-08-01
Dataset Import: Additional checks related to bucket encryption / KMS key


blitzmohit:
**Is your idea related to a problem? Please describe.**

1. While importing a dataset:

-  User can forget to specify the KMS key even if it is part of the bucket's default encryption -> All folder/bucket shares will succeed on the UI, however the reader role will not have access to decrypt the bucket contents
-  User can forget to grant the pivotRole access to modify the key policy -> All folder/bucket shares will fail when trying to update the key policy

2. When creating a share for an imported dataset:
-  User can create a cross account share request for a bucket encrypted with SSE-S3 key (amazon managed key) -> Cross-account shares for folders/bucket will succeed on the UI, however the reader role will not have access to decrypt the bucket contents as key policy cannot be modified for amazon managed keys.

**Describe the solution you'd like**
1. While importing a dataset we can the following additional checks:

- Display a warning/error if user is importing an S3 bucket that has a default encryption setup with a kms key but no key has been specified for import. This information can be fetched from the S3 bucket via boto3 (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/get_bucket_encryption.html)
- Check if the key policy for the kms key being imported provides the right level of permissions to the pivotRole 

1. While creating a share we can add an additional checks:
- If the dataset is using an imported bucket encrypted with SSE-S3 then cross-account share can either show a warning or failure message.

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

dlpzx:
Thanks for opening the issue, we can support on the design or after v2.0.0 add it to our roadmap

dlpzx:
Completed!

PR #848, opened 2023-10-31
Enabling S3 bucket share 

anushka-singh:
### Feature or Bugfix
- Feature

### Detail
- We want to enable bucket sharing along with access point share which already exists in data all right now. 
- A user will be able to request shares at bucket level and at the folder level with access points. 
- Please NOTE: There is some common code between Access point share managers and processors and S3 Bucket managers and processors. We will send out a separate PR for that refactoring work at a later time. 

### Relates
- https://github.com/awslabs/aws-dataall/issues/284
- https://github.com/awslabs/aws-dataall/pull/823
- https://github.com/awslabs/aws-dataall/pull/846/files#diff-c1f522a1f50d8bcf7b6e5b2e586e40a8de784caa80345f4e05a6329ae2a372d0

### Contributors:
- Contents of this PR have been contributed by @anushka-singh, @blitzmohit, @rbernotas, @TejasRGitHub

### Security

  - Is the input sanitized?


backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -207,7 +258,7 @@ def revoke_share(cls, engine: Engine, share_uri: str):
             log.info(f'Still remaining LF resources shared = {existing_shared_items}')
             if not existing_shared_items and revoked_tables:
                 log.info("Clean up LF remaining resources...")
-                clean_up_tables = processor.delete_shared_database()
+                clean_up_tables = processor.clean_up_share() and processor.delete_shared_database()
```
noah-paige, line 261: Is there a reason we add `processor.clean_up_share() and...` 

I do not believe we have a `clean_up_share()` method for `ProcessLFCrossAccountShare` or `ProcessLFSameAccountShare` classes

noah-paige: `clean_up_share()` was removed in `v2m1m0` - I think we can revert this back to `clean_up_tables=processor.delete_shared_database()`

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -217,4 +268,17 @@ def revoke_share(cls, engine: Engine, share_uri: str):
                 new_share_state = share_sm.run_transition(ShareObjectActions.Finish.value)
             share_sm.update_state(session, share, new_share_state)
 
-            return revoked_tables_succeed and revoked_folders_succeed
+            return revoked_folders_succeed and revoked_s3_buckets_succeed and revoked_tables_succeed
+
+    @staticmethod
+    def _handle_table_share_failure(session, share, table, share_item_status):
```
noah-paige, line 274: Is this method called anywhere `_handle_table_share_failure()`

backend/dataall/modules/datasets_base/db/dataset_models.py
```diff
@@ -141,3 +141,23 @@ class Dataset(Resource, Base):
...
+    AwsAccountId = Column(String, nullable=False)
+    S3BucketName = Column(String, nullable=False)
+    region = Column(String, default='eu-west-1')
+    partition = Column(String, default='aws')
+    KmsAlias = Column(String, nullable=False)
+    imported = Column(Boolean, default=False)
+    importedKmsKey = Column(Boolean, default=False)
+    userRoleForStorageBucket = query_expression()
+    projectPermission = query_expression()
+    environmentEndPoint = query_expression()
```
noah-paige, line 159: Are the following fields used in `DatasetBucket` table: `partition` , `userRoleForStorageBucket`, `projectPermission`, and `environmentEndPoint` ?

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -12,15 +12,17 @@
...
 from dataall.modules.dataset_sharing.services.dataset_alarm_service import DatasetAlarmService
 from dataall.modules.dataset_sharing.db.share_object_repositories import ShareObjectRepository
+from dataall.modules.dataset_sharing.services.share_managers.share_manager_utils import ShareManagerUtils
 
 from dataall.modules.datasets_base.db.dataset_models import DatasetStorageLocation, Dataset
 
 logger = logging.getLogger(__name__)
 ACCESS_POINT_CREATION_TIME = 30
 ACCESS_POINT_CREATION_RETRIES = 5
+IAM_ACCESS_POINT_ROLE_POLICY = "targetDatasetAccessControlPolicy"
```
noah-paige, line 22: Should we also add another variable `DATAALL_ALLOW_OWNER_SID="AllowAllToAdmin"` similar to have we do with `IAM_ACCESS_POINT_ROLE_POLICY` here and with `DATAALL_ALLOW_OWNER_SID` in `s3_bucket_share_manager.py`

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -0,0 +1,443 @@
...
+        principal_list = self.get_principal_list(statement)
+        if f"{target_requester_arn}" not in principal_list:
+            principal_list.append(f"{target_requester_arn}")
+        statement["Principal"]["AWS"] = principal_list
+        return statement
+
+    @staticmethod
+    def generate_owner_access_statement(s3_bucket_name, owner_roleId):
+        owner_policy_statement = {
+            "Sid": "AllowAllToAdmin",
```
noah-paige, line 224: Let's use `DATAALL_ALLOW_OWNER_SID` rather than hardcoded `AllowAllToAdmin` here?

noah-paige: And also maybe pass in for `generate_default_bucket_policy()` call

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
```diff
@@ -21,6 +21,7 @@ def __init__(
         target_environment: Environment,
         source_env_group: EnvironmentGroup,
         env_group: EnvironmentGroup,
+        existing_shared_buckets: bool = False
```
noah-paige, line 24: is this needed in the `__init__()` - could we not handle similar to how we handle in `ProcessS3BucketShare()` when passing in `existing_shared_folders`?

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -140,34 +142,53 @@ def grant_target_role_access_policy(self):
...
+            IAM_ACCESS_POINT_ROLE_POLICY,
         )
         if existing_policy:  # type dict
-            if self.bucket_name not in ",".join(existing_policy["Statement"][0]["Resource"]):
-                logger.info(
-                    f'targetDatasetAccessControlPolicy exists for IAM role {self.target_requester_IAMRoleName}, '
-                    f'but S3 Access point {self.access_point_name} is not included, updating...'
-                )
-                target_resources = [
+            s3_target_resources = [
```
noah-paige, line 155: I had to push back the indentation of this block to pass linting checks

noah-paige:
Hi @anushka-singh - thanks again for this PR, I just left a few comments after going through the code myself. Most of the comments are minor but please give a look over and let me know if any questions

I will also be deploying the above to my AWS data.all deployment today and running through some tests and will leave any remaining comments here when complete

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -356,31 +387,52 @@ def delete_target_role_access_policy(
...
+            ShareManagerUtils.remove_resource_from_statement(
+                existing_policy["Statement"][0],
+                s3_target_resources
+            )
+
+            kms_target_resources = [
+                f"arn:aws:kms:{dataset.region}:{dataset.AwsAccountId}:key/{kms_key_id}",
+                f"arn:aws:kms:{dataset.region}:{dataset.AwsAccountId}:key/{kms_key_id}/*"
+            ]
+            ShareManagerUtils.remove_resource_from_statement(
```
noah-paige, line 415: Can we add some conditional check here - if `len(existing_policy["Statement"]) > 1` or something of the like before trying to remove principals from the KMS Permissions section

If I have a dataset with SSE-S3 encryption that I have approved folder shares for - I would not have this second KMS statement because the KMS Key is managed by AWS and is not a KMS CMK. 

Then when trying to revoke the folder share - the share revoke process would fail here when cleaning up because existing_policy["Statement"][1] gets an `IndexError: list index out of range` (I ran into this during my testing)

The same would be good to add for bucket sharing revokes as well

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -356,31 +387,52 @@ def delete_target_role_access_policy(
...
+                f"arn:aws:s3:{dataset.region}:{dataset.AwsAccountId}:accesspoint/{access_point_name}/*"
+            ]
+            ShareManagerUtils.remove_resource_from_statement(
+                existing_policy["Statement"][0],
+                s3_target_resources
+            )
+
+            kms_target_resources = [
+                f"arn:aws:kms:{dataset.region}:{dataset.AwsAccountId}:key/{kms_key_id}",
+                f"arn:aws:kms:{dataset.region}:{dataset.AwsAccountId}:key/{kms_key_id}/*"
```
noah-paige, line 413: Unless I am mistaken - the kms target resource `f"arn:aws:kms:{dataset.region}:{dataset.AwsAccountId}:key/{kms_key_id}/*"` does not mean anything to IAM permissions. This type of formatting only applies to S3 Buckets and their object contents - having the S3Arn + the S3Arn`/*`

I think we should be okay to remove the second `kms_target_resource` that includes the trailing `/*` everywhere

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
anushka-singh: delete_shared_database only deletes shared database when share request is rejected. Clean_up_share does a bunch of other things like:
```
        1) deletes S3 access point for this share in this Dataset S3 Bucket
        2) delete_target_role_access_policy to access the above deleted access point
        3) delete_dataset_bucket_key_policy to remove access to the requester IAM role
```
Hence I have added both here.

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
anushka-singh: Nope removing. This will be added as a part of Central catalog support PR.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: Makes sense! Done.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
anushka-singh: Used DATAALL_ALLOW_OWNER_SID rather than hardcoded AllowAllToAdmin. Not sure what you mean by 'pass in for generate_default_bucket_policy() call' 
Its defined as a constant at the top of the file, so we dont need to pass it in.

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
anushka-singh: This is probably not needed at all. Will remove if not needed

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
noah-paige: Yes for folder share - `class ProcessS3AccessPointShare()` 

But the above is for Table Shares and I do not believe `def clean_up_share()` is defined for table shares anymore

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
noah-paige: If the bucket policy does not exist we call `generate_default_bucket_policy()` 

the `def generate_default_bucket_policy(s3_bucket_name, owner_role_Id)` has a piece where it adds a statement with `"Sid": "AllowAllToAdmin",`

Could we define the function as:

`def generate_default_bucket_policy(s3_bucket_name, owner_role_Id, allow_owner_sid)` 

...and when calling the function in bucket_share_manager have it as:

`S3ControlClient.generate_default_bucket_policy(self.bucket_name, exceptions_roleId, DATAALL_ALLOW_OWNER_SID)`

backend/dataall/modules/datasets_base/db/dataset_models.py
anushka-singh: I believe we had a discussion on this in weekly calls. It was decided that we would keep this class uniform with other classes in this file and it would be addressed by a database/table clean up that AWS would be undertaking separately.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: Good catch!

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: Yes, good catch! I hadn't thought about the use case where KMS is managed by AWS. 

I have addressed all the points here except the last one where we need to modify `add_missing_resources_to_policy_statement ` to not run into the list index out of range. This one will need some tinkering, so I have not addressed it yet.

backend/dataall/modules/dataset_sharing/aws/kms_client.py
```diff
@@ -69,3 +69,22 @@ def check_key_exists(self, key_alias: str):
             return None
         else:
             return key_exist
+
```
dlpzx, line 72: I like this method, but I don't see that it is being used at the moment. What feature are you trying to implement?

backend/dataall/modules/dataset_sharing/aws/s3_client.py
```diff
@@ -121,6 +121,51 @@ def generate_access_point_policy_template(
...
+                    "Effect": "Allow",
+                    "Principal": "*",
+                    "Action": "s3:*",
+                    "Resource": [
+                        f"arn:aws:s3:::{s3_bucket_name}",
+                        f"arn:aws:s3:::{s3_bucket_name}/*"
+                    ],
+                    "Condition": {
+                        "StringLike": {
+                            "aws:userId": owner_roleId
```
dlpzx, line 144: Tracking it back to where this list of owner_roleIds gets generated, i can see that we are passing role arns. From this [blogpost](https://aws.amazon.com/blogs/security/how-to-restrict-amazon-s3-bucket-access-to-a-specific-iam-role/) I though that we could only add actual user ids in the policy. Have we tested that the dataset role, the environment role and the pivot role can access the bucket?

backend/migrations/versions/8c79fb896983_add_table_for_buckets.py
```diff
@@ -0,0 +1,206 @@
...
+        for dataset in datasets:
+            shared_bucket_object: ShareObjectItem = session.query(ShareObjectItem).filter(
+                and_(
+                    ShareObjectItem.itemType == ShareableType.S3Bucket.value,
+                    ShareObjectItem.itemUri == dataset.datasetUri,
+                )
+            ).first()
+            dataset_bucket: DatasetBucket = session.query(DatasetBucket).filter(
+                DatasetBucket.datasetUri == dataset.datasetUri
+            ).first()
```
dlpzx, line 184: I think this is particular to your implementation that had another release. If possible I would remove it, but it is only cosmetic. I don't see it affects other deployments in any way

frontend/src/utils/share.js
```diff
@@ -0,0 +1,10 @@
+export const generateShareItemLabel = (itemType): string => {
+  switch (itemType) {
+    case 'Table':
+      return 'Table';
+    case 'S3Bucket':
+      return 'S3Bucket';
+    case 'StorageLocation':
+      return 'Folder';
+  }
+};
```
dlpzx, line 10: I am not sure this is the best place to place this function. It is only used by Shares/components and /views. I lean towards placing code closer to its usage, and if it is shared across multiple modules, then add it to the correspondent functional directory. Let me review with the frontend designers, because what usually applies for backend does not always apply to frontend.

frontend/src/modules/Shares/components/AddShareItemModal.js
```diff
@@ -20,6 +20,7 @@ import { Defaults, Pager, Scrollbar } from 'design';
 import { SET_ERROR, useDispatch } from 'globalErrors';
 import { useClient } from 'services';
 import { addSharedItem, getShareObject } from '../services';
+import { generateShareItemLabel } from '../../../utils/share';
```
dlpzx, line 23: Since v2.0 we avoid relative imports in the frontend and replace them with global imports. In case the share.js is places inside the utils, it needs to be added to the `index.js` of the utils and imported directly from `../utils`

frontend/src/modules/Shares/components/RevokeShareItemsModal.js
```diff
@@ -10,6 +10,7 @@ import { Defaults } from 'design';
 import { SET_ERROR, useDispatch } from 'globalErrors';
 import { useClient } from 'services';
 import { getShareObject, revokeItemsShareObject } from '../services';
+import { generateShareItemLabel } from '../../../utils/share';
```
dlpzx, line 13: Same, relative imports should be replaced by global imports

frontend/src/modules/Shares/views/ShareView.js
```diff
@@ -65,6 +65,7 @@ import {
   UpdateRejectReason,
   UpdateRequestReason
 } from '../components';
+import { generateShareItemLabel } from '../../../utils/share';
```
dlpzx, line 68: Relative imports should be replaced by global imports

backend/dataall/modules/dataset_sharing/aws/kms_client.py
anushka-singh: We needed this for an internal implementation. I will remove it as its not being used anywhere in open source.

frontend/src/modules/Shares/components/AddShareItemModal.js
anushka-singh: I added it to index.js. 
I will still need to import it like this:
`import { generateShareItemLabel } from '../../../utils';`
Does that look okay?

frontend/src/modules/Shares/components/RevokeShareItemsModal.js
anushka-singh: Done!

frontend/src/modules/Shares/views/ShareView.js
anushka-singh: Done!
Could you also tell me why this is best practice? Just curious!

frontend/src/utils/share.js
anushka-singh: Sounds good! Please let me know what the conclusion is and I can make the changes.

backend/dataall/modules/dataset_sharing/aws/s3_client.py
anushka-singh: We use the arn to only extract the role name from a role arn. We then pass it to get_role_id() which ultimately fetches the roleId. We basically return the list of Role ids for the list of arns existing within the provided aws account  number. We dont use Role arns in the policy directly.

backend/dataall/modules/datasets_base/db/dataset_models.py
```diff
@@ -141,3 +141,23 @@ class Dataset(Resource, Base):
     @classmethod
     def uri(cls):
         return cls.datasetUri
+
```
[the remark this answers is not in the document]
anushka-singh: I think that is a great idea! It would definitely allow more flexibility, scalability, and ease of analysis. We can have a Dataset Table (Fact Table) containing the core dataset metadata such as dataset uri, name, aws account, region, etc. We can then have other dimension tables around it and keep adding them as we expand our dataset properties. We will have to think of the design more deeply though. We might want to think of indexign and partitioning strategies to optimize performance too.

backend/dataall/modules/datasets_base/db/dataset_models.py
```diff
@@ -141,3 +141,23 @@ class Dataset(Resource, Base):
...
+    S3BucketName = Column(String, nullable=False)
+    region = Column(String, default='eu-west-1')
+    partition = Column(String, default='aws')
+    KmsAlias = Column(String, nullable=False)
+    imported = Column(Boolean, default=False)
+    importedKmsKey = Column(Boolean, default=False)
+    userRoleForStorageBucket = query_expression()
+    projectPermission = query_expression()
+    environmentEndPoint = query_expression()
+
```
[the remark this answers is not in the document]
anushka-singh: We had a discussion on this in one of our weekly calls. It was decided that we would keep this class uniform with other classes in this file and it would be addressed by a database/table clean up that AWS would be undertaking separately. Should we address this after 2.2 in the clean up phase? If you think otherwise, I can take care of it in the next iteration.

backend/migrations/versions/8c79fb896983_add_table_for_buckets.py
```diff
@@ -0,0 +1,206 @@
...
+
+def upgrade():
+    try:
+        envname = os.getenv('envname', 'local')
+        print('ENVNAME', envname)
+        engine = get_engine(envname=envname).engine
+        bind = op.get_bind()
+        session = orm.Session(bind=bind)
+        datasets: [Dataset] = session.query(Dataset).all()
+        if not has_table('dataset_bucket', engine):
```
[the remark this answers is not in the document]
anushka-singh: I agree! Created an enhancement request in github for this: https://github.com/awslabs/aws-dataall/issues/865

backend/migrations/versions/8c79fb896983_add_table_for_buckets.py
anushka-singh: Not particularly sure about what code you are talking about. 

I do see we have `if column_exists('dataset', 'dataSharingModel'):` here. I will remove that from migration script as datasharingModel never made it to open source. Please let me know if you were talking about some other change though!

backend/dataall/modules/datasets_base/db/dataset_models.py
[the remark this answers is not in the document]
dlpzx: We will take it as part of the clean-up

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
```diff
@@ -179,21 +186,33 @@ def clean_up_share(
...
+        clean_up_folder = cls(
+            session,
+            dataset,
+            share,
+            folder,
+            source_environment,
+            target_environment,
+            source_env_group,
+            env_group,
+            existing_shared_buckets
```
dlpzx, line 198: Do we need `existing_shared_buckets` here? I think we can remove it from here and from the init right?

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -0,0 +1,445 @@
...
+                    existing_policy_statement=existing_policy["Statement"][1],
+                    iam_role_policy_name=IAM_S3BUCKET_ROLE_POLICY
+                )
+
+            policy = existing_policy
+        else:
+            logger.info(
+                f'{IAM_S3BUCKET_ROLE_POLICY} does not exists for IAM role {self.target_requester_IAMRoleName}, creating...'
+            )
+            policy = {
```
dlpzx, line 121: The limit for managed IAM policies is 6144 characters and white spaces don't count. I counted the chars of the policy without replacing the bucket_name... and other parameters and we have a policy of ~370 characters. With each new bucket we will add around ~170 chars. Which means that each team can request access to around 35 buckets (very quick and dirty aproximation). Do you think that you will face this scenario?

In case of failure, the block is safely placed inside a try/except block in the process_approved_shares. So in the worst case scenario, the share will end up in the failure state

dlpzx (commented):
looks good, only some small clarifications

frontend/src/modules/Shares/views/ShareView.js
dlpzx: So it means that you can import directly from ../utils, like this:
`import { generateShareItemLabel } from '../utils';`

dlpzx: Absolute imports make the hell of ../../../../../ way easier (check this [post](https://medium.com/geekculture/making-life-easier-with-absolute-imports-react-in-javascript-and-typescript-bbdab8a8a3a1)).

In my short experience using them, it makes it easier to move around files knowing that no matter where they are placed the imports do not change. After 10 mins you already know your "packages" by heart and can reference them very fast. In the frontend for example, we always import components from `design`, `services`, `utils`. 

And as a positive side effect, it makes you reflect on where you place the code. We still use relative imports inside each of the modules, because it is code that is used only inside the module. Globally exposed code is meant to be re-used everywhere. This differentiation kind of makes it evident when module specific code is made global

dlpzx:
Final remark on the PR: @noah-paige @anmolsgandhi @anushka-singh 
At the moment all sharing mechanisms are enabled and there is no configuration on the config.json. So if a customer wants to restrict a particular way of sharing data in data.all it is not possible at the moment. What do you think?

- do we see the need of allowing configuration of the sharing mechanisms allowed?
- and how important is it?

In any case I think we can move forward with this PR and if needed implement the config in another PR

dlpzx:
Testing in AWS:

- [X] CICD from previous deployment succeeds

For each of the following checks we test the 3 different scenarios:
    - Created Dataset
    - Imported Dataset with SSE-S3 encryption
    - Imported Dataset with KMS encryption
 
- [ ] Before doing anything, we check that pre-existing tables and folders shares still work --> check in Athena and Access points calls
- [ ] We share the S3 Bucket and Pre-existing tables and folders shares still work --> check in Athena and Access points calls
- [ ] Check S3 Bucket share works
- [ ] Check removing S3 Bucket share
- [ ] Check removing folder share
- [ ] Check removing both
- [ ] Check re-adding bucket share
- [ ] Check re-adding folder share
- [ ] Check new S3 share
- [ ] Check new folder share

backend/migrations/versions/8c79fb896983_add_table_for_buckets.py
dlpzx: Yes, that and also the following block. For a regular customer, no share_bucket_objects are created at this time, so this block is unnecessary. It is not going to introduce any error as the value checked in the if-clause will be `None`. However, taking the point of view of another customer, I would find it strange and difficult to understand.
```
        for dataset in datasets:
            shared_bucket_object: ShareObjectItem = session.query(ShareObjectItem).filter(
                and_(
                    ShareObjectItem.itemType == ShareableType.S3Bucket.value,
                    ShareObjectItem.itemUri == dataset.datasetUri,
                )
            ).first()
            dataset_bucket: DatasetBucket = session.query(DatasetBucket).filter(
                DatasetBucket.datasetUri == dataset.datasetUri
            ).first()
            if shared_bucket_object is not None:
                shared_bucket_object.itemUri = dataset_bucket.bucketUri
                shared_bucket_object.itemName = dataset_bucket.S3BucketName
```

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
dlpzx: I think this is the source of error 2 reported in the testing below! @anushka-singh

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -140,34 +143,52 @@ def grant_target_role_access_policy(self):
...
-                existing_policy["Statement"][0]["Resource"].extend(target_resources)
-                policy = existing_policy
-            else:
-                logger.info(
-                    f'targetDatasetAccessControlPolicy exists for IAM role {self.target_requester_IAMRoleName} '
-                    f'and S3 Access point {self.access_point_name} is included, skipping...'
+                share_manager.add_missing_resources_to_policy_statement(
+                    kms_key_id,
+                    kms_target_resources,
+                    existing_policy["Statement"][1],
```
dlpzx, line 184: IAM roles that were requesters and already had a `IAM_ACCESS_POINT_ROLE_POLICY`, that policy won't contain the second statement and will therefore fail. (This is actually the cause of error 3 in the testing). We need to add this KMS-statement before adding resources to it

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -0,0 +1,445 @@
...
+from dataall.modules.dataset_sharing.services.share_managers.share_manager_utils import ShareManagerUtils
+from dataall.modules.dataset_sharing.services.dataset_alarm_service import DatasetAlarmService
+from dataall.modules.datasets_base.db.dataset_models import Dataset, DatasetBucket
+from dataall.modules.dataset_sharing.db.share_object_repositories import ShareObjectRepository
+
+logger = logging.getLogger(__name__)
+
+DATAALL_READ_ONLY_SID = "DataAll-Bucket-ReadOnly"
+DATAALL_ALLOW_OWNER_SID = "AllowAllToAdmin"
+IAM_S3BUCKET_ROLE_POLICY = "dataall-targetDatasetS3Bucket-AccessControlPolicy"
```
dlpzx, line 21: All the IAM policies are prefixed with the resource prefix of the environment. The default value is `dataall` but it could take other values. All resources created in the environment will be prefixed with this resource prefix and the permissions given to the pivot role will be limited to the resources prefixed with this particular prefix. 

Instead of hardcoding "dataall-targetDatasetS3Bucket-AccessControlPolicy" we need to define this string as "targetDatasetS3Bucket-AccessControlPolicy" and then append the policy to the prefix like: `f"{target_environment.resourcePrefix}-{IAM_S3BUCKET_ROLE_POLICY}"`

dlpzx: (We have the same issue with `IAM_ACCESS_POINT_ROLE_POLICY = "targetDatasetAccessControlPolicy"` and now is tricky to update them)

dlpzx:
I just realized:
1. the PR is pointing to `v2m1m0`. Changing the source branch results in some additional git-ing that we need to do
2. we need to update the documentation in the userguide. How much bandwidth do you have for these? I don't think it will be much

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: Good catch! Implemented solution in next revision.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
anushka-singh: I think we can very easily reach this limit specially for datasets which are to be "open" or used across teams. We should find a workaround for this. Maybe we can have multiple policies ending with policy-1, policy-2 etc as we do for dataallpivotrole right now..

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
anushka-singh: Yes we do. Its being used in the function below.
This was always working, I removed it by mistake while fixing something else. I just had to add existing_shared_buckets: bool = False in class ProcessS3AccessPointShare to fix it. 
Tested it on my local and it works fine now.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: Fixed this in the next revision.

frontend/src/modules/Shares/views/ShareView.js
anushka-singh: Thanks for the explanation! 
That was very helpful.

anushka-singh:
> Final remark on the PR: @noah-paige @anmolsgandhi @anushka-singh At the moment all sharing mechanisms are enabled and there is no configuration on the config.json. So if a customer wants to restrict a particular way of sharing data in data.all it is not possible at the moment. What do you think?
> 
> * do we see the need of allowing configuration of the sharing mechanisms allowed?
> * and how important is it?
> 
> In any case I think we can move forward with this PR and if needed implement the config in another PR

On this comment. 
We do have a config for disabling the folder sharing mechanism. 
`"file_uploads": false,
                "file_actions": false,`
Setting this to false, disables access point sharing. 
There is no provision for disabling s3 bucket sharing at the moment though. 
I think we dont need the option at the moment since the idea was every dataset should be able to be shared fully with bucket share or folder wise with access point. If in the future, we identify a need for restricting access, we can simply add a config in config.json for it.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
anushka-singh: I am not sure I understood this one completely. Lets discuss this in the next office hours or offline.

anushka-singh:
> I just realized:
> 
> 1. the PR is pointing to `v2m1m0`. Changing the source branch results in some additional git-ing that we need to do
> 2. we need to update the documentation in the userguide. How much bandwidth do you have for these? I don't think it will be much

1. Yes, that will be needed. I might need your help with it. 
2. I would like to update the documentation too. Do we have an estimate on how long it will take? I can then maybe try to put in some sprint planning.

anushka-singh:
> Testing in AWS:
> 
> * [x]  CICD from previous deployment succeeds. Migration script runs successfully.
> 
> For each of the following checks we test the 3 different scenarios: - Created Dataset - Imported Dataset with SSE-S3 encryption - Imported Dataset with KMS encryption - Imported Dataset from multiple buckets with KMS encryption
> 
> * [x]  Before doing anything, we check that pre-existing tables and folders shares still work --> check in Athena and Access points calls --> we see the changes in the UI, with the new command to list s3
> * [x]  We share the S3 Bucket and Pre-existing tables and folders shares still work --> check in Athena and Access points calls
> * [x]  Check S3 Bucket share works -  we see all share item and its status after share as SHARE_SUCCEEDED for the bucket and get-object downloading of files in the bucket works
> 
> Now, only the case for imported KMS vs SSE-S3 are checked: Check removing S3 Bucket share - get object gives an AccessDenied error. Share item status is REVOKE_SUCCEEDED.
> 
> * [x]  Imported Dataset with SSE-S3 encryption: succeeds but with some error logs (error 1)
> * [x]  Imported Dataset with KMS encryption
> 
> Check removing folder share - get object gives an AccessDenied error. Share item status is REVOKE_SUCCEEDED
> 
> * [ ]  Imported Dataset with SSE-S3 encryption: folders are in REVOKE_SUCCEDDED but share request stays in REVOKE_IN_PROGRESS (details and logs in error 2)
> * [ ]  Imported Dataset with KMS encryption, not tested yet (it will fail like the previous one)
> 
> Check re-adding bucket share only
> 
> * [x]  Imported Dataset with SSE-S3 encryption
> * [x]  Imported Dataset with KMS encryption
> 
> Check re-adding folder share only
> 
> * [ ]  Imported Dataset with SSE-S3 encryption, not tested yet (it will fail with error 3)
> * [ ]  Imported Dataset with KMS encryption, not tested yet (it will fail with error 3)
> 
> Edge cases. Some not tested yet. Check add folder + s3 share at the same time
> 
> * [x]  Created Dataset with KMS encryption --> checked with a role that did not have previous folder share requests to avoid error 3 (get_object works using s3 and s3 access point)
> * [ ]  Imported Dataset with SSE-S3 encryption
> 
> Check remove folder + s3 share at the same time --- failing error 2
> 
> * [ ]  Created Dataset with KMS encryption
> * [ ]  Imported Dataset with SSE-S3 encryption
> 
> Check adding one and removing the other at the same time
> 
> * [ ]  Created Dataset with KMS encryption
> * [ ]  Imported Dataset with SSE-S3 encryption
> 
> ### Errors found
> #### Error 1 ---> In another issue
> This error is just cosmetic and it will addressed in #614
> 
> * Approving S3 Bucket sharing log shows failure but the sharing works fine (share item ends in SHARE_SUCCEEDED and we can download objects from S3 bucket):`Failed to get kms key id of alias/SSE-S3 : An error occurred (NotFoundException) when calling the DescribeKey operation: Alias arn:aws:kms:eu-west-1:XXXXXXXX:alias/SSE-S3 is not found.`
> * Revoking S3 Bucket sharing log shows failure but the sharing revoke works fine (share item ends in REVOKE_SUCCEEDED and we cannot download objects from S3 bucket): `Failed to get kms key id of alias/SSE-S3 : An error occurred (NotFoundException) when calling the DescribeKey operation: Alias arn:aws:kms:eu-west-1:XXXXXXX:alias/SSE-S3 is not found.`
> 
> #### Error 2 ---> 🧑‍🏭 needs fixing
> This is a generic issue, it is not due to pre-existing resources. Imported Dataset with SSE-S3 encryption: folders are in REVOKE_SUCCEDDED but share request stays in REVOKE_IN_PROGRESS (error 2) Looking at the logs, it fails in `dataall/modules/dataset_sharing/services/data_sharing_service.py:199` when we pass the variable `existing_shared_buckets=existing_shared_buckets` to the initialization of the `ProcessS3AccessPointShare`, but that variable is not declared in the class
> 
> Screenshot 2023-11-13 at 09 25 26
> 
> #### Error 3 ---> 🧑‍🏭 needs fixing
> This is a backwards compatibility issue. It fails when the Bucket is encrypted with a KMS key. Looking at the code and the logs I think the issue is in this line `dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py:184`
> 
> ```
> share_manager.add_missing_resources_to_policy_statement(
>     kms_key_id,
>     kms_target_resources,
>     existing_policy["Statement"][1],
>     IAM_ACCESS_POINT_ROLE_POLICY
> )
> ```
> 
> We are assuming that the IAM role policy has 2 statements, but in pre-existing IAM roles there is one statement only.
> 
> Screenshot 2023-11-13 at 10 02 41
> 
> Screenshot 2023-11-13 at 09 47 12
> 
> #### Error 4 ---> In another issue
> Sometimes we face again the issue that we had in the past of an access point taking a long time to get created and the put access point policy failure. I just want to note it here, but it is unrelated to this PR and it should be solved in a separate PR.

Error 2 and 3 found here have been fixed in next revision

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
dlpzx: I think I formulated it wrong, we need the variable, but do we need to pass the variable to the cls in like 200?

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
dlpzx: Code-wise the solution looks good. I will test but I'll mark this issue as resolved :)

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
dlpzx: It is not crucial, we can leave it out of the PR

frontend/src/utils/share.js
dlpzx: I would move it to `frontend/modules/Shares`

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
anushka-singh: I agree. Lets file a separate github issue for this and address it there.

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
anushka-singh: Sure
Removed it from line 200

dlpzx:
Testing in AWS 2:

- [X] CICD from previous deployment succeeds. Migration script runs successfully.
----------
For each of the following checks we test the 3 different scenarios:
    - Created Dataset
    - Imported Dataset with SSE-S3 encryption
    - Imported Dataset with KMS encryption
 
- [ ] Before doing anything, we check that pre-existing folders shares still work --> check Access points calls --> we see the changes in the UI, with the new command to list s3
- [ ] We share the S3 Bucket and Pre-existing folders shares still work --> check Access points calls
- [ ] Check S3 Bucket share works -  we see all share item and its status after share as SHARE_SUCCEEDED for the bucket and get-object downloading of files in the bucket works
-----------------
Check removing S3 Bucket share - get object gives an AccessDenied error. Share item status is REVOKE_SUCCEEDED. 

- [ ] Imported Dataset with SSE-S3 encryption: succeeds but with some error logs (error 1)
- [ ] Imported Dataset with KMS encryption

Check removing folder share - get object gives an AccessDenied error. Share item status is REVOKE_SUCCEEDED

- [ ] Imported Dataset with SSE-S3 encryption: folders are in REVOKE_SUCCEDDED but share request stays in REVOKE_IN_PROGRESS (details and logs in error 2)
- [ ] Imported Dataset with KMS encryption, not tested yet (it will fail like the previous one)

Check re-adding bucket share only
- [ ] Imported Dataset with SSE-S3 encryption
- [ ] Imported Dataset with KMS encryption

Check re-adding folder share only
- [ ] Imported Dataset with SSE-S3 encryption, not tested yet (it will fail with error 3)
- [ ] Imported Dataset with KMS encryption, not tested yet (it will fail with error 3)

----------------------
Edge cases. Some not tested yet.
Check add folder + s3 share at the same time
- [ ] Created Dataset with KMS encryption --> checked with a role that did not have previous folder share requests to avoid error 3 (get_object works using s3 and s3 access point)
- [ ] Imported Dataset with SSE-S3 encryption


Check remove folder + s3 share at the same time --- failing error 2
- [ ] Created Dataset with KMS encryption
- [ ] Imported Dataset with SSE-S3 encryption


Check adding one and removing the other at the same time
- [ ] Created Dataset with KMS encryption
- [ ] Imported Dataset with SSE-S3 encryption


### Errors found

dlpzx:
Re-tested with the latest fixes @anushka-singh 

Backwards compatibility test
Error 1 test:
- [X] Created Dataset -- Error 1 fixed
- [X] Imported Dataset with KMS encryption 

Error 2 test:
- [ ] Imported Dataset with SSE-S3 encryption - Approve S3 bucket share - it does not contain an statement with kms/None

Error 3 test:
- [ ] Imported Dataset with SSE-S3 encryption - Revoke S3 Bucket share and access points - it successfully removes items in IAM policy

dlpzx (approved):
Looks good! All tests have passed. Great job @anushka-singh

PR #855, opened 2023-11-03
Add callback and dependency matrix useclient

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- On some updates React is not rendering the `setReAuth()` function in `useClient.js:useEffect()` and thus the ReAuth Modal is never showing after returning a 401 error from the backend

- To remediate
  - Added `setReAuth` as part of the useEffect depedencies
  - Added callback to `setReAuth` to prevent infinite loop of re-rendering the function

### Security
N/A

  - Is the input sanitized?


noah-paige:
Tested in an AWS Deployment and re-auth functionality still working as expected

dlpzx (approved):

dlpzx:
Can we close this PR? @noah-paige

noah-paige:
@dlpzx I would propose we keep this open and merge into `main` - while the ReAuth components still function without the changes here it resolves some of the warnings coming from `npm run lint` and I think helps with readability of the code to explicitly be setting the proper dependency matrices

Let me know your thoughts

dlpzx:
I think we should then merge it to `main`

PR #857, opened 2023-11-06
Fix: ssm prefix custom cdk policy

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
As part of the environment stack we deploy some SSM parameters prefixed with `dataall` prefix in all cases. The custom DataallCustomCDKPolicy provided to use in the `cdk bootstrap` command restricts SSM permissions to resource-prefixed parameters. As a result, when using an environment with a resource prefix different from the default `dataall` the stack fails to create and to delete because it cannot create or delete those SSM parameters.

- The first commit adds the generic `dataall` SSM permisssions to the custom policy. It open the permissions slightly
- The second and third commits rename the SSM parameters to use the resource prefix

Each commit would solve the issue on its own, so we don't really need both. However, there are arguments to keep both. Having generic permissions to dataall-SSM parameters is restrictive and might be useful for other cases when the toolkit is used. It is good to keep all created resources prefixed with the same prefix, so that users can easily track which resources belong to a data.all environment.

The only issue is that for number 2, we need users to update the environment stacks before creating more datasets (add in release notes)

### Relates
- v2.1 release

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/datasets/cdk/dataset_custom_resources_extension.py
+++ b/backend/dataall/modules/datasets/cdk/dataset_custom_resources_extension.py
@@ -88,14 +88,14 @@ class DatasetCustomResourcesExtension(EnvironmentStackExtension):
             setup,
             'LakeformationDefaultSettingsCustomeResourceFunctionArn',
             string_value=lf_default_settings_custom_resource.function_arn,
-            parameter_name=f'/dataall/{setup.environment().environmentUri}/cfn/lf/defaultsettings/lambda/arn',
+            parameter_name=f'/{setup.environment().resourcePrefix}/{setup.environment().environmentUri}/cfn/lf/defaultsettings/lambda/arn',
         )
 
         ssm.StringParameter(
             setup,
             'LakeformationDefaultSettingsCustomeResourceFunctionName',
             string_value=lf_default_settings_custom_resource.function_name,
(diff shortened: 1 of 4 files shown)


dlpzx:
Tested on AWS deployment. I wanted to know your opinion @noah-paige:
- Should we remove the permission to `dataall-` SSM params?
- Should we go ahead with the renaming? or is it an unnecessary change?
I do not have an strong opinion (only slight opinions) on it :)

backend/dataall/modules/datasets/cdk/dataset_custom_resources_extension.py
```diff
@@ -88,14 +88,14 @@ def extent(setup: EnvironmentSetup):
             setup,
             'LakeformationDefaultSettingsCustomeResourceFunctionArn',
             string_value=lf_default_settings_custom_resource.function_arn,
-            parameter_name=f'/dataall/{setup.environment().environmentUri}/cfn/lf/defaultsettings/lambda/arn',
```
noah-paige, line 91: Minor formatting - should we jsut define `_env = setup.environment()` and then call the `_env.resourcePrefix` and `_env.environmentUri` all the subsequent times?

dlpzx: Done :)

dlpzx:
Re-tested with the latest changes

PR #858, opened 2023-11-06
Fix Get Dataset Table Overview for Non-Dataset Owners

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- If a User who is not Dataset Owner, Env Admin, or Dataset Steward tries to view a dataset table from the Catalog they cannot view the any of the tabs for: `Preview`, `Overview`, `Columns`, or `Metrics` and receive an error: `An error occurred (UnauthorizedOperation) when calling GET_DATASET_TABLE operation`

- This PR:
  - Removes `GlueTableProperties` from being returned in the `getDatasetTable` query as it is not required for any Dataset Table operations on data.all's UI
    - Resolves Requester User can see Overview Tab
  - Adds Check on Dataset Confidentiality Tag to determine to show data in `Columns` Tab or not and removes checks on 
    - Resolves Requester User can see Column Tab (for Unclassified Datasets)
  - Removes Resource Permissions checks on `Metrics` Tab queries to only handle check on Dataset Confidentiality Tag
    - Resolves Requester User can see Metrics Tab (for Unclassified Datasets)
  - Fix bug in update column description
    - Resolves Dataset Owner being able to update Column Description

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/datasets/aws/glue_table_client.py
+++ b/backend/dataall/modules/datasets/aws/glue_table_client.py
@@ -34,7 +34,7 @@ class GlueTableClient:
     def update_table_for_column(self, column_name, table_input):
         try:
             response = self._client.update_table(
-                DatabaseName=self._table.name,
+                DatabaseName=self._table.GlueDatabaseName,
                 TableInput=table_input,
             )
             log.info(
--- a/backend/dataall/modules/datasets/services/dataset_profiling_service.py
+++ b/backend/dataall/modules/datasets/services/dataset_profiling_service.py
@@ -71,7 +71,6 @@ class DatasetProfilingService:
             return DatasetProfilingRepository.list_profiling_runs(session, uri)
 
     @classmethod
-    @has_resource_permission(GET_DATASET_TABLE)
     def get_dataset_table_profiling_run(cls, uri: str):
         with get_context().db_engine.scoped_session() as session:
             cls._check_preview_permissions_if_needed(session, table_uri=uri)
@@ -99,7 +98,6 @@ class DatasetProfilingService:
             return run
 
     @classmethod
-    @has_resource_permission(GET_DATASET_TABLE)
     def list_table_profiling_runs(cls, uri: str):
         with get_context().db_engine.scoped_session() as session:
             cls._check_preview_permissions_if_needed(session=session, table_uri=uri)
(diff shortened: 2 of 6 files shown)


noah-paige:
Tested in AWS:

Dataset Owner w/ Dataset Unclassified
- [ ] Can view Table Preview Tab
- [ ] Can view Table Overview Tab
- [ ] Can view Table Column Tab + Edit Column Description
- [ ] Can view Table Metrics Tab

Requestor User w/ Dataset Unclassified
- [ ] Can view Table Preview Tab
- [ ] Can view Table Overview Tab
- [ ] Can view Table Column Tab but NOT Edit Column Description
- [ ] Can view Table Metrics Tab but NOT start profiling run

Dataset Owner w/ Dataset Official
- [ ] Can view Table Preview Tab
- [ ] Can view Table Overview Tab
- [ ] Can view Table Column Tab + Edit Column Description
- [ ] Can view Table Metrics Tab

Requestor User w/ Dataset Official
- [ ] Can NOT view Table Preview Tab
- [ ] Can view Table Overview Tab
- [ ] Can NOT view Table Column Tab
- [ ] Can NOT view Table Metrics Tab

backend/dataall/modules/datasets/aws/glue_table_client.py
```diff
@@ -34,7 +34,7 @@ def get_table(self):
     def update_table_for_column(self, column_name, table_input):
         try:
             response = self._client.update_table(
-                DatabaseName=self._table.name,
```
dlpzx, line 37: good catch

backend/dataall/modules/datasets/services/dataset_profiling_service.py
```diff
@@ -71,7 +71,6 @@ def list_profiling_runs(uri):
             return DatasetProfilingRepository.list_profiling_runs(session, uri)
 
     @classmethod
-    @has_resource_permission(GET_DATASET_TABLE)
     def get_dataset_table_profiling_run(cls, uri: str):
         with get_context().db_engine.scoped_session() as session:
             cls._check_preview_permissions_if_needed(session, table_uri=uri)
```
dlpzx, line 76: If the permissions checks based on dataset type (or any other object type) increase we could transform the _check_preview_permission method into a decorator. I like the separation between the function permission logic and the function logic. What do you think?

dlpzx (approved):
As commented, I think a possible enhancement would be to create a permission checker with the dependency on object types, but it is not a must at the moment as the only use case is the PREVIEW_DATASET_TABLE

backend/dataall/modules/datasets/services/dataset_profiling_service.py
noah-paige: I agree - I think this would be a good enhancement that we could pick up later if/when we continue to extend the use of dataset permissions. I will merge this PR as is to `v2m1m0` for now though

PR #859, opened 2023-11-06
Use op.execute to alter column type in migration script

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
In theory op.alter_table can modify column data types. But in our validation testing alembic runs into issues when those data types are more complex, like the user-define data type that we are trying to modify in the latest migration script. 
The migration seems to succeed but the data type does not change from the Enum type. The problem is that if in the future customers introduce new notification types they will receive failures when writing data to RDS.

After digging a bit, I found that other projects have faced similar issues and the way to work around it is to directly use SQL statements to modify the data type.

The migration script has been tested in AWS. I set a limit of 100 characters which is more than the double of the longest notification type at the moment.

### Relates
V2.1 release

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/migrations/versions/4f3c1d84a628_modify_notifications_column_types.py
+++ b/backend/migrations/versions/4f3c1d84a628_modify_notifications_column_types.py
@@ -5,6 +5,8 @@ Revises: 917b923f74bd
 Create Date: 2023-10-20 15:04:15.061516
 
 """
+import os
+
 from alembic import op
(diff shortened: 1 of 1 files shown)


noah-paige (approved):
lgtm - approving

noah-paige:
Migration script tested in my AWS Deployment as well - `notifications.type` data_type changed to varchar

Issue #788, opened 2023-10-12
data.all in local environment starts in a broken db state


zsaltys:
### Describe the bug

With the current 2.0 branch of data.all if we start data.all in a local dev environment then it will start with HALF the tables missing. This is the first issue and requires running alembic migrations to fix.

Upon running alembic migrations we find that they don't run with the error:

> sqlalchemy.exc.ProgrammingError: (pg.ProgrammingError) ERROR:  column "privateSubnetIds" of relation "vpc" already exists

coming from:

> INFO  [alembic.runtime.migration] Running upgrade bd271a2780b2 -> 5d5102986ce5, Add subnet ids columns

We can then attempt to drop all the tables with:

python backend/migrations/drop_tables.py

And run the migrations again but then data.all won't start because permission table will be missing permissions such as MANAGE_GLOSSARY, MANAGE_DATASETS

### How to Reproduce

Start data.all in dev environment from scratch and attempt to upgrade the local database with latest migrations

### Expected behavior

1. 2.0 branch local dev environment should start with all the tables installed
2. If I drop all the tables and attempt to rerun all the migrations, this should work and data.all should start without issues
3. https://awslabs.github.io/aws-dataall/deploy-locally/ should be updated to include explaining the alembic migrations and how to run them to update to the latest state and how to add new ones.
4. Ideally there should be tests that test that data.all can start with migrations rebuilt from scratch

### OS

NA

### Python version

NA

### AWS data.all version

2.0

dlpzx:
Hi @zsaltys, thanks for opening the issue. I am trying to reproduce the problem.

dlpzx:
## Test 1: build docker images from scratch
After cleaning up all images and everything from Docker locally I recreated the `db` image. 

As a result I obtain 39 tables:
![image](https://github.com/awslabs/aws-dataall/assets/71252798/c481ddb1-1f5a-405e-adfe-d90e94737eea)

If I now go to an actual deployment of data.all for the current `main` branch and check the amount of tables, I see that the aurora database contains 49 tables:

![image](https://github.com/awslabs/aws-dataall/assets/71252798/76ce545d-f412-4317-a18a-b18fb74206bf)
![image](https://github.com/awslabs/aws-dataall/assets/71252798/9f1b2577-d112-4a0c-ac85-2cb686bd4ef3)
![image](https://github.com/awslabs/aws-dataall/assets/71252798/99fcd13f-dcf6-48b0-a7d7-8a27fa7ce09b)

[...]

PR #860, opened 2023-11-07
Added github workflows on release branches and fix migration workflow issues

dlpzx:
### Feature or Bugfix
- Feature
- Bugfix

### Detail
- Add all applicable GitHub workflows to PRs pointing at `v2m*` branches
- fix semgrep finding issue from GitHub workflows from migration script for notifications type --> added `nosemgrep` as no user input is passed to the SQL query and only code administrators will have access to the query.
- fix migration validation: this one is tricky as it succeeds when running it locally and on a real pipeline. It turns out that the issue was not on the migration script itself but on the way we dropped and updated tables in the validation migration stage. For dropping tables, we were using a different schema that the one used in upgrade database. This PR removes the schema_name variable and uses the envname as schema for all cases. One final note, this issue might be related to #788.

Here some screenshots of the resulting local schema for the notification table after running `make drop-tables` and `make upgrade-db`
image

### Relates
- V2.1 release

### Security

  - Is the input sanitized?

### DIFF
--- a/.github/workflows/bandit.yml
+++ b/.github/workflows/bandit.yml
@@ -8,6 +8,7 @@ on:
   pull_request:
     branches:
     - main
+    - v2m*
 
 permissions:
   contents: read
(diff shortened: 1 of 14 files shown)


dlpzx:
Tested in AWS to verify migrations and validate migrations succeed

PR #882, opened 2023-11-17
Kms explosion fix

anushka-singh:
### Feature or Bugfix
- Bugfix

### Detail
- DataAll currently creates one SID per role in the KMS policy attached to a bucket with RoleID as the SID name.
- We want to collapse these SIDs into one SID.
-  Access point and Bucket share will have different SIDs in KMS policy. 
- Use role ARN instead of role ID.
- NOTE: if KMS policy was previously created, it will remain the same. SID will be the user ID and not the KMS decrypt SID created in this PR. It will not impact any future shares though. 
- NOTE: This is to be merged after bucket share PR is merged. 

### Testing
- Tested this on local dev environment and KMS policy now has 1 statement with kms decrypt and using SID of KMS decrypt. 

### Security

  - Is the input sanitized?


dlpzx:
Hi @anushka-singh looking at the PR, it includes 2 main changes: (1) the compression of multiple statements into one and (2) the replacement of role id by role arns in the KMS policy.

The reason for (1) is very clear, you want to optimize the policy size, but it is not clear to me why you want to change the role ids. I am not saying it is wrong, I just want to better understand your motivation. The big advantage of role_arns is that they are human-readable, while role_ids need to be fetched. The advantage of role_ids is that they are unique. In case of role deletion, the deleted-role and the new-role will have the same role_arn but different role_ids. This is used to ensure that access is granted only to the "correct" role

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -267,30 +268,25 @@ def grant_dataset_bucket_key_policy(self):
...
-                                "aws:userId": f"{target_requester_id}:*"
-                            }
-                        }
-                    }
-                )
-                kms_client.put_key_policy(
-                    kms_key_id,
-                    json.dumps(policy)
-                )
+            existing_policy = json.loads(kms_client.get_key_policy(kms_key_id))
```
dlpzx, line 271: What happens if we run in the exception in `get_key_policy` and it returns None? it will go into the else in line 280 and generate the default statement and then we will try to get existing_policy["Statement"] in line 285 and that I think it fails because of None["Statement"]

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -499,3 +516,35 @@ def handle_revoke_failure(self, error: Exception) -> None:
...
+        return f"arn:aws:iam::{target_account_id}:role/{target_requester_IAMRoleName}"
+
+    @staticmethod
+    def generate_default_kms_decrypt_policy_statement(target_requester_arn):
+        return {
+            "Sid": f"{DATAALL_ACCESS_POINT_KMS_DECRYPT_SID}",
+            "Effect": "Allow",
+            "Principal": {
+                "AWS": [
+                    f"{target_requester_arn}"
```
dlpzx, line 531: Same as in the BUcket sharing, what happens for the case where no KMS key is found? or there is an error in the call/permissions

tests/modules/datasets/tasks/test_s3_access_point_share_manager.py
```diff
@@ -658,12 +648,23 @@ def test_update_dataset_bucket_key_policy_without_env_admin(
...
+        assert kms_key_policy["Statement"][0]["Sid"] == DATAALL_ACCESS_POINT_KMS_DECRYPT_SID
+        assert kms_key_policy["Statement"][0]["Action"] == "kms:Decrypt"
+
+        # Check if the "Principal" contains the added target_requester_arn
+        assert "Principal" in kms_key_policy["Statement"][0]
+        assert "AWS" in kms_key_policy["Statement"][0]["Principal"]
+        assert "different_env_admin_id" in kms_key_policy["Statement"][0]["Principal"]["AWS"]
+        assert kms_key_policy["Statement"][0]["Resource"] == "*"  # Resource should be "*"
+        assert f"arn:aws:iam::{target_environment.AwsAccountId}:role/{target_environment.EnvironmentDefaultIAMRoleName}" in \
+               kms_key_policy["Statement"][0]["Principal"]["AWS"]
```
dlpzx, line 667: Thank you :)

tests/modules/datasets/tasks/test_s3_access_point_share_manager.py
```diff
@@ -1347,18 +1339,20 @@ def test_delete_dataset_bucket_key_policy_existing_policy_with_no_additional_tar
     # Given
     kms_client = mock_kms_client(mocker)
     kms_client().get_key_id.return_value = "1"
+    target_requester_arn = f"arn:aws:iam::{TARGET_ACCOUNT_ENV}:role/{TARGET_ACCOUNT_ENV_ROLE_NAME}"
```
dlpzx, line 1342: very minor, we can use the manager to replace this line right? like `manager.get_role_arn(TARGET_ACCOUNT_ENV, TARGET_ACCOUNT_ENV_ROLE_NAME)` This way we also test that small method

tests/modules/datasets/tasks/test_s3_bucket_share_manager.py
```diff
@@ -114,17 +116,21 @@ def base_bucket_policy(dataset2):
     return bucket_policy
 
 
-def base_kms_key_policy(target_environment_samlGrpName: str):
+def base_kms_key_policy(target_requester_arn=None):
+    if target_requester_arn is None:
```
dlpzx, line 120: Should this be is not None?

dlpzx:
Very few comments, nice job @anushka-singh. I am testing the changes in AWS:
- [X] CICD updates successfully
- [ ] Create new bucket share
- [ ] ....

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
anushka-singh: We are assuming that the following kms key is always present:
```
{
            "Sid": "Enable IAM User Permissions",
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "arn:aws:iam::xxxxx:root",
                    "arn:aws:iam::xxxxx:role/dataallPivotRole"
                ]
            },
            "Action": "kms:*",
            "Resource": "*"
        },
```
The reason for the assumption is that this is a pre-req step to using SSE-KMS encryption keys for the share. We have to add dataallPivotRole in statement Enable IAM User Permissions while importing a dataset.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: We are assuming that the following kms key is always present:
```
{
            "Sid": "Enable IAM User Permissions",
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "arn:aws:iam::xxxxx:root",
                    "arn:aws:iam::xxxxx:role/dataallPivotRole"
                ]
            },
            "Action": "kms:*",
            "Resource": "*"
        },
```
The reason for the assumption is that this is a pre-req step to using SSE-KMS encryption keys for the share. We have to add dataallPivotRole in statement Enable IAM User Permissions while importing a dataset.

tests/modules/datasets/tasks/test_s3_access_point_share_manager.py
anushka-singh: Addressed this in next revision.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
anushka-singh: But this might cause issues when we use S3 encryption or someone deletes the policy by mistake. I have addressed this in next revision.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: But this might cause issues when we use S3 encryption or someone deletes the policy by mistake. I have addressed this in next revision.

anushka-singh:
> Hi @anushka-singh looking at the PR, it includes 2 main changes: (1) the compression of multiple statements into one and (2) the replacement of role id by role arns in the KMS policy.
> 
> The reason for (1) is very clear, you want to optimize the policy size, but it is not clear to me why you want to change the role ids. I am not saying it is wrong, I just want to better understand your motivation. The big advantage of role_arns is that they are human-readable, while role_ids need to be fetched. The advantage of role_ids is that they are unique. In case of role deletion, the deleted-role and the new-role will have the same role_arn but different role_ids. This is used to ensure that access is granted only to the "correct" role

Yes the reason is human readability. Users complained that they couldnt easily see who the role belonged to. If they transitioned from old kms policy where you manually managed it to a new one, theyd want to see all new roles added by data all and compare with old ones they had already.
The other thing is if youre troubleshooting or checking why access is not working, having role arns helps because you can instantly identify which role name is missing and proceed with solving the problem. Otherwise we would spend extra time to figure how to convert name to id etc.
These were the reasons that prompted us to move to role arn.

anushka-singh:
> Very few comments, nice job @anushka-singh. I am testing the changes in AWS:
> 
> * [x]  CICD updates successfully
> * [ ]  Approve Bucket share and Access point share in Dataset with SSE-S3 encryption
> * [x]  Approve Bucket share and Access point share in Dataset with KMS encryption
> * [ ]  Revoke Bucket share and Access point share in Dataset with SSE-S3 encryption
> * [ ]  Revoke Bucket share and Access point share in Dataset with KMS encryption --> Error
> 
> Notes:
> 
> * Regarding "NOTE: if KMS policy was previously created, it will remain the same. SID will be the user ID and not the KMS decrypt SID created in this PR. It will not impact any future shares though." I have seen that the policy does not delete the statements for the roleId, instead it adds the new `DataAll-Access-Point-KMS-Decrypt` and `DataAll-Bucket-KMS-Decrypt` statements
> 
> Errors: **Error 1** KMS policy in access point revoke still contains `DataAll-Access-Point-KMS-Decrypt` statements image
> 
> Line 212 still evaluates to `True` because there is an existing shared bucket. So even if there are no folders shared the function `delete_dataset_bucket_key_policy` won't be called

Addressed this comment in next revision, please take a look!

dlpzx:
I am first testing locally:
- [ ] Approve Bucket share and Access point share in Dataset with SSE-S3 encryption
- [ ] Approve Bucket share and Access point share in Dataset with KMS encryption
- [ ] Revoke Bucket share and Access point share in Dataset with SSE-S3 encryption
- [ ] Revoke Bucket share and Access point share in Dataset with KMS encryption

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -324,29 +330,40 @@ def update_dataset_bucket_key_policy(self):
...
-                    "Condition": {
-                        "StringLike": {
-                            "aws:userId": f"{target_requester_id}:*"
-                        }
-                    }
-                }
-            )
-            kms_client.put_key_policy(kms_key_id, json.dumps(policy))
+        existing_policy = json.loads(kms_client.get_key_policy(kms_key_id))
+        target_requester_arn = self.get_role_arn(self.target_account_id, self.target_requester_IAMRoleName)
```
dlpzx, line 336: Folder sharing for SSE-S3 encrypted datasets fails because of line 335:
In this case, `existing_policy = kms_client.get_key_policy(kms_key_id)` return `None`. If we apply `json.loads` to this before checking if the policy exists it is going to fail. I think we can move the json.loads inside the if block in line 340 and that should solve the issue

dlpzx:
Hi @anushka-singh! The changes look good! I will do a final round of testing in AWS and approve

anushka-singh:
> Hi @anushka-singh! The changes look good! I will do a final round of testing in AWS and approve

Thank you! Looking forward to it.

dlpzx:
Testing in AWS:
- [X] CICD successfully runs
- [ ] Approve Bucket share and Access point share in Dataset with SSE-S3 encryption ----> Error!
- [X] Approve Bucket share and Access point share in Dataset with KMS encryption
- [ ] Revoke Bucket share and Access point share in Dataset with SSE-S3 encryption
- [ ] Revoke Bucket share and Access point share in Dataset with KMS encryption

I tried all possible combinations when approving and revoking: only Bucket, only folder, both at the same time. With 2 different shares and consumption roles. So I also tested addition/deletion of one role in the KMS policy and the complete clean-up when both shares are revoked.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
dlpzx: In addition, for SSE-S3, existing_policy resolves to None. Which means that it tries to execute the else block in line 356, resulting in failure of folder sharing for SSE-S3. We need to add a check of dataset.imported

dlpzx: Let me rephrase it, because I think I was not very clear. We have several conditions to check:
If the dataset is encrypted with SSE-S3 or with KMS.
- If KMS yes, then if the KMS policy exists
     - If yes,....
     - If no, create new policy
- If SSE-S3 - do nothing


Maybe instead of implementing checks inside the function we can modify the call to the function and check in `dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py:93`

```
if not dataset.imported or dataset.importedKmsKey:
    sharing_folder.update_dataset_bucket_key_policy()
```
And the same applies to `dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py:212`. It should only be called if KMS is used for encryption.
```
if not dataset.imported or dataset.importedKmsKey:
   clean_up_folder.delete_dataset_bucket_key_policy(
                dataset=dataset
            )
```

anushka-singh: okay now that makes sense. Addressed in next revision. Please take a look!

Just to confirm, if we create/import a dataset in data all which has SSE-S3 encryption, we wont handle cleaning/updating/removing it via data all correct?

dlpzx (approved):
I tested the small fix locally one final time and all functionalities worked as expected (KMS and SSE-S3 Datasets sharing). Approving now

PR #883, opened 2023-11-17
Add additional checks for dataset importing

nikpodsh:
### Feature or Bugfix
Feature

### Detail
Added additional checks for dataset importing  (see #614 )

Implemented checks that fail if:
1) An importing dataset bucket is encrypted with AWS managed key, but KMS key provided
2) An importing dataset bucket is encrypted with KMS key, but KMS key not provided
3) An importing dataset bucket is encrypted with KMS key, but provided KMS key is wrong
4) An imported dataset bucket is encrypted with AWS managed key, but user is trying to create a share requests for different account than dataset's account

From #614:
> User can forget to grant the pivotRole access to modify the key policy -> All folder/bucket shares will fail when trying to update the key policy

The check for this is not implemented. Rationale for this:
There are two types of pivotRole: manual pivotRole (1) and cdk pivotRole (2). 
1) Manual pivotRole has  `kms:PutKeyPolicy` for '*' resources in the environment. No action needed
2) When dataset stack is deployed (for imported datasets), it triggers update of an environment stack. The environment stack updates policies for all imported keys and datasets. Eventually, it sets `kms:PutKeyPolicy` for the imported key that pivotRole can change the policy. Thus, the users can catch the error only if they try to create share before update of the stack is succeed 

### Relates
#614

### Testing

Added unit test for checks in this PR. Added example of testing data.all aws clients, by mocking  boto3 client. Usually, aws clients are mocked in the integration testing and this test can be used as an example how to mock boto3 instead. 

### Security

  - Adding `s3:GetEncryptionConfiguration` to the pivotRole to fetch the encryption configuration for env buckets. It's needed to implement additional checks. EncryptionConfiguration provides only information about encryption (encryption type and key) and doesn't allow pivotRole to read the content of the bucket. 


dlpzx:
We will pick up this task @nikpodsh

backend/dataall/base/db/exceptions.py
```diff
@@ -137,6 +137,18 @@ def __str__(self):
         return f'{self.message}'
 
 
+class AWSInsufficientPrivileges(Exception):
```
dlpzx, line 140: I renamed it to match the typical AWS error message Insufficient Permissions

backend/dataall/core/environment/cdk/pivot_role_core_policies/s3.py
```diff
@@ -18,6 +18,7 @@ def get_statements(self):
                     's3:ListAllMyBuckets',
                     's3:GetBucketLocation',
                     's3:PutBucketTagging'
+                    's3:GetEncryptionConfiguration'
```
dlpzx, line 21: We should restrict this permission to dataset buckets only. For data.all managed buckets this permission is already included in the Get*. For imported buckets I have added it to the datasets/cdk/pivot role policy

backend/dataall/core/environment/cdk/pivot_role_core_policies/s3.py
```diff
@@ -17,7 +17,8 @@ def get_statements(self):
                 actions=[
                     's3:ListAllMyBuckets',
                     's3:GetBucketLocation',
-                    's3:PutBucketTagging'
+                    's3:PutBucketTagging',
+                    's3:GetEncryptionConfiguration'
```
dlpzx, line 21: Disclaimer: I thought about restricting this policy and add it to the imported buckets only. The issue is that we check the Bucket encryption before we import it, we check the Bucket encryption before we add pivot role permissions to the Bucket. Since it is a read permission that only reads configuration and not data it is acceptable to be added to the pivot role

dlpzx:
Testing locally:
- [X] Create Dataset
- [X] Import SSE-S3 Bucket and adding KMS Alias throws error
- [X] Import SSE-S3 Bucket without adding KMS Alias succeeds
- [ ] Import KMS Bucket and without adding KMS Alias throws error
- [ ] Import KMS Bucket adding KMS Alias that does not exist throws error
- [ ] Import KMS Bucket adding KMS Alias that exist succeeds

backend/dataall/modules/dataset_sharing/services/share_object_service.py
```diff
@@ -48,6 +49,15 @@ def create_share_object(
             dataset: Dataset = DatasetRepository.get_dataset_by_uri(session, dataset_uri)
             environment = EnvironmentService.get_environment_by_uri(session, uri)
```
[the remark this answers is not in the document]
nikpodsh: Frankly, I haven't check it in my account and might be wrong too. 

The issue originally appeared  in #614 :
     
   > User can create a cross account share request for a bucket encrypted with SSE-S3 key (amazon managed key) -> Cross-account shares for folders/bucket will succeed on the UI, however the reader role will not have access to decrypt the bucket contents as key policy cannot be modified for amazon managed keys.

Here, some additional links that I found
1)  https://docs.aws.amazon.com/kms/latest/developerguide/key-policy-modifying.html
2)  https://stackoverflow.com/questions/59043525/s3-cross-account-access-with-default-kms-key

tests/modules/datasets/test_import_dataset_check_unit.py
```diff
@@ -0,0 +1,108 @@
...
+
+def test_s3_managed_bucket_import(mock_aws_client):
+    dataset = Dataset(KmsAlias=None)
+
+    mock_encryption_bucket(mock_aws_client, 'AES256', None)
+
+    assert DatasetService.check_imported_resources(dataset)
+
+
+def test_s3_managed_bucket_but_bucket_encrypted_with_ksm(mock_aws_client):
```
noah-paige, line 19: minor - typo `kms`

noah-paige:
Tested a similar set of actions in an AWS Deployment with this PR:
- [x] CICD pipeline succeeds
- [x] Pivot role without S3:GetBucketEncryption permissions throws error Data.all Environment Pivot Role does not have s3:GetEncryptionConfiguration Permission for xxxxx bucket: An error occurred (AccessDenied) when calling the GetBucketEncryption operation: Access Denied
- [x] Import SSE-S3 Bucket and adding KMS Alias throws error
- [x] Import SSE-S3 Bucket without adding KMS Alias succeeds
- [x] Import KMS Bucket and without adding KMS Alias throws error
- [x] Import KMS Bucket adding KMS Alias that does not exist throws error
- [x] Import KMS Bucket adding KMS Alias but key policy restricts pivotRole permissions throws error
- [x] Import KMS Bucket adding KMS Alias that exist succeeds
- [x] Dataset access button works for both datasets

All looks good on my end

noah-paige (approved):
lgtm - approving

Issue #842, opened 2023-10-30
environment admin teams cannot see organizations they created


zsaltys:
### Describe the bug

If you are an environment admin that has been invited to an organization to link environments but you do not have a permission MANAGE_ORGANIZATIONS then you will not be able to see the environments under the Organizations -> Environments tab.

This is because of this code:

```
  @staticmethod
  **@has_tenant_permission(permissions.MANAGE_ORGANIZATIONS)**
  @has_resource_permission(permissions.GET_ORGANIZATION)
  def paginated_organization_environments(session, uri, data=None) -> dict:
```

We do not grant environment admin teams MANAGE_ORGANIZATIONS permission because that would allow them to create new organizations and that is not their role.

We of course have to make sure that paginated_organization_environments does not return environments that the environment admin has no permissions to see.

My proposal would be to remove the check for: @has_tenant_permission(permissions.MANAGE_ORGANIZATIONS)

### How to Reproduce

1. create an environment team that does not have manage_organizations permission and attempt to see the environment you linked under the organization.

### Expected behavior

1. if i create an environment in an organization i am invited to then i should be able to see that environment in under the org environments tab.

### OS

NA

### Python version

NA

### AWS data.all version

2.0

dlpzx:
Hi @zsaltys thanks for opening an issue. As discussed offline, it looks like that permission check is an error. It does not make sense to check the `MANAGE_ORGANIZATIONS` at the get organization level. Nonetheless, we want to take some time to see if there are any implications of removing that permission check. Once we do that we are happy to take or make the PR ourselves after v2.1.0 release

zsaltys:
@dlpzx sounds good. Simpler if you make this change because it's very basic.

PR #887, opened 2023-11-21
Remove unnecessary MANAGE_ORGANIZATIONS check

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
- Remove unnecessary permission check in list environments in the organization

### Relates
- #842 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/core/organizations/api/queries.py
+++ b/backend/dataall/core/organizations/api/queries.py
@@ -23,15 +23,6 @@ listOrganizations = gql.QueryField(
     test_scope='Organization',
 )
 
-listOrganizationInvitedGroups = gql.QueryField(
-    name='listOrganizationInvitedGroups',
-    type=gql.Ref('GroupSearchResult'),
-    args=[
-        gql.Argument(name='organizationUri', type=gql.NonNullableType(gql.String)),
-        gql.Argument(name='filter', type=gql.Ref('GroupFilter')),
-    ],
-    resolver=list_organization_invited_groups,
-)
 
 listOrganizationGroups = gql.QueryField(
     name='listOrganizationGroups',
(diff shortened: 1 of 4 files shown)


noah-paige:
Tested in AWS that a user (UserA) in a group with no MANAGE_ORGANIZATIONS tenant permission is able to see environments tab for the organization they are apart of. A couple of considerations, 

1. UserA can see ALL environments that are linked to the organization (not just the ones they created as env Admins)

- do we want to filter to only show the environments that a user in envAdmin for?
- we would also have to keep a query that shows all environments linked to an org (i.e. the creator of the organization should be able to see all environments linked to their org)


2. UserA can not view any of the teams that are a part of the organization since we still include `@has_tenant_permission(permissions.MANAGE_ORGANIZATIONS)` checks on listing the organization groups. I think this is okay but wanted to call it out if the intention is to allow GETS/reads on the organization information without the need for MANAGE_ORGANIZATIONS tenant permission

dlpzx:
1. In my opinion if you are part of the Organization it is good to see which other environments belong to it. Seeing it from a user perspective, it can lead to collaboration between teams working in separate environments. Something like "we are aware that the X department has linked their account to data.all"
2. Very similar, seeing other teams in your organization increases the visibility of teams across the enterprise. I think we should change it. What do you think?

noah-paige:
> 1. In my opinion if you are part of the Organization it is good to see which other environments belong to it. Seeing it from a user perspective, it can lead to collaboration between teams working in separate environments. Something like "we are aware that the X department has linked their account to data.all"
> 
>     2. Very similar, seeing other teams in your organization increases the visibility of teams across the enterprise. I think we should change it. What do you think?

@dlpzx I do not see any major issues with a team that is part of an organization being able to see what other environments or teams are a part of the larger organization (obviously while ensuring no permissions allowing any other create, update, delete actions on teams/envs they are not a part of)

I think following the same structure with Org-Teams the way we are fixing now for Org-Environments tab make sense, but @zsaltys what are your thoughts on this? Do you see any risks here?

nikpodsh (approved):

PR #889, opened 2023-11-23
Update CodeBuild images to Linux2 standard5.0

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
[Amazon Linux 2 x86_64 standard:4.0 use node16 ](https://docs.aws.amazon.com/codebuild/latest/userguide/available-runtimes.html)which is already deprecated. In this PR we update the CodeBuild images to use Amazon Linux 2 x86_64 standard:5.0 instead

### Security
[OWASP 10](https://owasp.org/Top10/en/). N/A

  - Is the input sanitized?

### DIFF
--- a/deploy/stacks/dbmigration.py
+++ b/deploy/stacks/dbmigration.py
@@ -141,7 +141,7 @@ class DBMigrationStack(pyNestedClass):
             id=f'DBMigrationCBProject{envname}',
             project_name=f'{resource_prefix}-{envname}-dbmigration',
             environment=codebuild.BuildEnvironment(
-                build_image=codebuild.LinuxBuildImage.AMAZON_LINUX_2_3,
+                build_image=codebuild.LinuxBuildImage.AMAZON_LINUX_2_5,
             ),
             role=self.build_project_role,
             build_spec=codebuild.BuildSpec.from_object(
--- a/deploy/stacks/pipeline.py
+++ b/deploy/stacks/pipeline.py
@@ -137,7 +137,7 @@ class PipelineStack(Stack):
                 'Synth',
                 input=source,
                 build_environment=codebuild.BuildEnvironment(
-                    build_image=codebuild.LinuxBuildImage.AMAZON_LINUX_2_4,
+                    build_image=codebuild.LinuxBuildImage.AMAZON_LINUX_2_5,
                 ),
                 commands=[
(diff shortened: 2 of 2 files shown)


dlpzx:
@noah-paige yes! Definitely, I did not want to interfere with the work of the DDK pipelines but I will do now. I also realized that we might be using deprecated versions of node in ECS. I'll review

dlpzx:
I am currently testing the changes in AWS

dlpzx:
Do not approve yet because there are some issues with using Linux5. In order to use node18 in the Docker images, we also need to upgrade the images used in the docker files. The new Amazon Linux 2023 is a bit different from Linux2, it is more minimalistic and does not include all packages, plus it replaces `yum` by `dnf`. There is no hurry, so in my opinion we should do this right and make sure that node is using node18 everywhere

dlpzx:
Tested in AWS
- [X] CICD builds and registers ECS and Lambda images successfully
- [X] API calls (Lambda) work as expected
- [X] Creation of Dataset (ECS) works as expected
- [X] Frontend runs as expected
- [X] user guide runs as expedcted

Tested locally
- [X] `docker-compose up` creates all images and containers without issue
- [ ] API calls and frontend appear as expected
- [ ] Creation of Dataset (ECS) works as expected

backend/docker/prod/lambda/Dockerfile
```diff
@@ -1,34 +1,44 @@
...
-    yum clean packages; yum clean headers; yum clean metadata; yum clean all; rm -rfv /var/cache/yum
+    dnf autoremove -y; \
+    dnf clean all; rm -rfv /var/cache/dnf
 
-RUN yum -y install shadow-utils wget
-RUN yum -y install openssl-devel bzip2-devel libffi-devel postgresql-devel gcc unzip tar gzip
-RUN amazon-linux-extras install $PYTHON_VERSION
-RUN yum -y install python38-devel
+# Install libraries
+RUN dnf -y install -y \
```
dbalintx, line 14: nit: -y twice

dbalintx:
Deployed locally - images are built properly
Deployed to AWS - codebuild succeeds, images are built properly
Created dataset, clicked through the FE, checked the userguide - seems to be fine.

Left 2 nit comments, otherwise LGTM

dbalintx (approved):

backend/docker/prod/lambda/Dockerfile
dlpzx: Fixed

dbalintx (approved):

dbalintx (approved):

PR #897, opened 2023-12-01
Support For External IDP and External User Pool Provider In data.all

TejasRGitHub:
### Feature or Bugfix
- Feature

### Detail
Currently data.all uses Cognito for user login and also maintains the user pool for teams/groups in data.all. 

This feature adds support to add any external OIDC based IdP. Apart from that , any external service provider can be used for maintaining teams/groups information. 

This PR contains both backend and frontend changes for External IDP. 

**Note** - When deploying with External IDP the user guide won't be setup as a part of deployment. Created following github issue to track this incremental change - https://github.com/awslabs/aws-dataall/issues/898

### Testing

- [x]  Local Development environment with Local Auth Context 
- [x]  `make test` - Unit Tests 
- [x]  Deployment to AWS With Cognito 
- [ ]  Deployment to AWS with External IDP  

### Relates
Github Issue - https://github.com/awslabs/aws-dataall/issues/872

### Security

fetching data from storage outside the application (e.g. a database, an S3 bucket)? Yes
  - Is the input sanitized? Yes
- Does this PR introduce any functionality or component that requires authorization? Yes
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms? Yes
  - Are you logging failed auth attempts? Yes ( At External IDP ) 
  - Have you used the least-privilege principle? How? Yes


deploy/configs/frontend_config.py
```diff
@@ -63,15 +67,28 @@ def create_react_env_file(
         file_content = f"""GENERATE_SOURCEMAP=false
 REACT_APP_GRAPHQL_API={graphql_api_url}
 REACT_APP_SEARCH_API={search_api_url}
-REACT_APP_COGNITO_USER_POOL_ID={user_pool_id}
+REACT_APP_USERGUIDE_LINK=https://{user_guide_link}
```
dlpzx, line 70: At the moment this value will be https:// right? Looking into the code the variable is only used in `frontend/src/design/components/layout/DefaultSidebar.js:201`, which is conditional on `REACT_APP_CUSTOM_AUTH`. It is very minor, but I think we can remove it and create the variable only if not custom_auth.

deploy/configs/frontend_config.py
```diff
@@ -128,13 +145,30 @@ def create_react_env_file(
     region = os.environ.get('deployment_region', 'eu-west-1')
     enable_cw_rum = os.environ.get('enable_cw_rum', 'False')
     reauth_ttl = os.environ.get('reauth_ttl', '5')
```
dlpzx, line 147: The amount of variables in this block and in the one in `deploy/stacks/pipeline.py` is getting very long. For custom_auth we are reading CodeBuild env variables that are PLAINTEXT. Maybe it is time to review this process, as a proposal we could use PARAMETER_STORE type of CodeBuild env vars: https://docs.aws.amazon.com/codebuild/latest/APIReference/API_EnvironmentVariable.html We would store the cdk.json as an SSM parameter and make it available as environment variables or directly reading it in the frontend_config script. What do you think @noah-paige?

deploy/stacks/pipeline.py
```diff
@@ -637,14 +637,15 @@ def set_backend_stage(self, target_env, repository_name):
                 prod_sizing=target_env.get('prod_sizing', True),
                 quicksight_enabled=target_env.get('enable_quicksight_monitoring', False),
                 enable_cw_rum=target_env.get('enable_cw_rum', False),
-                enable_cw_canaries=target_env.get('enable_cw_canaries', False),
+                enable_cw_canaries=target_env.get('enable_cw_canaries', False) and target_env.get("custom_auth", None) == None,
```
dlpzx, line 640: I don't understand this line. Why do we need to add this guardrail?

deploy/stacks/cloudfront.py
```diff
@@ -365,7 +353,7 @@ def __init__(
                     route53_targets.CloudFrontTarget(cloudfront_distribution)
                 ),
             )
-        if userguide_alternate_domain:
+        if userguide_alternate_domain and custom_auth != None:
```
dlpzx, line 356: very very minor: let's be consistent and use `is not None` as we have done when evalutaing the variable in other places

dlpzx, line 356: As a suggestion, maybe we can simplify the way we are executing the custom_auth condition. At the moment we are evaluating the condition in multiple places, maybe we can group together the code that is relevant for the userguide. Something like:

```python
        if frontend_alternate_domain:
            frontend_record = route53.ARecord(
                self,
                'CloudFrontFrontendDomain',
                record_name=frontend_alternate_domain,
                zone=hosted_zone,
                target=route53.RecordTarget.from_alias(
                    route53_targets.CloudFrontTarget(cloudfront_distribution)
                ),
            )
        user_docs_bucket = None
        if custom_auth is None:
            userguide_docs_distribution, user_docs_bucket = self.build_static_site(
                f'userguide',
                acl,
                auth_at_edge,
                envname,
                resource_prefix,
                userguide_domain_names,
                certificate,
                ssl_support_method,
                security_policy,
                logging_bucket,
                custom_auth
            )
            self.user_docs_bucket = user_docs_bucket
            self.user_docs_distribution = userguide_docs_distribution

            if userguide_alternate_domain:
                userguide_record = route53.ARecord(
                    self,
                    'CloudFrontUserguideDomain',
                    record_name=userguide_alternate_domain,
                    zone=hosted_zone,
                    target=route53.RecordTarget.from_alias(
                        route53_targets.CloudFrontTarget(userguide_docs_distribution)
                    ),
                )

        if tooling_account_id:
            cross_account_deployment_role = iam.Role(
                self,
                f'S3DeploymentRole{envname}',
                role_name=f'{resource_prefix}-{envname}-S3DeploymentRole',
                assumed_by=iam.AccountPrincipal(tooling_account_id),
            )
            resources_for_cross_account = []
            resources_for_cross_account.append(f'{cloudfront_bucket.bucket_arn}/*')
            if user_docs_bucket:
                resources_for_cross_account.append(f'{user_docs_bucket.bucket_arn}/*')

```

dlpzx (commented):
I am reviewing in parts. At the moment I am reviewing the deployment part

deploy/stacks/lambda_api.py
```diff
@@ -150,6 +157,64 @@ def __init__(
...
+                    bundling=BundlingOptions(
+                        image=_lambda.Runtime.PYTHON_3_9.bundling_image,
+                        local=SolutionBundling(source_path=custom_authorizer_assets),
+                    ),
+                ),
+                memory_size=512 if prod_sizing else 256,
+                description='dataall Custom authorizer replacing cognito authorizer',
+                timeout=Duration.seconds(20),
+                environment=custom_lambda_env,
+                vpc=vpc,
```
dlpzx, line 199: Why are we not creating a security group for the Lambda? I am not a security expert, but I imagine a scenario where our Lambda is called by anyone in the internet and might be throttled (being very paranoid-minded)

deploy/stacks/lambda_api.py
```diff
@@ -425,15 +493,38 @@ def set_up_graphql_api_gateway(
...
+                'CognitoAuthorizer',
+                cognito_user_pools=[user_pool],
+                authorizer_name=f'{resource_prefix}-{envname}-cognito-authorizer',
+                identity_source='method.request.header.Authorization',
+                results_cache_ttl=Duration.minutes(60),
+            )
+        else:
+            #Create a custom Authorizer
+            custom_authorizer_role = iam.Role(self,
+                                              'custom-auth-role',
```
dlpzx, line 510: Can add the `role_name` parameter in CDK and set it to `{resource_prefix}-{envname}-custom....`? If we want to restrict CDK policies in the future having prefixed resources with controlled names will make our life easier

dlpzx (commented):
Some more comments. I still need to review the custom resources in the deployment

deploy/configs/frontend_config.py
TejasRGitHub: Agreed, will add this change

deploy/stacks/cloudfront.py
```diff
@@ -506,18 +495,30 @@ def build_static_site(
         ssl_support_method,
         security_policy,
         logging_bucket,
+        custom_auth
     ):
```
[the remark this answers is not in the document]
TejasRGitHub: Yes, you are right. Addressing this in the next update

deploy/stacks/pipeline.py
TejasRGitHub: cw_canaries create a cognito user in order to work. Thus, putting this guardrail to make sure that cw_canaries is not created when the custom auth is set in which cognito stack is not present

deploy/stacks/pipeline.py
```diff
@@ -941,15 +956,18 @@ def set_albfront_stage(self, target_env, repository_name):
...
-                self.cognito_config_action(target_env),
-            ]
+        post = []
+        if target_env.get('custom_auth') is None:
+            post.append(self.cognito_config_action(target_env))
+
+        if target_env.get('enable_cw_rum', False):
+            post.append(self.cw_rum_config_action(target_env))
+
+        if len(post) == 0:
```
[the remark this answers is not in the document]
TejasRGitHub: Yes, this codeBuild step is not doing anything important , I have added this because the `post` step in code pipeline expected a Codebuildstep. let me see if I can refactor the code to remove this

backend/api_handler.py
```diff
@@ -74,11 +75,17 @@ def get_groups(claims):
         groups: list = (
             saml_groups.replace('[', '').replace(']', '').replace(', ', ',').split(',')
         )
-    cognito_groups = claims.get('cognito:groups', '').split(',')
-    groups.extend(cognito_groups)
+    cognito_groups = claims.get('cognito:groups', '')
+    if len(cognito_groups):
```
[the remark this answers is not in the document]
TejasRGitHub: Either of them is fine to me . No preferences as such 😄

backend/api_handler.py
```diff
@@ -74,11 +75,17 @@ def get_groups(claims):
...
+        groups.extend(cognito_groups.split(','))
     return groups
 
 
+def get_custom_groups(user_id):
+    service_provider = ServiceProviderFactory.get_service_provider_instance()
+    return service_provider.get_groups_for_user(user_id)
+
```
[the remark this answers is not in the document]
TejasRGitHub: for 1. if using the custom auth, we can avoid calling this function. It still won't create any problems if we call it as it would return an empty list. But I will add code to avoid calling it. I was working on both cognito and okta and thus , missed adding conditions on calling the `get_groups` method. 
for 2. correct

deploy/stacks/lambda_api.py
TejasRGitHub: Yes. I think you are right. Added a security groups similar to other Lambdas

deploy/stacks/cloudfront.py
TejasRGitHub: Addressed this by clubbing changes together which are used for creating user guide

template_cdk.json
```diff
@@ -45,7 +45,20 @@
           "reauth_apis": "list_of_strings_OPERATION_NAMES_TO_REQUIRE_REAUTH_ON|DEFAULT=None",
           "ttl": "int_TIME_IN_MINUTES_TO_ALLOW_USER_TO_PERFORM_SENSITIVE_APIS_BEFORE_FORCING_REAUTH|DEFAULT=5"
         },
-        "cognito_user_session_timeout_inmins": "integer_COGNITO_USER_SESSION_TIMEOUT_INMINS|DEFAULT=43200"
+        "cognito_user_session_timeout_inmins": "integer_COGNITO_USER_SESSION_TIMEOUT_INMINS|DEFAULT=43200",
+        "custom_auth": {
```
dbalintx, line 49: were you able to check if custom auth works together with the re-auth feature? I believe for that the only requirement is to have the claims containing an `auth_time` field.

deploy/configs/frontend_config.py
```diff
@@ -63,15 +67,28 @@ def create_react_env_file(
         file_content = f"""GENERATE_SOURCEMAP=false
 REACT_APP_GRAPHQL_API={graphql_api_url}
 REACT_APP_SEARCH_API={search_api_url}
-REACT_APP_COGNITO_USER_POOL_ID={user_pool_id}
+REACT_APP_ENABLE_PIVOT_ROLE_AUTO_CREATE={pivot_role_auto_create}
+REACT_APP_REAUTH_TTL={reauth_ttl}
+"""
+        if custom_auth:
+            file_content = file_content + f"""REACT_APP_CUSTOM_AUTH={custom_auth.get("provider", "none")}
```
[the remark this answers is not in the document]
TejasRGitHub: Great catch. Let me take a look into the code and see if I can add a validation even before we some to this part of code.

deploy/stacks/pipeline.py
dlpzx: Ahh thanks!

deploy/stacks/pipeline.py
```diff
@@ -765,7 +775,11 @@ def set_cloudfront_stage(self, target_env):
...
                 vpc=self.vpc,
             ),
-            self.cognito_config_action(target_env),
+        )
+        if target_env.get('custom_auth', None) is None:
+            front_stage_actions = (
+                *front_stage_actions,
+                self.cognito_config_action(target_env),
         )
         if target_env.get('enable_cw_rum', False):
```
noah-paige, line 784: should we also a check here (L 784) similar to what we pass in BackendStage `target_env.get('enable_cw_canaries', False) and target_env.get("custom_auth", None) == None,` so cw rum actions only occur if cw rum enabled AND custom auth not specified

deploy/stacks/pipeline.py
```diff
@@ -742,6 +744,14 @@ def set_cloudfront_stage(self, target_env):
                     f'export enable_cw_rum={target_env.get("enable_cw_rum", False)}',
                     f'export resource_prefix={self.resource_prefix}',
                     f'export reauth_ttl={str(target_env.get("reauth_config", {}).get("ttl", 5))}',
+                    f'export custom_auth_provider={str(target_env.get("custom_auth", {}).get("provider", "None"))}',
```
noah-paige, line 747: do we not need to match the `export` statements in these commands with the same in the `set_albfront_stage()` and also only create the UserGuide if custom_auth is not present?

deploy/stacks/pipeline.py
```diff
@@ -936,21 +951,13 @@ def set_albfront_stage(self, target_env, repository_name):
...
-            post=[
-                self.cognito_config_action(target_env),
-                self.cw_rum_config_action(target_env),
-            ]
-        else:
-            post=[
-                self.cognito_config_action(target_env),
-            ]
-        return post
+        if target_env.get('enable_cw_rum', False):
```
noah-paige, line 959: Similar to above let's restrict based on both enable cw rum and custom auth is NOT present

deploy/stacks/lambda_api.py
TejasRGitHub: When a security group was not specifically mentioned, the lambda creation attached a security group by default with no inbound and outbound rules. Although, this takes care of having a security group but I have explicitly added security group for this custom authorizer lambda

dlpzx: I looked into the docs. A Lambda can be deployed without a VPC or in a VPC. If deployed in a VPC, it always has security groups, but as you pointed out, better to have control over the rules. I have check the new implementation, we can resolve this comment

deploy/configs/frontend_config.py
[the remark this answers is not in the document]
dlpzx: I tried looking for the code changes, so maybe this is unnecessary but, one place where this check would make sense is when the CDK templates are synthesized (CDK Synth). For that, we would need can include additional checks in `deploy.stacks.pipeline.PipelineStack.validate_deployment_params` function

backend/dataall/base/services/service_provider_factory.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class ServiceProviderFactory:
+
+    @staticmethod
+    def get_service_provider_instance():
+        if (os.environ.get("custom_auth", None)):
+            # Return instance of your service provider which implements the ServiceProvider interface
+            # Please take a look at the "Deploy to AWS" , External IDP section for steps
+            return None
```
dlpzx, line 13: If a customer implements custom_auth, the step to update the ServiceProviderFactory is a deployment requirement.

In the case a customer forgets to update this line with their own ServiceProvider, it will only fail once the api is called, returning an error like "Cannot call get_groups_for_user from None"
"""
def get_custom_groups(user_id):
    service_provider = ServiceProviderFactory.get_service_provider_instance()
    return service_provider.get_groups_for_user(user_id)
"""


To avoid this situation we could add more guardrails:
1. In this file, instead of returning None, we can raise and Exception with a descriptive error log
2. At the time of deployment (in the deploy/stack/pipeline.py), we can validate the `service_provider = ServiceProviderFactory.get_service_provider_instance()`

What do you think?

backend/dataall/core/groups/api/resolvers.py
```diff
@@ -71,3 +72,13 @@ def list_cognito_groups(context, source, filter: dict = None):
         if group not in invited_group_uris:
             res.append({"groupName": group})
     return res
+
+
```
dlpzx, line 76: Clarification: this is needed to setGroups in useGroups. One of the most used hooks. It is called by the user to get its groups. Because it should only allow the user to get ist own groups, I think we should add some check that makes sure that the userid used in the call is the context user id. Without the server-side check, anyone with a JWT token and the API url can call this API call for any user

frontend/src/authentication/components/GuestGuard.js
```diff
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import { Navigate } from 'react-router-dom';
-import { useAuth } from '../hooks';
+import { useAuth } from '../hooks/useAuth';
```
dlpzx, line 3: I think this change is not needed. Hooks already exports useAuth

frontend/src/services/graphql/Groups/listGroups.js
dlpzx, line 1: We need to clean-up and delete the listCognitoGroups right? It is unused at the moment

frontend/src/design/components/popovers/AccountPopover.js
```diff
@@ -12,10 +12,10 @@ import {
 import { useSnackbar } from 'notistack';
 import { useRef, useState } from 'react';
 import { Link as RouterLink, useNavigate } from 'react-router-dom';
-import { useAuth } from 'authentication';
 import { useGroups } from 'services';
 import { CogIcon } from '../../icons';
 import { TextAvatar } from '../TextAvatar';
+import { useAuth } from '../../../authentication';
```
dlpzx, line 18: Better to use the global import (`import { useAuth } from 'authentication';`) instead of the relative import (`import { useAuth } from '../../../authentication';`)

frontend/src/index.js
```diff
@@ -25,11 +24,9 @@ ReactDOM.render(
...
               <BrowserRouter>
                 <SnackbarProvider dense maxSnack={3} hideIconVariant>
-                  <RequestContextProvider>
-                    <AuthProvider>
-                      <App />
-                    </AuthProvider>
-                  </RequestContextProvider>
+                  <AuthProvider>
+                    <App />
+                  </AuthProvider>
```
dlpzx, line 29: @noah-paige does this interfere with re-auth?

frontend/src/authentication/contexts/CognitoAuthContext.js
```diff
@@ -23,147 +23,13 @@ Auth.configure({
...
-  const login = async () => {
-    Auth.federatedSignIn()
-      .then((user) => {
-        dispatch({
-          type: 'LOGIN',
-          payload: {
-            user: {
-              id: user.attributes.email,
-              email: user.attributes.email,
-              name: user.attributes.email
```
dlpzx, line 124: This info is now returned in GenericAuthContext right? But instead of returning `id` `email` and `name` we return the following in `frontend/src/authentication/contexts/GenericAuthContext.js:188`
```
        email: user.attributes.email,
        id_token: user.signInUserSession.idToken.jwtToken,
        short_id: 'none'
```

Does this influence any place where we use user.id or user.name? For example in `frontend/src/design/components/popovers/AccountPopover.js:64` we use `user.name`

tests/core/groups/test_group.py
```diff
@@ -0,0 +1,32 @@
...
+def test_list_groups_env(client, env_fixture, group, module_mocker):
+    mock_client = MagicMock()
+    module_mocker.patch(
+        'dataall.base.aws.cognito.Cognito',
+        return_value=mock_client
+    )
+    mock_client().list_groups.return_value = ['cognitos', 'testadmins']
+    module_mocker.patch(
+        'dataall.core.groups.api.resolvers.ServiceProviderFactory.get_service_provider_instance',
+        return_value=mock_client()
```
dlpzx, line 13: Why did we return Cognito() in `test_notification_service` and we return mock_client here?
```
    mocker.patch(
        'dataall.modules.notifications.services.ses_email_notification_service.ServiceProviderFactory.get_service_provider_instance',
        return_value=cognito_client()
    )
```

frontend/src/services/hooks/useGroups.js
```diff
@@ -1,16 +1,36 @@
 import { Auth } from 'aws-amplify';
 import { useEffect, useState } from 'react';
 import { SET_ERROR, useDispatch } from 'globalErrors';
+import { useClient } from './useClient';
+import { useAuth } from '../../authentication';
+import { getGroupsForUser } from '../../authentication/services/getServiceProviderInfo';
```
dlpzx, line 6: Use global imports

frontend/src/authentication/services/getServiceProviderInfo.js
dlpzx, line 1: I would keep this in the generic services. Can we move it to `src/services`?

dlpzx: https://github.com/awslabs/aws-dataall/issues/398

backend/dataall/core/groups/api/resolvers.py
```diff
@@ -71,3 +72,13 @@ def list_cognito_groups(context, source, filter: dict = None):
...
+def get_groups_for_user(context, source, userid):
+    envname = os.getenv('envname', 'local')
+    if envname in ['dkrcompose']:
```
dbalintx, line 79:
```suggestion
    if envname in ['local', 'dkrcompose']:
```

backend/dataall/base/aws/cognito.py
```diff
@@ -64,3 +63,6 @@ def list_cognito_groups(envname: str, region: str):
             )
             raise e
         return groups
+
+    def get_groups_for_user(self, user_id):
```
dbalintx, line 67: To make the created ServiceProvider abstraction meaningful for this function as well, how about moving the logic relevant to this function [from the api handler](https://github.com/awslabs/aws-dataall/pull/897/files#diff-d9614c6d1ad54c1883eb1a16f3f4e342f93939f7b02ca8529c60dc0e2a85c98aR66) to here? You could omit the user_id input from the base class to make it work

deploy/stacks/pipeline.py
TejasRGitHub: Yes, missed this part, Was of the opinion that Cognito stack is not needed for this one. But turns out thet cw_rum needs cognito identity pool to be enabled. 

Making `enable_cw_rum ` config dependent  on `custom_auth ` config and adding this check whereever needed

deploy/stacks/pipeline.py
TejasRGitHub: Thanks for noticing that. Added those exports as in the albfront stage

deploy/custom_resources/custom_authorizer/custom_authorizer_lambda.py
```diff
@@ -0,0 +1,47 @@
...
+    effect = 'Allow'
+    policy = AuthServices.generate_policy(verified_claims, effect, incoming_event['methodArn'])
+    logger.debug('Generated policy is ', policy)
+    return policy
+
+
+# Below Code only used for testing on local development IDE
+# the following is useful to make this script executable in both
+# AWS Lambda and any other local environments
+if __name__ == '__main__':
```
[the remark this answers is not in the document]
TejasRGitHub: That is just some dummy account Id used for testing purposes. Used this part of the code to test for functions used in the `lambda_handler` code.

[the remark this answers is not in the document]
TejasRGitHub: @dlpzx , I can remove the code completely if its causing a distraction when reading this file.

frontend/src/index.js
TejasRGitHub: Hi @dlpzx , I have tested the reAuth with Cognito Stack and External Idp . Works for both. Please test it once at your end to be sure.

frontend/src/authentication/contexts/CognitoAuthContext.js
TejasRGitHub: @noah-paige , you are right! . Yes we still have the user.name, user.email, user.id , and all three contains the `email` field for Cognito stack and also for the External Idp stack. 

Due to this this just works as expected with the `AccountPopover` and also whereever it is used elsewhere

deploy/stacks/pipeline.py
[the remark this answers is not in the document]
TejasRGitHub: Refactored this so that we avoid the `len(post)` check.

deploy/stacks/lambda_api.py
TejasRGitHub: Updated code with the `role_name` changes

frontend/src/authentication/components/GuestGuard.js
TejasRGitHub: Addressed this

frontend/src/authentication/contexts/GenericAuthContext.js
```diff
@@ -0,0 +1,303 @@
...
+});
+
+export const GenericAuthProvider = (props) => {
+  const { children } = props;
+  const [state, dispatch] = useReducer(reducer, initialState);
+  const auth = useAuth();
+  const isLoading = auth ? auth.isLoading : false;
+  const userProfile = auth ? auth.user : null;
+  const authEvents = auth ? auth.events : null;
+
```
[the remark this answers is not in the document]
TejasRGitHub: When using Amplify ( cognito ) In that case, the auth is undefined and in which case the isLoading, userProfile ( which is auth.user) , authEvents ( which is auth.events ) will be false, null and null respectively , and tuseEffects and they won't be running. 

Initially had though of creating useEffects conditionally on the REACT_APP_CUSTOM_AUTH environment variable, but linter was throwing errors that useEffects cannot be initialized conditionally and thus using this method disable and enable useEffects.

deploy/configs/frontend_config.py
noah-paige: Think we should definitely look into using SSM - can pick this up as an enhancement task

backend/dataall/core/groups/api/queries.py
```diff
@@ -0,0 +1,26 @@
+from dataall.base.api import gql
+from dataall.core.groups.api.resolvers import get_group, list_groups, get_groups_for_user
+
+getGroup = gql.QueryField(
```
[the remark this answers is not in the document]
TejasRGitHub: Yes, atleast from what I searched . It just seems to have been used in tests.

deploy/stacks/albfront_stack.py
```diff
@@ -220,65 +221,66 @@ def __init__(
...
+            )
+            userguide_alb.target_group.configure_health_check(
+                port='80',
+                path='/',
+                timeout=Duration.seconds(10),
+                healthy_threshold_count=2,
+                unhealthy_threshold_count=2,
+                interval=Duration.seconds(15),
+            )
+            self.allow_alb_access(userguide_alb, ip_ranges, vpc)
```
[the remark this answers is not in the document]
TejasRGitHub: Oh yes, that's correct. Missed that. Will update this

backend/dataall/base/services/service_provider_factory.py
TejasRGitHub: I think having a guardrail for this would be very helpful. 

For 1. I agree I will raise an exception instead of returning None. Thus, the logs can clearly state what went wrong. 

For 2. I don't think so this would be easily doable as the code depends on setting of an environment variable `custom_auth` which is set in the lambda code. 

In order to do such a thing, while in the backend stack, this environment variable has to be set and then an instance of service_provider could be created and checked like 
```python
        if custom_auth is not None:
            try:
                os.environ['custom_auth'] = 'externalIDP'
                ServiceProviderFactory.get_service_provider_instance()
            except Exception as e:
                if str(e) == 'Service Provider not implemented when using custom auth configuration. Please implement and the again deploy backend stack':
                    del os.environ['custom_auth']
                    raise Exception('Service Provider not implemented when using custom auth configuration. Please implement and the again deploy backend stack')
```

But in order for the `ServiceProviderFactory.get_service_provider_instance()` to be imported in the `deploy` the `backend` folder has to be made package with an `__init__.py` file exporting the `dataall` package. 

Please let me know if there have another better way to do this check. Also, what do you think of the above way of adding this guardrail

backend/dataall/core/groups/api/resolvers.py
TejasRGitHub: Addressed this in the next update. Thanks!

backend/dataall/base/aws/cognito.py
TejasRGitHub: Currently, this function is not used when using the cognito stack. Thus for now, this function just returns nothing. The function that you were aluding to - [in the api handler](https://github.com/awslabs/aws-dataall/pull/897/files#diff-d9614c6d1ad54c1883eb1a16f3f4e342f93939f7b02ca8529c60dc0e2a85c98aR66) , would be getting groups information from the claims provided by the Cognito Authorizer. I am not sure if sections of this function would be a good fit here. Please let me know if the explanation I provided answers your question . Maybe I am not getting question 😅

frontend/src/reauthentication/contexts/RequestContext.js
```diff
@@ -52,6 +53,8 @@ export const RequestContextProvider = (props) => {
   const [requestInfo, setRequestInfo] = useState(null);
   const navigate = useNavigate();
   const client = useClient();
+  const token = useToken();
```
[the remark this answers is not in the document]
TejasRGitHub: Unfortunately, the `use...` hooks cannot be initialized conditionally. 

For the `useToken` hook which is added. The token information is used for both Cognito and any external IDP. This adds another check to the reAuth, by checking if the token is a newer token which is retrieved after the reAuth. 

For the `useAuth` hook. Based on if the frontend is using cognito or external IDP , this auth object will contain appropriate context. The context is what is described in the GenericAuthContext.js file. For the external IDP as there doesn't exist a Amplify like method ( Auth.currentAuthenticatedUser();) to get the user details , this is being used.

backend/dataall/core/groups/api/resolvers.py
TejasRGitHub: That's another great catch. You are right, this check should be done. 

Option1.
The user_id is not available in the context. In order to do this change, an additional field will be added, to the context. 
In case of external IDP, the user_id will be loaded in this context. This then can be used to check the user_id in the resolver by checking if the context.user_id == user_id. 
In case of Cognito IDP, the user_id will be same as username ( which is the email field ) in the context. 

Option2.
Another way of doing this check is , in the `api_handler.py` file, only for this specific resolver, the query params can be parsed and then the user_id from the `event` can be checked with the query params. 

I personally like option 1, but would like to know what option you find would be the best

backend/dataall/base/services/service_provider_factory.py
dlpzx: I overlooked the os.environ. mmm I see that it makes it harder to implement. In that case I think the error message and the documentation will be enough.

frontend/src/authentication/contexts/GenericAuthContext.js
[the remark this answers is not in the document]
dlpzx: ahhh, thanks for the clarification. I thought the same hahahhah

frontend/src/index.js
dlpzx: Thanks for the clarification and for the testing

frontend/src/reauthentication/contexts/RequestContext.js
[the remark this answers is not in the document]
dlpzx: No problem, we will leave them as is

backend/dataall/core/groups/api/resolvers.py
dlpzx: I prefer option 1 too. Option 2 is too particular for this API call. At the end we will have a context with the user_id. And we will check the same way that we check other permissions against the username

```
class RequestContext:
    """Contains API for every graphql request"""
    db_engine: Engine
    username: str
    groups: List[str]
    user_id: str
```

deploy/configs/frontend_config.py
dlpzx: I added it as a comment in the issue https://github.com/awslabs/aws-dataall/issues/899

backend/dataall/base/aws/cognito.py
dbalintx: > The function that you were aluding to - [in the api handler](https://github.com/awslabs/aws-dataall/pull/897/files#diff-d9614c6d1ad54c1883eb1a16f3f4e342f93939f7b02ca8529c60dc0e2a85c98aR66) , would be getting groups information from the claims provided by the Cognito Authorizer.

But isn't it exactly what this function is supposed to do? To get the groups for the user - in case of Cognito auth by extracting them from the claims - in case of external Idp via some other custom implementation.

This change has no functional values, but since you created an abstraction via the Service provider factory, I think it would be cleaner to make use of its abstract methods.

deploy/stacks/pipeline.py
```diff
@@ -894,8 +942,17 @@ def set_albfront_stage(self, target_env, repository_name):
...
+                        f'export reauth_ttl={str(target_env.get("reauth_config", {}).get("ttl", 5))}',
+                        f'export custom_auth_provider={str(target_env.get("custom_auth", {}).get("provider", "None"))}',
+                        f'export custom_auth_url={str(target_env.get("custom_auth", {}).get("url", "None"))}',
+                        f'export custom_auth_redirect_url={str(target_env.get("custom_auth", {}).get("redirect_url", "None"))}',
+                        f'export custom_auth_client_id={str(target_env.get("custom_auth", {}).get("client_id", "None"))}',
+                        f'export custom_auth_response_types={str(target_env.get("custom_auth", {}).get("response_types", "None"))}',
+                        f'export custom_auth_scopes={str(target_env.get("custom_auth", {}).get("scopes", "None"))}',
+                        f'export custom_auth_claims_mapping_email={str(target_env.get("custom_auth", {}).get("claims_mapping", {}).get("email", "None"))}',
+                        f'export custom_auth_claims_mapping_user_id={str(target_env.get("custom_auth", {}).get("claims_mapping", {}).get("user_id", "None"))}',
```
noah-paige, line 955: Lines 956 - 961 causes the CICD Pipelien to fail for VPC Facing deployment with custom auth:

```
        "mkdir ~/.aws/ && touch ~/.aws/config",
        "echo \"[profile buildprofile]\" > ~/.aws/config",
        "echo \"role_arn = arn:aws:iam::ACCOUNT_ID:role/RESOURCE_PREFIX-ENVNAME-cognito-config-role\" >> ~/.aws/config",
        "echo \"credential_source = EcsContainer\" >> ~/.aws/config",
        "aws sts get-caller-identity --profile buildprofile",
        "export AWS_PROFILE=buildprofile",
```

This is because the `...-cognito-config-role` only gets created in `IdPStack()` if `custom_auth` 

This role needs to be created regardless of `custom_auth` or not to allow tooling account codeBuild to assume it and read from SSM in the deployment account in `frontend_config.py` python script

noah-paige: Since `...-cognito-config-role` is also used in `cognito_config_action()` for NON custom_auth deployments - it may be best if `custom_auth` instead of deploying the `IdPStack()` create a new role with the same permissions as the `...-cognito-config-role` minus the following:
- Cognito Permissions + Resources
- CW Rum Permissions + Resources
- Secrets Manager Permissions + Resources (used for CW Canaries)
So pretty much a new role in backend with SSM permissions that the tooling account can assume

backend/dataall/base/aws/cognito.py
dbalintx: yes exactly. but as these are no functional changes it could also be in a followup not to block the merging

deploy/stacks/pipeline.py
TejasRGitHub: Hi @noah-paige , Thanks for pointing this out. Can you tell if this is the policy you had in mind ?

```python
 actions=[
                    'secretsmanager:DescribeSecret',
                    'secretsmanager:GetSecretValue',
                    'ssm:GetParameterHistory',
                    'ssm:GetParameters',
                    'ssm:GetParameter',
                    'ssm:GetParametersByPath',
                    'kms:Decrypt',
                    'kms:GenerateDataKey',
                    'kms:DescribeKey'
                ],
 resources=[
                    f'arn:aws:ssm:*:{self.account}:parameter/*{resource_prefix}*',
                    f'arn:aws:ssm:*:{self.account}:parameter/*dataall*',
                    f'arn:aws:secretsmanager:{self.region}:{self.account}:secret:*{resource_prefix}*',
                    f'arn:aws:secretsmanager:{self.region}:{self.account}:secret:*dataall* 
                ],
```

noah-paige:
Testing Custom Auth + VPC Facing (updating as I test):

- [ ] CICD Pipeline Deploys --> Only After fixing `...-cognito-config-role` issue
  - [ ] No User Guide Resources Created
  - [ ] Custom Authorizer Created
- [ ] Log In via External IdP
- [ ] Open UI, first API call to api handler listEnvironments successfully returns the environments of my user
- [ ] open catalog, API call to search handler successfully returns all data items
- [ ] UI shows user name in the top right corner
- [ ] ReAuth on ListOrg successful and successfully lists orgs and re-directs to page after providing credentials again
- [ ] User Log Out from UI successfully

backend/dataall/base/aws/cognito.py
TejasRGitHub: Yes I think we can implement the function at a later time if that's not needed right now. Commented here. https://github.com/awslabs/aws-dataall/issues/899

deploy/stacks/pipeline.py
noah-paige: No KMS permissions should be needed either - just SSM

TejasRGitHub: Updated the code. with the policy needed for frontend deploy

dlpzx:
Testing in AWS internet-facing no custom-auth:

- [x] CICD Pipeline Deploys 
- [x] User Guide Created
- [x] Log In via Cognito
- [X] Open UI, first API call to api handler listEnvironments successfully returns the environments of my user
- [X] open catalog, API call to search handler successfully returns all data items
- [X] UI shows user name in the top right corner
- [X] User Log Out from UI successfully

dlpzx (approved):
Good to merge!

backend/dataall/core/groups/api/queries.py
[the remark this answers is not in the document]
TejasRGitHub: @noah-paige is there any clean-up issue that is already filed ? This can be added into it

PR #908, opened 2023-12-12
Fix Migration Script for New Deployment

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- When deploying from scratch - the migration script for byo vpc for mlstudio had a line of code `if has_table('sagemaker_studio_domain')` which was evaluating to `False` and skipping necessary migration code
  - Removed the `if has_table()` and just throw Exception in try/catch block for upgrade migration script

- Also as part of this PR I made sure the security group gets created in the mlstudio_extension stack no matter what type of VPC we use for ML Studio Domain
  - Without the Security Group VpcOnly Deployments of ML Studio Domain with `security_group=[]` will never be able to connect to internet without manual changes
  - Tested with Imported VPC and provisioned Studio User Profile - can access internet via NAT Gateway as prescribed with VPCOnly deployments if internet access is required
  - Tested changes to ML Studio Domain Security Group does NOT cause new creation of Domain (only UpdateDomain no replacement) --> Environment Stack Updates Successfully

### Relates
- https://github.com/awslabs/aws-dataall/pull/894

### Security
N/A

  - Is the input sanitized?


dlpzx (approved):
Looks good and tested!

PR #920, opened 2023-12-15
fixing s3 bucket sharing for federated roles

zsaltys:
### Feature or Bugfix

- Bugfix

### Detail
- Creating share requests for S3 buckets with roles like arn:aws:iam::1234:role/sso/foo will fail the share because /sso/ will be removed by backend resulting in an exception with an invalid principal. This was fixed before for access points etc but was missed in S3 bucket sharing.

### Relates
- https://github.com/awslabs/aws-dataall/issues/921

### Security
N/A


noah-paige:
Tested these changes in AWS Deployment with a service role as share principal:

- [x] CICD Pipeline successful + Quality Gate Passes
- [x] Folder Share Successful
- [x] Bucket Share Successful
- [x] Folder Share Revoke Successful
- [x] Bucket Share Revoke Successful

noah-paige (approved):
Tested in AWS - lgtm, approving!

PR #933, opened 2023-12-19
Bugfix 932: Investigate why some shares did not go to failed state, but remained stuck or in-progress

anushka-singh:
### Feature or Bugfix
- Bugfix

### Detail
- Investigate why the shares stayed in progress even if they actually must have failed because of missing permissions or any other reason.
- Please look at the ISSUE linked to read more about the cause. 
 - Also includes some linting fixes. 

### Resolution:

I have changed the Subject to be constructed like this:
subject = f'ALARM: DATAALL S3 Bucket {bucket.S3BucketName} Sharing Failure'[:100]

This will ensure any unnecessary words from Subject are removed and that we always guarantee a 100 character limit. The message body is further created properly and with a lot of details in case the Subject line gets truncated.
I will repeat this pattern for revoke shares too and for folder and table shares

### Relates
- https://github.com/awslabs/aws-dataall/issues/932

### Security

  - Is the input sanitized?


backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
```diff
@@ -19,9 +19,7 @@ def trigger_table_sharing_failure_alarm(
             target_environment: Environment,
     ):
         log.info('Triggering share failure alarm...')
-        subject = (
-            f'ALARM: DATAALL Table {table.GlueTableName} Sharing Failure Notification'
-        )
+        subject = f'ALARM: DATAALL Table {table.GlueTableName} Sharing Failure Notification'[:100]
```
zsaltys, line 22: I think we can be smarter here. Let's actually remove the dynamic part from the subject so it remains static and nothing ever needs to be cut. I think people can normally create filters/label by sender to identify it's a data.all email and we can see the glue table name inside the text itself? I would apply this for all alarms.

dlpzx:
HI @anushka-singh thanks for the PR! The issue is as you described: "small but important". I have read through the PR and the changes look good. The only issue are the lines at the end of files that were removed and are making the linting tests fail.

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
zsaltys, line 22: If we switch to static subjects then I would remove the 100 character truncate and I would update this code in share processors:

```
except Exception as e:
                self.handle_revoke_failure(share_item=share_item, table=table, error=e)
                new_state = revoked_item_SM.run_transition(ShareItemActions.Failure.value)
                revoked_item_SM.update_state_single_item(self.session, share_item, new_state)
                success = False
```

At the very minimum the code inside self.handle_revoke_failure should be in try catch.. Email sending even if the entire service is down should never leave shares in progress. We can just log that sending email failed... Thats far better than leaving a share blocked.

I would argue that even table state updates should be in their own try/catch with retries if the aurora service is down or behaving erratically. 

How I would resolve this:

1. put the code inside  self.handle_revoke_failure into a try/catch so that this method can never affect state transitions and just log if it fails .. it's not important enough to break transitions.

2. revoked_item_SM.run_transition does not throw exceptions so we're fine with it. no changes needed.

3. revoked_item_SM.update_state_single_item this could be wrapped inside in a try/catch or even outside .. but after catching it should LOG and re-raise the exception.

@anushka-singh One thing I don't understand .. If an exception happens in this EXCEPT block does it get logged or not?

anushka-singh: I thought about that too. 
I prefer having the information of the table or bucket name in the Subject itself. 
Think about the scenario when users have a tonne of requests to approve and want to prioritize the important ones ones from the failed (or successful) shares. They will have to open each email or notification individually and read through the information to get their answer. If we just put it in the subject, even if it gets truncated, they will atleast be able to read some of it and decide which ones to prioritize. Hence having the 100 character limit makes more sense to me. 

@dlpzx what do you think?

anushka-singh:
> HI @anushka-singh thanks for the PR! The issue is as you described: "small but important". I have read through the PR and the changes look good. The only issue are the lines at the end of files that were removed and are making the linting tests fail.

Not sure where the lint is coming from! I will change it and send the PR

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -266,4 +266,4 @@ def revoke_share(cls, engine: Engine, share_uri: str):
                 new_share_state = share_sm.run_transition(ShareObjectActions.Finish.value)
             share_sm.update_state(session, share, new_share_state)
 
-            return revoked_folders_succeed and revoked_s3_buckets_succeed and revoked_tables_succeed
+            return revoked_folders_succeed and revoked_s3_buckets_succeed and revoked_tables_succeed
```
dlpzx, line 269: Linting

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -553,4 +553,4 @@ def get_principal_list(statement):
         principal_list = statement["Principal"]["AWS"]
         if isinstance(principal_list, str):
             principal_list = [principal_list]
-        return principal_list
+        return principal_list
```
dlpzx, line 556: linting

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -481,4 +481,4 @@ def generate_default_kms_decrypt_policy_statement(target_requester_arn):
             },
             "Action": "kms:Decrypt",
             "Resource": "*"
-        }
+        }
```
dlpzx, line 484: linting

backend/dataall/modules/dataset_sharing/services/share_processors/s3_bucket_process_share.py
```diff
@@ -166,4 +166,4 @@ def process_revoked_shares(
                 revoked_item_SM.update_state_single_item(session, removing_item, new_state)
                 success = False
 
-        return success
+        return success
```
dlpzx, line 169: linting

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
```diff
@@ -212,4 +212,4 @@ def clean_up_share(
             if not dataset.imported or dataset.importedKmsKey:
                 clean_up_folder.delete_dataset_bucket_key_policy(dataset=dataset)
 
-        return True
+        return True
```
dlpzx, line 215: linting

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
dlpzx: I see the value of having the table name in the subject because of the capability to search by table name (or at least first part of the name). But maybe we can move the dynamic part to the end of the string so that we truncate the table/folder/bucket name and not the message itself. Something like:

`subject = f'ALARM: DATAALL Sharing Failure Notification for Table {table.GlueTableName} '[:100]` what do you think?

dlpzx: Good catch Zilvinas! One solution is the proposed: adding a try/catch inside the `handle_revoke_failure` (and approve) in `dataall/modules/dataset_sharing/services/share_managers/lf_share_manager.py` to ensure that no matter what, when there is a failure the item ends up in `Share_Failed` status. 


Another option is to move the code a bit and place the notification after the update of the item (which makes more sense logically)

```
            except Exception as e:
                new_state = revoked_item_SM.run_transition(ShareItemActions.Failure.value)
                revoked_item_SM.update_state_single_item(self.session, share_item, new_state)
                success = False
                self.handle_revoke_failure(share_item, table, e)

```
Let me know your thoughts. We can also implement both

anushka-singh: I agree. Will write it this way in next revision.

anushka-singh: I have implemented both solutions in next revision. Please check @dlpzx

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
anushka-singh: Fixed

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: Fixed

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
anushka-singh: Fixed

backend/dataall/modules/dataset_sharing/services/share_processors/s3_bucket_process_share.py
anushka-singh: Fixed

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
anushka-singh: Fixed

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
zsaltys: I think it's a fine compromise to put it at the end..

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
```diff
@@ -19,9 +19,7 @@ def trigger_table_sharing_failure_alarm(
             target_environment: Environment,
     ):
         log.info('Triggering share failure alarm...')
-        subject = (
-            f'ALARM: DATAALL Table {table.GlueTableName} Sharing Failure Notification'
-        )
+        subject = f'ALARM: DATAALL Sharing Failure Notification for Table {table.GlueTableName}'[:100]
```
zsaltys, line 22: I would fix all of these to:

- remove ALARM: (don't think it's useful)
- rename DATAALL to data.all
- not use word notification and change to more concise version: "data.all share failure for table {table.GlueTableName}"

backend/dataall/modules/dataset_sharing/services/share_managers/lf_share_manager.py
```diff
@@ -480,17 +480,21 @@ def handle_share_failure(
...
-        DatasetAlarmService().trigger_table_sharing_failure_alarm(
-            table, self.share, self.target_environment
-        )
-        return True
+            DatasetAlarmService().trigger_table_sharing_failure_alarm(
+                table, self.share, self.target_environment
+            )
+            return True
+        except Exception as e:
+            logger.error(f"Exception during handle_share_failure: {e}")
```
zsaltys, line 496: This could be more meaningful: "Could not process dataset alarms" because otherwise this exception message is useless because we will see where it happened in the exception stack trace.. Also there's a nicer way to do it like this:

logger.error("Could not process dataset alarms: " , exc_info=True)

backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
```diff
@@ -288,14 +288,18 @@ def update_state(self, session, share_uri, new_state):
...
-        ShareObjectRepository.update_share_item_status(
-            session=session,
-            uri=share_item.shareItemUri,
-            status=new_state
-        )
-        self._state = new_state
-        return True
+        try:
+            logger.info(f"Updating share item in DB {share_item.shareItemUri} status to {new_state}")
+            ShareObjectRepository.update_share_item_status(
```
zsaltys, line 293: Imo returning FALSE here is not OK because it is a critical failure which cannot and should not be ignored or swallowed.. Because this line is the line that marks in-progress to FAILED... So I would change this to:

logger.error("Could not update share item status: " , exc_info=True)
raise e

zsaltys (changes requested):

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
anushka-singh: @zsaltys 
if there's an issue within self.handle_revoke_failure that raises another exception, it won't be caught by the except Exception as e block in process_revoked_shares. Instead, it will propagate up the call stack, and if not caught elsewhere, it could lead to the program terminating abruptly. To handle exceptions within self.handle_revoke_failure, we could include a try-except block within that method itself. Something like this:
```
def handle_revoke_failure(
        self,
        table: DatasetTable,
        share_item: ShareObjectItem,
        error: Exception,
) -> bool:
    try:
        logger.error(
            f'Failed to revoke S3 permissions to table {table.GlueTableName} '
            f'from source account {self.source_accountid}//{self.source_region} '
            f'with target account {self.target_environment.AwsAccountId}/{self.target_environment.region} '
            f'due to: {error}'
        )
        DatasetAlarmService().trigger_revoke_table_sharing_failure_alarm(
            table, self.share, self.target_environment
        )
        return True
    except Exception as inner_exception:
        # Log the inner exception, handle it, or raise it again if needed
        logger.error(f'Error in handle_revoke_failure: {inner_exception}')
        # Handle or raise the exception as needed

```

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
anushka-singh: Sure

backend/dataall/modules/dataset_sharing/services/share_managers/lf_share_manager.py
anushka-singh: Alright!

backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
anushka-singh: I agree

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -510,22 +510,27 @@ def handle_share_failure(self, error: Exception) -> None:
...
+                f'with target account {self.target_environment.AwsAccountId}/{self.target_environment.region} '
+                f'due to: {error}'
+            )
+            DatasetAlarmService().trigger_revoke_folder_sharing_failure_alarm(
+                self.target_folder, self.share, self.target_environment
+            )
+            return True
+        except Exception as e:
+            logger.error(f"Failed to trigger revoke folder sharing failure alarm due to: {e}")
+            raise e
```
zsaltys, line 533: Change to:

logger.error("Could not process dataset alarms: ", exc_info=True)
return False

backend/dataall/modules/dataset_sharing/services/share_managers/lf_share_manager.py
```diff
@@ -504,16 +508,20 @@ def handle_revoke_failure(
...
+                f'from source account {self.source_environment.AwsAccountId}//{self.source_environment.region} '
+                f'with target account {self.target_environment.AwsAccountId}/{self.target_environment.region} '
+                f'due to: {error}'
+            )
+            DatasetAlarmService().trigger_revoke_table_sharing_failure_alarm(
+                table, self.share, self.target_environment
+            )
+            return True
+        except Exception as e:
+            logger.error(f"Exception during handle_revoke_failure: {e}")
```
zsaltys, line 523: Change to:

logger.error("Could not process dataset alarms: ", exc_info=True)
return False

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -438,16 +438,20 @@ def handle_revoke_failure(self, error: Exception) -> bool:
...
+                f'Failed to revoke S3 permissions to bucket {self.bucket_name} '
+                f'from source account {self.source_environment.AwsAccountId}//{self.source_environment.region} '
+                f'with target account {self.target_environment.AwsAccountId}/{self.target_environment.region} '
+                f'due to: {error}'
+            )
+            DatasetAlarmService().trigger_revoke_s3_bucket_sharing_failure_alarm(
+                self.target_bucket, self.share, self.target_environment
+            )
+            return True
+        except Exception as e:
```
zsaltys, line 452: Change to:

logger.error("Could not process dataset alarms: ", exc_info=True)
return False

backend/dataall/modules/dataset_sharing/services/share_processors/lf_process_same_account_share.py
```diff
@@ -95,10 +95,10 @@ def process_approved_shares(self) -> bool:
                 shared_item_SM.update_state_single_item(self.session, share_item, new_state)
 
             except Exception as e:
-                self.handle_share_failure(table, share_item, e)
                 new_state = shared_item_SM.run_transition(ShareItemActions.Failure.value)
                 shared_item_SM.update_state_single_item(self.session, share_item, new_state)
                 success = False
+                self.handle_share_failure(table, share_item, e)
```
zsaltys, line 101: I would revert these swaps.. because if transition fails the alarm will never be sent because it throws an exception.. If we leave it as it was alarm will be sent or fail quietly with logging regardless whether the share state transition fails or not...

zsaltys (changes requested):

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: Alright!

backend/dataall/modules/dataset_sharing/services/share_processors/lf_process_same_account_share.py
anushka-singh: Done!

backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
zsaltys: we can resolve

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
```diff
@@ -51,9 +49,9 @@ def trigger_revoke_table_sharing_failure_alarm(
             target_environment: Environment,
     ):
         log.info('Triggering share failure alarm...')
-        subject = f'ALARM: DATAALL Table {table.GlueTableName} Revoking LF permissions Failure Notification'
+        subject = f'Data.all Revoke LF permissions Failure for Table {table.GlueTableName}'[:100]
         message = f"""
-    You are receiving this email because your DATAALL {self.envname} environment in the {self.region} region has entered the ALARM state, because it failed to revoke Lake Formation permissions for table {table.GlueTableName} with Lake Formation.
+    You are receiving this email because your Data.all {self.envname} environment in the {self.region} region has entered the ALARM state, because it failed to revoke Lake Formation permissions for table {table.GlueTableName} with Lake Formation.
```
zsaltys, line 54: lowercase data.all

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
```diff
@@ -19,11 +19,9 @@ def trigger_table_sharing_failure_alarm(
             target_environment: Environment,
     ):
         log.info('Triggering share failure alarm...')
-        subject = (
-            f'ALARM: DATAALL Table {table.GlueTableName} Sharing Failure Notification'
-        )
+        subject = f'Data.all Share Failure for Table {table.GlueTableName}'[:100]
```
zsaltys, line 22: lowercase data.all

zsaltys: do not capitalize words because the subjects are inconsistent.. for example in one subject word 'permissions' is not capitalized.. I'd just keep it normal english and only capitalize the first letter but because data.all always starts with lower d I'd just go with:

f'data.all share failure for table {table.GlueTableName}'[:100]

zsaltys (changes requested):

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
anushka-singh: I just did a quick search with case type match across all the code. 
Data.all is clearly the most used option over data.all.
data.all is only used in logs in quicksight.py. Everywhere else, Data.all is used. To be consistent across most files, I prefer using Data.all

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
anushka-singh: I wanted to be consistent with OS original Camel case usage in Subject. 
I dont mind changing the lowercase permissions to Permissions to be consistent across all Subjects in the alarms.

Additionally, I just did a quick search with case type match across all the code. 
Data.all is clearly the most used option over data.all.
data.all is only used in logs in quicksight.py. Everywhere else, Data.all is used. To be consistent across most files, I prefer using Data.all

zsaltys: Sounds good, if that's what the project is using then i'm OK with it.. I assumed they would use data.all. we can resolve.

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
```diff
@@ -51,9 +49,9 @@ def trigger_revoke_table_sharing_failure_alarm(
             target_environment: Environment,
     ):
         log.info('Triggering share failure alarm...')
-        subject = f'ALARM: DATAALL Table {table.GlueTableName} Revoking LF permissions Failure Notification'
+        subject = f'Data.all Revoke LF permissions Failure for Table {table.GlueTableName}'[:100]
```
zsaltys, line 52: update to Permissions to be consistent with capitalization

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
zsaltys: agreed

zsaltys (changes requested):

backend/dataall/modules/dataset_sharing/services/dataset_alarm_service.py
anushka-singh: Done!

backend/dataall/modules/dataset_sharing/services/share_processors/s3_bucket_process_share.py
```diff
@@ -161,9 +161,9 @@ def process_revoked_shares(
                 revoked_item_SM.update_state_single_item(session, removing_item, new_state)
 
             except Exception as e:
-                removing_bucket.handle_revoke_failure(e)
                 new_state = revoked_item_SM.run_transition(ShareItemActions.Failure.value)
                 revoked_item_SM.update_state_single_item(session, removing_item, new_state)
                 success = False
+                removing_bucket.handle_revoke_failure(e)
```
zsaltys, line 167: should move back where it was

zsaltys (changes requested):

zsaltys (approved):

zsaltys:
@dlpzx @noah-paige you may want to SQUASH this before merging...

backend/dataall/modules/dataset_sharing/services/share_processors/lf_process_cross_account_share.py
```diff
@@ -108,13 +108,15 @@ def process_approved_shares(self) -> bool:
                     self.create_resource_link(**data)
 
                     new_state = shared_item_SM.run_transition(ShareItemActions.Success.value)
-                    shared_item_SM.update_state_single_item(self.session, share_item, new_state)
 
                 except Exception as e:
+                    # Always keep success=False as the first line in this except in order to make finally work properly
+                    success = False
                     self.handle_share_failure(table=table, share_item=share_item, error=e)
                     new_state = shared_item_SM.run_transition(ShareItemActions.Failure.value)
```
zsaltys, line 116: I would suggest to change this and other parts to:

```
except Exception as e:
    # must run first to ensure state transitions to failed
    new_state = shared_item_SM.run_transition(ShareItemActions.Failure.value)
    success = False
    
    # statements which can throw exceptions but are not critical
    self.handle_share_failure(table=table, share_item=share_item, error=e)
    
finally:
    shared_item_SM.update_state_single_item(self.session, share_item, new_state)
```

zsaltys (changes requested):

zsaltys (approved):

noah-paige:
Tested in AWS Account:

- [x] CICD Pipeline Update Successful
- [x] Share Folder, Bucket, Table Successful
- [x] Revoke Folder, Bucket, Table Successful
- [x] Force Failed Share --> Alarm Notification Sent to SNS, Share not stuck
- [x] Force Failed Revoke --> Alarm Notification Sent to SNS, Share not stuck

noah-paige (approved):
Reviewed code and tested, looks good!

PR #934, opened 2023-12-19
feat: support for custom env linking text with sanitization

zsaltys:
**Feature or Bugfix**
- Feature

**Detail**

Adding a way to override env linking prerequisites for those that do not want to use manual steps and have other ways to achieve the same thing.

**Security**
The only security concern is that there's an html value coming out of config.json which gets set as innerHTML. I don't think this would be an issue unless someone would find a way to override the config. value as an attack.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -32,6 +32,7 @@
         "classnames": "^2.3.1",
         "date-fns": "^2.28.0",
         "dayjs": "^1.11.0",
+        "dompurify": "^3.0.6",
         "formik": "^2.2.9",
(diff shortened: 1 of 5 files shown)


noah-paige:
Thanks for re-opening this PR with resolved conflicts @zsaltys - I am reviewing this now

noah-paige:
Tested in AWS Deployment:

- [x] CICD Pipeline Succeeds
- [x] Link Environment Shows Correctly (Both with and without `custom_env_linking_text` in `config.json`)

noah-paige (approved):
Looks good to me - approving

PR #948, opened 2024-01-05
Fix when migrating from Manually Created Pivot Role to Auto Create Pivot Role

TejasRGitHub:
### Feature or Bugfix
- Bugfix

### Detail

- Changes in the way KMS keys of datasets( S3 buckets ) are specified in the environment role .

### Testing 

- Created an environment with manual pivot role and migrated to auto create pivot role. 
- Checked if the datasets previously present in the environment get updated properly. 
- Imported another dataset on this environment successfully. 

### Relates
- https://ouryahoo.atlassian.net/browse/DATA-416

### Security

  - Is the input sanitized?
  - Have you used the least-privilege principle? How? **Yes, only access to KMS keys which are present on the environment**


backend/dataall/modules/datasets/cdk/env_role_dataset_s3_policy.py
```diff
@@ -70,18 +70,16 @@ def _generate_dataset_statements(datasets: List[Dataset]):
 
     @staticmethod
     def _set_allowed_kms_keys_statements(datasets):
-        allowed_buckets_kms_keys = []
+        imported_kms_alias = []
+        # Datasets belonging to a team and an environment are present in same region and aws account
+        imported_dataset_resources = [f"arn:aws:kms:{datasets[0].region}:{datasets[0].AwsAccountId}:key/*"]
```
dlpzx, line 75: I think this can cause errors if there are no datasets. Can we move line 75 after we check if there are datasets in line 76?

TejasRGitHub: From the code flow no need according to me . Because this function is called at line 66. Which is only called when there are some datasets which are present. 
But terms of making this function more clean and error resistant if its called somewhere else I will change this as per you recommendation

TejasRGitHub (commented):
Updating code to add imported section after checking if datasets are present

dlpzx (approved):
reviewed and tested

Issue #956, opened 2024-01-09
Dataset sharing not working for imported dataset after enabling auto create pivot Role 


anushka-singh:
### Describe the bug

Error: Failed to attach policy to KMS key xxxxx on yyyyy : An error occurred (MalformedPolicyDocumentException) when calling the PutKeyPolicy operation: The new key policy will not allow you to update the key policy in the future.

### How to Reproduce

```
Enable auto create pivot role and try a share and it will fail
```

### OS

Mac

### Python version

3.8

### AWS data.all version

2.2

PR #961, opened 2024-01-10
Bugfix 956

anushka-singh:
### Feature or Bugfix
- Bugfix

### Detail
Error: Failed to attach policy to KMS key xxxxx on yyyy : An error occurred (MalformedPolicyDocumentException) when calling the PutKeyPolicy operation: The new key policy will not allow you to update the key policy in the future.

### Relates
- https://github.com/data-dot-all/dataall/issues/956

### Security

  - Is the input sanitized?


backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -485,6 +499,8 @@ def delete_dataset_bucket_key_policy(
                 principal_list.remove(f"{target_requester_arn}")
                 if len(principal_list) == 0:
                     statements.pop(DATAALL_ACCESS_POINT_KMS_DECRYPT_SID)
+                    if DATAALL_ACCESS_POINT_KMS_DECRYPT_SID not in statements.keys():
```
noah-paige, line 502: do we need this if statement? If there are no principals in `principal_list` can we just remove `DATAALL_ACCESS_POINT_KMS_DECRYPT_SID`

We may need to keep `DATAALL_ACCESS_POINT_ENABLE_PIVOT_ROLE_PERMISSIONS_SID` Statement to avoid `MalformedPolicyDocument` error

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -541,6 +558,29 @@ def generate_default_kms_decrypt_policy_statement(target_requester_arn):
...
         }
 
+    @staticmethod
+    def generate_enable_pivot_role_permissions_policy_statement(dataset_account_id):
+        return {
+            "Sid": f"{DATAALL_ACCESS_POINT_ENABLE_PIVOT_ROLE_PERMISSIONS_SID}",
+            "Effect": "Allow",
+            "Principal": {
+                "AWS": [
+                    f"arn:aws:iam::{dataset_account_id}:role/dataallPivotRole"
```
noah-paige, line 568: pivot role could also be `dataallPivotRole-cdk` - should look up from SSM what the pivotRoleName is

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -405,6 +419,8 @@ def delete_target_role_bucket_key_policy(
                     principal_list.remove(f"{target_requester_arn}")
                     if len(principal_list) == 0:
                         statements.pop(DATAALL_BUCKET_KMS_DECRYPT_SID)
+                        if DATAALL_BUCKET_KMS_DECRYPT_SID not in statements.keys():
```
noah-paige, line 422: Same as Access point processor

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -482,3 +498,32 @@ def generate_default_kms_decrypt_policy_statement(target_requester_arn):
...
         }
+
+    @staticmethod
+    def generate_enable_pivot_role_permissions_policy_statement(source_account_id):
+        return {
+            "Sid": f"{DATAALL_BUCKET_ENABLE_PIVOT_ROLE_PERMISSIONS_SID}",
+            "Effect": "Allow",
+            "Principal": {
+                "AWS": [
+                    f"arn:aws:iam::{source_account_id}:role/dataallPivotRole"
```
noah-paige, line 509: Same as Access point processor

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -482,3 +498,32 @@ def generate_default_kms_decrypt_policy_statement(target_requester_arn):
...
+                "kms:GenerateDataKey*",
+                "kms:PutKeyPolicy",
+                "kms:GetKeyPolicy",
+                "kms:ReEncrypt*",
+                "kms:TagResource",
+                "kms:UntagResource",
+                "kms:DescribeKey"
+               ],
+            "Resource": "*",
+            "Condition": {
```
noah-paige, line 524: Do not think Condition needed in Key Policy

anushka-singh: Removed

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: I am popping both Statements if principal list is 0 in the next revision. 
I dont think it will lead to MalformedPolicyDocument.. But we can test it out.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
anushka-singh: Done

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
anushka-singh: Done

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -23,6 +23,7 @@
 IAM_ACCESS_POINT_ROLE_POLICY = "targetDatasetAccessControlPolicy"
 DATAALL_ALLOW_OWNER_SID = "AllowAllToAdmin"
 DATAALL_ACCESS_POINT_KMS_DECRYPT_SID = "DataAll-Access-Point-KMS-Decrypt"
+DATAALL_ACCESS_POINT_ENABLE_PIVOT_ROLE_PERMISSIONS_SID = "DataAll-Access-Point-Enable-Pivot-Role-Permissions"
```
noah-paige, line 26: One thing I am realizing is that if a dataset has a folder share and a bucket share then there will be 2 statements enabling PivotRole Permissions in the Key Policy 

Would be curious to see if there's a way to not have duplicate here but also understand having them separate is far easier to manage in terms of checking if exists and removing if no more folder and/or bucket shares

... something to think about but deploying now to test as is and confirm the above suspicion

anushka-singh: I had the same thought. 
Here is how I think about it: Eventually we will move to having refactored code where we just have 1 file with all these functions consolidated in it. Currently there is a lot of common code between access point and bucket share that can be pulled out into a separate class. At that time, we can have just one of these pivot role auto create permissions in place.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
noah-paige: testing the above while processing a bucket share worked with this PR, I did receive `MalformedPolicyDocumentException` when revoking a share 

Do we see any issue with checking if `DATAALL_ACCESS_POINT_ENABLE_PIVOT_ROLE_PERMISSIONS_SID` statement exists and if not just creating (i.e. never removing the permissions from the KMS Key Policy)

These are the required permissions we already implicitly give to the pivotRole in IAM with our conditional statements... I do not see an issue with doing as such but let me know your thoughts @anushka-singh @dlpzx

noah-paige: If we do want to also keep the remove logic then can we provide a default value when we do `.pop()`, such as:

`statements.pop(DATAALL_BUCKET_ENABLE_PIVOT_ROLE_PERMISSIONS_SID, True)`

Otherwise we may run into `KeyErrors` for existing shares after they include the PR and they try to run a revoke

noah-paige:
One other thing to note is that for created datasets we already create a KMS Key and explicitly give the pivotRole permission in the Key Policy (check out L140-161 in `dataall/backend/dataall/modules/datasets/cdk/dataset_stack.py`)

My suggestion would be the following to avoid duplicate permissions from different statements:

- Change `DATAALL_BUCKET_ENABLE_PIVOT_ROLE_PERMISSIONS_SID` and `DATAALL_ACCESS_POINT_ENABLE_PIVOT_ROLE_PERMISSIONS_SID` to be the same constant `KMSPivotRolePermissions` (which is also the same `SID` used in `dataset_stack.py`)
- Only have the logic that checks in the SID exists and if not adds the explicit KMS Permissions
- Remove the logic that removes the SID if there are no more principals

I think the above would be the cleanest and most straightforward way to resolve the bug but open to other ideas as well!

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
dlpzx: Something like a `share_managers/kms_policy_utils` similar to the `share_manager_utils`? We could have a single function that handles `update_dataset_bucket_key_policy` for both. I like that it reduces the places where we modify the policy to a single one

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
dlpzx: In my opinion it makes sense to not remove it, because the pivot role access to the key is similar to a configuration requirement of a dataset that we apply in a lazy mode. 

If we keep it we avoid issues with simultaneous revoke/shares that create and delete it and with Malformedpolicy errors

anushka-singh: Okay I agree!
Good find @noah-paige 

In the remove functions, I will not pop `DATAALL_ACCESS_POINT_ENABLE_PIVOT_ROLE_PERMISSIONS_SID`

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
anushka-singh: @dlpzx I 100% agree we should consolidate the functions into a single utils file. However, I want to do that as a part of clean up outside of this PR. If thats okay with you, I can keep update_dataset_bucket_key_policy in both access point and bucket share for now and we can extract common functions out as a part of different issue I will create.

anushka-singh: In the new revision I have used KMSPivotRolePermissions as Noah advised in both the share managers.

backend/dataall/core/stacks/db/enums.py
```diff
@@ -6,7 +6,7 @@ class StackStatus(Enum):
     CREATE_COMPLETE = 'CREATE_COMPLETE'
     CREATE_FAILED = 'CREATE_FAILED'
     DELETE_IN_PROGRESS = 'DELETE_IN_PROGRESS'
-    DELETE_COMPLETE = 'DELETE_FAILED'
```
anushka-singh, line 9: I spotted this bug while debugging something else. Hence fixed it here. Please let me know if it looks good, otherwise I can remove this change!

dlpzx: It is very minor, we can include it here

noah-paige:
Tested in AWS
- [x] CICD Pipeline Succeeds

For created dataset:
- [x] Revoke existing Bucket Share - successful
- [x] Process Bucket Share (share succeeds, no additional kms statement for pivotRole)
- [x] Revoke Bucket Share (share revoked)
- [x] Process Folder Share (share succeeds, no additional kms statement for pivotRole)
- [x] Revoke Folder Share (share revoked)

For imported dataset with imported KMS Key:
- [x] Process Bucket Share (share succeeds, adds kms statement for pivotRole)
- [x] Revoke Bucket Share (share revoked, pivotRole still has permissions)
- [x] Process Folder Share (share succeeds, adds kms statement for pivotRole)
- [x] Revoke Folder Share (share revoked, pivotRole still has permissions)

noah-paige:
I finished my testing and everything is working perfectly - my one minor commment is that I am not a huge fan of swapping `kms_client.get_key_id(key_alias)` to `kms_client.get_key_id_using_list_aliases(key_alias)`

It feels a littly hacky or just not as straightforward to list the key aliases, iterate through to find the matching one and then parse out the associated KMS Key Id from there.

I know @anushka-singh there were some issues earlier in testing with `kms_client.get_key_id(key_alias)` but I would be inclined to try out the above code with reverting back to `kms_client.get_key_id(key_alias)` to see if it works - I plan to do so now to see

If you look at some of the earlier functions called in the S3 Bucket Share Processor - such as `grant_s3_iam_access()` - the methods use the `kms_client.get_key_id(key_alias)` rather than the other way of getting Key Id

so I do not think it should be an issue to do the same when updating the key policy in `update_dataset_bucket_key_policy()` or the equivalent delete method


All that being said if @anushka-singh or @dlpzx you think the above is not a requirement I can approve as is and have the above as a later enhancement as required

anushka-singh:
> I finished my testing and everything is working perfectly - my one minor commment is that I am not a huge fan of swapping `kms_client.get_key_id(key_alias)` to `kms_client.get_key_id_using_list_aliases(key_alias)`
> 
> It feels a littly hacky or just not as straightforward to list the key aliases, iterate through to find the matching one and then parse out the associated KMS Key Id from there.
> 
> I know @anushka-singh there were some issues earlier in testing with `kms_client.get_key_id(key_alias)` but I would be inclined to try out the above code with reverting back to `kms_client.get_key_id(key_alias)` to see if it works - I plan to do so now to see
> 
> If you look at some of the earlier functions called in the S3 Bucket Share Processor - such as `grant_s3_iam_access()` - the methods use the `kms_client.get_key_id(key_alias)` rather than the other way of getting Key Id
> 
> so I do not think it should be an issue to do the same when updating the key policy in `update_dataset_bucket_key_policy()` or the equivalent delete method
> 
> All that being said if @anushka-singh or @dlpzx you think the above is not a requirement I can approve as is and have the above as a later enhancement as required

We did try out using `kms_client.get_key_id(key_alias)`. The issue was a chicken and egg problem. To get key id to then get the existing policy, pivot role needed access to kms. But we were adding access later in the share manager. The error was: `Failed to get kms key id of alias/di2: An error occurred (AccessDeniedException) when calling the DescribeKey operation: User: arn:aws:sts::xxxx:assumed-role/dataallPivotRole/dataallPivotRole is not authorized to perform: kms:DescribeKey on resource: arn:aws:kms:us-east-1:xxxx:key/yyyy because no resource-based policy allows the kms:DescribeKey action`

Describe key is used in get_key_id function and needs some extra privileges. ListAliases however did not need those extra permissions - which is why Adriana and I decided to use listAliases which we already use in check_key_exists function in a similar pattern already. 
Having said that, please let me know if you are able to make it work with just get_key_id in your trial and we can implement that.

noah-paige:
Understood the above - but then in a hypothetical bucket share how does it work since the step `grant_s3_iam_access()` uses the `KMSClient.get_key_id()` function and is executed before calling `grant_dataset_bucket_key_policy()` which updates the KMS Key Policy

The pivotRole should have `kms:DescribeKey` permissions on all KMS Keys (it is one of the few wildcard permissions we permit) so as long as the pivotRole has the default KMS Key Policy (i.e. delegating IAM to manage permissions) than I would presume it should be okay - will report back with testing findings thought!

noah-paige:
Replacing `get_key_id_using_list_aliases()` with `get_key_id()` I tested in AWS again:

Tested in AWS
- [x] CICD Pipeline Succeeds

For created dataset:
- [x] Process Bucket Share (share succeeds, no additional kms statement for pivotRole)
- [x] Revoke Bucket Share (share revoked)
- [x] Process Folder Share (share succeeds, no additional kms statement for pivotRole)
- [x] Revoke Folder Share (share revoked)

For imported dataset with imported KMS Key:
- [x] Process Bucket Share (share succeeds, adds kms statement for pivotRole)
- [x] Revoke Bucket Share (share revoked, pivotRole still has permissions)
- [x] Process Folder Share (share succeeds, adds kms statement for pivotRole)
- [x] Revoke Folder Share (share revoked, pivotRole still has permissions)

anushka-singh:
> Replacing `get_key_id_using_list_aliases()` with `get_key_id()` I tested in AWS again:
> 
> Tested in AWS
> 
> * [x]  CICD Pipeline Succeeds
> 
> For created dataset:
> 
> * [x]  Process Bucket Share (share succeeds, no additional kms statement for pivotRole)
> * [x]  Revoke Bucket Share (share revoked)
> * [x]  Process Folder Share (share succeeds, no additional kms statement for pivotRole)
> * [x]  Revoke Folder Share (share revoked)
> 
> For imported dataset with imported KMS Key:
> 
> * [x]  Process Bucket Share (share succeeds, adds kms statement for pivotRole)
> * [x]  Revoke Bucket Share (share revoked, pivotRole still has permissions)
> * [x]  Process Folder Share (share succeeds, adds kms statement for pivotRole)
> * [x]  Revoke Folder Share (share revoked, pivotRole still has permissions)
> 
> @anushka-singh if you would have some time to validate the same on your end and if all works I would encourage we switch back to `get_key_id()`, remove the `get_key_id_using_list_aliases()` function and then I am happy to approve from there

Hey Noah!
Thanks for testing this theory out. I was able to replicate the same behavior on my end. I used `get_key_id()`, instead of the `get_key_id_using_list_aliases()` function and was able to have successful share and revoke operations. I have pushed another revision with that version. Please approve and merge if you see fit.

noah-paige (approved):
Tested the changes in AWS and all looks good, approving now!

Issue #902, opened 2023-12-04
Add WAF rules to Cognito user pool


dlpzx:
**Is your idea related to a problem? Please describe.**
Currently we use WAF to protect the API Gateway and CloudFront distribution. Cognito released in 2022 the integration with WAF, which adds WAF rules and protection to the Cognito user pool. We have the opportunity to increase security by implementing WAF rules in data.all Cognito.

**Describe the solution you'd like**
Add WAF rules to the Cognito user pool in data.all

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

PR #976, opened 2024-01-14
Add WAF ACL to Cognito User Pool

noah-paige:
### Feature or Bugfix

- Feature

### Detail
- Add WAF ACL to Cognito

### Relates
- https://github.com/data-dot-all/dataall/issues/902

### Security

  - Is the input sanitized?

### DIFF
--- a/deploy/stacks/cognito.py
+++ b/deploy/stacks/cognito.py
@@ -15,6 +15,7 @@ from aws_cdk import (
 
 from .pyNestedStack import pyNestedClass
 from .solution_bundling import SolutionBundling
+from .waf_rules import get_waf_rules
 
 
 class IdpStack(pyNestedClass):
@@ -77,7 +78,7 @@ class IdpStack(pyNestedClass):
                 metric_name='waf-cognito',
                 sampled_requests_enabled=True,
             ),
-            rules=self.get_waf_rules(envname, custom_waf_rules, ip_set_regional),
+            rules=get_waf_rules(envname, 'Cognito', custom_waf_rules, ip_set_regional),
         )
 
         wafv2.CfnWebACLAssociation(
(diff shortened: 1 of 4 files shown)


deploy/stacks/cognito.py
```diff
@@ -50,6 +53,41 @@ def __init__(
...
             advanced_security_mode='ENFORCED'
         )
+
+        # Create IP set if IP filtering enabled in CDK.json
+        ip_set_regional = None
+        if custom_waf_rules and custom_waf_rules.get('allowed_ip_list'):
+            ip_set_regional = wafv2.CfnIPSet(
+                self,
+                'DataallRegionalIPSet-Cognito',
+                name=f'{resource_prefix}-{envname}-ipset-regional',
```
dlpzx, line 63: The name of the IP set is the same one as for the WAF of API GW in `lambda_api.py`. This will cause `WAFDuplicateItemException: AWS WAF couldn’t perform the operation because some resource in your request is a duplicate of an existing one.` when `allowed_ip_list` is defined

dlpzx: We can create 2 different IP sets (as we are trying to do now) or we can think of how to merge it in one. I personally see the value fo having 2 separate IP sets that allows this feature to be extended and restrict IP addresses differently for Cognito and for API GW

deploy/stacks/waf_rules.py
```diff
@@ -0,0 +1,179 @@
...
+                sampled_requests_enabled=True,
+                cloud_watch_metrics_enabled=True,
+                metric_name='AWS-AWSManagedRulesKnownBadInputsRuleSet',
+            ),
+            priority=priority,
+            override_action=wafv2.CfnWebACL.OverrideActionProperty(none={}),
+        )
+    )
+    priority += 1
+    waf_rules.append(
```
dlpzx, line 127: I have been searching for recommended WAF rules for Cognito in particular. Because it is a recent launch the resources are a bit limited, but in general geo-fencing, IP-allow-list, rate limits and the typical AWSManagedRulesCommonRuleSet appear in demos and blogs.
From a cost perspective, WAF currently adds $23/month to the data.all bill. But if you see the docs: https://aws.amazon.com/waf/pricing/ going over 1500 WCUs affects the costs slightly `The total WCUs for a web ACL can't exceed 5000. Using over 1500 WCUs affects your costs.` we are currently at 1427.

In short:
- there are no additional WAF rules to add at the moment
- we might remove the SQL or the linux rules, but it is not a big deal because the costs are very low. wdyt?

dlpzx (changes requested):
I will test with custom_waf_rules once the issue of the names of the IP sets is resolved

deploy/stacks/cognito.py
noah-paige: good catch - updating the PR now

deploy/stacks/waf_rules.py
noah-paige: Agreed - I could not find any prescriptive guidance on best practice WAF rules for Cognito. I think we are okay to keep rules the same for now and if needed break it out so each WebACL has its own unique rule set at a later point

dlpzx:
Overall looks good! Testing in AWS with custom_waf_rules:

- [ ] CICD Pipeline Succeeds
- [ ] Added WAF Rules to Cognito User Pool (no updates to Cloudfront or API GW)
- [ ] Log In, Log Out, and Access Data.all UI

deploy/stacks/cognito.py
```diff
@@ -26,6 +28,7 @@ def __init__(
         vpc=None,
         prod_sizing=False,
         internet_facing=True,
+        custom_waf_rules=None,
```
dlpzx, line 31: I have been tracking down this parameter and it seems unused upstream. It is not defined neither in the backend stack nor in the cdk.json file. Should we fix it in this PR or should we open a separate issue for it?

dlpzx, line 31: I have been tracking this parameter upstream and I cannot see it neither in the `cdk.json` nor in the backend stack. Should we re-enable it and add it to the cdk.json template and docs? In such case, maybe it is worth opening a separate PR. wdyt?

dlpzx:
I created another issue for the `custom_waf_rules` : https://github.com/data-dot-all/dataall/issues/995

PR #978, opened 2024-01-16
Enum refactoring

SofiaSazonova:
### Feature or Bugfix
- Refactoring

### Detail
- Get rid of redefinitions of the same Enums. Used class GraphQLEnumMapper where it's required
- Unhardcoded enum-constants for migrations and tests

### Security N/A


backend/__init__.py
dlpzx, line 1: Why are we adding this file?

backend/dataall/modules/dataset_sharing/db/share_object_models.py
```diff
@@ -5,7 +5,7 @@
 from sqlalchemy.orm import query_expression
 
 from dataall.base.db import Base, utils
-from dataall.modules.dataset_sharing.db.enums import ShareObjectStatus, ShareItemStatus
+from ..common.enums import ShareObjectStatus, ShareItemStatus
```
dbalintx, line 8: I suggest using absolute imports with full path to stick to the convention. Also in some other files

backend/dataall/modules/datasets/api/dataset/resolvers.py
```diff
@@ -52,15 +52,15 @@ def resolve_user_role(context: Context, source: Dataset, **kwargs):
                 .first()
             )
             if share and (
-                share.owner == context.username or share.principalId in context.groups
+                    share.owner == context.username or share.principalId in context.groups
```
dbalintx, line 55: not sure if these formatting changes are needed in this file

backend/dataall/modules/datasets/api/table/enums.py
```diff
@@ -1,12 +1,6 @@
 from dataall.base.api.constants import GraphQLEnumMapper
 
 
-class DatasetSortField(GraphQLEnumMapper):
-    created = 'created'
-    updated = 'updated'
-    label = 'label'
-
-
 class DatasetTablePreviewStatus(GraphQLEnumMapper):
```
dbalintx, line 4: maybe move this enum to `datasets_base/constants/enums.py ` and get rid of this file?

backend/dataall/modules/dataset_sharing/common/enums.py
```diff
@@ -1,3 +1,4 @@
+
```
dlpzx, line 1: remove empty line

backend/dataall/modules/dataset_sharing/common/enums.py
```diff
@@ -1,3 +1,4 @@
...
 from dataall.base.api.constants import GraphQLEnumMapper
```
dlpzx, line 4: Why did we move it out of `api`? All other GraphQLEnumMapper enums are inside the `api` subpackage of the module

dlpzx (approved):

backend/migrations/versions/5e5c84138af7_backfill_confidentiality.py
```diff
@@ -74,8 +76,8 @@ def upgrade():
         print('Updating datasets...')
         datasets: [Dataset] = session.query(Dataset).all()
         for dataset in datasets:
-            if dataset.confidentiality not in ['Unclassified', 'Official', 'Secret']:
-                dataset.confidentiality = 'Unclassified'
+            if dataset.confidentiality not in ConfidentialityClassification:
```
dbalintx, line 79: are you sure this is working?
you could do something like:

if not any(value == dataset.confidentiality for value in ConfidentialityClassification.__dict__.values()):

noah-paige:
Hi @SofiaSazonova @dlpzx - what is left to be done before we can push forward testing and approving this PR...

I see we have some `common/enums.py` files still in the repo for `dataset_sharing` and datasets_base` - did we decide best to move under `api/` folder or do we want to keep it as is now?

I am open for alternative approaches / further discussion - want to align so I can help review and push this PR through

SofiaSazonova:
> Hi @SofiaSazonova @dlpzx - what is left to be done before we can push forward testing and approving this PR...
> 
> I see we have some `common/enums.py` files still in the repo for `dataset_sharing` and datasets_base`- did we decide best to move under`api/` folder or do we want to keep it as is now?
> 
> I am open for alternative approaches / further discussion - want to align so I can help review and push this PR through

@noah-paige  I moved all enums to "common" folder since they are used not only in api, but also in models and etc. If you or @dlpzx insist I can move them back to api folder, but in my opinion it's odd))

tests/modules/datasets/conftest.py
```diff
@@ -40,12 +40,12 @@ def dataset(client, patch_es, patch_dataset_methods):
...
-        name: str,
-        owner: str,
-        group: str,
-        confidentiality: str = None
+            org: Organization,
+            env: Environment,
+            name: str,
+            owner: str,
+            group: str,
+            confidentiality: str = None
```
dlpzx, line 48: remove additional indentation

tests/modules/datasets/conftest.py
```diff
@@ -265,9 +268,9 @@ def folder_fixture(db, dataset_fixture):
 @pytest.fixture(scope="module")
 def dataset_model(db):
     def factory(
-        organization: Organization,
-        environment: Environment,
-        label: str
+            organization: Organization,
+            environment: Environment,
+            label: str
```
dlpzx, line 273: Remove additional indentation

dlpzx:
> > Hi @SofiaSazonova @dlpzx - what is left to be done before we can push forward testing and approving this PR...
> > I see we have some `common/enums.py` files still in the repo for `dataset_sharing` and datasets_base`- did we decide best to move under`api/` folder or do we want to keep it as is now?
> > I am open for alternative approaches / further discussion - want to align so I can help review and push this PR through
> 
> @noah-paige I moved all enums to "common" folder since they are used not only in api, but also in models and etc. If you or @dlpzx insist I can move them back to api folder, but in my opinion it's odd))

In my opinion there is no need to create a new layer `common` in each module. `db`, `api`, `services` already import each other, it is not like they are independent sub-modules. I pictured layers as functional components with clearly defined purposes, `common` is a bit too generic and has the risk of being a "here we can store anything" layer. That's why I would add the constants to an existing layer. 

Now, which one should it be? We can argue that since it is business logic they should be defined as part of `services`  in a file such as `ml_studio_enums.py`. Another option would be to add it into `api` because the enums inherit the class `GraphQLEnumMapper`.  I am 50% 50% with both options. 

Another idea that I got is to rename the enums, instead of `enum.py` name them as `modulename_enums.py` to make it easier to search and manage each module separately. Similar to `<modulename_models.py` in `db` or `mlstudio_service.py` in `services`. What do you think?

SofiaSazonova:
I agree, that naming `modulename_enums.py` is much better. 
As for storage of the file...I think may be then models is better? Like it's a part of field definition as well.

SofiaSazonova:
Also, there is no `api` folder in `datasets_base`. So, for unification I tend more to `services`, as you proposed

SofiaSazonova:
So, I moved enums to services and renamed them as `module_enums`

dlpzx:
Overall  changes look good @SofiaSazonova. I am doing a last test in AWS before approving
- [ ] CICD completes
- [ ] Share Enum - share functionality works as expected
- [ ] Dataset Enums - creation of Dataset
- [ ] Organization Enums - create Organization

dlpzx (approved):
Tested in AWS

Issue #377, opened 2023-03-16
Add parameter to show data.all current version


dlpzx:
**Is your idea related to a problem? Please describe.**
I do not like going to the commits and looking for the release dates to establish the version of data.all that I currently have deployed. 

**Describe the solution you'd like**
I would like to look up this information in a more straight forward way. For example we could implement an base file or an SSM parameter to keep this information.

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

noah-paige:
Closing this issue as it has been implemented in PR #991

PR #991, opened 2024-01-19
377 version param

SofiaSazonova:
### Feature or Bugfix

- Feature

### Detail
version.json in repo root contains information about version. This json-string is uploaded to SSM parameter /dataall/{env}/versin. Also this file is copied to frontend folder and version number is displayed near "dataall" header 

### Relates
https://github.com/data-dot-all/dataall/issues/377

### Security

fetching data from storage outside the application (e.g. a database, an S3 bucket)? NA
  - Is the input sanitized?
- Does this PR introduce any functionality or component that requires authorization? NA
- Are you using or adding any cryptographic features? NA
- Are you introducing any new policies/roles/users?NA


SofiaSazonova:
As long, as @dlpzx required simple way to store version info, I made it as obvious as possible.
Version info is updated only in one file - version.json
We will both have ssm-param for that and static file in the front.

**What can be different?**
We can introduce another API route (smth like /metadata) to fetch version from backend and display at the frontend.
In this case it will be 100% the same at front and back, but maybe in future we want to have different versions for front and back? Also, I decided not to make another API section just for one version-param.

SofiaSazonova:
Screenshot 2024-01-19 at 20 01 48
Screenshot 2024-01-19 at 20 02 05
Screenshot 2024-01-19 at 20 03 00

deploy/stacks/pipeline.py
```diff
@@ -558,6 +559,7 @@ def set_quality_gate_stage(self):
                         'mkdir -p source_build',
                         'mv backend ./source_build/',
                         'mv config.json ./source_build/',
+                        'mv version.json ./source_build/',
```
noah-paige, line 562: Do we need this line for `UploadCodeToS3`? I think we can remove as the `source_build.zip` is used primarily for DBMigration step only which should not be dependent on version

noah-paige:
Left a minor comment - otherwise code looks good and a very much needed enhancement 🙌 

To finish up testing and push through by Monday:

Deployed to AWS (Internet Facing)
- [ ] CICD Successful
- [ ] Version present in login logo + webpage

Deployed to AWS (VPC Facing)
- [ ] CICD Successful
- [ ] Version present in login logo + webpage

Deployed Local
- [ ] Logo present in webpage

SofiaSazonova:
Now version is tooltip
image

deploy/stacks/pipeline.py
noah-paige: Let's remove this line `mv version.json ./source_build/` - I tested to ensure it is not needed

I think best to keep hardcoding of `version.json` minimal to only when necessary

Will approve right after this change

SofiaSazonova: Done

noah-paige (approved):
Looks good - approving!

PR #1020, opened 2024-01-30
adding update consumption role capability

### LINKED ISSUE #870: Add capability on Data.all to allow changing the owner for a consumption role without deleting and recreating the role.
**Is your idea related to a problem? Please describe.**
Data.all currently does not support any updates for the owners on the consumption role. For any change, the consumption role has to be deleted and recreated with the new owner association.

1. Navigate to Environments > Teams and add consumption role.
2. The role will prompt to select an owner from the drop down menu. Select an owner and click on "Add Consumption Role"

The role gets created. To re assign a different owner for the role, there is no provision to edit the consumption role. The only option currently is to delete the consumption role and recreate it by choosing the new owner from the drop down menu again.

**Describe the solution you'd like**
Add capability on the UI to allow the admins to edit the consumption role and map it to a different owner.

petrkalos:
### Feature or Bugfix

- Feature

### Detail
- Add capability to update consumption role owner and name
- Setting the paradigm to use [DataGrid (with edit/CRUD ops)](https://v5.mui.com/x/react-data-grid/editing/#full-featured-crud-component) instead of Table

### Relates
- Resolves #870 

### Security

### Testing

Tested with multiple accounts and roles
* Locally
* AWS account
* Added unit-tests
  * Mocked the entire `sts` client as the tests were sending request to AWS

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/core/environment/api/resolvers.py
+++ b/backend/dataall/core/environment/api/resolvers.py
@@ -222,6 +222,17 @@ def remove_consumption_role(context: Context, source, environmentUri=None, consu
     return status
 
 
+def update_consumption_role(context: Context, source, environmentUri=None, consumptionRoleUri=None, input={}):
+    with context.engine.scoped_session() as session:
+        status = EnvironmentService.update_consumption_role(
+            session=session,
+            uri=consumptionRoleUri,
+            env_uri=environmentUri,
+            input=input,
+        )
+    return status
+
+
 def list_environment_invited_groups(
     context: Context, source, environmentUri=None, filter=None
 ):
--- a/backend/dataall/core/environment/api/resolvers.py
+++ b/backend/dataall/core/environment/api/resolvers.py
@@ -224,13 +224,13 @@ def remove_consumption_role(context: Context, source, environmentUri=None, consu
 
 def update_consumption_role(context: Context, source, environmentUri=None, consumptionRoleUri=None, input={}):
     with context.engine.scoped_session() as session:
-        status = EnvironmentService.update_consumption_role(
+        consumption_role = EnvironmentService.update_consumption_role(
             session=session,
             uri=consumptionRoleUri,
             env_uri=environmentUri,
             input=input,
         )
-    return status
+    return consumption_role
 
 
 def list_environment_invited_groups(
--- a/backend/dataall/core/environment/services/environment_service.py
+++ b/backend/dataall/core/environment/services/environment_service.py
@@ -463,6 +463,12 @@ class EnvironmentService:
     @has_tenant_permission(permissions.MANAGE_ENVIRONMENTS)
     @has_resource_permission(permissions.REMOVE_ENVIRONMENT_CONSUMPTION_ROLE)
     def update_consumption_role(session, uri, env_uri, input):
+        if not input:
+            raise exceptions.RequiredParameter('input')
+        if not input.get('groupUri'):
+            raise exceptions.RequiredParameter('groupUri')
+        if not input.get('consumptionRoleName'):
+            raise exceptions.RequiredParameter('consumptionRoleName')
         role_query = session.query(ConsumptionRole).filter(
             (
                 and_(
--- a/backend/dataall/core/environment/services/environment_service.py
+++ b/backend/dataall/core/environment/services/environment_service.py
@@ -469,15 +469,7 @@ class EnvironmentService:
             raise exceptions.RequiredParameter('groupUri')
         if not input.get('consumptionRoleName'):
             raise exceptions.RequiredParameter('consumptionRoleName')
-        role_query = session.query(ConsumptionRole).filter(
-            (
-                and_(
-                    ConsumptionRole.consumptionRoleUri == uri,
-                    ConsumptionRole.environmentUri == env_uri,
-                )
-            )
-        )
-        consumption_role = role_query.first()
+        consumption_role = EnvironmentService.get_environment_consumption_role(session, uri, env_uri)
         if consumption_role:
             ResourcePolicy.update_resource_policy(
                 session=session,
@@ -487,9 +479,10 @@ class EnvironmentService:
                 new_group=input['groupUri'],
                 new_permissions=permissions.CONSUMPTION_ROLE_ALL
             )
-            role_query.update(input)
+            for key, value in input.items():
+                setattr(consumption_role, key, value)
             session.commit()
-        return role_query.first()
+        return consumption_role
 
     @staticmethod
     def query_user_environments(session, username, groups, filter) -> Query:
@@ -849,7 +842,7 @@ class EnvironmentService:
         return env_group
 
     @staticmethod
-    def get_environment_consumption_role(session, role_uri, environment_uri):
+    def get_environment_consumption_role(session, role_uri, environment_uri) -> ConsumptionRole:
         role = (
             session.query(ConsumptionRole)
             .filter(
--- a/backend/dataall/core/environment/services/environment_service.py
+++ b/backend/dataall/core/environment/services/environment_service.py
@@ -460,6 +460,32 @@ class EnvironmentService:
         return True
 
     @staticmethod
+    @has_tenant_permission(permissions.MANAGE_ENVIRONMENTS)
+    @has_resource_permission(permissions.REMOVE_ENVIRONMENT_CONSUMPTION_ROLE)
+    def update_consumption_role(session, uri, env_uri, input):
+        role_query = session.query(ConsumptionRole).filter(
+            (
+                and_(
+                    ConsumptionRole.consumptionRoleUri == uri,
+                    ConsumptionRole.environmentUri == env_uri,
+                )
+            )
+        )
+        consumption_role = role_query.first()
+        if consumption_role:
+            ResourcePolicy.update_resource_policy(
+                session=session,
+                resource_uri=uri,
+                resource_type=ConsumptionRole.__name__,
+                old_group=consumption_role.groupUri,
+                new_group=input['groupUri'],
+                new_permissions=permissions.CONSUMPTION_ROLE_ALL
+            )
+            role_query.update(input)
+            session.commit()
+        return role_query.first()
+
+    @staticmethod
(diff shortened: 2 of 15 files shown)


noah-paige:
Testing this PR now:

- [ ] Update Existing Consumption Role
- [ ] Create New Consumption Role
- [ ] Update New Consumption Role
- [ ] Delete Consumption Role

frontend/src/modules/Environments/components/EnvironmentRoleAddForm.js
```diff
@@ -12,53 +12,18 @@ import {
...
 import PropTypes from 'prop-types';
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import * as Yup from 'yup';
-import { Defaults } from 'design';
 import { SET_ERROR, useDispatch } from 'globalErrors';
-import { listEnvironmentGroups, useClient } from 'services';
+import { useClient } from 'services';
 import { addConsumptionRoleToEnvironment } from '../services';
+import { useFetchGroups } from '../../../utils/api';
```
noah-paige, line 20: Should we change away from relative import here to keep consistent with how we set import statements, mainly:
- Edit `dataall/frontend/src/utils/index.js` to export * form `api.js`
- Edit import here as `import { useFetchGroups } from 'utils';`

frontend/src/modules/Environments/components/EnvironmentTeams.js
```diff
@@ -325,6 +332,31 @@ export const EnvironmentTeams = ({ environment }) => {
...
+      enqueueSnackbar('Consumption Role was updated', {
+        anchorOrigin: {
+          horizontal: 'right',
+          vertical: 'top'
+        },
+        variant: 'success'
+      });
+      fetchRoles();
+    } else {
+      throw new Error(response.errors[0].message);
```
noah-paige, line 356: Is there a reason we throw Error here instead of calling `dispatch(SET_ERROR ...)` ?

frontend/src/modules/Environments/components/EnvironmentTeams.js
```diff
@@ -368,12 +400,49 @@ export const EnvironmentTeams = ({ environment }) => {
...
-    if (value <= roles.pages && value !== roles.page) {
-      await setFilterRoles({ ...filterRoles, page: value });
+  const handlePageChangeRoles = async (page) => {
+    page += 1; //expecting 1-indexing
+    if (page <= roles.pages && page !== roles.page) {
+      await setFilterRoles({ ...filterRoles, page: page });
     }
   };
 
+  const [rowModesModel, setRowModesModel] = React.useState({});
```
noah-paige, line 410: Can we use `useState()` that is already imported rather than `React.useState()`?

frontend/src/utils/api.js
```diff
@@ -0,0 +1,50 @@
+import { listEnvironmentGroups, useClient } from '../services';
```
noah-paige, line 1: I like that we moved this to a generic location that can then be used whenever we call `listEnvironmentGroups()`

Should this be placed in `utils/api.js` though or somewhere else, such as under `modules/Shared/...`?

frontend/src/modules/Environments/components/EnvironmentTeams.js
petrkalos: That's correct, `onProcessRowUpdateError` will call dispatch for everything that goes wrong

frontend/src/modules/Environments/components/EnvironmentRoleAddForm.js
petrkalos: You are right, I also despise relative imports, IDE auto-generated it :D.

I am no JS expert but couldn't we create a mapping (in `package.json`) and then import like `@app/utils/api.js`? That way we can completely remove the `index.js` files from the project

dlpzx: At the moment there are absolute imports across modules and relative imports inside each module. That's why you are seeing relative imports and sometimes we import from `../services` and sometimes form `services`

dlpzx: I have never worked with JS mappings, but from what I am reading it still needs to specify each mapping in the json right? So it means replacing the `index.js` files in each package by entries in the json right? Maybe we can ask @maryamkhidir and @itsmo-amzn

petrkalos: I don't think so, we just have to provide the path to the top directory and then do full imports from there.

backend/dataall/core/environment/services/environment_service.py
```diff
@@ -459,6 +459,32 @@ def remove_consumption_role(session, uri, env_uri):
...
 
+    @staticmethod
+    @has_tenant_permission(permissions.MANAGE_ENVIRONMENTS)
+    @has_resource_permission(permissions.REMOVE_ENVIRONMENT_CONSUMPTION_ROLE)
+    def update_consumption_role(session, uri, env_uri, input):
+        role_query = session.query(ConsumptionRole).filter(
+            (
+                and_(
+                    ConsumptionRole.consumptionRoleUri == uri,
+                    ConsumptionRole.environmentUri == env_uri,
```
dlpzx, line 470: The consumptionRoleUri is unique, we do not need the environmentUri filter, which means that in the mutation we can just have `consumptionUri` and `updateConsumptionUriInput` as inputs

backend/dataall/core/environment/api/resolvers.py
```diff
@@ -222,6 +222,17 @@ def remove_consumption_role(context: Context, source, environmentUri=None, consu
...
 
+def update_consumption_role(context: Context, source, environmentUri=None, consumptionRoleUri=None, input={}):
+    with context.engine.scoped_session() as session:
+        status = EnvironmentService.update_consumption_role(
+            session=session,
+            uri=consumptionRoleUri,
+            env_uri=environmentUri,
+            input=input,
+        )
+    return status
```
dlpzx, line 233: ultra minor: can we rename the variable `status` to `consumption_role`

frontend/src/utils/api.js
dlpzx: I agree that `utils/api.js` is not the best place. Recaping, the frontend structure is divided into "functional" components:
- `authentication`
- `design` - UI components, contexts, hooks
- `globalErrors`
- `modules`. Each having:
  - `components`- UI components used in views
  - `views` - UI views
  - `services` - API services used in the module only
- `reauthentication`
- `services` 
     - `graphql` API services used across modules
     - `hooks` 
- `utils`

dlpzx: Inside `modules/Shared` there are 3 modules used in other modules. I personally don't like this folder because shared code is sometimes defined at the top-level modules (e.g. `services`) and in this case inside modules.

In any case, I think the `useFetchGroups` constant fits better in `frontend/src/services/hooks/useFetchGroups.js`

frontend/src/modules/Environments/components/EnvironmentTeams.js
```diff
@@ -368,12 +400,49 @@ export const EnvironmentTeams = ({ environment }) => {
...
+
+  const handleCancelClick = (id) => () => {
+    setRowModesModel({
+      ...rowModesModel,
+      [id]: { mode: GridRowModes.View, ignoreModifications: true }
+    });
+  };
+
+  const processRowUpdate = async (newRow, oldRow) => {
+    await updateConsumptionRoleHandler(newRow);
```
dlpzx, line 440: `oldRow` is not used

frontend/src/modules/Environments/components/EnvironmentTeams.js
```diff
@@ -325,6 +332,31 @@ export const EnvironmentTeams = ({ environment }) => {
     }
   };
 
+  const updateConsumptionRoleHandler = async (newRow) => {
+    const response = await client.mutate(
+      updateConsumptionRole({
```
dlpzx, line 337: Are we validating the input parameters or ensuring that they are not empty? groupUri and consumptionRoleName cannot be empty

backend/dataall/core/environment/services/environment_service.py
petrkalos: Make sense, I will remove it from the entire stack

petrkalos: actually are we sure it's unique? [get_environment_consumption_role](https://github.com/data-dot-all/dataall/blob/823e8b11b483a0eb5d8d80f4cb665355a95eb56b/backend/dataall/core/environment/services/environment_service.py#L843) doesn't make that assumption? I think I will reuse this method as well (I should have anyway) for safety

frontend/src/modules/Environments/components/EnvironmentRoleAddForm.js
itsmo-amzn: Hi @petrkalos, we recently conducted modularization for the frontend: https://github.com/data-dot-all/dataall/issues/398ff, in which we structured our project in terms of meaningful standalone modules instead of the typical functional approach (components, hooks, context, utils, ..).

So, we defined our main modules for absolute imports here: https://github.com/data-dot-all/dataall/blob/main/frontend/jsconfig.json

And for importing components or services within the same module, we went with relative imports.

It seemed a little far-fetched to import same-module files using absolute path instead of relative imports.

If you have a strong opinion on this, you can follow the docs here: https://create-react-app.dev/docs/importing-a-component/#absolute-imports to enable absolute imports for the entire project.

backend/dataall/core/environment/services/environment_service.py
noah-paige: good call - let's re-use that method `get_environment_consumption_role`

frontend/src/modules/Environments/components/EnvironmentRoleAddForm.js
petrkalos: Certainly a new PR, created an issue #1041

noah-paige (approved):
looks good - approved!

PR #1033, opened 2024-02-05
788 alembic

SofiaSazonova:
### Feature or Bugfix
- Bugfix
- Refactoring

### Detail
1. Alembic migrations are now synchronised with the current code-state
2. `make generate-migrations` command is added. It can be applied right after the docker container with psql is started
3. Readme about migrations

### Relates
[data.all in local environment starts in a broken db state #788](https://github.com/data-dot-all/dataall/issues/788)

### Security
N/A


backend/migrations/README.md
```diff
@@ -0,0 +1,84 @@
+# Managing DB with Alembic locally
+
+When we run ```docker-compose up``` the postgres container is created with no tables or schemas.
```
dlpzx, line 3: Maybe we should change the command here to `docker-compose build` because `docker-compose up` builds and starts the containers, so after running it the tables will be created with `declarative_base`

backend/migrations/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+Upon start of GraphQL container, sqlalchemy ```declarative_base``` is used to create all tables with this function: 
+```Base.metadata.create_all(engine.engine)```. **The number of tables depends on the modules that are enabled in ```config.json``` (in the root of the project).**
+
+As the database is created from scratch, it has no current information about migration state, so, first we need to run database upgrade.
+After that alembic will be able to generate the further migrations localy.
```
dlpzx, line 9: typo: locally

backend/migrations/README.md
```diff
@@ -0,0 +1,84 @@
...
+```bash
+make upgrade-db 
+```
+or
+```bash
+export PYTHONPATH=backend
+export envname=local
+alembic -c backend/alembic.ini upgrade head
+```
+This command will apply all migrations, and syncronize the DB state with local list of migraitions.
```
dlpzx, line 21: typo: migrations (migraitions)

backend/migrations/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+When we run ```docker-compose up``` the postgres container is created with no tables or schemas.
+
+Upon start of GraphQL container, sqlalchemy ```declarative_base``` is used to create all tables with this function: 
+```Base.metadata.create_all(engine.engine)```. **The number of tables depends on the modules that are enabled in ```config.json``` (in the root of the project).**
+
+As the database is created from scratch, it has no current information about migration state, so, first we need to run database upgrade.
+After that alembic will be able to generate the further migrations localy.
+
+To upgrade database without generating migrations use
```
dlpzx, line 11: "upgrade database" might lead to confusion, maybe we can write here the sentence "This command will apply all migrations, and syncronize the DB state with local alembic historyt of migrations."

backend/migrations/README.md
dlpzx, line 1: Maybe we could add more info about the generic setup including things like:
- you need to be running a `db` locally in order to run migrations. At a minimum you need to build the docker image for the `db` docker image (right?, I am not 100% sure)
- the environment variable `envname` represents the schema name. It is used by alembic to create an schema in the database
- in the real-life RDS database, `envname` takes the value of the environment (e.g. `dev`. `test`...). 
- Locally, these are the by default created schemas: `local`, `...` (I don't know, we need to check). The dockerized application runs in the environment `dkrcompose`. In the following we recommend you to use the envname `local` to validate migrations

backend/migrations/env.py
```diff
@@ -1,6 +1,21 @@
 from __future__ import with_statement
 from alembic import context
 from logging.config import fileConfig
+import re
+
+
+# DO NOT DELETE
+# import additional models here
+# they are not used directly in env.py, but these imports are important for alembic
+
```
dlpzx, line 10: We don't know the root cause of alembic ignoring these models right?

backend/migrations/versions/a375771e4d0f__alembic_sync_state.py
```diff
@@ -0,0 +1,185 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table('item_tags')
+    op.drop_table('dataset_table_profiling_job')
+    op.drop_table('redshiftcluster')
+    op.drop_table('dataset_quality_rule')
+    op.drop_table('redshiftcluster_datasettable')
+    op.drop_table('redshiftcluster_dataset')
+    op.drop_table('tag')
+    op.drop_table('group_member')
```
dlpzx, line 28: The drops I agree, but the alterations need to be considered one by one. 
- For example, in the `dataset_bucket` table instead of dropping the foreign key in RDS I think the issue is in the definition in the code. We should add the foreign key in `dataall.modules.datasets_base.db.dataset_models.DatasetBucket`

backend/migrations/versions/a375771e4d0f__alembic_sync_state.py
```diff
@@ -0,0 +1,185 @@
...
+    op.alter_column('datapipelineenvironments', 'region',
+                    existing_type=sa.VARCHAR(),
+                    nullable=True)
+    op.alter_column('dataset_bucket', 'partition',
+                    existing_type=sa.VARCHAR(),
+                    nullable=True)
+    op.drop_constraint('dataset_bucket_datasetUri_fkey', 'dataset_bucket', type_='foreignkey')
```
dlpzx, line 35: In the `dataset_bucket` table instead of dropping the foreign key in RDS I think the issue is in the definition in the code. We should add the foreign key in `dataall.modules.datasets_base.db.dataset_models.DatasetBucket`

dlpzx (changes requested):
Some changes needed

backend/migrations/env.py
SofiaSazonova: No, I couldn't find it. I found only a guide, where it was said that from multiple files you should explicitly import not only Base, but the models as well.
But honestly I couldn't understand, why exactly this models are unseen without explicit import.

backend/migrations/README.md
```diff
@@ -0,0 +1,111 @@
...
+```bash
+docker compose up
+```
+2. Specify the location of database model descriptions. If you are at the project's root, the target folder path is backend.
+```bash
+export PYTHONPATH=backend
+```
+3. The containers initiated in the first step will default to using the schema named `local`. Alembic relies on the environmental variable `envname` to determine the schema. Set it to `local` with the following command:
+```bash
+export envname=local
```
TejasRGitHub, line 28: I am afraid that the containers will use the schema named `dkrcompose` and not `local` as per the docker-compose.yaml file

backend/migrations/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+When we run ```docker compose build``` the postgres container is created with no tables or schemas.
+
+Upon start of GraphQL container, sqlalchemy ```declarative_base``` is used to create all tables with this function: 
+```Base.metadata.create_all(engine.engine)```. **The number of tables depends on the modules that are enabled in ```config.json``` (in the root of the project).**
+
+As the database is created from scratch, it has no current information about migration state, so, first we need to run database upgrade.
+After that alembic will be able to generate the further migrations locally.
+
+This command will apply all migrations, and syncronize the DB state with local alembic historyt of migrations.
```
TejasRGitHub, line 43: Just a small typo with "historyt"

backend/migrations/README.md
dlpzx: Yes, if using the docker containers the `envname` is `dkrcompose` because it is specified in the graphql container:
image

backend/migrations/README.md
```diff
@@ -1,24 +1,53 @@
 # Managing DB with Alembic locally
+In data.all we use [Alembic](https://alembic.sqlalchemy.org/en/latest/)  -- a lightweight database migration tool for usage with the SQLAlchemy Database Toolkit for Python.
+Alembic relies on the database's current state to generate and apply migrations accurately.
+Alembic determines the changes to be made by comparing the current state of the database with the desired state specified in your SQLAlchemy models. This process, known as schema diffing, requires an existing database to identify differences.
+Alembic generates migration scripts based on the detected differences between the current database schema and the desired schema defined in your application code. This generation is dependent on the actual structure and content of the database.
 
-When we run ```docker-compose up``` the postgres container is created with no tables or schemas.
+In order to create and test migrations locally you will have to create a local database.
+
```
dlpzx, line 8: I think we are mixing 2 ways of running migrations. One is the `dkrcompose` environment that uses declarative base and the other is if we create a new schema maybe called "local". I would argue that both are useful, "local" allows users to play and break the schema as much as they want. "dkrcompose" allows them to do more complex testing (with populated data from a previous version)

backend/migrations/versions/442dd8f223a9__alembic_sync.py
```diff
@@ -1,34 +1,36 @@
...
     op.drop_table('dataset_table_profiling_job')
+    op.drop_table('group_member')
+    op.drop_table('redshiftcluster')
     op.drop_table('dataset_quality_rule')
+    op.drop_table('tag')
+    op.drop_table('redshiftcluster_datasettable')
+    op.drop_table('item_tags')
     op.drop_table('redshiftcluster_dataset')
-    op.drop_table('group_member')
     op.alter_column('datapipelineenvironments', 'region',
```
dlpzx, line 29: We should fix the original table, this field should not be updated to be nullable. Same for line 36 and 40

dlpzx (commented):
Left some final touches and comments on the actual migration script

backend/migrations/README.md
noah-paige: Agree with some of the earlier comments - I think we should to explicitly call out the different schemas created (`dkrcompose` to be used in local deploy of data.all vs. `local` for isolated instance of db schema to play with)

- Running data.all following local deploy commands will have db schema built using declarative_base (I believe from the line - `Base.metadata.create_all(engine.engine)` 

- Running `alembic -c backend/alembic.ini upgrade head` with a new envname (such as `local`) to build schema from scratch

Let's maybe separate the 2 above and highlight the use case for working with each?

Also, a call out but when I run an upgrade command `alembic -c backend/alembic.ini upgrade head` with envname `dkrcompose` to make changes to local deploy dataall db changes - I need to also edit the hostname to `localhost` because of this line in `get_engine()` - `hostname = 'db' if envname == 'dkrcompose' else 'localhost'` (and when I change hostname then other docker containers can not connect until I revert back to `db`)

...not sure if others do the same / something we could make easier as part of this PR

TejasRGitHub: Hi @noah-paige , Agree on the point that we need to explicitly mentioned which schema does what. On your last point, I have faced that issue while upgrading the dockercompose db and use the same trick as you do. It would be very helpful if we solve this in this PR

backend/migrations/versions/442dd8f223a9__alembic_sync.py
```diff
@@ -0,0 +1,181 @@
...
+    op.alter_column('datapipelineenvironments', 'region',
+                    existing_type=sa.VARCHAR(),
+                    nullable=True)
+    op.drop_constraint('dataset_bucket_datasetUri_fkey', 'dataset_bucket', type_='foreignkey')
+    op.create_foreign_key(None, 'dataset_bucket', 'dataset', ['datasetUri'], ['datasetUri'], ondelete='CASCADE')
+    op.alter_column('environment_parameters', 'paramValue',
+                    existing_type=sa.VARCHAR(),
+                    nullable=True)
+    op.create_foreign_key(None, 'environment_parameters', 'environment', ['environmentUri'], ['environmentUri'])
+    op.alter_column('sagemaker_studio_domain', 'environmentUri',
```
noah-paige, line 38: Think we can change our model to make environmentUri not nullable for `sagemaker_studio_domain`

Makefile
```diff
@@ -85,7 +85,13 @@ upgrade-db: upgrade-pip install-backend
 	export PYTHONPATH=./backend && \
 	alembic -c backend/alembic.ini upgrade head
 
-clean:
```
dlpzx, line 88: Is this intended? Even if `clean` is not used in the pipelines it is a useful command for local development

SofiaSazonova: Oh, I think I missed it during merge. There were some 'removing edits' from main.
Just a sec, I get it back

backend/migrations/versions/6c9a8afee4e4__alembic_sync.py
```diff
@@ -0,0 +1,163 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table('redshiftcluster_datasettable')
+    op.drop_table('dataset_quality_rule')
+    op.drop_table('tag')
+    op.drop_table('redshiftcluster_dataset')
+    op.drop_table('item_tags')
+    op.drop_table('dataset_table_profiling_job')
+    op.drop_table('redshiftcluster')
+    op.drop_table('group_member')
```
dlpzx, line 28: Aren't line 29 and line 30 doing the same?

SofiaSazonova: If you use alembic lib to create a foreign key in model, It will result create_foreign_key. Constraint is more general think. I would leave it this way

backend/migrations/README.md
```diff
@@ -0,0 +1,111 @@
...
+The type comparison logic is fully extensible as well; see [Comparing Types](https://alembic.sqlalchemy.org/en/latest/autogenerate.html#compare-types) for details.
+
+ - Change of server default. This will occur if you set the `EnvironmentContext.configure.compare_server_default` parameter to `True`, or to a custom callable function. This feature works well for simple cases but cannot always produce accurate results. The Postgresql backend will actually invoke the “detected” and “metadata” values against the database to determine equivalence. The feature is off by default so that it can be tested on the target schema first. Like type comparison, it can also be customized by passing a callable; see the function’s documentation for details.
+
+**Autogenerate can not detect:**
```
dlpzx, line 82: Maybe we should add a comment on the modules that need to be imported explictly, just to make sure that clients don't trust alembic blindly

SofiaSazonova: What imports do you mean?

dlpzx:
From a code perspective I think this PR is good to go. I will do a final test by deploying the PR to an AWS deployment to verify that the migrations run successfully on pre-existing AWS deployments.

noah-paige (approved):
I think this PR looks good to merge! Thanks @SofiaSazonova for dealing with a bunch of iterations on this one 😄

dlpzx (approved):

Issue #805, opened 2023-10-16
Showing multiple shared dbs for same shared dataset for a single user in worksheets view


noah-paige:
### Describe the bug

A User (UserA) belongs to Group1, Group2, and owns consumption role (CR-1) all within EnvX. When all 3 Groups request access and get a share successfully approved to 1 Dataset (DatasetZ), UserA can see 2 shared DBs show but only 1 works querying the data in Worksheets 

* I believe this is because of filtering to the shared Datasets (in searchEnvironmentDataItems) is not filtering for the group that owns the worksheet so a shared DB is showing for both Group1 and Group2 but when querying the data it is using the role mapped to the group that owns the worksheet

### How to Reproduce

* Create EnvX with Owner Group1
* Invite Group2 and create a consumption role (CR-1) in EnvX
* Have UserA belong to Group1 and Group2
* Create a share request to a Dataset’s tables to Group1, Group2, and CR-1
* As UserA, create a new Worksheet and try to query data shared to UserA in EnvX
    * Multiple DBs should appear but not all will work

### Expected behavior

Only show 1 shared DB that is able to query data for the Group that owns the Worksheet

### OS

N/A

### Python version

N/A

### AWS data.all version

v2.0

dlpzx:
This should be solved with #1063

PR #1063, opened 2024-02-14
Update Worksheet database names in UI for new simplified db names

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
After implementing #1016 the names displayed for the databases in Worksheets won't contain the unique identifier.
In addition this PR solves https://github.com/data-dot-all/dataall/issues/805 by removing duplicates also in FE.

Here is an screenshot of a local test:

Screenshot 2024-02-14 at 17 34 16

Update:
Because there will be a mix of old shares with Glue database names ending in`shared_URI` and shares with database names suffixed by `shared` only, this PR introduces a new field in the GraphQL type returned by the searchDataItems query. This field resolves the name of the shared Glue database. @noah-paige @TejasRGitHub in this [commit](https://github.com/data-dot-all/dataall/pull/1063/commits/3321109705976b66fe33fd9645c9e2bf5140b546)

At first I tried implementing a separate resolver for the Worksheets, but I think we can fix step by step and first focus on the database name and then on the group of the Worksheet vs the group chosen inside the Worksheet. In any case I left the [commit ](https://github.com/data-dot-all/dataall/pull/1063/commits/82e4ed3f83f6138f5b605b9e20faa1b5e0034e72)to have some reference.

### Relates
- #1016 
- #805 

### Security

  - Is the input sanitized?


frontend/src/modules/Worksheets/views/WorksheetView.js
```diff
@@ -181,11 +181,19 @@ const WorksheetView = () => {
           response.data.searchEnvironmentDataItems.nodes.map((d) => ({
             datasetUri: d.datasetUri,
             value: d.datasetUri,
-            label: `${d.GlueDatabaseName}_shared_${d.shareUri}`,
-            GlueDatabaseName:
-              `${d.GlueDatabaseName}_shared_${d.shareUri}`.substring(0, 254),
+            label: `${d.GlueDatabaseName}_shared`,
+            GlueDatabaseName: `${d.GlueDatabaseName}_shared`,
```
noah-paige, line 185: So this will always show `${d.GlueDatabaseName}_shared` now - even if it is an old share correct?

noah-paige:
The PR looks good just 2 things to call out:
1. For old shares using `shareUri` in suffix we will now be showing the wrong db name --> if an easy way possible should we try to fix this, maybe by doing similar to what we do with ShareView and using the resolved consumption data to get shared DB name `${share.consumptionData.sharedGlueDatabase}`?

2. I think we should drill down Team to be the same as the team that created the Worksheet and not have it as an optional parameter - this can be a separate PR

frontend/src/modules/Worksheets/views/WorksheetView.js
dlpzx: oops, completely blocked that from my mind. Fixing it

TejasRGitHub:
Hi @dlpzx , @noah-paige , 

I agree with @noah-paige on both the points. 

If possible it would be great to display the database name ( old or new ) which is actually present in the account where it is shared. 

For the second point, I agree that we should not have a drop down to select the team and only show the databases and tables which are accessible to the team which created the worksheet and also because the querying is happening with the worksheet admin-group's role and not the role of the team selected in dropdown ( please correct me if I am wrong )

frontend/src/modules/Worksheets/views/WorksheetView.js
dlpzx: Updated (see description of PR)

noah-paige (approved):
lgtm!

backend/dataall/modules/dataset_sharing/api/resolvers.py
```diff
@@ -218,6 +218,25 @@ def list_shareable_objects(
...
 
 
+def resolve_shared_database_name(
+    context: Context, source
+):
+    if not source:
+        return None
+    old_shared_db_name = (source.GlueDatabaseName + '_shared_' + source.shareUri)[:254]
+    with context.engine.scoped_session() as session:
+        share = ShareObjectService.get_share_object_in_environment(uri=source.environmentUri, shareUri=source.shareUri)
```
TejasRGitHub, line 228: Why did we get the share item from this new method instead of getting it from get_share_object() ? Did we just want to confirm if the user has the permissions to access that env ?

frontend/src/modules/Worksheets/views/WorksheetView.js
```diff
@@ -181,11 +181,18 @@ const WorksheetView = () => {
...
             value: d.datasetUri,
-            label: `${d.GlueDatabaseName}_shared_${d.shareUri}`,
-            GlueDatabaseName:
-              `${d.GlueDatabaseName}_shared_${d.shareUri}`.substring(0, 254),
+            label: d.sharedGlueDatabaseName,
+            GlueDatabaseName: d.sharedGlueDatabaseName,
             environmentUri: d.environmentUri
           }));
+        // Remove duplicates based on GlueDatabaseName
+        sharedWithDatabases = sharedWithDatabases.filter(
```
TejasRGitHub, line 189: Could we potentially remove this duplication in the query itself ? `dataall.modules.dataset_sharing.db.share_object_repositories.ShareObjectRepository.paginate_shared_datasets` in here

backend/dataall/modules/dataset_sharing/api/resolvers.py
noah-paige: believe it is exactly for that reason because of decorator `@has_resource_permission(GET_ENVIRONMENT)` but will leave it open for @dlpzx  to confirm

frontend/src/modules/Worksheets/views/WorksheetView.js
noah-paige: Not sure if we want to remove duplication of DB names on the backend because this query `searchEnvironmentDataItems` is also used at `frontend/src/modules/Environments/components/EnvironmentSharedDatasets.js` where we show each share item that is shared with the given environment and for what respective team in that environment (in this case we would want to show all shares duplicate dbs or not)

Another note (out of scope for this PR): We currently do not handle shared items of type `S3Bucket` for the table mentioned above ...

TejasRGitHub: Oh okay I get it. Thanks @noah-paige . 

I didn't quite get your what you meant by the note and which table you are mentioning

dlpzx: There are 2 reasons not to remove the duplicates, one is that the query is used in other places as Noah pointed out and the second reason was that it is difficult to predict which shares are using the old method and the new method, and that is retrieved in the resolver that is a field of the type. Once we deprecate the old type of share names I think we can redesign the remove-duplicates and the separation between plot the shared items in environment and list the shared databases in Worksheets. I tried implementing a new resolver (check the commits), but I found it a bit complex to fulfill the purpose of this PR

backend/dataall/modules/dataset_sharing/api/resolvers.py
TejasRGitHub: Thanks @noah-paige , @dlpzx for the clarification

frontend/src/modules/Worksheets/views/WorksheetView.js
TejasRGitHub: Yeah makes sense. Thanks @dlpzx

Issue #1048, opened 2024-02-09
Maximum policy size of 10240 bytes exceeded


sandeephs1:
### Describe the bug

Linking new environments is erroring -

``` Resource handler returned message: "Maximum policy size of 10240 bytes exceeded for role dataallPivotRole (Service: Iam, Status Code: 409, 
Request ID: <>)" (RequestToken: <>, HandlerErrorCode: ServiceLimitExceeded)```

### How to Reproduce

```
*P.S. Please do not attach files as it's considered a security risk. Add code snippets directly in the message body as much as possible.*
```
Try to link new environment

### Expected behavior

Successfully should be able to link new environments

### Your project

_No response_

### Screenshots

_No response_

### OS

Mac Sonoma 14.3

### Python version

3.9.13

### AWS data.all version

2.0.0

### Additional context

_No response_

sandeephs1:
Error is - 
Resource handler returned message: "Maximum policy size of 10240 bytes exceeded for role dataallPivotRole (Service: Iam, Status Code: 409, 
Request ID: <>)" (RequestToken: <>, HandlerErrorCode: ServiceLimitExceeded)

dlpzx:
Hi @sandeephs, by the name `dataallPivotRole` I am guessing you are using the manually created pivot roles right? And the issue you are raising might come from creating multiple data.all Environments in the same AWS account. CDK automatically adds permissions to the environment CloudFormation custom resources SQS queues (and their KMS keys) to the pivot role. These permissions are added as inline policies, which in case of many environments might result in a policy size limit.

At the moment we support the one environment per AWS pattern, but if you confirm that this is your issue we can work on the multiple-environments in the same account approach. That way customers could have one environment per AWS account AND region.

We can prioritize https://github.com/data-dot-all/dataall/issues/936.

sandeephs1:
Hi @dlpzx Yes we have requirement to link multiple environments with the same AWS account and a different region. also issue #936 is the request from us only. would request to prioritize the same

dlpzx:
There was another customer with a similar issue. We have decided to include this task as an stretch goal for 2.3 (target date mid March) @sandeephs1

sandeephs1:
Hello @dlpzx once the environment is deleted, Env policy associated to "dataallPivotRole" is not cleaned up. Whats the best way to find out the Envpolicies which are still associated but Environment is deleted so that it can be manually cleaned up and free up space.
[...]

Issue #936, opened 2023-12-22
Issue with enable_pivot_role_auto_create When Deploying Multiple Environments in a Single AWS Account


asifma:
### Describe the bug

We have encountered a challenge when deploying multiple environments within a single AWS account, specifically related to the enable_pivot_role_auto_create setting. Currently, when this setting is set to True, it leads to an unexpected behavior that hinders the successful deployment of multiple environments.

The issue stems from the handling of the pivot role in the backend/dataall/core/environment/cdk/environment_stack.py file. Here's a snippet for reference:

`

        pivot_role_as_part_of_environment_stack = ParameterStoreManager.get_parameter_value(
            region=os.getenv('AWS_REGION', 'eu-west-1'),
            parameter_path=f"/dataall/{os.getenv('envname', 'local')}/pivotRole/enablePivotRoleAutoCreate"
        )
        self.create_pivot_role = True if pivot_role_as_part_of_environment_stack == "True" else False
        self.engine = self.get_engine()

        self._environment = self.get_target(target_uri=target_uri)

        self.environment_groups: [EnvironmentGroup] = self.get_environment_groups(
            self.engine, environment=self._environment
        )

        self.environment_admins_group: EnvironmentGroup = self.get_environment_admins_group(
            self.engine, self._environment
        )

        # Create or import Pivot role
        if self.create_pivot_role is True:
            config = {
                'roleName': self.pivot_role_name,
                'accountId': self.dataall_central_account,
                'externalId': self.external_id,
                'resourcePrefix': self._environment.resourcePrefix,
                'environmentUri': self._environment.environmentUri
            }
            pivot_role_stack = PivotRole(self, 'PivotRoleStack', config)
            self.pivot_role = iam.Role.from_role_arn(
                self,
                f'PivotRole{self._environment.environmentUri}',
                pivot_role_stack.pivot_role.role_arn,
            )
        else:
            self.pivot_role = iam.Role.from_role_arn(
                self,
                f'PivotRole{self._environment.environmentUri}',
                f'arn:aws:iam::{self._environment.AwsAccountId}:role/{self.pivot_role_name}',
            )
`

### How to Reproduce

1. Set "enable_pivot_role_auto_create": true, in your DeploymentEnvironment in cdk.json
1. Create an environment
2. Try to create an additional environment in the same AWS Account
3. Second environment will fail to deploy, see CloudFormation in AWS Console for error

### Expected behavior

You should be able to create multiple environments in any region in the same AWS Account

[...]

PR #1064, opened 2024-02-15
Allow multiple environments in the same account with cdk-pivot role

dlpzx:
### Feature or Bugfix
IMPORTANT

- ⚠️ ⚠️ ⚠️ IMPORTANT: Users MUST upgrade their data.all Environments and Datasets if they want to enable multiple_environments_same_account pivot roles! We need to make it very clear in the release notes. Otherwise a good hint that indicates if an environment is updated is that when opening the Environment/Stack tab they get an error of the type: `An error occurred (AWSResourceNotFound) when calling GET_STACK_LOGS operation: Logs could not be found for this stack`

- ⚠️ ⚠️ ⚠️ IMPORTANT: For data.all **CREATED** datasets, this PR introduces a change in the IaC that BREAKS the KMS policy for folder and bucket sharing. The share will appear as unhealthy if we verify it and will be fixed by re-applying the share. To automate this process we recommend users to:
       - 1. Update data.all to new version
       - 2. Update all Environment and Dataset stacks 
       - 3. trigger the `share-verifier` task in ECS
       - 4. trigger the `share-reapplier` task in ECS introduced in PR #1151 

- ⚠️ After upgrading, there will be some useless LakeFormation permissions granting permissions to the deleted `dataallPivotRole-cdk` because LakeFormation does not delete the permissions of deleted IAM roles. In addition, Lake Formation keeps the old pivot role as datalake admin flagged as invalid (see pic below). The permissions do not really harm anyone and can be deleted manually. If you are a user and are struggling cleaning-up or wish more support, please open a GitHub issue referencing this PR and we will address your challenges.
image

#### Problem statement
data.all =<2.2.0 is not designed to allow the creation of multiple data.all Environments in the same AWS account. However, there are customers that store and work with data in the same AWS account but in different regions. This PR is meant to enable that use-case solving its 2 blockers:
- If using cdk-pivot roles, the second environment stack fails to deploy because there is already an existing dataallPivotRole-cdk IAM role created in the first environment stack. Explained in detail: #936 
- If using manually created pivot roles, there is a limit ~5regions that can be onboarded. The failure is caused by the inline policies that CDK adds to the pivot role for the DLQ SQS queues and X-trace. Explained in detail: #1048 

## Design
Our recommendation is to use cdk-created pivot roles because their policies can be more restrictive and manual errors of deploying CFN stacks are avoided. To solve #936 and use CDK pivot role in a multi-environment same account setup we have 2 options:
1. re-use the same IAM pivot role for all environments in the account: environment 1 deploys the pivot role nested stack and the subsequent environments detect that there is an IAM pivot role already created and use the existing one. The main advantage is its simplicity, we keep using the same IAM role and there is no need for backwards compatibility or changes in the code to assume a different role. But it has some important drawbacks:
 - it does not solve #1048
 - issues with deletion of environment 1 ---> handle environment 2 updates. Environments depend on each other :'(
 - permissions of the pivot role refer to environment datasets.... ---> how will the pivot role get updated for other environments?? Not as simple as it looks.
2. deploy one IAM pivot role for each environment. Each environment deploys a separate pivot role stack. All role assumption boto3 calls need to be updates to use a different pivot role depending on the environment. It keeps each pivot role restricted to its environment, there is no coupling between environments and no #1048 policy limit issues. As disadvantages we can point out:
 - more code changes, passing info between calls and boto3 methods
 - backwards compatibility with original `dataallPivotRole-cdk` role. 

#### Design decisions

Despite having disadvantages, this PR implements option 2 because it is more robust in the long-term. It also has advantages in terms of security as it "isolates" regions; we can take it further an define pivot role policies with the specified region. In addition, we reduce the impact of IAM size service quotas as each role is dedicated to one environment only.

To minimize the code changes and impact, I have used the `region` to differentiate between pivot roles in the same account. At first I added the `environmentUri` to the pivot role name, but it made it very difficult to use in the boto3 `aws` clients. We would need to modify all the client calls to input the `environmentUri`. That's why I re-used the `region` which was already a parameter of most clients.

This PR does not handle multiple environments using the manual pivot role. Meaning that it does not expect that each environment has a dedicated `dataallPivotRole-region`. It works as usual with one `dataallPivotRole` for all environments. We want to favor CDK-pivot roles and slowly move away from manual pivot roles.

#### Backwards compatibility

As preventive measurement, this PR introduces a `config.json` parameter to configure the usage of `cdk_pivot_role_multiple_environments_same_account`. The default value is `False`, thus the changes introduced in this PR are not mandatory and customers can migrate whenever they have the need. UPDATE FROM REVIEW: If we see that the backwards compatibility is smooth we will discard the parameter. This is a good enhancement and does not remove any functionality for customers.

## TESTING
On an existing deployment with 2 environments and datasets (imported and created) and multiple shares, update the codebase. All modules are active and enabled in the environments.

Part 1: `dataallPivotRole-cdk` continues working
- [X] Update environment with no parameter `cdk_pivot_role_multiple_environments_same_account`--> check it updates and keeps deploying the `dataallPivotRole-cdk`
- [X] Update environment with `cdk_pivot_role_multiple_environments_same_account`= `False` --> check it keeps deploying the `dataallPivotRole-cdk`
- [X] Check IAM permissions of dataallPivotRole-cdk`(still contain KMS and S3 statements relative to imported dataset) and ensure it is the LF admin and has dataset permissions in LF.
- [x] Create new dataset --> check new permissions are added to `dataallPivotRole-cdk`
- [X] CloudFormation get stack (in Environment view) --> boto3 calls use the remote_session pointing at `dataallPivotRole-cdk`

Part 2: `dataallPivotRole-cdk-region` migration infrastructure
- [X] change `cdk_pivot_role_multiple_environments_same_account`to `True` and update environment. Environment updates the pivot role stack and creates NEW `dataallPivotRole-cdk-region` - as expected, in CloudFormation we see the message `Requested update requires the creation of a new physical resource; hence creating one`
- [X] IAM permissions of `dataallPivotRole-cdk-region` --> check that imported datasets in the environment are present in the policy statements
- [X] Ensure `dataallPivotRole-cdk-region` is a LF admin - it happens right after environment update
- [X] Update created dataset stack and check that `dataallPivotRole-cdk-region`  has LF permissions to the dataset. Check that KMS key policy includes the new pivot role. Finally check that the trust policy of the dataset IAM role includes the new pivot role: image
- [X] Update imported dataset stack and check that `dataallPivotRole-cdk-region`  has LF permissions to the dataset. Bucket policy and KMS policies are not managed by data.all in the creation of dataset. Finally check that the trust policy of the dataset IAM role includes the new pivot role image

Part 3: `dataallPivotRole-cdk-region` migration data sharing
- [X] Click on verify permissions for one of the shares (imported dataset) ---> ⚠️ verification shows table as unhealthy, folder appears as healthy. When users click on re-apply share, it grants permissions to the new pivot role in the cross account database and resource link and it appears as healthy again. See detail in Issues found section.
- [X] Add new table to share in the same share --> check that share is successful
- [X] Add bucket to share --> check that share is successful - KMS policy contains new pivot role!
- [X] Revoke one table  --> check share is successful - to ensure LF permissions work
- [X] Revoke one folder  --> check share is successful - to ensure LF permissions work
- [X] Revoke one bucket  --> check share is successful - to ensure LF permissions work
- [X] click on verify permissions in the other share (created dataset) --> ⚠️  in the dataset_stack, the KMS key policy has been wiped out, so bucket and access point shares are unhealthy! image (This part is fixed, see details below) After re-applying shares: 
image
` Failed to get kms key id of alias/loc-new-uqo3m5oh: An error occurred (AccessDeniedException) when calling the DescribeKey operation: User: arn:aws:sts::796821004569:assumed-role/dataallPivotRole-cdk-eu-west-1/dataallPivotRole-cdk-eu-west-1 is not authorized to perform: kms:DescribeKey on resource: arn:aws:kms:eu-west-1:796821004569:key/d00a9763-7fd0-4845-81d6-863d20714888 because no resource-based policy allows the kms:DescribeKey action`

Part4: `dataallPivotRole-cdk-region` in boto3 calls - especially IAM and Quicksight that required some additional development.
- [X] IAM - Add consumption role and check its policies attached
- [X] CloudFormation - get stack logs in any Stack tab - check client
- [X] Athena dataset client - Preview table
- [X] Glue dataset client - Start crawler
- [X] Glue profiling client - Run profiling job --> it triggers the profiling job but there is an issue unrelated to this PR related to the profiling script int he S3 bucket of the environment: `key: profiling/code/jars/deequ-2.0.0-spark-3.1.jar.The specified key does not exist`
- [X] S3 location client - Create folder - check that s3 prefix gets created in S3
- [X] S3 dataset client - Get Dataset AWS creds - also checks that the dataset role can be assumed by the new pivot role --> for backwards compatibility dataset stack MUST be updated.
- [X] LF dataset clients +  S3 bucket policy client - Create Dataset
- [X] IAM + data sharing clients (KMS, LF, RAM, S3, GLUE) - Share new items, revoke items, create new share. --> same error on KMS key for bucket share.
- [x] Quicksight - Create Quicksight session in dashboards - it checks not only the client but the Quicksight identity region
- [ ] SageMaker Notebooks - create and open Notebook 🔴 Error: 'Could not find any VPCs matching...' it fails on creation
- [X] SageMaker ML Studio + EC2 base client - create and open ML Studio user
- [X] Athena - run query in Worksheets
- [ ] Remote session in Pipelines - Create Pipeline 2 types: CDK and non-CDK (tested below)

Additional testing in AWS + support PRs (Update April 8th) - deploy https://github.com/data-dot-all/dataall/pull/1151 as well
- [X] Deploy in AWS CICD pipeline succeeds
- [x] Revoke shared bucket without verify/reapply (after update environments and datasets)
- [x] Revoke shared folder without verify/reapply (after update environments and datasets)
- [x] Revoke shared table without verify/reapply (after update environments and datasets) - I checked the logs and before the revoke it grants the pivot role permissions to the db
- [X] Verify shared table succeeds and does not fail because of missing pivot role permissions (test Fix in PR: #1149 )
- [X] Deploy #1150 - Verify shared folder and bucket shows unhealthy for data.all created datasets, admins go to data.all and trigger ECS task share-reapplier, verify shared folder and bucket again, it should be healthy (Test #1150) 🚀 

I triggered the share-verifier task manually and only the expected broken shares (created datasets - folder/bucket sharing) appeared as unhealthy

Other less important items - open issues if needed
- [X] Test Notebooks --> all good, it must have been something related to my local environment
- [X] Test Pipelines --> all good we can get credentials
- [ ] Test profiling jobs ---> another different error `ImportError: cannot import name 'ColumnProfilerRunner' from 'awsglue.transforms' (/opt/amazon/lib/python3.6/site-packages/awsglue/transforms/__init__.py)` --> open github issue

### Open Issues, remarks and questions

- [X] ISSUE TO FIX:  Folder and bucket permissions get wiped out when the dataset stacks get updated with the new KMS statements (see the issues above). Users need to re-apply the shares because after the upgrade they will be broken. ⚠️ in order for this PR to be released we need to implement some ECS task or automation to re-apply all failed shares ---> [Issue](https://github.com/data-dot-all/dataall/issues/1150) Using PR https://github.com/data-dot-all/dataall/pull/1151 this issue can be remediated

- [X] TO DECIDE: Should we make this feature configurable? If so, what is the deprecation timeline for `dataallPivotRole-cdk`? ---> At the moment: Yes, this can be configured so that users can decide when to do the upgrade. We will use the maintenance window implementation to update users about this change.

- [X] ISSUE TO FIX: Bucket sharing or folder sharing in testing part3 will fail for imported datasets because the statement in the KMS policy `DATAALL_KMS_PIVOT_ROLE_PERMISSIONS_SID` is skipped in `dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py:535` keeping the reference to the old pivot role. One quick fix is to rename the Sid, but that will leave orphan resources referencing the old pivot role. A better fix is to ensure that the role is that sid is the one of the pivot role. --> 🟢 fixed

- [X] ISSUE TO FIX: Missing `kms:DescribeKey` permissions for pivot role from KMS key policy making the share fail. I am not sure how this issue did not surface before. I think it has to due with the by default permissions in KMS vs the ones applied by data.all.  `Failed to get kms key id of alias/loc-new-2-eo336u1r: An error occurred (AccessDeniedException) when calling the DescribeKey operation: User: arn:aws:sts::796821004569:assumed-role/dataallPivotRole-cdk-eu-west-1/dataallPivotRole-cdk-eu-west-1 is not authorized to perform: kms:DescribeKey on resource: arn:aws:kms:eu-west-1:796821004569:key/bb37c05a-0588-45f2-a9e4-4075392ee5e1 because no resource-based policy allows the kms:DescribeKey action ` --> permissions added in this [commit](https://github.com/data-dot-all/dataall/pull/1064/commits/0ba9a246035cb1ec9373a0d2cdc24f0f7050bd13)

- [X] ISSUE TO FIX: Table health verification shows unhealthy after update: `Requestor arn:aws:iam::SOURCE_ACCOUNT:role/dataallPivotRole-cdk-eu-west-1 missing LF permissions: ALL for Glue DB Target: loc_imported_c2_kms_URI
Requestor arn:aws:iam::TARGET_ACCOUNT:role/dataallPivotRole-cdk-eu-west-1 missing LF permissions: ALL for Glue DB Target: loc_imported_c2_kms_URI_shared ` The issue is solved when we click on `Re-Apply share` but it is a bit annoying that customers need to re-apply all their shares. I think we should introduce an ECS task to re-apply all shares. @anushka-singh already suggested this feature for other purposes --->based on @noah-paige's review, I think the best fix will be to ensure that the pivot role gets granted all the permissions in the verification task, that way the verification won't fail and the re-apply won't be necessary for cases in which the missing pivot role permissions are the source problem. ---> Fix in PR: #1149 

-------- 
Less important
- [X] ISSUE UNRELATED TO THIS PR: Run profiling job --> it triggers the profiling job but there is an issue unrelated to this PR related to the profiling script in the S3 bucket of the environment: `key: profiling/code/jars/deequ-2.0.0-spark-3.1.jar.The specified key does not exist` ---> confirm and create issue

- [X] ISSUE TO INVESTIGATE: the permissions of the pivot role after the shares are reapplied seem incomplete:
image  --> 🟢 It does not seem to have an impact on shares, all table sharing keeps on working as expected.

### Relates
- #936 
- #1048 

### Security

fetching data from storage outside the application (e.g. a database, an S3 bucket)? `No`
  - Is the input sanitized? `n/a`
  - What precautions are you taking before deserializing the data you consume? `n/a`
  - Is injection prevented by parametrizing queries? n/a
  - Have you ensured no `eval` or similar functions are used? `n/a`
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms? `n/a`
  - Are you logging failed auth attempts? `n/a`
- Are you using or adding any cryptographic features? `No`
  - Do you use a standard proven implementations?`n/a`
  - Are the used keys controlled by the customer? Where are they stored?`n/a`
  - Have you used the least-privilege principle? How? -----> 🖐️ We use a dedicated IAM role for each data.all Environment onboarded to data.all. This role can access data in its region only. This way the blast radius of any attack is limited to a single AWS region. This PR improves the user experience to use cdk-Pivot role, which encourages users to avoid manual pivot role that has more open permissions and is more error prone. -----> This PR also introduces a tiny change on the IAM permissions of the Lambdas and ECS tasks of the backend. They can now assume `dataallPivotRole-cdk-*` to account for the multiple region roles. This change is still restrictive enough


backend/dataall/base/aws/sts.py
```diff
@@ -364,6 +361,7 @@ def generate_console_url(credentials, session_duration=None, region='eu-west-1',
 
     @staticmethod
     def is_assumable_pivot_role(accountid):
+        # TODO: region
```
noah-paige, line 364: need to pass region here as well right?

backend/dataall/core/environment/api/resolvers.py
```diff
@@ -677,7 +678,7 @@ def get_pivot_role_name(context: Context, source, organizationUri=None):
             resource_uri=organizationUri,
             permission_name=permissions.GET_ORGANIZATION,
         )
-        pivot_role_name = SessionHelper.get_delegation_role_name()
+        pivot_role_name = SessionHelper.get_delegation_role_name(region='<REGION>')
```
noah-paige, line 681: when a user copies pivotrole name it will now show as `dataallPivotRole-<REGION>` on link env UI yes?

backend/dataall/modules/datasets/aws/glue_dataset_client.py
```diff
@@ -9,8 +9,8 @@
 
 class DatasetCrawler:
     def __init__(self, dataset: Dataset):
-        session = SessionHelper.remote_session(accountid=dataset.AwsAccountId)
         region = dataset.region if dataset.region else 'eu-west-1'
```
noah-paige, line 12: is dataset every nullable? do we need this line (L12)

noah-paige:
Went through the PR and left some minor comments. Overall the PR makes sense to me, my comments for the Issues to Decide below:


- (1) Health verification shows unhealthy after update... 
Confirming my understanding first:
  - This would not just be for tables but also for folders and buckets right?
    - For Tables - it shows unhealthy but the shares still work (missing PivotRole permissions that can be fixed on next reapply)
    - For Folder/Buckets - these share items fail once dataset_stack updated and KMS wiped, right? (fixed on reapply as. well)

Question:
    - Have we tested switch to new pivotRole naming and then immediately revoke table share item (I think should work but worth testing?)
 
Ideas:
 - I do agree I think we should have a reapply ECS task that could be run once for all shares in unhealthy status rather than handling each one individually. I think we need to better understand how this would be used, as in should this be ad-hoc triggered, scheduled,  paired with share verify task (maybe initially keep it as its own ECS task triggered on-demand?)

- Secondly, should we even be checking PivotRole permissions for `verify_shares` table items? PivotRole missing LF permissions should not interfere with the table item share and we tend to grant PivotRole LF permissions (if missing) before adding, removing, or re-applying table items?

---
- (2) Orphaned LF Permissions to Old Pivot Role
  - I am a bit surprised that we do not overwrite the old pivotRole name as admin based off of the logic in `dataall/modules/datasets/cdk/assets/lakeformationdefaultsettings/index.py` (this CR depends on pivot_role meaning the new pivot role must exist and the old should not and thus the old pivot role name should not be a `DataLakeAdmins` passed to `put_data_lake_settings()` 
    - Maybe we can fix logic there with a 1-liner to remove invalid role as Admin 
    - I think that would suffice for this PR

noah-paige:
For notebook stack creation failing with `Error: 'Could not find any VPCs matching...'` - would this not be due to missing permissions or similar on the cdk execution role if anything (and not the pivot role)?

backend/dataall/base/aws/sts.py
dlpzx: My bad, I needed a final review of my TODOs. Fixed! Yes, `lf_share_manager` is the only place where it is called (it is a function used for central catalog feature)

backend/dataall/core/environment/api/resolvers.py
dlpzx: There are multiple scenarios for pivot roles:
- manual pivot role -->`get_delegation_role_name` returns `dataallPivotRole`
- CDK SINGLE pivot role, `core.features.cdk_pivot_role_multiple_environments_same_account` is set to False -->`get_delegation_role_name` returns `dataallPivotRole-cdk` 
- CDK MULTIPLE pivot roles, `core.features.cdk_pivot_role_multiple_environments_same_account` is set to True -->`get_delegation_role_name` returns `dataallPivotRole-cdk-<REGION>` 

The resolver `get_pivot_role_name` is called in `getPivotRoleName` query that is only used in the EnvironmentCreateForm. For CDK created roles, the users do not even see any name in the UI, but I left it coded in case we want to plot anything in the UI down the line. For the manual pivot role they will be able to see and copy `dataallPivotRole`

backend/dataall/modules/datasets/aws/glue_dataset_client.py
dlpzx: Maybe at some point it was, but region has been present in datasets since we open-sourced, so I am going to remove the if conditions. Good catch!

backend/dataall/base/aws/sts.py
```diff
@@ -186,23 +195,11 @@ def get_cdk_exec_role_arn(cls, accountid, region):
         return 'arn:aws:iam::{}:role/cdk-hnb659fds-cfn-exec-role-{}-{}'.format(accountid, accountid, region)
```
dlpzx, line 196: Removed unused method `get_delegation_role_id`

backend/dataall/base/aws/sts.py
dlpzx, line 1: There are some unused methods in SessionHelper --> to be cleaned-up in follow-up PR

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -392,17 +406,18 @@ def grant_dataset_bucket_key_policy(self):
```
dlpzx, line 406: I changed the way we were checking the `DATAALL_KMS_PIVOT_ROLE_PERMISSIONS_SID` because it was not updating the statement if it already existed

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -602,6 +626,7 @@ def generate_enable_pivot_role_permissions_policy_statement(pivot_role_name, sou
                 'kms:ReEncrypt*',
                 'kms:TagResource',
                 'kms:UntagResource',
+                'kms:DescribeKey',
```
dlpzx, line 629: The pivot role IAM policies include:
- "kms:List*", to ALL resources
- "kms:DescribeKey", to ALL resources

And then for data.all-created keys + imported keys:
```
            "Action": [
                "kms:Decrypt",
                "kms:Encrypt",
                "kms:GenerateDataKey*",
                "kms:GetKeyPolicy",
                "kms:PutKeyPolicy",
                "kms:ReEncrypt*",
                "kms:TagResource",
                "kms:UntagResource"
```

I think we should include the same permissions in the KMS resource policy: in s3_bucket_share_manager, in s3_access_points_manager and in the dataset_stack!

noah-paige: even better we define this once as a constant somewhere in dataset module and reference the permissions block in all 3 locations

dlpzx: I will have a look at it, but it can only be placed in `datasets_base` because otherwise it creates a circular dependency. Can I make a comment about this and solve it in the design of #1123 ?

sandeephs1:
Hello @dlpzx Thanks for prioritising and the details.

Attached image is my understanding about the feature

image

About `open issues, remarks and questions`

TO DECIDE: Should we make this feature configurable? If so, what is the deprecation timeline for dataallPivotRole-cdk?
>> Yes should be made configurable. 
>> As soon as the feature is ready we are planning to adapt it. 
>> By end of Q3 we should have upgraded all our work environments so as long is it is available till Q3 end we are Good

ISSUE TO FIX: Folder and bucket permissions get wiped out when the dataset stacks get updated with the new KMS statements (see the issues above). Users need to re-apply the shares because after the upgrade they will be broken. ⚠️ in order for this PR to be released we need to implement some ECS task or automation to re-apply all failed shares ---> Issue

>> By Q2 end we have plans to rollout for stakeholder use so if we get it before Q2 End should be Good

ISSUE TO DECIDE: orphan permissions pointing at the old cdk role in LF: LakeFormation does not delete the permissions of deleted IAM roles, so the old pivot role keeps having LF permissions and is still the datalake admin after it has been deleted. It appears as invalid. Should we do anything about it? Users can just clean-up the permissions and remove the deleted role from their accounts, but it is true that it is a bit annoying. We can at least offer some sort of script or at least instructions

>> We haven't faced this issue, It is Low priority for us. Instructions should suffice

dlpzx:
Hi @sandeephs1 thanks for the diagram :) Yes, that is exactly the expected behavior. We will consider your feedback, with the feature as is now I think your requirements for Q2-Q3 are satisfied

deploy/stacks/container.py
```diff
@@ -445,7 +445,7 @@ def create_task_role(self, envname, resource_prefix, pivot_role_name):
                         'sts:AssumeRole',
                     ],
                     resources=[
-                        f'arn:aws:iam::*:role/{pivot_role_name}',
+                        f'arn:aws:iam::*:role/{pivot_role_name}-*',
```
noah-paige, line 448: what if pivot_role_name is `dataallPivotRole-cdk` with no regional suffix (feature flag is disabled in `config.json`)? Should this instead be `f'arn:aws:iam::*:role/{pivot_role_name}*'` (with no `-`)

deploy/stacks/lambda_api.py
```diff
@@ -316,7 +316,7 @@ def create_function_role(self, envname, resource_prefix, fn_name, pivot_role_nam
                         'sts:AssumeRole',
                     ],
                     resources=[
-                        f'arn:aws:iam::*:role/{pivot_role_name}',
+                        f'arn:aws:iam::*:role/{pivot_role_name}-*',
```
noah-paige, line 319: same as AssumeRle on `container.py` comment

noah-paige:
Tested deploying this PR and ensuring:

- [x] CICD Pipeline Successful
- [x] Env Updates Successful (on stack updater task run)
- [x] Dataset Updates Successful (on stack updater task run)
- [x] Table Shares still work
- [x] Folder Share works after verify + reapply 
- [x] New Env in same account, dif. region works


Left one minor comment on IAM policy permissions - once addressed I think good to approve!

deploy/stacks/lambda_api.py
dlpzx: Fixed in https://github.com/data-dot-all/dataall/pull/1064/commits/66e65b0b54db09614cd23177e82f93249983db55 Great catch

deploy/stacks/container.py
dlpzx: Fixed in https://github.com/data-dot-all/dataall/pull/1064/commits/66e65b0b54db09614cd23177e82f93249983db55 Great catch

Issue #922, opened 2023-12-15
Consumer IAM role updates can get wiped out by CloudFormation


zsaltys:
When we request shared access using an IAM role, data.all updates the role with necessary permissions. While this is a useful feature, it poses a problem. Consumer roles are usually managed by tools like CloudFormation and are not created or controlled by data.all. If a consumer role is managed through CloudFormation, any modifications to its policies will overwrite changes made to it by data.all.

This can result in the following scenario:

1) a request is made for access using an IAM consumer role
2) access is approved
3) access is verified as working and a production application is deployed with the new role
4) production engineering updates the IAM role using cloudformation being unaware that data.all has made modifications to it's policy
5) cloudformation wipes out any changes made by data.all
6) production application looses access causing an outage

With this enhancement we want to figure out how to prevent such issues. In our specific enterprise we do not want data.all to update IAM roles for us because we cannot not allow our devops to continue managing them via CloudFormation. One way to do that would be to have a feature flag that disables data.all from modifying consumer IAM roles itself.

However this would not solve the whole problem because then users would be left confused if they were granted access and it simply doesn't work because the IAM role does not have access and they would be left to figure out what modifications they need to make. 

The simplest thing we could do is provide clear UI instructions how the IAM role needs to be modified to grant it access. We could do this by providing JSON policy statements and a CloudFormation snippet that can be copy pasted. 

Another idea would be to create a policy for the IAM consumer role when it gets registered on the environment. Then all shares will always update this policy by triggering a change on the environment stack. We would then provide clear instructions during share registration that this policy will need to be attached to the role by the user itself using CloudFormation etc or if they wants they can tick a checkbox and let data.all do it for them (in case they don't manage roles with CF). We could even go further and show a validation check on each consumer role if it has the policy attached and if it doesn't then do not allow to use that consumer role in share requests.

This last idea would guarantee that a consumer role cannot be used if it's not properly configured and it would allow data.all to manage the permissions the IAM role has without the danger of CF wiping out changes data.all has made.

zsaltys:
@noah-paige @dlpzx Please review. I would like to prioritize this highly for 2.3 as priority nr 3 after unique shares fix (prio 2) and ability to fix permissions on datasets (prio 1).

dlpzx:
Hi @zsaltys thanks for opening the issue! We have thought about this issue offline and here is our summary and proposal

In short, the goal is:

* To introduce a configurable feature that allows external management of IAM roles and their data policies
[...]

PR #1062, opened 2024-02-14
827 share verify and reapply fix

noah-paige:
### Feature or Bugfix

- Feature

### Detail
- Add Ability to Verify Health Status of Share Requests
- Add Ability to Re-apply share request processing if in an unhealthy state

### Relates
- #827 

### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)?
  - Is the input sanitized?
  - What precautions are you taking before deserializing the data you consume?
  - Is injection prevented by parametrizing queries?
  - Have you ensured no `eval` or similar functions are used?
- Does this PR introduce any functionality or component that requires authorization?
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms?
  - Are you logging failed auth attempts?
- Are you using or adding any cryptographic features?
  - Do you use a standard proven implementations?
  - Are the used keys controlled by the customer? Where are they stored?
- Are you introducing any new policies/roles/users?
  - Have you used the least-privilege principle? How?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1068, opened 2024-02-21
922 -  Replace IAM inline policies by configurable Managed Policies for folder and bucket sharing 

SofiaSazonova:
### Feature or Bugfix
- Feature

### Detail
- For each consumption role and group role, invited to the env 1 managed policy is created
- User can choose, if the consumption role is data.all managed or not
- If data.all managed, the created policy is automatically attached to role
- without this policy attached user can not create a share for this consumption role
- policy attachment and managed options are displayed in Env->Teams
- bucket and accesspoint shares are managed through this managed policies

### Relates
- [922](https://github.com/data-dot-all/dataall/issues/922)

### SecurityN/A

How to test:
See backwards compatibility and local testing comments below.
 
@dlpzx Update:
- ⚠️ When upgrading customers need to update their environment stacks to update the pivot role if created through CDK. If the pivot role is manually created, they need to update its permissions manually with the latest version
- ⚠️ For backwards compatibility: for all existing roles both created and imported by data.all. The share-policy will be created in any of the next 3 scenarios:
a) if a new share request is created
b) if new items are added to an existing share request 
c) if items are revoked from an existing share request 

More details in the comments


SofiaSazonova:
ToDoS:
Frontend:
1. 'Copy' icon in the Consumption role tables now inherits the palette (light|dark theme). In dark theme it appears orange, but it should be white as in the table above
2. Copy no only the Policy name, but code-snippets for IaC
3. Tooltips or/and explanations, what "customer/data.all managed" means

BackEnd:
1. More elegant way to store 'empty' permissions. Right now it's implemented with 
` fake_statement = "arn:aws:s3:::initial-fake-empty-bucket" `

backend/migrations/versions/af0437dab922__add_dataallManaged_flag.py
```diff
@@ -0,0 +1,24 @@
+"""_add_dataallManaged_flag
+
+Revision ID: af0437dab922
+Revises: f6cd4ba7dd8d
+Create Date: 2024-02-15 10:42:06.833990
+
```
dlpzx, line 6: We need to make sure that this migration script runs after the one in PR for alembic and keep a consistent alembic migration history

backend/migrations/versions/af0437dab922__add_dataallManaged_flag.py
```diff
@@ -0,0 +1,24 @@
...
+# revision identifiers, used by Alembic.
+revision = 'af0437dab922'
+down_revision = 'f6cd4ba7dd8d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('consumptionrole', sa.Column('dataallManaged', sa.Boolean(), nullable=False))
```
dlpzx, line 20: How are we going to handle pre-existing consumption roles? Maybe we need to backfill the values of this column or set a default = `True` for existing roles. I think with the condition `nullable=False` we are going to run into RDS consistency problems otherwise

backend/dataall/core/environment/api/types.py
```diff
@@ -177,5 +189,15 @@
         gql.Field(name='hasNext', type=gql.Boolean),
         gql.Field(name='hasPrevious', type=gql.Boolean),
         gql.Field(name='nodes', type=gql.ArrayType(ConsumptionRole)),
+        gql.Field(
```
dlpzx, line 192: I think we do not need these fields, they are already part of `ConsumptionRole` which is the type for each of the `nodes`

backend/dataall/core/environment/api/resolvers.py
```diff
@@ -360,6 +361,17 @@ def get_parent_organization(context: Context, source, **kwargs):
     return org
```
dlpzx, line 363: One note out of the scope of this PR: `core` is not following the standard we have set for the `modules`, there is no a clear separation between the `api`, `services` and `db` layer. Balint refactored `organizations` and I refactored `vpc` we still need to refactor `environments`, `groups`, `permissions`, `stacks`

backend/dataall/core/environment/db/environment_models.py
```diff
@@ -84,6 +84,14 @@ class ConsumptionRole(Base):
     groupUri = Column(String, nullable=False)
     IAMRoleName = Column(String, nullable=False)
     IAMRoleArn = Column(String, nullable=False)
+    dataallManaged = Column(Boolean, nullable=False, default=False)
```
dlpzx, line 87: Why False as default?

backend/dataall/modules/dataset_sharing/api/input_types.py
```diff
@@ -10,10 +9,10 @@
         gql.Argument(name='principalId', type=gql.NonNullableType(gql.String)),
         gql.Argument(name='principalType', type=gql.NonNullableType(gql.String)),
         gql.Argument(name='requestPurpose', type=gql.String),
+        gql.Argument(name='attachMissingPolicies', type=gql.Boolean)
```
dlpzx, line 12: Minor: this field should be NonNullable right?

backend/dataall/core/environment/services/environment_service.py
```diff
@@ -447,8 +465,62 @@ def add_consumption_role(session, uri, data=None) -> (Environment, EnvironmentGr
             permissions=permissions.CONSUMPTION_ROLE_ALL,
             resource_type=ConsumptionRole.__name__,
         )
+
+        EnvironmentService._generate_managed_policies_for_consumption_role(environment, consumption_role)
+
         return consumption_role
 
+    @staticmethod
+    def _generate_managed_policies_for_consumption_role(environment, consumption_role):
```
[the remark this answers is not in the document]
noah-paige: +1 to the above - I think it is important we make this feature more generic rather than requiring ConsumptionRole() to be created for each EnvGroup Role now. Also in the future if we extend this feature of `dataallManaged` to other IAM entities I think it would get very confusing

backend/dataall/core/environment/db/environment_models.py
```diff
@@ -84,6 +84,14 @@ class ConsumptionRole(Base):
...
     created = Column(DateTime, default=datetime.datetime.now)
     updated = Column(DateTime, onupdate=datetime.datetime.now)
     deleted = Column(DateTime)
+
+    def get_managed_share_policy_name(self):
+        return ConsumptionRole.generate_policy_name(self.environmentUri, self.IAMRoleName)
+
+    @staticmethod
+    def generate_policy_name(environment_uri, role_name):
```
noah-paige, line 96: Maybe we should prefix with `dataall-env-...` or with `{resource_prefix}-env-...` similar to how we do for other resources created in an environment?

backend/dataall/modules/dataset_sharing/cdk/pivot_role_data_sharing_policy.py
```diff
@@ -16,7 +16,14 @@ def get_statements(self):
                 effect=iam.Effect.ALLOW,
                 actions=[
                     'iam:PutRolePolicy',
-                    'iam:DeleteRolePolicy'
+                    'iam:DeleteRolePolicy',
```
noah-paige, line 19: Based on the way we name the managed policies could we drill down the resources to something like `arn:aws:iam::aws:policy/{resource_prefix}-env-{environment_uri}-share-*` ?

If so we then may not have to add any additional excpetions to `.checkov.baseline` too

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -146,97 +146,71 @@ def grant_target_role_access_policy(self):
...
+            self.source_environment,
+            self.target_environment,
+            self.source_env_group,
+            self.env_group
+        )
+        share_manager.add_missing_resources_to_policy_statement(
+            self.bucket_name,
+            s3_target_resources,
+            policy_document["Statement"][0],
+            IAM_ACCESS_POINT_ROLE_POLICY
```
noah-paige, line 180: Passing `IAM_ACCESS_POINT_ROLE_POLICY` here is only for logging statements but is to be `iam_role_policy_name` which is no longer the case with this PR - I think we should update what we pass to `add_missing_resources_to_policy_statement(...)` and may no longer need `IAM_ACCESS_POINT_ROLE_POLICY`

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -146,97 +146,71 @@ def grant_target_role_access_policy(self):
...
+        if kms_key_id:
+            kms_target_resources = [
+                f"arn:aws:kms:{self.dataset_region}:{self.dataset_account_id}:key/{kms_key_id}"
+            ]
+            if len(policy_document["Statement"]) > 1:
+                share_manager.add_missing_resources_to_policy_statement(
+                    kms_key_id,
+                    kms_target_resources,
+                    policy_document["Statement"][1],
+                    IAM_ACCESS_POINT_ROLE_POLICY
```
noah-paige, line 197: Same as above

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -431,54 +408,65 @@ def delete_target_role_access_policy(
         logger.info(
             'Deleting target role IAM policy...'
         )
-        access_point_name = S3AccessPointShareManager.build_access_point_name(share)
-        existing_policy = IAM.get_role_policy(
+
+        accesspoint_policy_name = ConsumptionRole.generate_policy_name(target_environment.environmentUri,
```
noah-paige, line 412: can we name this the same as above to avoid confusion `share_resource_policy_name`

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -72,95 +73,71 @@ def grant_s3_iam_access(self):
         logger.info(
             f'Grant target role {self.target_requester_IAMRoleName} access policy'
         )
-        existing_policy = IAM.get_role_policy(
+        bucket_policy_name = ConsumptionRole.generate_policy_name(self.target_environment.environmentUri,
```
noah-paige, line 76: can we name `bucket_iam_policy_name` or something that indicates this is the iam policy not the s3 bucket policy

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -72,95 +73,71 @@ def grant_s3_iam_access(self):
...
-                resource_type=self.bucket_name,
-                target_resources=s3_target_resources,
-                existing_policy_statement=existing_policy["Statement"][0],
-                iam_role_policy_name=IAM_S3BUCKET_ROLE_POLICY
-            )
+        share_manager.add_missing_resources_to_policy_statement(
+            resource_type=self.bucket_name,
+            target_resources=s3_target_resources,
+            existing_policy_statement=policy_document["Statement"][0],
+            iam_role_policy_name=IAM_S3BUCKET_ROLE_POLICY
```
noah-paige, line 107: Similar to above comment with access points and `IAM_ACCESS_POINT_ROLE_POLICY`

backend/dataall/modules/dataset_sharing/services/share_object_service.py
```diff
@@ -65,13 +100,26 @@ def create_share_object(
                     environment.environmentUri
                 )
                 principal_iam_role_name = consumption_role.IAMRoleName
+                ShareObjectService.check_if_target_role_has_policies_attached(
+                    consumption_role.IAMRoleName,
```
noah-paige, line 104: minor: pass `principal_iam_role_name` here instead of `consumption_role.IAMRoleName`

deploy/pivot_role/pivotRole.yaml
```diff
@@ -443,6 +443,13 @@ Resources:
             Action:
               - 'iam:PutRolePolicy'
               - 'iam:DeleteRolePolicy'
+              - 'iam:CreatePolicy'
```
noah-paige, line 446: same comment as with cdk pivot role to restrict resource here

frontend/src/modules/Environments/components/EnvironmentRoleAddForm.js
```diff
@@ -91,7 +94,8 @@ export const EnvironmentRoleAddForm = (props) => {
         <Box sx={{ p: 3 }}>
           <Formik
             initialValues={{
-              groupUri: ''
+              groupUri: '',
+              dataallManaged: false
```
noah-paige, line 98: I would prefer `dataallManaged` to be default as `True` for both new roles via frontend form here and as default in backend (mentioned on an earlier comment by Adriana)

Open for discussion on this - but would think it would help in terms of backwards compatibility for pre-existing roles as well

frontend/src/modules/Environments/components/EnvironmentTeams.js
```diff
@@ -647,6 +647,58 @@ export const EnvironmentTeams = ({ environment }) => {
...
+                                variant: 'success'
+                              }
+                            );
+                          }}
+                        >
+                          <CopyAllOutlined />
+                        </LoadingButton>
+                      </Box>
+                    )
+                  },
```
noah-paige, line 701: Have yet to test these changes but this looks super!

noah-paige:
Left some comments throughout but will leave high level comments here:

1. I think the biggest thing we need to address to backwards compatibility. If I have pre-existing EnvGroups and Consumption Roles - I need a way to one-time transfer the policy document from `targetDatasetAccessControlPolicy` and `dataall-targetDatasetS3Bucket-AccessControlPolicy` (if they exist only) to the new managed policy role (and also create this new managed policy for pre-existing roles)
  - Maybe we can introduce this logic during processing of shares to do a check? 


2. We also are now generating 1 managed policy for both bucket share and folder (i.e. access point) share - I do not believe this should cause any issues but important to note that these sharing mechanisms would now both rely on managing the same underlying IAM policy

SofiaSazonova:
Backward compatibility is a good call. In the middle of the work I forgot about it (((

I previously introduced an idea of "banner" in the home page, where the user if notified about the changes. On this banner we can place 'Detect problems' and "Fix problems" buttons, that will go through all the Consumer and Group roles and create the required policies with the required resources. 

This banner can be off in a month e.g. or smth like that. Or we can store 'dont ever remind me again' param. 
@noah-paige  @dlpzx  What do you think?

SofiaSazonova:
Ah. No.
Everything is much-much easier.
The resources already shared will be healthy as they use inline policies.
We just need to ensure, that for existing CR, if there are no policies yet, they will be created.

SofiaSazonova:
And, also, when access is revoked, we also will do the right thing)

backend/migrations/versions/af0437dab922__add_dataallManaged_flag.py
SofiaSazonova: I merged migrations

backend/dataall/core/environment/db/environment_models.py
SofiaSazonova: changed

backend/dataall/core/environment/services/environment_service.py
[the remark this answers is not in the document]
SofiaSazonova: I put it all in separate environment-related service

backend/dataall/modules/dataset_sharing/api/input_types.py
SofiaSazonova: changed

backend/dataall/core/environment/db/environment_models.py
SofiaSazonova: it's already so

backend/dataall/modules/dataset_sharing/cdk/pivot_role_data_sharing_policy.py
SofiaSazonova: AttachPolicy permissions should be set for the consumption role. Name of the consumption role can be random

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
SofiaSazonova: changed

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
SofiaSazonova: changed

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
SofiaSazonova: changed

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
SofiaSazonova: changed

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
SofiaSazonova: changed

backend/dataall/modules/dataset_sharing/services/share_object_service.py
SofiaSazonova: changed

frontend/src/modules/Environments/components/EnvironmentRoleAddForm.js
SofiaSazonova: done

backend/dataall/modules/dataset_sharing/cdk/pivot_role_data_sharing_policy.py
dlpzx: What we can do is restrict the Create/DeletePolicy, Create/DeletePolicyVersion

dlpzx (commented):
> Ah. No. Everything is much-much easier. The resources already shared will be healthy as they use inline policies. We just need to ensure, that for existing CR, if there are no policies yet, they will be created.

Unfortunately I don't think it is that straight forward. For pre-existing shares we will have a mix of inline policies and managed policies, so the pre-existing shares the inline policies will continue working after the upgrade. But we will face issues

SofiaSazonova:
> > Ah. No. Everything is much-much easier. The resources already shared will be healthy as they use inline policies. We just need to ensure, that for existing CR, if there are no policies yet, they will be created.
> 
> Unfortunately I don't think it is that straight forward. For pre-existing consumption roles with share requests, we had inline policies, that will continue to work after the upgrade if nothing happens on the role but we could face issues:
> 
> * in case the permissions get wiped out by CloudFormation: we need to check how the [PR for share verify](https://github.com/data-dot-all/dataall/pull/1062/files) would handle it. @noah-paige does it check the statements? or does it check that the policy is attached to the role? In the second case it will appear as "successful share" but it won't be as the statements needed are not part of the IAM policy
> * in case we try to revoke: with the new code we will try to update the new IAM policy, not the previous IAM inline statements.

I took care of it. When I attach managed policies, I copy resources from inline policy and delete inline policy. 

Just now I realised, a shouldn’t copy only resources, but kms policy as well, if it exists. I will add it also,

backend/dataall/core/environment/services/env_share_policy_service.py
dlpzx, line 1: Right now we have 2 services, the environment service with the teams and consumption role handling and then this SharePolicy service. I agree that EnvironmentService is getting messy and that we could break it down into smaller services (consumptionRoleService, TeamsService...) but right now it looks a bit patchy that only this SharePolicy is on its own, which I think is a service that would make more sense inside a ConsumptionRole service.

dlpzx: I also see a bit of coupling between core/Environment and modules/data_sharing. In my opinion the code to manage shares of the consumption roles should be all in data_sharing if possible. From core we should think of abstract ways of using the modules logic. But I need to get a closer look at the code

backend/dataall/core/environment/db/environment_models.py
dlpzx: It is currently defined as `return f'dataall-env-{environment_uri}-share-{iam_role_name}'` but maybe it should be using the resource_prefix instead of hardcoding dataall

dlpzx:
TODO:
- [ ] Scope down permissions for create/delete IAM policies
- [ ] Review backwards compatibility changes
- [ ] refactor SharePolicyService and decouple from core
- [ ] Rename policies to use resource_prefix
- [ ] Call create_and_attach on data share processing not on submit

dlpzx:
Backwards compatibility cases:
1. Create new ShareObject for an existing consumption role - > check in CreateShareObject API that creates the policy. It does not attach the policy. It throws error if it is customer managed so that customers attach the policy.
2. Add more items to share request with existing consumption role:
   - if role is dataallManaged - > then we could directly create the policy and attach it in `grant_target_role_access_policy`
   - if role is customer managed - > we can follow the same approach and create and attach in `grant_target_role_access_policy`, another option is to throw an exception as in `CreateShareObject` but I think it is over-complicated and the pain point of shares being broken will be fixed by #1062  @noah-paige what do you think?
3. Revoke items from share request with existing consumption role

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -0,0 +1,106 @@
...
+        for policy in policies:
+            policy_name = policy.generate_policy_name(self)
+            all_policies.append(IAM.is_policy_attached(self.account, policy_name, self.role_name))
+        return all_policies
+
+    def generate_policy_name(self) -> str:
+        """
+        Returns string and needs to be implemented in the ManagedPolicies inherited classes
+        """
+        return NotImplementedError
```
petrkalos, line 100:
```suggestion
        raise NotImplementedError
```

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def generate_empty_policy(self) -> dict:
+        """
+        Returns dict and needs to be implemented in the ManagedPolicies inherited classes
+        """
+        return NotImplementedError
```
petrkalos, line 106:
```suggestion
        raise NotImplementedError
```

backend/dataall/core/environment/services/managed_iam_policies.py
petrkalos, line 1: Consider using [abc](https://docs.python.org/3/library/abc.html) to create an Abstract Class and enforce the implementation of the methods.

dlpzx:
Testing locally:

## Backwards compatibility





## Newly added consumption roles/teams


1. Environment Group Policy
- [ ] Invite group to environment -> IAM policy for shares is created with fake resources. Policy is NOT attached to role.
- [ ] Create Share request ->  it creates successfully. Policy is attached to role.
- [ ] Add folders and bucket, process Share request -> IAM policy is updated with dataset, fake resources are removed from policy. Policy is still attached to role.
- [ ] Revoke folder of Share request -> IAM policy is updated and access point resource is removed, S3 bucket resource stays. Policy is still attached to role.
- [ ] Revoke bucket of Share request -> IAM policy is updated and s3 resources are removed. Policy is still valid. Policy is still attached to role.
- [ ] Remove group from environment -> IAM policy for shares is deleted
 
 2. Data.all managed consumption roles
- [ ] Add role to environment with "Let data.all attach policies" = `True`  -> IAM policy for shares is created with fake resources. In Teams tab, that role is displayed as Data.all managed with attached policies
- [ ] Create Share request ->  it creates successfully. Policy is attached to role.
- [ ] Add folders and bucket, process Share request -> IAM policy is updated with dataset, fake resources are removed from policy. Policy is still attached to role.
- [ ] Revoke folder of Share request -> IAM policy is updated and access point resource is removed, S3 bucket resource stays. Policy is still attached to role.
- [ ] Revoke bucket of Share request -> IAM policy is updated and s3 resources are removed. Policy is still valid. Policy is still attached to role.
- [ ] Remove consumption tole from environment -> IAM policy for shares is deleted

 3. Customer managed consumption roles (no need to process share again as it is the same as test2)
- [ ] Add role to environment with "Let data.all attach policies" = `False`  -> IAM policy for shares is created with fake resources. In Teams tab, that role is displayed as Customer managed with NO attached policies
- [ ] Create Share request with flag  "Let data.all attach policies" = `False`->  Throws error
- [ ] Create Share request with flag  "Let data.all attach policies" = `False` + directly in IAM manually attach the policy for sharing to the role -> Before creating the share, Check that in the Teams tab the policy appears as attached. Share request creates successfully.
- [ ] Delete the previous share and remove the policy that you attached manually, Create Share request with flag  "Let data.all attach policies" = `True`->  it creates successfully and the policy is attached to the role

dlpzx:
I just realized that for pre-existing consumption roles, data.all will always assume they are data.all managed. I want to verify that is what we want @noah-paige @TejasRGitHub @anushka-singh @zsaltys is that alright?

backend/dataall/core/environment/services/managed_iam_policies.py
dlpzx: Done in this [commit](https://github.com/data-dot-all/dataall/pull/1068/commits/92b98d3180c6a619731d87e1dc1cbdd6d4385627)

backend/dataall/core/environment/services/managed_iam_policies.py
dlpzx: Can you have a look at this [commit](https://github.com/data-dot-all/dataall/pull/1068/commits/92b98d3180c6a619731d87e1dc1cbdd6d4385627). I think the 2 classes approach is the safest, but we end up with a bunch of initializations, wdyt?

anushka-singh:
> I just realized that for pre-existing consumption roles, data.all will always assume they are data.all managed. I want to verify that is what we want @noah-paige @TejasRGitHub @anushka-singh @zsaltys is that alright?

Is there anything admins will have to do as an additional manual step to manage the roles properly?

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -19,17 +22,54 @@ def __init__(
...
         Manager that registers and calls all policies created by data.all modules and that
         need to be created for consumption roles and team roles
         """
-
-        policies = ManagedPolicy.__subclasses__()
-        logger.info(f'Found {len(policies)} subclasses of ManagedPolicy')
-        for policy in policies:
-            empty_policy = policy.generate_empty_policy(self)
-            policy_name = policy.generate_policy_name(self)
+        for Policy in self.initializedPolicies:
```
petrkalos, line 101: why `Policy` and not `policy`? this might confuse someone thinking that this is a PolicyClass object (similar to the rest of the places you used capital P)

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -19,17 +22,54 @@ def __init__(
...
+            role_name,
+            account,
+            environmentUri,
+            resource_prefix,
+    ):
+        self.role_name = role_name
+        self.account = account
+        self.environmentUri = environmentUri
+        self.resource_prefix = resource_prefix
+        self.ManagedPolicies = ManagedPolicy.__subclasses__()
```
petrkalos, line 84: * As we discussed offline `__subclasses__` will return only those classes that have imported in the interpreter prior to this call. How do we ensure that this happens?
* Since self.ManagedPolicies isn't used anywhere else in the class I wouldn't make it a member but a local var. Also I'd most likely name in `ManagedPoliciesCls`

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -19,17 +22,54 @@ def __init__(
...
+        logger.info(f'Found {len(self.ManagedPolicies)} subclasses of ManagedPolicy')
+        self.initializedPolicies = [self._initialize_policy(policy) for policy in self.ManagedPolicies]
```
petrkalos, line 86: `initializedPolicies` makes me think that some extra processing/initialisation took place yet you only call the ctor. I'd name it `self.policies`

backend/dataall/core/environment/services/managed_iam_policies.py
dlpzx: It is a Policy class, that's why I defined it with uppercase

backend/dataall/core/environment/services/managed_iam_policies.py
dlpzx: - we can ensure that it is initialized by adding it to the `__init__` of `dataall/modules/dataset_sharing/__init__.py` using the `(ModuleInterface)`

dlpzx:
## Backwards compatibility
1. Environment Group Policy - IAM role created by data.all - on a team that already has a share request with s3 bucket with a dataset using KMS.
- [ ] click  on update Environment stack -> the stack is updated, the managed policy is created containing the S3 and KMS resources. Policy is attached to role.

ADD ITEMS, REVOKE and GRANT listed below are the same for data.all created roles and for imported roles.

2. Imported - Environment Group Policy GRANT - on a team that already has a share request with access points and s3 bucket with a dataset using KMS.
- [ ] Create NEW Share request ->  it creates successfully. Policy is created containing the S3 and KMS resources of the inline policies of the previous shares. Policy is attached to role.
- [ ] Add folders and bucket, process Share request -> IAM policy is updated with dataset. Policy is still attached to role.

3.  Imported - Environment Group Policy ADD ITEMS - on a team that already has a share request with access points and s3 bucket with a dataset using KMS.
- [ ] add more folders to share from the share request -> Policy is created containing the S3 and KMS resources of the inline policies of the previous shares, plus the added folders. Policy should be attached to role.
 

4.  Imported - Environment Group Policy REVOKE - on a team that already has a share request with access points and s3 bucket with a dataset using KMS.
- [ ] Revoke folders from the share request -> Policy is created containing the S3 and KMS resources of the inline policies of the previous shares, minus the revoked folders. Policy should be attached to role.


 
5. Data.all managed consumption roles GRANT - on a team that already has a share request with access points and s3 bucket with a dataset NOT using KMS.
- [ ] Create NEW Share request ->  it creates successfully. Policy is created containing the S3 ONLY resources of the inline policies of the previous shares. Policy is attached to role.
- [ ] Add folders and bucket, process Share request -> IAM policy is updated with dataset (and new KMS statement). Policy is still attached to role.

6. Data.all managed consumption roles ADD ITEMS - on a team that already has a share request with access points and s3 bucket with a dataset NOT using KMS.
- [ ] add more folders to share from the share request -> Policy is created containing the S3 ONLY resources of the inline policies of the previous shares, plus the added folders. Policy should be attached to role.

7. Data.all managed consumption roles REVOKE - on a team that already has a share request with access points and s3 bucket with a dataset NOT using KMS.
- [ ] Revoke folders from the share request -> Policy is created containing the S3 ONLY resources of the inline policies of the previous shares, minus the revoked folders. Policy should be attached to role.

backend/dataall/core/environment/cdk/environment_stack.py
```diff
@@ -431,6 +432,28 @@ def create_group_environment_role(self, group: EnvironmentGroup, id: str):
...
         ).generate_policies()
 
+        external_managed_policies = []
+        policy_manager = PolicyManager(
+            environmentUri=self._environment.environmentUri,
+            resource_prefix=self._environment.resourcePrefix,
+            role_name=group.environmentIAMRoleName,
+            account=self._environment.AwsAccountId
+        )
+        for policy in policy_manager.get_all_policies():
```
noah-paige, line 442: so if share policy does not exist we create it from inline (backwards compatibility) - but we do not attach it to the role (at least not on that specific update)

but if we then update env stack again the managed policies will get attached to the role

OR if we invite a new team to the env their policies will get created during API call and be attached to role on env stack update

…correct?

backend/dataall/core/environment/cdk/environment_stack.py
```diff
@@ -431,6 +432,28 @@ def create_group_environment_role(self, group: EnvironmentGroup, id: str):
             permissions=group_permissions,
         ).generate_policies()
 
+        external_managed_policies = []
```
noah-paige, line 435: Not from this PR but on L408 I think we are missing a `group_roles.append(...)` for imported env group roles

backend/dataall/core/environment/services/environment_service.py
```diff
@@ -341,6 +353,12 @@ def remove_group(session, uri, group):
             resource_uri=environment.environmentUri,
             resource_type=Environment.__name__,
         )
+        PolicyManager(
+            role_name=group_membership.environmentIAMRoleName,
+            environmentUri=environment.environmentUri,
+            account=environment.AwsAccountId,
+            resource_prefix=environment.resourcePrefix
+        ).delete_all_policies()
```
noah-paige, line 361: Should these lines be moved to above L346-348 (before we delete the group_membership object)

Not sure if reference to `group_membership.environmentIAMRoleName` breaks after than but getting local IDE flag `"environmentIAMRoleName" is not a known member of "None"`

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -0,0 +1,156 @@
...
+        return (share_policy is not None)
+
+    def check_if_policy_attached(self):
+        policy_name = self.generate_policy_name()
+        return IAM.is_policy_attached(self.account, policy_name, self.role_name)
+
+    def attach_policy(self):
+        policy_arn = f"arn:aws:iam::{self.account}:policy/{self.generate_policy_name()}"
+        try:
+            IAM.attach_role_policy(
```
noah-paige, line 62: does `attach_role_policy()` ever raise an exception?

backend/dataall/modules/dataset_sharing/api/input_types.py
```diff
@@ -29,7 +30,7 @@
     name='ShareItemSelectorInput',
     arguments=[
         gql.Argument(name='shareUri', type=gql.NonNullableType(gql.String)),
-        gql.Argument(name='itemUris', type=gql.NonNullableType(gql.ArrayType(gql.String))),
+        gql.Argument(name='revokedItemUris', type=gql.NonNullableType(gql.ArrayType(gql.String))),
```
noah-paige, line 33: change back to `itemUris`

backend/dataall/modules/dataset_sharing/cdk/pivot_role_data_sharing_policy.py
noah-paige: On second thought we will have to also include delete policy permissions for the inline policies we are removing, 
`dataall-targetDatasetS3Bucket-AccessControlPolicy` and `targetDatasetAccessControlPolicy`

backend/dataall/modules/dataset_sharing/services/managed_share_policy_service.py
```diff
@@ -0,0 +1,238 @@
...
+        ).build_compliant_name()
+
+    def generate_empty_policy(self) -> dict:
+        return {
+            "Version": "2012-10-17",
+            "Statement": [
+                {
+                    "Sid": EMPTY_STATEMENT_SID,
+                    "Effect": "Allow",
+                    "Action": "none:null",
```
noah-paige, line 52: did not know this was possible... cool!

backend/dataall/modules/dataset_sharing/services/managed_share_policy_service.py
```diff
@@ -0,0 +1,238 @@
...
+from dataall.base.aws.iam import IAM
+from dataall.base.utils.naming_convention import NamingConventionService, NamingConventionPattern
+from dataall.core.environment.services.managed_iam_policies import ManagedPolicy
+import logging
+
+log = logging.getLogger(__name__)
+
+OLD_IAM_ACCESS_POINT_ROLE_POLICY = "targetDatasetAccessControlPolicy"
+OLD_IAM_S3BUCKET_ROLE_POLICY = "dataall-targetDatasetS3Bucket-AccessControlPolicy"
+FAKE_S3_PLACEHOLDER = "arn:aws:s3:::initial-fake-empty-bucket"
```
noah-paige, line 11: Remove `FAKE_S3_PLACEHOLDER`

backend/dataall/base/aws/iam.py
```diff
@@ -101,3 +100,147 @@ def delete_role_policy(
...
+            response = iamcli.get_policy(PolicyArn=arn)
+            versionId = response['Policy']['DefaultVersionId']
+            policyVersion = iamcli.get_policy_version(PolicyArn=arn, VersionId=versionId)
+            policyDocument = policyVersion['PolicyVersion']['Document']
+            return versionId, policyDocument
+        except Exception as e:
+            log.error(
+                f'Failed to get policy {policy_name} : {e}'
+            )
+            return None
```
noah-paige, line 158: change to `return None, None` to return tuple - otherwise may run into unpacking error

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -172,6 +169,7 @@ def check_target_role_access_policy(self) -> None:
         and add to folder errors if check fails
         :return: None
         """
+        # TODO: review after merge
```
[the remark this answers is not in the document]
noah-paige: Please check my work on the above - just threw something together but likely can be optimized / re-use similar method calls for check and add permissions

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -148,76 +145,69 @@ def grant_s3_iam_access(self):
         Updates requester IAM role policy to include requested S3 bucket and kms key
```
noah-paige, line 145: Proposed change to `def check_s3_iam_access()`

```
    def check_s3_iam_access(self) -> None:
        """
        Checks if requester IAM role policy includes requested S3 bucket and kms key permissions
        and add to bucket errors if check fails
        :return: None
        """

        logger.info(f"Verify target role {self.target_requester_IAMRoleName} access policy")

        share_policy_service = SharePolicyService(
            environmentUri=self.target_environment.environmentUri,
            account=self.target_environment.AwsAccountId,
            role_name=self.target_requester_IAMRoleName,
            resource_prefix=self.target_environment.resourcePrefix
        )

        share_resource_policy_name = share_policy_service.generate_policy_name()

        logger.info(f'Share policy name is {share_resource_policy_name}')
        version_id, policy_document = IAM.get_managed_policy_default_version(
            self.target_account_id,
            share_resource_policy_name
        )

        if not policy_document:
            self.bucket_errors.append(
                ShareErrorFormatter.dne_error_msg(share_resource_policy_name, self.target_requester_IAMRoleName)
            )
            return

        key_alias = f"alias/{self.dataset.KmsAlias}"
        kms_client = KmsClient(self.dataset_account_id, self.source_environment.region)
        kms_key_id = kms_client.get_key_id(key_alias)

        s3_target_resources = [
            f"arn:aws:s3:::{self.bucket_name}",
            f"arn:aws:s3:::{self.bucket_name}/*"
        ]

        if not share_policy_service.check_resource_in_policy_statement(
            target_resources=s3_target_resources,
            statement_sid=f"{IAM_S3_BUCKETS_STATEMENT_SID}S3",
            policy_document=policy_document
        ):
            self.bucket_errors.append(
                ShareErrorFormatter.missing_permission_error_msg(
                    self.target_requester_IAMRoleName,
                    "IAM Policy",
                    f"{IAM_S3_BUCKETS_STATEMENT_SID}S3",
                    "S3 Bucket",
                    f"{self.bucket_name}",
                )
            )

        if kms_key_id:
            kms_target_resources = [f"arn:aws:kms:{self.dataset_region}:{self.dataset_account_id}:key/{kms_key_id}"]
            if not share_policy_service.check_resource_in_policy_statement(
                target_resources=kms_target_resources,
                statement_sid=f"{IAM_S3_BUCKETS_STATEMENT_SID}KMS",
                policy_document=policy_document
            ):
                self.bucket_errors.append(
                    ShareErrorFormatter.missing_permission_error_msg(
                        self.target_requester_IAMRoleName,
                        "IAM Policy",
                        f"{IAM_S3_BUCKETS_STATEMENT_SID}KMS",
                        "KMS Key",
                        f"{kms_key_id}",
                    )
                )
        return
```

backend/dataall/modules/dataset_sharing/services/share_managers/share_manager_utils.py
```diff
@@ -1,11 +1,6 @@
 import abc
```
noah-paige, line 1: unused import

deploy/pivot_role/pivotRole.yaml
```diff
@@ -439,12 +439,25 @@ Resources:
...
+              - !Sub 'arn:aws:iam::${AWS::AccountId}:role/*'
+          - Sid: IAMRolePolicy2
+            Action:
+              - 'iam:CreatePolicy'
+              - 'iam:DeletePolicy'
+              - 'iam:CreatePolicyVersion'
+              - 'iam:DeletePolicyVersion'
+            Effect: Allow
+            Resource:
+              - !Sub 'arn:aws:iam::${AWS::AccountId}:policy/${EnvironmentResourcePrefix}*'
```
noah-paige, line 460: same comment as data sharing pivot role policy to include delete policy permissions for the inline policies we are removing,
`dataall-targetDatasetS3Bucket-AccessControlPolicy` and `targetDatasetAccessControlPolicy`

.checkov.baseline
```diff
@@ -66,7 +66,8 @@
                 {
                     "resource": "AWS::IAM::ManagedPolicy.PivotRolepolicy3",
                     "check_ids": [
-                        "CKV_AWS_109"
+                        "CKV_AWS_109",
```
noah-paige, line 69: just tested and re-running chekcov command and creating a new baseline will likely remove this change ... would wait until we finalize pivotRole policies though before we run it one last time before PR merge

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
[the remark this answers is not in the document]
dlpzx: I worked on this before looking at the PR again, damm it!

backend/dataall/core/environment/services/managed_iam_policies.py
dlpzx: I am not sure I am following you, the initalizedPolicies are ManagedPolicies initialized with values

backend/dataall/base/aws/iam.py
dlpzx: Good catch, this leads me to think that we need to test some not-happy paths

backend/dataall/core/environment/services/managed_iam_policies.py
dlpzx: In this [commit](https://github.com/data-dot-all/dataall/pull/1068/commits/65cf21a507086e3587ecfa0e8e22d52e8bf7601a) I added error handling to be propagated up to the API. I also restructured the resolvers a bit so that they will fail if they cannot create the policies before creating the RDS rows

backend/dataall/modules/dataset_sharing/cdk/pivot_role_data_sharing_policy.py
dlpzx: I have scoped down the permissions for the first case but not for the old inline policies, I will test with what we ahve now and see how the backwards comaptibility works

backend/dataall/core/environment/cdk/environment_stack.py
dlpzx: Fixed, it only affected subscriptions that's why we did not spot it

dlpzx:
# Local testing [ With latest changes Mon March 4th]

## Newly added consumption roles/teams

1.a Environment Group Policy - IAM role created by data.all
- [ ] Invite group to environment -> IAM policy for shares is created with fake resources (update: with 2 statements one for access points and the other for buckets). Role is created in the environment stack where the policy is attached to role.
- [ ] Create Share request ->  it creates successfully. Policy is still attached to role. 
- [ ] Add folders and bucket, process Share request -> IAM policy is updated with dataset, fake resources are removed from policy. there are 4 statements. Policy is still attached to role. 
- [ ] Revoke folder of Share request -> it shows fake resources policy for access points but S3Bucket s3 and kms statements remain.
- [ ] Revoke bucket of Share request -> IAM policy is updated and s3 resources are removed. Policy is still valid and contains fake resources for BucketSharing and AccessPoint sharing. Policy is still attached to role.
- [ ] Remove group from environment -> IAM policy for shares is deleted
- [ ] IAM role is deleted from environment stack safely

1.b Environment Group Policy - IAM role created by data.all as ENVIRONMENT Admin
In this test use a resource prefix different from the standard `dataall`
- [ ] Create new environment -> IAM policy for shares is created with fake resources (update: with 2 statements one for access points and the other for buckets). Role is created in the environment stack where the policy is attached to role. -> Issue found 🟥 It does not create nor attach the policy because it cannot assume the pivot role because it does not exist at the time we hit create_environment. The role is able to open a share request thanks to the backwards compatibility code that checks if there the policy exists, it will try to create a backfilled policy, but it will only contain the fake statements

2.a Environment Group Policy - IAM role IMPORTED
- [ ] Invite group to environment importing role -> IAM policy for shares is created with fake resources. Policy is attached to role.
- [ ] Create Share request ->  it creates successfully. Policy is still attached to role. 
- [ ] Add folders ONLY process Share request -> IAM policy is updated with dataset, fake resources for accesspoints are replaced in the policy. Bucket fake resources stay. Policy is still attached to role.
- [ ] Revoke folder of Share request -> IAM policy is updated and access point resource is replaced by fake resource, kms statement is removed, S3 bucket fake resource stays. Policy is still attached to role.
- [ ] Remove group from environment -> IAM policy for shares is deleted and removed from role

2.b Environment Group Policy -IAM role IMPORTED as ENVIRONMENT Admin
- [ ] Create new environment  importing role -> IAM policy for shares is created with fake resources (update: with 2 statements one for access points and the other for buckets). Role is "REFERENCED" in the environment stack where the policy is attached to role.
 
 3. Data.all managed consumption roles
- [ ] Add role to environment with "Let data.all attach policies" = `True`  -> IAM policy for shares is created with fake resources. In Teams tab, that role is displayed as Data.all managed with attached policies. Policy is attached to the role.
- [ ] Create Share request ->  it creates successfully. Policy is still attached to role.
- [ ] Add folders and bucket, process Share request -> IAM policy is updated with dataset, fake resources are removed from policy. Policy is still attached to role.
- [ ] Revoke folder of Share request -> IAM policy is updated and access point resource is removed, S3 bucket resource stays. Policy is still attached to role.
- [ ] Revoke bucket of Share request -> IAM policy is updated and s3 resources are removed. Policy is still valid. Policy is still attached to role.
- [ ] Remove consumption tole from environment -> IAM policy for shares is deleted

4. Customer managed consumption roles (no need to process share again as it is the same as test2)
- [ ] Add role to environment with "Let data.all attach policies" = `False`  -> IAM policy for shares is created with fake resources. In Teams tab, that role is displayed as Customer managed with NO attached policies
- [ ] Attach the policy manually -> Check that in the Teams tab the policy appears as attached. 
- [ ] Create Share request with flag  "Let data.all attach policies" = `False`->  Throws error
- [ ] Create Share request with flag  "Let data.all attach policies" = `False` + directly in IAM manually attach the policy for sharing to the role -> Share request creates successfully.
- [ ] Delete the previous share and remove the policy that you attached manually, Create Share request with flag  "Let data.all attach policies" = `True`->  it creates successfully and the policy is attached to the role

## data sharing module is disabled
- no policies are created for any role
- UI does not break
- TBD

For testing in AWS add:
- [ ] CICD pipeline completes
- [ ] Migration successfully add consumption role `dataallManaged` column and backfills existing rows with `True`


## Remarks:
In this PR:
- at the moment the empty IAM policy statement refers a non-existent AWS service. In the rare case a service called `none` with action `null` is launched we will need to think of other ways of implementing an empty IAM policy placeholder.
- Support for the case where a user accidentally deletes the IAM policy: at the moment it is taken care by the backwards compatibility code. In the future, we will need to adapt the code and in the case the policy does not exist, we call `generate_empty_policy` instead of `create_managed_policy_from_inline_and_delete_inline`. If there were existing shares, the user will be able to verify and reapply the unhealthy shares with PR #1062 
- Some KMS statements will be shared between access points and S3 bucket sharing. I think the benefit of keeping both sharing mechanisms isolated overcomes the downside of duplication in the KMS statements

Out of scope for this PR:
- S3 access points are granting permissions to the Bucket. I think we should restrict permissions to the access point only
- S3 access points sharing still faces some issues with the lag between access point creation and putPolicy operation.

deploy/pivot_role/pivotRole.yaml
dlpzx: Confirmed: ` Failed to delete policy targetDatasetAccessControlPolicy of role mar-science-march-a2-xbvhdms8 : An error occurred (NoSuchEntity) when calling the DeleteRolePolicy operation: The role policy with name targetDatasetAccessControlPolicy cannot be found.
cdkproxy-1       | Failed to delete policy dataall-targetDatasetS3Bucket-AccessControlPolicy of role mar-science-march-a2-xbvhdms8 : An error occurred (NoSuchEntity) when calling the DeleteRolePolicy operation: The role policy with name dataall-targetDatasetS3Bucket-AccessControlPolicy cannot be found.
` I'll fx it

backend/dataall/base/aws/iam.py
```diff
@@ -33,71 +35,247 @@ def get_role(account_id: str, role_arn: str, role=None):
...
+            client = IAM.client(account_id)
+            response = client.attach_role_policy(
+                RoleName=role_name,
+                PolicyArn=policy_arn
+            )
+            return True
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'AccessDenied':
+                raise Exception(
+                    f'Data.all Environment Pivot Role does not have permissionsto to attach policy {policy_arn} to the role {role_name}: {e}')
```
noah-paige, line 259: Super minor - missing space "... permissions to ..."

.checkov.baseline
noah-paige: I believe when re-running checkov command and creating a new baseline this exception on `PivotRolepolicy3` is no longer present

noah-paige (approved):
lgtm - approving

Issue #941, opened 2024-01-02
Two share requests running in parallel can override each other


zsaltys:
### Describe the bug

We had a situation where a dataset owner was approving a lot of share requests very quickly. What ended up happening is that we see in CloudTrail that the pivotRole was putting a new bucket policy on the same bucket **twice** at the exact same time. What ended up happening is that one share request overrode the other because both requests ran in parallel, figured out how the policy needs to be updated and tried to update it. In the end one request overrode the other.

From my understanding share requests run as ecs tasks. There is probably no easy way to ensure that only one such task can be running at a time for 1 dataset. Therefore we probably need to introduce locking into the sharing mechanism so that other tasks have to wait until others are done before they attempt to read the existing bucket policy and make changes.

### How to Reproduce

It is not trivial to do but we've seen this happen multiple times already. The result with S3 bucket sharing is that access request is granted and the share works but the bucket policy is missing permissions.

### Expected behavior

Share requests running in parallel should never override one another.

### OS

N/A

### Python version

N/A

### AWS data.all version

2.2

dlpzx:
Hi @zsaltys, thanks for the issue. It is an important task and we will look into it. We can work together on the design in this issue

dlpzx:
⚠️ ⚠️ [WIP - not ready for review yet]

Issue: Share processes executed in parallel on the SAME dataset override each other and cause sharing failures.
Solution: We need a way to ensure that shares on the SAME dataset are executed sequentially.

# Design

## Alternatives
- Evaluate and wait in ECS task - Change logic in ECS task, wait if there are shares on the same dataset in a parallel task (what @zsaltys suggested above)
- Metadata with queue order, wait in ECS task - Add order in share request metadata, change logic in ECS task, wait depending on the queue order
- Evaluate and Re-queue in Worker Lambda - Change logic in Worker Lambda or add another Lambda that requeues the SQS task if the target dataset is the same one.
- Metadata with queue order,  and Re-queue in Worker Lambda - Change logic in Worker Lambda or add another Lambda that requeues the SQS task if the target dataset is the same one.
- ECS task to manage sharing tasks
- Batch same-dataset shares and process them in the same ECS task.
- Embed ECS tasks in Step Functions - A little of an overkill for an infrequent issue.
- Place Eventbridge Pipe in front of sharing ECS task -> Might be an [issue with execution time limitations](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-pipes-batching-concurrency.html).

[...]

PR #1072, opened 2024-02-22
Feature 941

anushka-singh:
### Feature or Bugfix
- Feature

### Detail
Share requests running in parallel override one another.

### Relates
- https://github.com/data-dot-all/dataall/issues/941

### Testing
- 2 simultaneous shares got processed successfully with locking mechanism.
- Here is what the dataset_lock DB looks like and the values in it.
Screenshot 2024-02-14 at 5 12 54 PM
- acquiredBy column show the share which last acquired the lock for the particular dataset

### Security

  - Is the input sanitized?


backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -226,3 +246,108 @@ def revoke_share(cls, engine: Engine, share_uri: str):
...
+
+        except Exception as e:
+            session.rollback()
+            log.error("Error occurred while acquiring lock:", e)
+            return False
+
+    @staticmethod
+    def acquire_lock_with_retry(dataset_uri, session, share_uri):
+        max_retries = 10
+        retry_interval = 60
```
anushka-singh, line 296: These 2 numbers are open to discussion. 
@dlpzx @noah-paige What do you think of these numbers?

anushka-singh:
Filed another issue for opening a discussion on another locking usecase: https://github.com/data-dot-all/dataall/issues/1074.

backend/dataall/modules/datasets_base/db/dataset_models.py
```diff
@@ -164,3 +164,14 @@ class DatasetBucket(Resource, Base):
...
     def uri(cls):
         return cls.bucketUri
+
+
+class DatasetLock(Base):
+    __tablename__ = 'dataset_lock'
+    datasetUri = Column(String, nullable=False, primary_key=True)
+    isLocked = Column(Boolean, default=False)
+    acquiredBy = Column(String, nullable=False)
+
```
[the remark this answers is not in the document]
noah-paige: The one thing that comes to mind is maybe we should have a foreign key relation between this table and datasetUri in dataset table?

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
noah-paige: I think these numbers make sense - will do some testing to get a better sense of user experience but think 10 minute max wait seems good

Should we make these global variables at the top of this file, such as 
```
MAX_RETRIES = 10
RETRY_INTERVAL = 60
```

anushka-singh: Yes I will make them global variables. 

We could also make the max_retries 20 and keep the retry_interval 30. That would still give us 10 min coverage but only a 30 sec delay between retries. However, this could cause heavy throttling.

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -38,71 +43,86 @@ def approve_share(cls, engine: Engine, share_uri: str) -> bool:
...
+                    dataset,
+                    share,
+                    source_environment,
+                    target_environment,
+                ) = ShareObjectRepository.get_share_data(session, share_uri)
+
+                lock_acquired = cls.acquire_lock_with_retry(dataset.datasetUri, session, share.shareUri)
+                if not lock_acquired:
+                    log.error(f"Failed to acquire lock for dataset {dataset.datasetUri}. Exiting...")
+                    return False
```
anushka-singh, line 60: Currently, if a share is unable to acquire a lock, it remains in share_approved state. It does not go into share_failed - which is problematic since the user will not be able to revoke or delete the share to recreate it later. 
Although its an extremely rare situation to occur that within 10 mins multiple shares are approved and are unable to acquire locks, we must add a safeguard in place. 

I wanted to hear your thoughts on how to achieve this here.
I could create a new SM and use transition states to transition to share_failed. Not sure if thats the best approach though.

backend/dataall/modules/datasets_base/db/dataset_models.py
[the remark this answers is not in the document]
anushka-singh: Sure, I can create that.

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
noah-paige: When share is approved...
- Share Object status goes from `Submitted` --> `Approved`
- Share Item(s) statuses go from `Pending_Approval` --> `Share_Approved`

I think that would be best as you proposed, have a new Transition that would bring 
- Share Object status from `Approved` --> `Processed`
- Each Share Item(s) statuses from `Share_Approved` --> `Share_Failed`
if the timeout is reached and can not acquire the lock

In the case above then we should not have any share items or share object stuck in `Share_Approved` or `Approved` , respectively and the use than retry to process the share items at a later time

anushka-singh:
Alembic script failing with :
```
ERROR [alembic.util.messaging] Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
FAILED: Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
```

I am not sure how to specify which head version alembic should run upgrade with in the script. @noah-paige Can you suggest?

noah-paige:
> Alembic script failing with :
> 
> ```
> ERROR [alembic.util.messaging] Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
> FAILED: Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
> ```
> 
> I am not sure how to specify which head version alembic should run upgrade with in the script. @noah-paige Can you suggest?

@anushka-singh - I believe it is because there are 2 alembic migration scripts building off of the same `down_revision`

A PR we merged last week has revision id of `6c9a8afee4e4` which should be the latest in OS and the migration script in this PR should be updated to have that revision id as its down revision

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -226,3 +249,106 @@ def revoke_share(cls, engine: Engine, share_uri: str):
...
+                        DatasetLock.isLocked == True,
+                        DatasetLock.acquiredBy == share_uri
+                    )
+                )
+            )
+
+            query.update(
+                {
+                    "isLocked": False,
+                    "acquiredBy": share_uri
```
noah-paige, line 345: do we need to update  `"acquiredBy": share_uri` - is it not already this share_uri and we can just update `isLocked`. If anything I think it makes most sense to update `"acquiredBy": ''` if the lock is being released (only if `acquiredBy` can be nullable

anushka-singh: The acquiredBy column was more to indicate lock last acquired by which share. 
However, I can make it nullable and reset it to '' when we release lock too. 
I just thought knowing the last share that acquired it, might be useful in debugging. 

I have pushed a revision where acquiredBy can be null and acquiredBy = '' while releasing lock.

backend/migrations/versions/a1049d854e29_implement_dataset_locking_mechanism.py
```diff
@@ -0,0 +1,152 @@
...
+                sa.Column('acquiredBy', String, nullable=False),
+                sa.PrimaryKeyConstraint('datasetUri')
+            )
+
+            print('Creating a new row for each existing dataset in dataset_lock table')
+            for dataset in datasets:
+                dataset_lock = DatasetLock(
+                    datasetUri=dataset.datasetUri,
+                    isLocked=False,
+                    acquiredBy=''
```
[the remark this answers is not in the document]
anushka-singh: Sure! Made the change

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
anushka-singh: Done!

backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
```diff
@@ -1307,3 +1307,61 @@ def count_role_principal_shares(session, principal_id: str, principal_type: Prin
             )
             .count()
         )
+
+
+class DatasetLockShareItemManagementSM:
```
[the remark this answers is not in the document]
anushka-singh: I thought it best to keep it different from ShareItemSM and ShareObjectSM since the transformations in those are very different from the ones we wanted. For example, we are going directly from Approved to Processed - which might confuse some customers.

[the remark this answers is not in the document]
noah-paige: I think we should opt to have 1 SM manage the entire status lifecycle of the share object / share items rather than 2 separate SM classes that would manage different transitions of the same status attribute

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -226,3 +277,127 @@ def revoke_share(cls, engine: Engine, share_uri: str):
...
+                    "isLocked": False,
+                    "acquiredBy": ''
+                },
+                synchronize_session=False
+            )
+
+            session.commit()
+            return True
+        except Exception as e:
+            log.error("Error occurred while releasing lock:", e)
```
TejasRGitHub, line 381: Should there be a similar session.rollback() in case anything fails ?

backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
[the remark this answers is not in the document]
noah-paige: I think best to just add a new `ShareObjectActions` of `FailedStart` or `FailedAcquireLock` or similar name and have that start the transitions we need using the pre-existing SM classes

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -226,3 +277,127 @@ def revoke_share(cls, engine: Engine, share_uri: str):
...
+            session (sqlalchemy.orm.Session): The SQLAlchemy session object used for interacting with the database.
+            share_uri: The ID of the share that is attempting to acquire the lock.
+
+        Returns:
+            bool: True if the lock is successfully acquired, False otherwise.
+        """
+        try:
+            # Execute the query to get the DatasetLock object
+            dataset_lock = (
+                session.query(DatasetLock)
```
TejasRGitHub, line 297: Can there be a situation in which two ECS processes start at the same time and arrive here and read the same state . And then update the dataset_lock table on L308 - 311. 

Similar to how it is mentioned here - https://dev.to/ivankwongtszfung/safe-update-operation-in-postgresql-using-sqlalchemy-3ela

noah-paige: Reading the above docs it seems like this is true - should we include `.with_for_update().first()` as specified in the docs @TejasRGitHub posted cc: @anushka-singh 

I believe that will resolve as no other session can update that record and will return an OperationalError that we can handle in exceptions

anushka-singh: @TejasRGitHub, In my code, I had added .first() to fetch only the first entry. So, I dont think a process will be able to read 2 states at the same time. 

I have anyway added the `.with_for_update().first()` guardrail to be safe. Thanks for the suggestion!

backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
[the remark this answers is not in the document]
anushka-singh: Sure, I have included the change in the next revision.

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
anushka-singh: Added!

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -226,3 +275,124 @@ def revoke_share(cls, engine: Engine, share_uri: str):
...
+                .filter(
+                    and_(
+                        DatasetLock.datasetUri == dataset_uri,
+                        DatasetLock.isLocked == True,
+                        DatasetLock.acquiredBy == share_uri
+                    )
+                )
+                .with_for_update().one()
+            )
```
noah-paige, line 367: @anushka-singh - sorry for all the back and forth ...

with the latest changes to add `.with_for_update().one()` I am now receiving error: `"Arguments: (AttributeError("'DatasetLock' object has no attribute 'update'"),)"`


I think for we have 2 options:

- (1) For `def release_lock()` we can handle the update the same as  for acquire lock 
```
            dataset_lock = (
                session.query(DatasetLock)
                .filter(
                    and_(
                        DatasetLock.datasetUri == dataset_uri,
                        DatasetLock.isLocked == True,
                        DatasetLock.acquiredBy == share_uri
                    )
                )
                .with_for_update().one()
            )
           if dataset_lock:
                dataset_lock.isLocked = False
                dataset_lock.acquiredBy = ''
...
```

- (2) OR perform query and update all in 1:
```
            query = (
                session.query(DatasetLock)
                .filter(
                    and_(
                        DatasetLock.datasetUri == dataset_uri,
                        DatasetLock.isLocked == True,
                        DatasetLock.acquiredBy == share_uri
                    )
                )
                .update(
                {
                    "isLocked": False,
                    "acquiredBy": ''
                },
                )
            )
``` (and likely can handle `acquire_lock()` similarly)

anushka-singh: Hmm, that is really weird. I did not come across this issue in my testing!

But I have pushed out another revision with solution (1). Please LMK if its fixed now. I will also try it out again on my end.

noah-paige: will do - trying again now

noah-paige:
One (..hopefully last) issue I am running into with `def acquire_lock()` (and potentially `def release_lock()`) - When running multiple shares all at once the ones that need to wait to acquire lock run into an error similar to below once beginning to process the share:
`sqlalchemy.orm.exc.DetachedInstanceError: Instance <Dataset at 0x7f8f546c4190> is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: http://sqlalche.me/e/13/bhk3)`

I believe this is because one the first iteration when the share tries to acquire the lock and fails - since we are using the `.one()` and not `.first()` (or `.one_or_none()`) for the query, we get an error `sqlalchemy.orm.exc.NoResultFound` and thus go to the excpetion block and run `session.rollback()`

Now `session.rollback()` closes the session and makes the previous objects retrieved from the session stale and "boundless" (i.e. dataset, share object, etc. all from `get_share_data()` and `get_share_data_items()`)

To resolve we have a couple of options:
- (1) use `.first()` or similar in the query so `dataset_lock` can be `None` and we do not always go to exception block if no results found
- (2) use `session.expunge_all()` before we run `session.rollback()` (definitely required for `acquire_lock()` - maybe not for `release_lock()` as we are done processing share at this point)

I would think best if we implement both 1 and 2 above - should both be quick changes :)

anushka-singh:
> One (..hopefully last) issue I am running into with `def acquire_lock()` (and potentially `def release_lock()`) - When running multiple shares all at once the ones that need to wait to acquire lock run into an error similar to below once beginning to process the share: `sqlalchemy.orm.exc.DetachedInstanceError: Instance <Dataset at 0x7f8f546c4190> is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: http://sqlalche.me/e/13/bhk3)`
> 
> I believe this is because one the first iteration when the share tries to acquire the lock and fails - since we are using the `.one()` and not `.first()` (or `.one_or_none()`) for the query, we get an error `sqlalchemy.orm.exc.NoResultFound` and thus go to the excpetion block and run `session.rollback()`
> 
> Now `session.rollback()` closes the session and makes the previous objects retrieved from the session stale and "boundless" (i.e. dataset, share object, etc. all from `get_share_data()` and `get_share_data_items()`)
> 
> To resolve we have a couple of options:
> 
> * (1) use `.first()` or similar in the query so `dataset_lock` can be `None` and we do not always go to exception block if no results found
> * (2) use `session.expunge_all()` before we run `session.rollback()` (definitely required for `acquire_lock()` - maybe not for `release_lock()` as we are done processing share at this point)
> 
> I would think best if we implement both 1 and 2 above - should both be quick changes :)
> 
> Also please let me know if you are experiencing similar

In my testing, I have not come across this issue. I am not sure why you are facing this..

I have implemented both 1 and 2 solutions. 
Implemented `session.expunge_all()` in release_lock as well to add guardrails.

noah-paige (approved):
Tested in AWS for approving and revoking 10+ shares at the same time - all works perfectly, thanks for your contribution @anushka-singh

PR #1084, opened 2024-03-01
Add guardrails to alembic sync upgrade/downgrade

noah-paige:
### Feature or Bugfix
- Bugfix

### Detail
- When testing out some of the upgrade/downgrade paths of the alembic migration scripts for DB - I ran into an downgrade migration error where I could not drop a ForeignKey Constraint with no name provided
  - Made updates to specify a foreign key constraint name on create and delete which aligns with best practice
  - Added additional guardrails to check if table exists before creating or deleting it

### Relates
- N/A

### Security

  - Is the input sanitized?


backend/migrations/versions/6c9a8afee4e4__alembic_sync.py
```diff
@@ -17,147 +19,169 @@
 
 
 def upgrade():
+    envname = os.getenv('envname', 'local')
```
SofiaSazonova, line 22: By default we use `dckrcompose`, don't we?

SofiaSazonova (commented):
Successfully downgraded and upgraded 

- [x] Up-to-date database
- [x] Up-to-date database with deleted table `redshiftcluster_datasettable`

backend/migrations/versions/6c9a8afee4e4__alembic_sync.py
noah-paige: For AWS deployment - `envname` will be the `envname` of the particular deployment account

For local deployment with docker - `envname` will be `dkrcompose`

If not set thought best to default to `local` - as we do with other migration scripts such as:
-  `dataall/backend/migrations/versions/4f3c1d84a628_modify_notifications_column_types.py`
- `dataall/backend/migrations/versions/a1049d854e29_implement_dataset_locking_mechanism.py`
etc.

SofiaSazonova (approved):

Issue #1092, opened 2024-03-07
Redirect to correct URL instead of landing page


TejasRGitHub:
**Is your idea related to a problem? Please describe.**
When URL pointing to a specific resource like share , dataset is pasted on the UI and when the user is not logged in then in that case the user is taken to an external Idp like Cognito/ Custom Idp, etc and then the user is redirected to the landing page instead of the specified URL.  Then the user again has to paste the URL where it was intended to go. 

**Describe the solution you'd like**
The user should be redirected to the correct URL instead of the landing page / base domain URL. 

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

dlpzx:
Thanks @TejasRGitHub for the issue and for the PR! 🚀  Way to go! We will review the PR as soon as possible.

noah-paige:
closing as completed in #1094

PR #1094, opened 2024-03-07
[GH-1092] Redirect the user to correct URL after login 

TejasRGitHub:
### Feature or Bugfix
- Feature

### Detail

- Code which uses session storage to redirect user to the correct URL 
- As the code only contains window path ( not the domain / entire link ) and also there are checks to see if the window path mostly comprises of / and alphanumeric it it robust from security stand point

### Relates
- https://github.com/data-dot-all/dataall/issues/1092

### Security

  - Is the input sanitized?
- Does this PR introduce any functionality or component that requires authorization? Yes
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms? Yes
- Are you introducing any new policies/roles/users? No


frontend/src/authentication/components/AuthGuard.js
```diff
@@ -15,6 +16,13 @@ export const AuthGuard = (props) => {
       setRequestedLocation(location.pathname);
     }
 
+    if (
+      !sessionStorage.getItem('window-location') &&
+      location.pathname !== '/'
```
TejasRGitHub, line 21: Checking this explicitly so that we do not set the '/' which will happen in case a user logsout and then pastes the URL in the browser and again tries to logback with the new URL . As the user logged out the redirect URL is the base URL and then this gets stored in the window-location which should not be the case. We only want to store window location path pointing to specific path

SofiaSazonova:
As a suggestion for enhancement, related to the window path.

There can be a situation:
1. UserB send a link to UserA to a data.all resource
2. UserA wants to login and be redirected to this resource, but they are not authorised to see it. So, they see an error(which can disappear quickly, it's unconvenient to copy the error text) and a blank page. 

May be UserA should be redirected to another "Access denied" page?

frontend/src/authentication/components/AuthGuard.js
```diff
@@ -23,6 +31,22 @@ export const AuthGuard = (props) => {
     return <Navigate to={requestedLocation} />;
   }
 
+  if (
+    sessionStorage.getItem('window-location') &&
```
SofiaSazonova, line 35: Could you please add short comments above these two if-statements to show to explain, what exactly we check this way?
Smth like 
`if the current url is not the same as the target one`

frontend/src/authentication/components/AuthGuard.js
```diff
@@ -23,6 +31,22 @@ export const AuthGuard = (props) => {
...
+    location.pathname !== sessionStorage.getItem('window-location')
+  ) {
+    const windowPathLocation = sessionStorage.getItem('window-location');
+    sessionStorage.removeItem('window-location');
+    // Check if the window-location only contains alphanumeric and / in it and its not tampered
+    if (!regexToValidateWindowPathName.test(windowPathLocation))
+      return <>{children}</>;
+    // A guardrail to limit the string of the pathname to a certain characters
+    if (windowPathLocation.length > 50) return <>{children}</>;
```
SofiaSazonova, line 44: What's wrong with long URLs? Why 50? Maybe it's better to make it a param, not magic constant.

frontend/src/authentication/contexts/GenericAuthContext.js
```diff
@@ -279,6 +282,10 @@ export const GenericAuthProvider = (props) => {
         }
       });
     }
+
+    if (sessionStorage.getItem('window-location')) {
```
SofiaSazonova, line 286: According to [specification](https://developer.mozilla.org/en-US/docs/Web/API/Storage/removeItem), `removeItem` is safe, you don't have to check, if the item exists.

SofiaSazonova (changes requested):

frontend/src/authentication/components/AuthGuard.js
TejasRGitHub: Hi @SofiaSazonova , Thanks for the review. I added this number by heuristically checking the length of the windowpath for the longest URL. I think it was the URL to table in a dataset and it was approx 33 - 37 characters. 

I can make it a param and store it in the contants.js . What are your thoughts ?

SofiaSazonova: I'd rather make it a setting, rather then a constant. 
The maximum URL length is something the customer may want to define before the deployment, and it's more an attribute of solution rather then code constant.
But maybe @dlpzx can correct me here.

SofiaSazonova (changes requested):
Just the tiny thing left: max URL length as a constant/setting

frontend/src/authentication/components/AuthGuard.js
SofiaSazonova: @TejasRGitHub in this case, `constants` is a good place for it.

TejasRGitHub: Hi @SofiaSazonova , Updated code to make it a constant . Please let me know if it looks good

SofiaSazonova (approved):

SofiaSazonova:
@TejasRGitHub Thanks!

PR #1095, opened 2024-03-07
Fix verify share table items and access point share no bucket policy

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Verify Share throws errors on verifying table shares since introducing the new way to `initialize_clients()` for LF Shares when determining the source Glue DB
  - This PR includes the `initialize_client()` call before starting to verify the table share item
  - This PR fixes verify share checks on table shares where the glue DB is hosted in a separate catalog account

- When processing a folder item share - it will fail if there is no pre-existing bucket policy on the S3 bucket
  - This PR adds checks similar to how we handle with bucket sharing method `get_bucket_policy_or_default()`

### Relates
- N/A

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
+++ b/backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
@@ -126,7 +126,6 @@ class S3AccessPointShareManager:
         """
         logger.info(f"Manage Bucket policy for {self.bucket_name}")
 
-        s3_client = S3Client(self.source_account_id, self.source_environment.region)
         bucket_policy = self.get_bucket_policy_or_default()
         counter = count()
         statements = {item.get("Sid", next(counter)): item for item in bucket_policy.get("Statement", {})}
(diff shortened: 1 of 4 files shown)


noah-paige:
Tested this PR for the following:

- Create Share for only Folder Share Item on Imported Dataset with no S3 Bucket Policy
- [x] Bucket Policy created correctly
- [x] Share Succeeds
- [x] Can Access Folder as Requestor Principal IAM Role


- Start Verify Share Items on Bucket and Table Items with Glue DB hosted in a separate Catalog Account in Data.all
- [x] Share Verify successfully runs and determines share remains in healthy status (as expected)

dlpzx:
Thanks for the PR @noah-paige. I confirm both errors and I am testing it. I will approve and merge once it is reviewed and tested

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -127,10 +127,54 @@ def manage_bucket_policy(self):
         logger.info(f"Manage Bucket policy for {self.bucket_name}")
 
         s3_client = S3Client(self.source_account_id, self.source_environment.region)
```
dlpzx, line 129: Minor: this client is never used, it is duplicated in line 154

dlpzx (approved):
tested!

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
noah-paige: good catch - removed

PR #1099, opened 2024-03-11
Handle Error on clean up share and not get stuck in IN_PROGRESS status

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Scenario
  - Folder share item exists in an existing share object in DA<=v2.2
  - DA upgrade to v2.3
  - Before updating environment/pivot role - user removes folder item from existing share object

- Issue
  - Share tries to create the managed policies in `def clean_up_share()` and remove the inline policies
  - Function fails since pivotRole does not have proper IAM permissions
  - Exception raised and share gets stuck in a REVOKE_IN_PROGRESS state + Dataset Lock Table never releases Lock for next share to process

- Resolution
  - Wrap `def clean_up_share()` method for revoke folder item in a try/except block
  - ++ Remove some unused additional code

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
+++ b/backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
@@ -267,13 +267,8 @@ class DataSharingService:
                     share_uri,
                     ShareableType.StorageLocation.value
                 )
-                existing_shared_buckets = ShareObjectRepository.check_existing_shared_items_of_type(
-                    session,
-                    share_uri,
-                    ShareableType.S3Bucket.value
-                )
-                existing_shared_items = existing_shared_folders or existing_shared_buckets
-                log.info(f'Still remaining S3 resources shared = {existing_shared_items}')
+
+                log.info(f'Still remaining S3 folder resources shared = {existing_shared_folders}')
                 if not existing_shared_folders and revoked_folders:
                     log.info("Clean up S3 access points...")
                     clean_up_folders = ProcessS3AccessPointShare.clean_up_share(
--- a/backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
+++ b/backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
@@ -215,16 +215,20 @@ class ProcessS3AccessPointShare(S3AccessPointShareManager):
             source_env_group,
             env_group,
         )
-        clean_up = clean_up_folder.delete_access_point(share=share, dataset=dataset)
-
-        if clean_up:
-            clean_up_folder.delete_target_role_access_policy(
-                share=share, dataset=dataset, target_environment=target_environment
-            )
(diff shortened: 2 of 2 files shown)


backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -267,13 +267,8 @@ def revoke_share(cls, engine: Engine, share_uri: str):
                     share_uri,
                     ShareableType.StorageLocation.value
                 )
-                existing_shared_buckets = ShareObjectRepository.check_existing_shared_items_of_type(
```
noah-paige, line 270: Removed since unused

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
```diff
@@ -215,16 +215,20 @@ def clean_up_share(
...
-
-        if clean_up:
-            clean_up_folder.delete_target_role_access_policy(
-                share=share, dataset=dataset, target_environment=target_environment
-            )
-            if not dataset.imported or dataset.importedKmsKey:
-                clean_up_folder.delete_dataset_bucket_key_policy(dataset=dataset)
-
-        return True
+        log.info("##### Cleaning up folder share resources #######")
```
noah-paige, line 218: This is a quick fix - we may want to make folder revoke better and add the clean up steps as part of the `def process_revoked_share()` 

Otherwise, revoking a folder item may not clean up all of the resources before reaching a `REVOKE_SUCCEEDED` state

For now leaving design as is and just implementing fix so we don't run into larger issues when upgrading DA and revoking items

dlpzx:
Testing it in AWS and provoking a failure by removing pivot role permissions. 
When checking the logs of the ECS share task I can see the error: `12-Mar-24 13:57:25 - INFO - clean_up_share - Failed to clean up folder share resources due to: Data.all Environment Pivot Role does not have permissions to update policy dataall-env-u22f6tn3-share-policy-dataall-test-environmentb1-u22f6tn3: An error occurred (AccessDenied) when calling the CreatePolicyVersion operation: User: arn:aws:sts::XXXXXXX:assumed-role/dataallPivotRole-cdk/dataallPivotRole-cdk is not authorized to perform: iam:CreatePolicyVersion on resource: policy arn:aws:iam::XXXXXXX:policy/dataall-env-u22f6tn3-share-policy-dataall-test-environmentb1-u22f6tn3 because no identity-based policy allows the iam:CreatePolicyVersion action  ` and `Clean up S3 successful = False `

But then, the revoke as a whole succeeds and the share item appears as revoke succeeded (because the revoke for each folder individually was indeed a success in `revoked_folders_succeed`. I think @noah-paige you are right at pointing out that we need to review how the clean-up is done. Maybe we should open an issue wdyt?

noah-paige:
New Issue to pick up later to improve the clean up of folders - https://github.com/data-dot-all/dataall/issues/1101

Issue #1076, opened 2024-02-26
Automatic code formatting


petrkalos:
**Is your feature request related to a problem? Please describe.**
Inevitable when we are working on new features we are touching a lot of files and our IDEs might reformat code unrelated to our change. This is not ideal because it makes reviewing much more difficult (hard to find the important changes) but also might introduce unnecessary conflicts. 

**Describe the solution you'd like**
I suggest we introduce automated formatting (i.e using black)
* Run it as part of CI/CD
* Urge contributors to use it as a pre-commit hook

**Describe alternatives you've considered**
The other alternative is to ask for people to refrain from doing such changes but this relies on good intentions it won't always work.

jaidisido:
In favour, however I recommend using [Ruff](https://docs.astral.sh/ruff/) instead of black as it's significantly faster and covers most linters in one. We used to use black within aws-sdk-pandas but recently made the switch to [ruff](https://github.com/aws/aws-sdk-pandas/blob/main/pyproject.toml#L141) and are happy customers

PR #1105, opened 2024-03-13
Ruff code auto-format

SofiaSazonova:
### Feature or Bugfix

- Refactoring

### Detail
Ruff tool configured for the project.
All files are formatted 
In pipelines instead of "make lint" now "make ruff" is used

### Relates
 https://github.com/data-dot-all/dataall/issues/1076

### Security


SofiaSazonova:
Flake8 fails, but it's fine by now. It's only a first iteration of introducing Ruff.
Ruff rules are slightly different

Makefile
```diff
@@ -33,9 +33,9 @@ install-cdkproxy:
 install-tests:
 	pip install -r tests/requirements.txt
 
-lint:
```
petrkalos, line 36: I wouldn't makefile target tool specific, let's keep it as `lint`

Makefile
```diff
@@ -33,9 +33,9 @@ install-cdkproxy:
...
-	pip install flake8
-	python -m flake8 --exclude cdk.out,blueprints --ignore E402,E501,F841,W503,F405,F403,F401,E712,E203 backend/
+ruff:
+	pip install ruff
+	ruff check --fix
```
petrkalos, line 38: shouldn't we run `ruff format` here as well?

pyproject.toml
```diff
@@ -1,21 +1,32 @@
-[tool.black]
+[tool.ruff]
 line-length = 120
-target-version = ["py37", "py38", "py39", "py310"]
```
petrkalos, line 3: Should we keep `target-version` as 39 which is the lambda runtime version we use?

pyproject.toml
```diff
@@ -1,21 +1,32 @@
...
-  | dev
-  | .coverage
-)/
-'''
\ No newline at end of file
+src = ["backend", "deploy"]
+
+
+[tool.ruff.lint]
+exclude = [ "backend/migrations","tests/**", "**/input_types.py", "**/mutations.py", "**/queries.py", "**/types.py", "**/api/__init__.py"]
```
petrkalos, line 7: why exclude migrations?

petrkalos (commented):
Since this is python and even format changes can easily break functionality can you describe what short of testing did you perform to validate the PR?

Makefile
SofiaSazonova: May be we can keep both? I don't want the name to be confusing, since it's just an alias

SofiaSazonova:
> Since this is python and even format changes can easily break functionality can you describe what short of testing did you perform to validate the PR?

1. Run 'make test' to ensure, that integration test are fine
2. Run 'cdk list' to ensure, that deployment is build without errors (there were some troubles with imports, that i caught that way)
3. Perform migrations (drop and upgrade) for local database
4. Check that CICD pipeline completed

Makefile
petrkalos: Previously we were running `flake8` and we called the target `lint` and not `flake8` which I think is correct because it's abstracting whatever underlying linter we use and also that way we wouldn't have to change the pipelines code.

petrkalos:
> And the problem with flake8 testing of this PR is `E704 multiple statements on one line (def)` It's a known disagreement, e.g. [discussed here](https://github.com/pyvista/pyvista/issues/5591) May be we should reconfigure the Github testing then.

Oh I thought we were going to drop flake8 on a next PR. 
I don't think it makes sense to keep 2 formatters/linters.

petrkalos:
> > Since this is python and even format changes can easily break functionality can you describe what short of testing did you perform to validate the PR?
> 
> 1. Run 'make test' to ensure, that integration test are fine
> 2. Run 'cdk list' to ensure, that deployment is build without errors (there were some troubles with imports, that i caught that way)
> 3. Perform migrations (drop and upgrade) for local database
> 4. Check that CICD pipeline completed

I think another thing you could run is to make some requests through the cdkproxy, an action that will create a new stack (i.e create a new dataset or environment etc)

pyproject.toml
SofiaSazonova: I enabled all formatting back. And added block-comment for current unused imports. In future, if something like this comes up again, we can decide on-the-go, if any other re-configuration is needed

dlpzx (approved):
Looks good to me

Issue #1101, opened 2024-03-12
Handle Entire Folder Revoke + Clean Up Before Share Process Succeeds


noah-paige:
### Describe the bug

When revoking a folder share item 
- The share item updates to `Revoke_Succeeded` once the access point policy is updated to remove the folder prefix for the principal IAM role

But if there are no additional folder shares for the same principal we perform a number of other steps, including:
- Delete Access Point
- Remove access point permissions from princinpal IAM Role share policy 
- Updates to the KMS policy permissions to remove principal IAM Role

The above actions all happen after the folder item is already updated to `Revoke_Succeeded` but theoretically the revoke process could fail in those steps and then there are pending resources remaining in the account that should have been cleaned up - and no way for data.all to notify the user or try again to clean up currently

### How to Reproduce

```
*P.S. Please do not attach files as it's considered a security risk. Add code snippets directly in the message body as much as possible.*
```

### Expected behavior

data.all updated to a Revoke_Succeeded status only when it is done cleaning up all resources related to the share item and/or notifies the user when it fails to clean up some particular resource

### OS

N/A

### Python version

N/A

### AWS data.all version

v2.2.0

dlpzx:
Merged!

Issue #997, opened 2024-01-24
remove AllowAllToAdmin statement from bucket policies


zsaltys:
Currently when a share is created for an S3 bucket (using bucket policies) data.all will add the following statement to the bucket policy:

```
{
			"Sid": "AllowAllToAdmin",
			"Effect": "Allow",
			"Principal": "*",
			"Action": "s3:*",
			"Resource": [
				"arn:aws:s3:::BUCKET_NAME",
				"arn:aws:s3:::BUCKET_NAME/*"
			],
			"Condition": {
				"StringLike": {
					"aws:userId": [
						"ROLE_ID_DATASET_ADMIN:*",
						"ROLE_ID_ENV:*",
						"ROLE_ID_PIVOT_ROLE:*"
					]
				}
			}
		},
```

The idea here is that dataset admin role, environment role and the pivot role have access to the bucket. However all those roles already have access to the bucket and the only way they would not have access if the bucket policy had statements restricting access to other roles which most people won't do as it's not necessary because it is enough to control permissions on the role within same account. Even if someone restricted their bucket policy to all other roles on the account then data.all would have no way to add this statement itself as no permissions would allow that. Importing such buckets would simply fail.

We have tested this and data.all functions completely fine with this block removed. Additionally the logic managing this statement has bugs. For example if you reimport a dataset then the dataset role admin id will not be updated on the bucket but everything will still continue working. The same happens if you unimport, delete an environment and reimport the bucket on a new environment the environment role id will not be updated.

Same idea is used for access points but there I'm not sure if this block is required or not. This should be checked.

I propose that we fix this by releasing a version which will clean up this block for everyone if they have it during dataset update cycle. And then during some version remove the cleanup logic as well.

noah-paige:
I performed some quick testing on the above by removing `AllowAllToAdmin` Sid for both a created and an imported dataset and tested (for each):
- [x] Existing Bucket Share and Folder Share Still Works
- [x] Dataset Role still can upload Data


Good catch @zsaltys - the `AllowAllToAdmin` block of code may not be needed. 

Also while in OS code we add `"Sid": "AllowAllToAdmin"` ( if it does not already exist on the bucket policy) we never update/delete it once added. This makes us more prone to errors especially during deletes / clean up of resources.

I am inclined that we pick this up in an upcoming release to remove this statement from the OS code

dlpzx:
Merged!

PR #1106, opened 2024-03-14
Fix/remove allow all bucket statement

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
- Remove unnecessary AllowAllToAdmin statement in bucket policy and in access point policies
- For the deletion of access points in existing shares I kept a reference to `AllowAllToAdmin`. We could change this, I am not strong opinionated.
- For existing bucket policies and access point policies `AllowAllToAdmin` will be deleted when all folders are revoked. It is unnecessary but it does not harm anyone to leave it as it is for existing shares.

[UPDATE]
Since the testing uncovered issues derived from removing  `AllowAllToAdmin` with the current clean-up actions I took the issue #1101 and included it in this PR. On top of the aforementioned changes, this PR:
- Remove unnecessary `clean_up` functions from revoke_folders and move the clean_up logic into the `process_revoked_folders` function (from the `data_sharing_service.py`)
- Break `delete_access_point_policy` into smaller functions with less responsibilities: `revoke_access_in_access_point_policy` that generates the new access point policy; `attach_new_access_point_policy` that attaches a given policy to an access point; `delete_access_point` that deletes an access point. 
- Logic on what should be deleted moved to `backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py`. Only if the resulting policy contains no statements the access point is deleted, the IAM permissions revoked from the requester IAM role and the KMS key policy revoked.
- Bonus point: fixed small bug on the clean-up (it was assuming prefix_list is a list, but for a single folder shared it can be a string)

### Relates
- #997 
- #1101 

### Security

  - Is the input sanitized?


SofiaSazonova (approved):

dlpzx:
Testing (I waited to have the PR for linting merged before doing proper testing)
- [X] Revoking all folders deletes access point with `AllowAllToAdmins`
- [X] Revoke Bucket does not change the `AllowAllToAdmin`, but works as expected
- [X] Sharing folder creates new access point and access point policy without `AllowAllToAdmins` statement
- [ ] Revoking the folder from previous step deletes access point --> error `Failed to revoke S3 permissions to folder folder1 from source account 496481071459//eu-west-1 with target account 443950722698/eu-west-1 due to: An error occurred (MalformedPolicy) when calling the PutAccessPointPolicy operation: Could not parse the policy: Statement is empty!`
- [ ] Share bucket with empty policy succeeds and adds SSL secure transport statement only

dlpzx:
I am picking up #1101 alongside this issue because they are very much related

petrkalos (approved):
lgtm

Issue #741, opened 2023-09-08
Standardize core modules and cleanup dead code


dlpzx:
**Is your idea related to a problem? Please describe.**
Data.all V2 modules in the backend follow a particular structure. Division of resolvers, db, services... At the moment core modules such as environments and organizations follow the old structure. In addition, there dead code in core that has not been removed.

Additionally, some of the operation names on the front end are to be changed to a standard naming convention (i.e. camel case, matching with backend name, etc.)

**Describe the solution you'd like**
Clean code that follows the same standard

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

dlpzx:
Completed

dlpzx:
Some modules are still note refactored, re-opening issue:
- environments
- groups
- permissions
- stacks

SofiaSazonova:
- [ ] environments 
- [ ] groups https://github.com/data-dot-all/dataall/pull/1113
- [ ] permissions
- [ ] stacks

dlpzx:
@SofiaSazonova the only thing left in this issue is refactoring environment resolvers right?

anmolsgandhi:
Hi @SofiaSazonova can this issue be closed ?

dlpzx:
Completed in 2.5.0

PR #1113, opened 2024-03-18
741 refactoring. dataall.core.groups

SofiaSazonova:
### Feature or Bugfix
- Refactoring

### Detail
- logic of group-listing was moved to services.group_service.py

### Relates
- [741](https://github.com/data-dot-all/dataall/issues/741)

### Security

### DIFF
--- a/backend/dataall/core/groups/api/resolvers.py
+++ b/backend/dataall/core/groups/api/resolvers.py
@@ -39,8 +39,9 @@ def get_group(context, source, groupUri):
 
 
 def list_groups(context, source, filter: dict = None):
-    with context.engine.scoped_session() as session:
-        return GroupService.list_groups_without_invited(session, filter)
+    if not filter:
+        filter = {}
+    return GroupService.list_groups_without_invited(filter)
 
 
 def get_groups_for_user(context, source, userid):
--- a/backend/dataall/core/groups/api/resolvers.py
+++ b/backend/dataall/core/groups/api/resolvers.py
@@ -2,10 +2,9 @@ import os
 import logging
 
 from dataall.base.context import get_context
-from dataall.base.services.service_provider_factory import ServiceProviderFactory
 from dataall.core.groups.db.group_models import Group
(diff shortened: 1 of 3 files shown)


dlpzx:
Overall it looks great. There is one additional thing that we should modify on the resolvers `dataall/core/groups/api/resolvers.py`. As you might have noticed, we are passing `context` as input to all API calls, sometimes it is needed and sometimes it is not, but we still have to pass it to all calls. In addition, we are opening a session directly in the resolver and pass it to the service which then passes it on the db layer. The api layer should not start a session, should only validate input/output, then service layer validates permissions and then we can start a session. I think it is easier to see with an example. In this function, we are opening the session, before validating the permissions in the `service` layer. (for groups there are no permissions, but in general `services` validates permissions.

```
def list_groups(context, source, filter: dict = None):
        return GroupService.list_groups_without_invited(session, filter)
```

To avoid 1. passing the context as input to all APIs and 2. starting a session before validating permissions, we released a way of accessing the context of the API at any moment. With that in mind, we should change the resolver to:


```
def list_groups(context, source, filter: dict = None):
      if not filter:
        filter = {}
    return GroupService.list_groups_without_invited(filter)
```

Note that I have added some validation for `filter`, because if it is `None` it might fail in `dataall.core.groups.services.group_service.GroupService.list_invited_groups`
```
    def list_invited_groups(session, filter: dict = None):
        category, category_uri = filter.get('type'), filter.get('uri')
```

Then in `dataall/core/groups/services/group_service.py` we should import `from dataall.base.context import get_context`
and use it to open a sql alchemy session.

```
    @staticmethod
    def list_groups_without_invited(session, filter: dict = None):
            cognito_groups = GroupService.list_cognito_groups()
            invited_groups = GroupService.list_invited_groups(session, filter)
    
            groups = []
            for group in cognito_groups:
                if group not in invited_groups:
                    groups.append({'groupName': group})
    
            return groups
```

dlpzx (approved):
Re-reviewed! It is ready to be merged :)

Issue #945, opened 2024-01-05
Restricting Team/User Access to Worksheets


sandeephs1:
Team should only be allowed to query dataset via worksheet. Team is applied with Environment Permissions - `Create dataset Share requests for this environment` and `Run Worksheet Athena queries on this environment`

But it ends up with the below error 

`Insufficient permissions to execute the query. User: arn:aws:sts::123456789:assumed-role/<rolename> is not authorized to perform: lakeformation:GetDataAccess on resource: arn:aws:glue:usregion:5678912345:table/<tablename> because no identity-based policy allows the lakeformation:GetDataAccess action`

If team is added with the Env Permission `Create datasets on this environment`, user is able to query via worksheet but user can also create a dataset, which is not the requirement.

Any suggestion

noah-paige:
Hi @sandeephs1 - thank you for opening this issue. You may have found a bug in the way we assign permissions for invited teams to an environment based on the set of env-level permissions given to the invited team.

I will try to reproduce the above scenario to ensure that I see the same behavior - from there can work on a fix

At a high level - the cause of the above is that we add a subset of glue and lakeformation permissiosn to the team's environment role in `/backend/dataall/modules/datasets/cdk/env_role_dataset_glue_policy.py` ONLY IF the team has the `CREATE_DATASET` permissions (which is the `Create datasets on this environment` Env Permission)

Removing that if statement will ensure the required permissions are added to the team's env role regardless of CREATE_DATASET or not

PR #1137, opened 2024-04-02
Restricting Team/User Access to Worksheets

SofiaSazonova:
### Feature or Bugfix

- Bugfix

### Detail
Glue-permissions were added to the team's env role only if the role had CREATE_DATASET permissions. But `glue-permissions are about the get-operations, so i removed that if, so the required permissions are added to the team's env role regardless of CREATE_DATASET 

### Relates
- [945](https://github.com/data-dot-all/dataall/issues/945)

### Security

### DIFF
--- a/backend/dataall/modules/datasets/indexers/location_indexer.py
+++ b/backend/dataall/modules/datasets/indexers/location_indexer.py
@@ -1,4 +1,5 @@
 """Indexes DatasetStorageLocation in OpenSearch"""
+
 import re
 
 from dataall.core.environment.services.environment_service import EnvironmentService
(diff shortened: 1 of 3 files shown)


backend/dataall/modules/datasets/indexers/location_indexer.py
```diff
@@ -1,4 +1,5 @@
 """Indexes DatasetStorageLocation in OpenSearch"""
+
```
dlpzx, line 2: This is because of the linting right? It was also flagged for me while executing ruff locally

dlpzx:
Hi @SofiaSazonova how have you tested this PR? 

When this is merged the users will need to update the environment cloudformation stack for the changes to be effective. We should include some more instructions and clarification on what this PR does

dlpzx (approved):

backend/dataall/modules/datasets/indexers/location_indexer.py
SofiaSazonova: Yes, it's linting.

SofiaSazonova:
@dlpzx Yes, It's tested.

Issue #944, opened 2024-01-05
Missing permissions when first enabling Worksheets module


rbernotas:
### Describe the bug

With a deployed instance of data.all that never had Worksheets enabled, if you then enable Worksheets in the config.json, the following permissions do not exist, and are not applied to tenants and resources that already exist:

- GET_WORKSHEET
- DELETE_WORKSHEET
- UPDATE_WORKSHEET
- RUN_WORKSHEET_QUERY
- RUN_ATHENA_QUERY
- MANAGE_WORKSHEETS

This causes the following issues:

- Worksheets can be created, but not used, due to the lack of the GET_WORKSHEET permission being applied to the resource.
- In the admin view, if you try to view the permissions for any given tenant team, it leads to a blank screen and you can no longer set permissions for the teams (this is due to the lack of the MANAGE_WORKSHEETS tenant permission).
- In the environment team invite popup, the permission for running athena queries is not available or set.  Even if you had GET_WORKSHEET available, you would not be able to run the query in the sheet.
- Worksheets can be created, but not deleted or updated.
- There may be other issues, but these are the most apparent.

The following manual remediations need to take place:

1.  Add the permissions to the RDS db like (your permissionUris and created timestamps may need to differ, and where "dataalldb" is replaced with the name of your db):

--INSERT INTO dataalldb.permission ("permissionUri", "name", "type", "description", "created") VALUES ('mv46k9bt', 'GET_WORKSHEET', 'RESOURCE', 'GET_WORKSHEET', '2024-01-04 02:42:05.916708');                                       
--INSERT INTO dataalldb.permission ("permissionUri", "name", "type", "description", "created") VALUES ('mpqkwrq4', 'DELETE_WORKSHEET', 'RESOURCE', 'DELETE_WORKSHEET', '2024-01-04 02:42:05.916708');
--INSERT INTO dataalldb.permission ("permissionUri", "name", "type", "description", "created") VALUES ('eiw8iujo', 'UPDATE_WORKSHEET', 'RESOURCE', 'UPDATE_WORKSHEET', '2024-01-04 02:42:05.916708');
--INSERT INTO dataalldb.permission ("permissionUri", "name", "type", "description", "created") VALUES ('f0ngt5mi', 'RUN_WORKSHEET_QUERY', 'RESOURCE', 'RUN_WORKSHEET_QUERY', '2024-01-04 02:42:05.916708');
[...]

PR #1145, opened 2024-04-04
initialise db once per deployment

petrkalos:
### Feature or Bugfix
Feature

### Detail
* Introduce a TriggerFunction that will be executed before every deployment of the ApiGateway stack, we can use multiple instances of the same to introduce other one-off tasks (i.e migrations)
* Initialise the permissions by checking which ones are missing one by one (drop the length logic)
* Warn users if there are duplicate entries in the DB
* Warn users if there are osbolete entries in the DB (entries that exist in the DB but no longer exist in the app)

### Relates
Solves #944

### DIFF
--- a/backend/api_handler.py
+++ b/backend/api_handler.py
@@ -37,8 +37,6 @@ ENVNAME = os.getenv('envname', 'local')
 ENGINE = get_engine(envname=ENVNAME)
 Worker.queue = SqsQueue.send
 
-TenantPolicyService.save_permissions_with_tenant(ENGINE)
-
 
 def resolver_adapter(resolver):
     def adapted(obj, info, **kwargs):
--- a/backend/dataall/core/permissions/services/permission_service.py
+++ b/backend/dataall/core/permissions/services/permission_service.py
@@ -1,11 +1,14 @@
-from dataall.core.permissions.api.enums import PermissionType
-from dataall.core.permissions.db.permission.permission_repositories import PermissionRepository
+import logging
+from collections import Counter
+
+from sqlalchemy.orm import Session
+
 from dataall.base.db import exceptions
+from dataall.core.permissions.api.enums import PermissionType
 from dataall.core.permissions.db.permission.permission_models import Permission
+from dataall.core.permissions.db.permission.permission_repositories import PermissionRepository
 from dataall.core.permissions.services.resources_permissions import RESOURCES_ALL_WITH_DESC
 from dataall.core.permissions.services.tenant_permissions import TENANT_ALL_WITH_DESC
-
-import logging
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


backend/dataall/core/permissions/db/permission_repositories.py
```diff
@@ -73,48 +75,54 @@ def save_permission(session, name: str, description: str, permission_type: str)
...
+        perms = []
+        logger.info(f'db  perms {sorted(db_perms)}')
+        logger.info(f'app perms {sorted(app_perms)}')
 
-        count_tenant_permissions = (
-            session.query(models.Permission).filter(models.Permission.type == PermissionType.TENANT.name).count()
-        )
+        for perm, freq in Counter(db_perms).items():
+            if freq > 1:
+                logger.error('%50s permission appears %3d times in the db', perm, freq)
```
dlpzx, line 112: I would also avoid deleting permissions from RDS because we do not know if those permissions are used down the line, is this what you were thinking when you implemented the warnings?

petrkalos: I thought about deleting directly but we would have to make sure no URI is used somewhere else and if it does we need to update it to the one we are planning to keep. One thought is to write a DB migration script that will always execute and clean up the dips. Another one is to write one that converts all the URIs to be equal with the permission name, that way we both deduplicate the existing store but also future proofing it (uri is the primary key so we won't be able to insert dups)

backend/bootstrap_handler.py
```diff
@@ -0,0 +1,19 @@
...
+
+from dataall.base.db import get_engine
+from dataall.core.permissions.db import save_permissions_with_tenant
+
+logger = logging.getLogger()
+logger.setLevel(os.environ.get('LOG_LEVEL', 'INFO'))
+log = logging.getLogger(__name__)
+
+
+def save_permissions(event, context):
```
dlpzx, line 16: In real deployment we have a trigger Lambda that executes this handler before the api_handler. For local deployment, we still keep the save_permissions_with_tenant in the local_graphql_server. That is alright right?

deploy/stacks/bootstrap_stack.py
```diff
@@ -0,0 +1,201 @@
...
+
+        self.trigger_function.connections.allow_to(
+            ec2.Peer.any_ipv4(), ec2.Port.tcp(443), 'Allow NAT Internet Access SG Egress'
+        )
+
+        for connection in connections:
+            connection.allow_default_port_from(
+                self.trigger_function, f'Allow dataall {self.trigger_function.function_name}'
+            )
+
```
dlpzx, line 71: We can scope down the permissions of the role of the `save_permissions`, it needs fewer permissions than the generic `api_handler` role. Especially important is to restrict the permissions to other IAM roles, e.g. we do not need to assume the pivot role or the cdk exec role in other accounts. Similar with Pass role, we should aim to have least-minimum-permissions in the save_permissions_role. I think we barely need a couple of permissions, I only identified parameter store and secrets manager permissions to get_engine. Maybe some KMS permissions for the encryption + some logs, xray

dlpzx (commented):
A more generic question: Do you think we should move alembic logic to a trigger Lambda like for the initialization of permissions? Or at least that the trigger of the CodeBuild project was in the backend stack and not in the CICD pipeline?

backend/bootstrap_handler.py
petrkalos: I deliberately didn't touch the local_server because it execute the save_permissions upon starting not on the first request, hence it makes no difference.

deploy/stacks/bootstrap_stack.py
```diff
@@ -0,0 +1,201 @@
...
+        **kwargs,
+    ):
+        super().__init__(scope, id, **kwargs)
+
+        if self.node.try_get_context('image_tag'):
+            image_tag = self.node.try_get_context('image_tag')
+        image_tag = f'lambdas-{image_tag}'
+
+        env = {'envname': envname, 'LOG_LEVEL': 'INFO'}
+
```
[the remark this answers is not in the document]
petrkalos: Good catch and I already tried it, the problem is that this creates a cycling dependency because you correctly noted that CDK already detects the dependency implicitly.

deploy/stacks/bootstrap_stack.py
petrkalos: I guess this comment goes to `create_function_role`, right? and yes you are absolutely right, I copy/pasted the perms from the api_handler and forgot to checkout the trimmed down version. Will update the commit.

dlpzx: Resolved now!

dlpzx (approved):
looks good!

Issue #1022, opened 2024-01-31
ML Studio - In order to use for example athena in datawrangler several permissions are missing


enr0c:
### Describe the bug

When using Sagemaker Studios on glue tables that one has access to via for example athena directly, inside Sagemaker studios it throws errors. They can be clustered into two areas:

1. List operations for athena: I have identified three missing pieces: athena:ListDataCatalogs, athena:ListDatabases and athena:ListTableMetadata
2. Priviliges on S3 storage location for athenaqueries is missing

### How to Reproduce

1. Create ML Studio
2. Create a data wrangler flow
3. Add athena as connection

### Expected behavior

It should work, respecting the role the user has, as its working for example in worksheets or from a jupyter notebook

### OS

not relevant

### Python version

not relevant

### AWS data.all version

2.2.0

ozggumus-aws:
@dlpzx this is the issue we have been discussed before. Currently we solve this problem by adding the inline policy below to the respective role manually: 

`{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "Statement1",
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::s3bucket/athenaqueries/dataall-role-name/*",
                "arn:aws:s3:::s3bucket/athenaqueries/dataall-role-name/"
            ]
        }
    ]
}`

enr0c:
plus the workaround for the athena: issues

`{
            "Action": [
                "athena:ListDataCatalogs"
            ],
            "Resource": [
                "*"
            ],
            "Effect": "Allow"
        },
        {
            "Action": [
                "athena:ListDatabases",
                "athena:ListTableMetadata"
            ],
            "Resource": [
                "arn:aws:athena:eu-central-1:726955877276:datacatalog/*"
            ],
            "Effect": "Allow"
        },`

dlpzx:
Hi @enr0c we are implementing this feature as part of 2.4

One thing that I want to leave referenced is that the AWS SDK for Pandas read_sql command that we typically use has a parameter called `ctas_approach` that if set to True(default), creates a temporary table in the database. For shared databases users get READ-ONLY permissions, so this type of queries would fail. The solution is to set that parameter to False (link to [docs](https://aws-sdk-pandas.readthedocs.io/en/stable/tutorials/006%20-%20Amazon%20Athena.html)).

```
df = wr.athena.read_sql_table(
    table="table",
[...]

PR #1155, opened 2024-04-09
Add Athena List permissions to use AWS SDK for Pandas in SageMaker

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
Gran `athena:List:*` permissions to environment-team-roles. We could have added the following permissions one by one, but since we are dangerously close to IAM service quotas on managed policies per IAM role this PR grants athena:List:* instead. 

```
{ "Action": [ "athena:ListDataCatalogs" ], "Resource": [ "*" ], "Effect": "Allow" }, { "Action": [ "athena:ListDatabases", "athena:ListTableMetadata" ], "Resource": [ "arn:aws:athena:eu-central-1:<account>:datacatalog/*" ], "Effect": "Allow" },
```

The issue also reports missing S3 permissions when using the SDK for pandas. These issues cannot be reproduced if using `ctas_approach` set to False in the read_sql statement (link to [docs](https://aws-sdk-pandas.readthedocs.io/en/stable/tutorials/006%20-%20Amazon%20Athena.html)). So no additional S3 permissions have been granted.

```
df = wr.athena.read_sql_table(
    table="table",
    database="dataall_DATABASE_shared",
    workgroup= "dataall-GROUP",
    ctas_approach=False
)
```

### Relates
- #1022 

### Security

fetching data from storage outside the application (e.g. a database, an S3 bucket)? `NO`
  - Is the input sanitized?
- Does this PR introduce any functionality or component that requires authorization? `NO`
- Are you using or adding any cryptographic features? `NO`
  - Have you used the least-privilege principle? How? `Yes, the user only gets List permissions to Athena resources`

### DIFF
--- a/backend/dataall/core/environment/cdk/env_role_core_policies/athena.py
+++ b/backend/dataall/core/environment/cdk/env_role_core_policies/athena.py
@@ -14,7 +14,10 @@ class Athena(ServicePolicy):
         statements = [
             iam.PolicyStatement(
                 # sid="ListAthena",
-                actions=['athena:ListWorkGroups', 'athena:ListTagsForResource', 'athena:GetWorkgroup'],
+                actions=[
+                    'athena:List*',
+                    'athena:GetWorkgroup'
+                ],
                 effect=iam.Effect.ALLOW,
                 resources=['*'],
             ),
@@ -23,7 +26,6 @@ class Athena(ServicePolicy):
                 actions=[
                     'athena:Get*',
                     'athena:BatchGet*',
-                    'athena:List*',
                     'athena:StartQueryExecution',
                     'athena:StopQueryExecution',
                     'athena:CreateNamedQuery',


petrkalos (approved):
* The documentation states that when `ctas_approach=False` that nested types are not handled. Do we know if there is a need for that?
* Do we have a long term plan to fix the IAM Role policy limits?

dlpzx:
> * The documentation states that when `ctas_approach=False` that nested types are not handled. Do we know if there is a need for that?
> 
If customers use `unload_approach=True and ctas_approach=False` they can still query nested data, the comment is for both being false. 
>     * Do we have a long term plan to fix the IAM Role policy limits? 
We can discuss offline. When it comes to IAM roles, we already optimized the size of the IAM policies in the way they are splitted. For the long-run we are assessing IAM Identity Center

PR #1159, opened 2024-04-09
fail if ruff (check or format) finds a difference

petrkalos:
### Feature or Bugfix
- Bugfix

### Detail
fail if ruff (check or format) finds a difference


dlpzx (approved):
looks good

Issue #1141, opened 2024-04-03
share revoke should not fail when a role is deleted


zsaltys:
We've ran into multiple situations where our users delete a role which is actively used in data.all. We can detect this (with custom scripts) and data.all will let you revoke the S3 share via UI but it will fail trying to revoke a LakeFormation share with an error like:

`Failed to get role ROLE_NAME due to: An error occurred (NoSuchEntity) when calling the GetRole operation: The role with name ROLE_NAME cannot be found.`

This can only be fixed by manually updating the share to FAILED in RDS and then the share can be deleted via UI. This is very inconvenient because only data.all engineering team can fix such issues rather than making it a responsibility of the share owner.

I suspect that this issue also affects the "share validator" released in 2.3 and that it fails to fix a share when a role is deleted. I propose the following enhancement for this problem:

When attempting to revoke a share detect that a role has been deleted/missing and simply put the share into FAILED mode so that the share can be deleted via UI. This should also happen when share fixer is trying to fix the share either via manual request or when it runs in the background.

This fix is related to this issue https://github.com/data-dot-all/dataall/issues/1029. The above error was observed on 2.2 release.

anushka-singh:
#### Confirming re-apply shares wont work if role was deleted for a successful share:

I had a successful share, then I deleted the consumption role. 
Verify shares, marked share as unhealthy with following errors:
![Screenshot 2024-04-03 at 9 37 09 AM](https://github.com/data-dot-all/dataall/assets/26413731/f4e3f9f7-53f2-4b2e-8a5f-17507f65c5c0)
When I hit re-apply shares, it does work and fails with error:
<img width="1455" alt="Screenshot 2024-04-03 at 9 41 17 AM" src="https://github.com/data-dot-all/dataall/assets/26413731/81b8e230-b7ae-4790-84ca-eaccc0a20278">

SofiaSazonova:
I think, the most logical way to handle this is:
1. Check the existence of the role during the healthcheck. Make it the first important statement in the result.
2. If during the healthcheck the role is missing, dataall must revoke all resource permissions and delete dataall managed policies, the share must become 'failed'
3. Revoke/Reapply share must lead to status 'Share failed', if there is no consumption role. Additionaly dataall must revoke all resource permissions and delete dataall managed policies
@zsaltys @dlpzx @anushka-singh what do you think?

dlpzx:
Completed in #1161

PR #1161, opened 2024-04-10
1141 revoke share fix

SofiaSazonova:
### Feature or Bugfix
- Bugfix

### Detail
- Health check is now started with check if the principal role exists
- If Role is missing:
-- add share item will fail (Share failed) and then the user will be able delete it
-- revoke of the item will be successful ( SHARE_SUCCEEDED) 
- If Role was missing, but then restored with the same name and attached policies (to keep arn), the share can be revoked and applied again successfuly

### Relates
- [1141](https://github.com/data-dot-all/dataall/issues/1141)

### Security

  - Is the input sanitized?


SofiaSazonova:
How to test:

- [ ] Create consumption role for the env and share request for this role
- [ ] After all items are shared successfully, delete IAM role via AWS Console
- [ ] Check share health. Shares must become unhealthy with the message 'Principal Role is not found'
- [ ] Revoke shares and delete them
- [ ] Delete share request and consumption role
- [ ] Create another consumption role
- [ ] Create share for this new role
- [ ] Share all items,  they must be shared successfully

backend/dataall/modules/dataset_sharing/aws/kms_client.py
```diff
@@ -1,10 +1,38 @@
...
+        if statement.get('Sid', 'no-sid') in [
+            DATAALL_BUCKET_KMS_DECRYPT_SID,
+            DATAALL_KMS_PIVOT_ROLE_PERMISSIONS_SID,
+            DATAALL_ACCESS_POINT_KMS_DECRYPT_SID,
+        ]:
+            principal_list = statement['Principal']['AWS']
+            if isinstance(principal_list, str):
+                principal_list = [principal_list]
+            new_principal_list = principal_list[:]
+            for p_id in principal_list:
```
dlpzx, line 28: I don't understand how the malformed principals are removed here. `AROA` principals are the IAM role canonical names, that does not mean that are "bad principals" right?

backend/dataall/modules/dataset_sharing/aws/s3_client.py
```diff
@@ -112,13 +119,37 @@ def generate_default_bucket_policy(s3_bucket_name: str):
         return policy
```
dlpzx, line 121: Same comment as with KMS, this is not an AWS boto3 call, seems more suitable to place it in the share managers in the service layer

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -111,50 +111,70 @@ def approve_share(cls, engine: Engine, share_uri: str) -> bool:
                     share_object_SM.update_state(session, share, new_object_state)
                     return False
```
dlpzx, line 113: Maybe we can place the verification of the principal role before the dataset lock. In line 79 `lock_acquired = cls.acquire_lock_with_retry(dataset.datasetUri, session, share.shareUri)` we might be waiting and locking the dataset in vain for a share that is going to fail either way

backend/dataall/modules/dataset_sharing/services/share_object_service.py
```diff
@@ -42,12 +44,33 @@
...
+        if not principal_role:
+            for item in ShareObjectRepository.get_all_sharable_items(session, share.shareUri):
+                ShareObjectRepository.update_share_item_health_status(
+                    session,
+                    share_item=item,
+                    healthStatus=ShareItemHealthStatus.Unhealthy.value,
+                    healthMessage=f'Share principal role {role_name} not found.',
+                    timestamp=datetime.now(),
+                )
+            return False
```
dlpzx, line 71: This function is doing 2 things: it is verifying the principal IAM role and then it is updating the health status, this second part might be problematic. Health-statuses are only changed in the verification workflows of the item. If we start adding changes in other parts of the code, it will soon start be difficult to follow (very similar to what happened with the regular share workflows in the past).

backend/dataall/modules/dataset_sharing/services/share_object_service.py
```diff
@@ -42,12 +44,33 @@
...
 log = logging.getLogger(__name__)
 
 
 class ShareObjectService:
+    @staticmethod
+    def verify_principal_role(session, share: ShareObject) -> bool:
+        role_name = share.principalIAMRoleName
+        account_id = SessionHelper.get_account()
+        delegation_role = SessionHelper.get_delegation_role_arn(account_id)
+        principal_role = IAM.get_role_arn_by_name(account_id, role_name, delegation_role)
```
dlpzx, line 61: We do not need to send the delegation_role as it defaults to it

backend/dataall/modules/dataset_sharing/services/share_object_service.py
dlpzx: With that in mind, I would break it down and only do the verification of the "does IAM role exist?" here. Which leads to the idea of, do we even need to make it a service function of sharing? We already have `base/aws.iam.IAM.get_role_by_name` that we can call directly, and as explained before it is just a:

```
environment = EnvironmentService.get_environment_by_uri(session, share.environmentUri)
principal_role = IAM.get_role_arn_by_name(
                account=environment.AwsAccountId,
                region=environment.region,
                role_name=share.principalIAMRoleName
                )
```

dlpzx:
Let me first confirm my understanding on how this PR interacts with the current workflows. For data sharing, the Share.Status and ShareItem.Status should move in this statemachine:
image

- `AddItem` = No changes, it is alright if the users add items, it is just info in RDS in `pending state`
- `Submit` => Should we allow requesters to submit a request when the IAM role does not exist?
- `Approve` => Should we allow approvers to approve a request when the IAM role does not exist?
- `Reject` => No changes, it is alright if the users reject, it is just info in RDS in `rejected`
- One the share ECS task `starts` => yes, as first item it should check if the principal IAM role exists:
      -  if the task is a `approve_share` task and the IAM role does not exist => treat is as another `Failure` (more or less what is implemented at the moment)
      - 👀  if the task is a `revoke_share` => clean-up the rest of the share and ensure the revoke completes => This is the most important part of the PR. If needed we could just implement this item. 👀 
- `RevokeItems` = No changes, it is alright if the users revokes items, data.all must ensure the share revoke won't fail, same for `DeleteItems`

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -448,14 +448,18 @@ def grant_dataset_bucket_key_policy(self):
...
-    def delete_target_role_bucket_policy(self):
+    def delete_target_role_bucket_policy(self, principal_exist):
         logger.info(f'Deleting target role from bucket policy for bucket {self.bucket_name}...')
         try:
             s3_client = S3Client(self.source_account_id, self.source_environment.region)
             bucket_policy = json.loads(s3_client.get_bucket_policy(self.bucket_name))
-            target_requester_arn = IAM.get_role_arn_by_name(
-                self.target_account_id, self.target_environment.region, self.target_requester_IAMRoleName
-            )
+            if principal_exist:
```
dlpzx, line 456: I don't think we need to pass the `principal_exist`, we can just use the function `IAM.get_role_arn_by_name` if it returns `None` then the principal does not exist. Like what is implemented in line 549

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -26,10 +35,6 @@
 logger = logging.getLogger(__name__)
 ACCESS_POINT_CREATION_TIME = 30
 ACCESS_POINT_CREATION_RETRIES = 5
-DATAALL_ALLOW_OWNER_SID = 'AllowAllToAdmin'
-DATAALL_ACCESS_POINT_KMS_DECRYPT_SID = 'DataAll-Access-Point-KMS-Decrypt'
-DATAALL_KMS_PIVOT_ROLE_PERMISSIONS_SID = 'KMSPivotRolePermissions'
-DATAALL_DELEGATE_TO_ACCESS_POINT = 'DelegateAccessToAccessPoint'
```
dlpzx, line 38: I think there are a couple of places where a share revoke could fail here. For example in `delete_dataset_bucket_policy` we call:
```
target_requester_arn = IAM.get_role_arn_by_name(
            self.target_account_id, self.target_environment.region, self.target_requester_IAMRoleName
        )
```
Which won't return anything, I would verify all calls to ` IAM.get_role_arn_by_name` in the data sharing functions

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
```diff
@@ -129,6 +129,7 @@ def process_revoked_shares(
         target_environment: Environment,
         source_env_group: EnvironmentGroup,
         env_group: EnvironmentGroup,
+        principal_exist: bool,
```
dlpzx, line 132: Instead of passing the `principal_exist` variable I would check directly where it is used, so internally in `revoke_target_role_access_policy`

dlpzx: Checking `revoke_target_role_access_policy` I realize that even if the role does not exist, the policy will exist and we need to clean it up

backend/dataall/modules/dataset_sharing/services/share_processors/s3_bucket_process_share.py
```diff
@@ -88,6 +89,8 @@ def process_approved_shares(
                 env_group,
             )
             try:
+                if not ShareObjectService.verify_principal_role(session, share):
```
dlpzx, line 92: Minor. The exception message is a bit misleading. We can only revoke it has been approved. For users it is impossible to revoke this share

backend/dataall/modules/dataset_sharing/services/share_processors/s3_bucket_process_share.py
```diff
@@ -123,6 +126,7 @@ def process_revoked_shares(
         target_environment: Environment,
         source_env_group: EnvironmentGroup,
         env_group: EnvironmentGroup,
+        principal_exist: bool,
```
dlpzx, line 129: Similar to the comment in the access point shares, I suggest we move the parameter calculation closer to where it is actually used instead of passing it from the top. (check if principal exists directly inside `delete_target_role_bucket_policy`)

backend/dataall/modules/dataset_sharing/aws/kms_client.py
SofiaSazonova: For these policies everywhere in dataall we use Arns to identify the principal. 
But if the principal is deleted, automatically (i think, it's IAM behaviour) those Arns are replaced by RoleIds.  So, if in data-all created S3 (and KMS) policies you encounter RoleId instead of Arn, it 99% percent means, that the role was deleted.

Of course, to be 100% sure, In case of Malformed policy exception, I need to check all principals by their IDs. Unfortunately boto3 has no instrument to list roles by RoleId (only aws cli has). So, in order to make thinfs easier, I made it that way.

backend/dataall/modules/dataset_sharing/aws/kms_client.py
```diff
@@ -1,10 +1,38 @@
...
 
 from dataall.base.aws.sts import SessionHelper
 from botocore.exceptions import ClientError
 
 log = logging.getLogger(__name__)
 
+DATAALL_BUCKET_KMS_DECRYPT_SID = 'DataAll-Bucket-KMS-Decrypt'
+DATAALL_KMS_PIVOT_ROLE_PERMISSIONS_SID = 'KMSPivotRolePermissions'
+DATAALL_ACCESS_POINT_KMS_DECRYPT_SID = 'DataAll-Access-Point-KMS-Decrypt'
+
```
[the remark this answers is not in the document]
SofiaSazonova: I thought about it. 
BUT
1. This method is only used here.
2. We call create_policy in many places, so, it's easy to forget to call _remove_malformed_principal before it. Here, on the other hand, it makes sense: if we have ClientError, we try to fix it.

I don't insist, but I think, it's more elegant to leave it here.

SofiaSazonova:
@dlpzx 
> * `Submit` => Should we allow requesters to submit a request when the IAM role does not exist?
> * `Approve` => Should we allow approvers to approve a request when the IAM role does not exist?

Good point, we can terminate it right there. 

All other cases are already covered in PR.

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
SofiaSazonova: Here it doesn't matter if it is None or not. The further actions are the same

SofiaSazonova: Issue adressed in [30b9d65](https://github.com/data-dot-all/dataall/pull/1161/commits/30b9d65b21969c27d2cea993bfd0bf2a86fa360d)

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
SofiaSazonova: Well, surprisingly, if I delete role via console, all policies are also deleted with it.
I am not sure, that we will be able to delete this role and keep policies

SofiaSazonova: Anyway, it won't hurt to check

backend/dataall/modules/dataset_sharing/services/share_processors/s3_bucket_process_share.py
SofiaSazonova: adressed in [b60f12d](https://github.com/data-dot-all/dataall/pull/1161/commits/b60f12d84c0cef198310dc6b022753d925816b7c)

backend/dataall/modules/dataset_sharing/services/share_processors/s3_access_point_process_share.py
SofiaSazonova: adressed in [b60f12d](https://github.com/data-dot-all/dataall/pull/1161/commits/b60f12d84c0cef198310dc6b022753d925816b7c)

backend/dataall/modules/dataset_sharing/services/share_managers/s3_bucket_share_manager.py
SofiaSazonova: adressed in [b60f12d](https://github.com/data-dot-all/dataall/pull/1161/commits/b60f12d84c0cef198310dc6b022753d925816b7c)

backend/dataall/modules/dataset_sharing/services/share_object_service.py
SofiaSazonova: adressed in [6e5951e](https://github.com/data-dot-all/dataall/pull/1161/commits/6e5951e7a849cdb0230f27379d3523fe2ace96b8)

backend/dataall/modules/dataset_sharing/services/share_object_service.py
SofiaSazonova: adressed in [e6d8](https://github.com/data-dot-all/dataall/pull/1161/commits/127e6d87e4368ce0c8e761f0ac7a66ab6ae04da9)

backend/dataall/modules/dataset_sharing/services/share_processors/s3_bucket_process_share.py
SofiaSazonova: addressed in [d216e9a](https://github.com/data-dot-all/dataall/pull/1161/commits/d216e9a3cc8e9024b759afadae86d6a20145582b)

backend/dataall/modules/dataset_sharing/aws/kms_client.py
dlpzx: Initially we used to write principals with roleIds, but we moved to arns as they are easier to read and understand by end-users. That's why I thought you were changing this. Because we used to work with role ids a lot, there are/were methods to get the role ids directly in the code. You can read the role id from a role by calling the IAM boto3 client:

```
    def get_role_id(accountid, name):
        session = SessionHelper.remote_session(accountid=accountid)
        client = session.client('iam')
        try:
            response = client.get_role(RoleName=name)
            return response['Role']['RoleId']
        except ClientError:
            return None
```
I think we should not clean-up all role ids because the majority of customers use "imported" Buckets and KMS keys and we can wipe out permissions set outside of data.all.

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
dlpzx: The lock is for the dataset resources, we want to prevent that the Bucket and KMS policies being updated during a share task are being updated simultaneously by another share task. A share item is unique for each share, there is no way another share is simultaneously updating it, plus RDS supports ACID transactions no?

backend/dataall/modules/dataset_sharing/aws/kms_client.py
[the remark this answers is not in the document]
dlpzx: Let me have a look at how the code looks after changes because it might be good now. I would be strict with the definition of layers, because we can end up mixing responsibilities and having a code that are just exceptions to the rule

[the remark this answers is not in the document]
dlpzx: So, we are calling `put_key_policy` and that's when in case of MalformedPolicy we remove the malformed principals. We are doing all the policy manipulation in the s3_X_share_managers except for this piece that is a bit hidden in the aws layer. I propose to go for:
- kms client = put_key_policy that just calls boto3 KMS put_key_policy and rasies errors

There are 4 places (+tests) where we call that function, we could insert there the logic of put_policy -> raise exception -> remove malformed -> try again to put_policy. e.g. `dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py:723`

```
                try:
                    kms_client.put_key_policy(kms_key_id, json.dumps(existing_policy))
                except 'MalformedPolicyDocumentException': #catch error better
                    clean_policy = self._remove_malformed_principals(existing_policy)
                    kms_client.put_key_policy(kms_key_id, json.dumps(clean_policy))
```

backend/dataall/modules/dataset_sharing/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -525,6 +531,13 @@ def update_dataset_bucket_key_policy(self):
         target_requester_arn = IAM.get_role_arn_by_name(
             self.target_account_id, self.target_environment.region, self.target_requester_IAMRoleName
         )
+
+        if not target_requester_arn:
+            raise PrincipalRoleNotFound(
+                'update dataset bucket key policy',
+                f'Principal role {self.target_requester_IAMRoleName} is not found. Failed to update MS key policy',
```
dlpzx, line 538: typo: KMS key policy

backend/dataall/modules/dataset_sharing/services/share_processors/s3_bucket_process_share.py
```diff
@@ -88,6 +90,11 @@ def process_approved_shares(
                 env_group,
             )
             try:
```
dlpzx, line 92: Why do we add this check in the buckets `process_approved_shares` but we do not do it for access points or tables? I would make it fail before triggering the ECS task, like the next steps of adding a check in the approve of the share

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -554,6 +585,17 @@ def reapply_share(cls, engine: Engine, share_uri: str):
...
+            # If principal role doesn't exist, all share items are unhealthy, no use of further checks
+            if not ShareObjectService.verify_principal_role(session, share):
+                ShareObjectService.update_all_share_items_status(
+                    session,
+                    share_uri,
+                    ShareItemHealthStatus.Unhealthy,
+                    f'Share principal Role {share.principalIAMRoleName} is not found.',
+                )
+                return False
+
```
dlpzx, line 598: We cannot update all share items `HealthStatus`  to `Unhealthy`, we can only update those whose `Status` is in `Share_Succeed`. Those items in other sharing states (e.g. `PendingApproval` do not have a share to be healthy or unhealthy)

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -454,6 +474,17 @@ def verify_share(
                 session, share_uri, status=status, healthStatus=healthStatus
             )
 
+        log.info(f'Verifying principal IAM Role {share.principalIAMRoleName}')
+        # If principal role doesn't exist, all share items are unhealthy, no use of further checks
+        if not ShareObjectService.verify_principal_role(session, share):
```
dlpzx, line 479: IMPORTANT: In approve shares and revoke shares we are procssing share requests, we are not verifying the healthstatus (check the big comment with the picture). In approve_share, if the role is not present, the `Status` of the item should go to failure. We do not manage the unhealthyness of a share, because it has not happened yet. We could argue that it is our chance to update the customer and tell them that OTHER share items already in `Share_Succeed` are unhealthy

dlpzx (changes requested):
Some comments and one important note on what is the shareStatus, healthStatus lifecycle. The big comment with the picture contains all the details

backend/dataall/modules/dataset_sharing/aws/kms_client.py
SofiaSazonova: Another option: store consumption role IDs in DB. What do you think?

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
SofiaSazonova: Update Health message and status is our only way to show the user in UI, WHY did share fail. Otherwise it will be unclear.

backend/dataall/modules/dataset_sharing/aws/kms_client.py
SofiaSazonova: On the other hand, there can be several deleted roles at once, so it's better to wipe them all by checking principal ID.

backend/dataall/modules/dataset_sharing/aws/s3_client.py
```diff
@@ -116,9 +126,11 @@ class S3Client:
     def __init__(self, account_id, region):
         session = SessionHelper.remote_session(accountid=account_id, region=region)
         self._client = session.client('s3', region_name=region)
+        self.region = region
         self._account_id = account_id
```
dlpzx, line 131: This comment is outdated

backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
```diff
@@ -820,6 +820,26 @@ def get_share_data(session, share_uri):
             target_environment,
         )
 
+    @staticmethod
+    def get_all_sharable_items(session, share_uri, status=None, healthStatus=None):
```
dlpzx, line 824: small typo: shareable

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -76,85 +76,107 @@ def approve_share(cls, engine: Engine, share_uri: str) -> bool:
...
+                        session,
+                        dataset,
+                        share,
+                        shared_tables,
+                        source_environment,
+                        target_environment,
+                        env_group,
+                    ).process_approved_shares()
+                    log.info(f'sharing tables succeeded = {approved_tables_succeed}')
+
```
dlpzx, line 157: Very minor: principal_healthy is a weird variable name to represent the success of the shares

backend/dataall/modules/dataset_sharing/services/share_object_service.py
```diff
@@ -42,12 +44,36 @@
...
+        principal_role = IAM.get_role_arn_by_name(account_id=env.AwsAccountId, region=env.region, role_name=role_name)
+        return principal_role is not None
+
+    @staticmethod
+    def update_all_share_items_status(
+        session, shareUri, new_health_status: str, message, previous_health_status: str = None
+    ):
+        for item in ShareObjectRepository.get_all_sharable_items(
+            session, shareUri, healthStatus=previous_health_status
+        ):
```
dlpzx, line 68: We should only sent to Unhealthy those items that are currently shared. If an item is in PendingState status it should not be in any state yet

backend/dataall/modules/dataset_sharing/services/share_processors/s3_bucket_process_share.py
dlpzx: ^^

dlpzx (commented):
Some additional comments

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
dlpzx: After discussions offline, we have decided to implement better UX for share failed errors in https://github.com/data-dot-all/dataall/issues/1168

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -76,85 +76,107 @@ def approve_share(cls, engine: Engine, share_uri: str) -> bool:
...
-                            share_uri,
-                            folder.locationUri,
-                        )
-                        cls.handle_share_items_failure_during_locking(
-                            session, share_item, ShareItemStatus.Share_Approved.value
-                        )
+                log.info(f'Verifying principal IAM Role {share.principalIAMRoleName}')
+                share_successfull = ShareObjectService.verify_principal_role(session, share)
+                # If principal role doesn't exist, all share items are unhealthy, no use of further checks
+                if share_successfull:
```
dlpzx, line 82: typo: successful

backend/dataall/modules/dataset_sharing/services/data_sharing_service.py
```diff
@@ -554,6 +582,11 @@ def reapply_share(cls, engine: Engine, share_uri: str):
                         )
                     return False
 
+            log.info(f'Verifying principal IAM Role {share.principalIAMRoleName}')
```
dlpzx, line 585: Move check in line 587 before the dataset_lock logic --> to line 553

dlpzx:
I think we need to add some test cases

- [ ] Create consumption role for the env and share request for this role ---> all items are shared successfully
- [ ] Now, delete the IAM role and click verify items and verify some of the items. ---> The verified items only must become unhealthy with the message 'Principal Role is not found'
- [ ] Revoke all items from the share ---> the unhealthy as well as the healthy items must be revoked successfully
       - [ ] Removed from KMS policy
       - [ ] Removed from S3 policy
       - [ ] Managed IAM policies deleted
- [ ] Re-add the items and submit the share request --> When share is submitted it should throw an error
- [ ] Re-create the consumption role, re-add the items, submit the request, delete the consumption role and the approvers click on approve --> When share is approved, it should throw an error

SofiaSazonova:
all these tests were made successfully)

dlpzx (approved):
Looks great! Thanks!

Issue #373, opened 2023-03-15
Inviting Team to Organization Sliders Disabled


noah-paige:
### Describe the bug

When an organization admin is inviting a team to the organization there are 2 slides for Organization Permissions that the invite can define:
1. 'Link environments to this organization' permissions
2. 'Invite teams to this organization' permissions

Both of these sliders are enabled by default and are not editable. 

When a team is invited to an org it has the permissions to link environments (1 above) but is not able to invite additional teams (2 above).

### How to Reproduce

```
*P.S. Please do not attach files as it's considered a security risk. Add code snippets directly in the message body as much as possible.*
```

### OS

N/A

### Python version

N/A

### AWS data.all version

1.4

dlpzx:
Hi @noah-paige, this is the intended behavior. If a team is invited to the Organization it means that it can perform all actions at the Organization-level = link environments and invite other teams to the orga. To disable this permissions you just need to remove the team from the Organization. 

Maybe the confusion came because the organization groups multiple environments and it can be assumed that "in order to work in environments I need to be invited to the organization" but that is NOT true. Any team can be invited to an environment without the need of it to be in the Organization-invited-teams.

We can work on UI enhancements to avoid this misunderstanding, so happy to hear your thoughts on how you would picture the whole user experience

noah-paige:
Hi @dlpzx - the above makes sense to me but when I am testing locally and inviting another team to my Organization that invited team cannot then invite additional teams to the org.

Also, I think some confusion from my end is coming because on the frontend when I try to invite a team to my organization the sliders shown below are not editable and I am not seeing the second slider (Invite Teams) permission to be granted

Is this the designed behavior and/or are you able to reproduce what I am seeing? 
 
![Screenshot 2023-03-16 at 12 25 43 PM](https://user-images.githubusercontent.com/69586985/225686872-f51804f6-f7a3-466e-8e99-bfc07369ed97.png)

PR #1162, opened 2024-04-10
Invite to Orginzation permissions

SofiaSazonova:
### Feature or Bugfix
- Bugfix

### Detail
- Add permissions to add and remove groups to Organization into ORGANIZATION_INVITED permission set
⚠️ ⚠️ ⚠️ IMPORTANT ⚠️ ⚠️ ⚠️
Apply migrations in order to receive new permissions.

### Relates
- [373](https://github.com/data-dot-all/dataall/issues/373)

### Security

  - Is the input sanitized?


dlpzx:
Is there not a way we can grant those permissions in the migration? I don't think it is scalable that users need to re-add groups to organizations

backend/dataall/core/permissions/services/organization_permissions.py
```diff
@@ -18,4 +18,4 @@
     INVITE_ORGANIZATION_GROUP,
     REMOVE_ORGANIZATION_GROUP,
 ]
```
dlpzx, line 20: `REMOVE_ORGANIZATION_GROUP,` is not used, I would avoid dead code

SofiaSazonova: it's used in `organization_service.py` , line 210
`    @staticmethod
    @TenantPolicyService.has_tenant_permission(MANAGE_ORGANIZATIONS)
    @ResourcePolicyService.has_resource_permission(REMOVE_ORGANIZATION_GROUP)
    def remove_group(uri, group):`

If I accidentally added the wrong group I would want to delete it as well.

dlpzx: I meant, we can use INVITE_GROUP, but it is ok I guess to have 2 permissions. I did not want to have this permission added here because we might want to introduce `REMOVE_ORGANIZATION_GROUP` relative to groupUri not organizationUri on the future

SofiaSazonova: I can do the following
1. Rename INVITE_ORGANIZATION_GROUP -> MANAGE_ORGANIZATION_GROUP
2. Make database migration for it
3. Use it for both add and remove groups

backend/migrations/versions/9dcc2f2b8858_add_organization_group_permissions.py
```diff
@@ -0,0 +1,46 @@
...
+branch_labels = None
+depends_on = None
+
+
+def get_session():
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+    return session
+
+
```
dlpzx, line 30: I am testing it at the moment but the issue might be that the permission does not exist yet, we cannot attach `INVITE_ORGANIZATION_GROUP, REMOVE_ORGANIZATION_GROUP` because they do not exist in RDS. Let me check

dlpzx: This would be solved by adding a call to `PermissionService.init_permissions(session)`

dlpzx: I just tested it and it fails with the error `An error occurred (ResourceNotFound):
            Resource Permission with URI INVITE_ORGANIZATION_GROUP not found` --->
Here is a version that I have tested and it does the trick :) 
```
def upgrade():
    session = get_session()
    print(f"Adding organization group permissions...")
    invite_permission = PermissionService.save_permission(
        session=session,
        name=INVITE_ORGANIZATION_GROUP,
        description='INVITE_ORGANIZATION_GROUP',
        permission_type=PermissionType.RESOURCE.name,
    )
    remove_permission = PermissionService.save_permission(
        session=session,
        name=REMOVE_ORGANIZATION_GROUP,
        description='REMOVE_ORGANIZATION_GROUP',
        permission_type=PermissionType.RESOURCE.name,
    )
    all_org_groups = session.query(OrganizationGroup).all()
    for org_group in all_org_groups:
        ResourcePolicyService.attach_resource_policy(
            session=session,
            group=org_group.groupUri,
            resource_uri=org_group.organizationUri,
            permissions=[INVITE_ORGANIZATION_GROUP, REMOVE_ORGANIZATION_GROUP],
            resource_type=Organization.__name__,
        )
```

SofiaSazonova: Ahhha, that's exactly the type of error I was concerned about with 'unclean' deployment)

dlpzx (approved):

Issue #804, opened 2023-10-16
Share Requestor can not view Dataset folders in UI even after share is complete


noah-paige:
### Describe the bug

When UserA is shared a dataset with all of its folders and clicks on the shared Dataset from the Dataset view

* UserA can not click on more information for the folders, receives the following error instead

 An error occurred (UnauthorizedOperation) when calling LIST_DATASET_FOLDERS operation: `User: XXXXX is not authorized to perform: LIST_DATASET_FOLDERS on resource: uri_here`

### How to Reproduce

* UserA requests access to a dataset including all of its tables and folders
* Dataset Owner approves the share request and share gets successfully processed
* UserA navigates to the Datasets Tab of the shared dataset and clicks on one of the folders it has been shared

### Expected behavior

UserA should be able to click on the Folder and receive some additional information related to the information (whether it should be the same amount of information as Dataset owner or a subset of that information is TBD)

### OS

N/A

### Python version

N/A

### AWS data.all version

2.0

dlpzx:
I also noticed that in the table service the `DatasetTableService.get_table` should also be decorated with GET_TABLE permissions

dlpzx:
Completed in #1163

PR #1163, opened 2024-04-11
Fix missing GET_FOLDER permissions

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
- Adds a new type of RESOURCE permission: `GET_DATASET_FOLDER`
- check `GET_DATASET_FOLDER` against the folder when we execute the `get_storage_location` service function
- attach `GET_DATASET_FOLDER` permission to the dataset owner and stewards when the folder is created
- remove `GET_DATASET_FOLDER` permissions to the dataset owner and stewards when the folder is deleted
- attach `GET_DATASET_FOLDER` permissions to the share requester group when the folder is shared
- remove `GET_DATASET_FOLDER` permissions to the share requester group when the folder is revoked
- Migration script that creates the new permission type and creates records backfilling the existing datasets and shares corresponding permissions.

NEXT STEPS: 
- ⚠️ This PR only grants TABLE and FOLDER READ permissions to share requesters when they make a share as a GROUP. It does not grant permissions to the group if the share request is done on behalf of a CONSUMPTION ROLE. If we want to implement this feature, which we have to decide, we can do it in a separate PR --> Issue https://github.com/data-dot-all/dataall/issues/1174 and Issue https://github.com/data-dot-all/dataall/issues/1173
- get_table service method is not decorated --> Issue https://github.com/data-dot-all/dataall/issues/1171

Testing (locally)
- [X] Create folder and as dataset owner check folder details
- [X] Request access to folder and as requester check folder details
- [X] Revoke the access, the group cannot see the folder details anymore 
- [X] Delete folder and Check database - permissions to folder are deleted
- [X] Request access to folder with a consumption role - the requester group does not have access to the folder 

Backwards compatibility (checked in database as well)
- [X] After upgrading with migration script, the dataset owner can check folder details
- [X] and the share request owner can check the folder details

image

### Relates
- #804 

### Security

  - Is the input sanitized?


backend/dataall/modules/dataset_sharing/services/share_item_service.py
```diff
@@ -101,16 +102,10 @@ def revoke_items_share_object(uri, revoked_uris):
...
-                revoke_table_items = ShareObjectRepository.find_all_share_items(
-                    session, uri, ShareableType.Table.value, [ShareItemStatus.Revoke_Approved.value]
-                )
-                for item in revoke_table_items:
-                    ResourcePolicyService.delete_resource_policy(
-                        session=session,
-                        group=share.groupUri,
-                        resource_uri=item.itemUri,
-                    )
+            if share.groupUri != dataset.SamlAdminGroupName and share.principalType == PrincipalType.Group.value:
```
noah-paige, line 105: something to consider is should this permission removal be handled here or AFTER revoke is successful (i.e. at the end of share manager task)? 

For the case when revoke does fail and in a "limbo" state where in this case user would not be able to get the folder or table from data.all UI but the share could potentially still be functional

... something to consider - can be a separate PR

backend/dataall/modules/dataset_sharing/services/share_item_service.py
```diff
@@ -101,16 +102,10 @@ def revoke_items_share_object(uri, revoked_uris):
...
+                log.info('Deleting TABLE/FOLDER READ permissions...')
+                ShareItemService._delete_dataset_table_read_permission(session, share)
```
noah-paige, line 107: For tables - I believe there may be a subset of table read permissions associate to CR shares that will never get deleted now  (maybe something we should handle in migration script?)

backend/dataall/modules/datasets/services/dataset_table_service.py
```diff
@@ -164,11 +164,11 @@ def sync_existing_tables(session, dataset_uri, glue_tables=None):
...
     @staticmethod
     def _attach_dataset_table_permission(session, dataset: Dataset, table_uri):
-        # ADD DATASET TABLE PERMISSIONS
-        env = EnvironmentService.get_environment_by_uri(session, dataset.environmentUri)
+        """
+        Attach Table permissions to dataset groups
+        """
         permission_group = {
             dataset.SamlAdminGroupName,
-            env.SamlGroupName,
```
noah-paige, line 171: I see we are removing env admins from getting table permissions for any new dataset tables - but do we clean up the existing permissions for previous tables created?

UPDATE: I actually see we are not cleaning up permissions for tables (maybe a separate PR should address this?). Looking at the way we handle "deleting" tables:

- If `sync()` determines table no longer exists in glue - we just update the table status to `Deleted` and do not show on UI (table item remains along with associated permissions 
  - OK, but could be made better
  - If table is shared a bunch and then sync determines it no longer exists and it disappears it is not straightforward
  - Maybe should keep on UI with a `Deprecated` or `Deleted` Flag telling user to delete table and all associated shares?

- If user deletes table from UI - we are deleting table if no shares exists on that table but not deleting the associated permissions
  - Not as OK - should be cleaning up orphaned permissions

backend/dataall/modules/dataset_sharing/services/share_item_service.py
dlpzx: That is what I thought, but surprise surprise there is only one permission that gets created for each group. If a group creates shares with 3 CRs there will only be one permission

backend/dataall/modules/dataset_sharing/services/share_item_service.py
noah-paige: makes sense think this is fine as is

backend/dataall/modules/dataset_sharing/services/share_item_service.py
noah-paige: Okay I see - so for all new CR shares the share requestor Group will not `DATASET_TABLE_READ` (unless they also have access to the table) 

But if there were pre-existing table item shares the  requestor Group would have `DATASET_TABLE_READ`. 

I guess I am just  wondering if we should enforce it one way or the other - but the imapct is minimal i believe

backend/dataall/modules/datasets/services/dataset_table_service.py
noah-paige: New Issue Created - https://github.com/data-dot-all/dataall/issues/1171

backend/dataall/modules/dataset_sharing/services/share_item_service.py
noah-paige: After doing some testing - I also see the same can be the case of an approval failure (i.e. `Share_Failed), where:

- Share Approved
- Requestor Group gets Permissions entry in DB for DatasetStorageLocation
- Share Failed --> Only option is to delete share
- Now if approver tries to approve again and all works this is fine (no dups) but at this moment the permission entry is orphaned
  - Requestor Group can still get folder even though Share_Failed

noah-paige:
Did some testing locally on this PR as well including:

- [x] Backfill Folder permissions
- [x] Add/Delete New Folders + Associated Permissions
- [x] Test access folders Dataset Owner
- [x] Folder Share + Associated Permissions and Revoke

Just have one pending comment from code review curious to hear thoughts on @dlpzx

backend/dataall/modules/dataset_sharing/services/share_item_service.py
dlpzx: Yes, I think we need to do some thinking on 1. lifecycle of "SHARE" permissions 2. duplicates on share permissions. I did not want to add anything to this PR to reduce the scope of the testing. I have created [one issue](https://github.com/data-dot-all/dataall/issues/1173) for the lifecycle of permissions and [another issue](https://github.com/data-dot-all/dataall/issues/1174) for the duplicates coming form consumption roles or groups in multiple environments.

noah-paige:
Just need to resolve conflict - then can approve this PR and can use other issues to address additional fixes/enhancements

noah-paige (approved):
looks good!

Issue #563, opened 2023-07-11
HealthOmics module in "Play" tools


dlpzx:
I want to use HealthOmics in a more ready-to-use solution and use it alongside data.all as a new "Play" tool

**Describe alternatives you've considered**
Credits to @ironspur5 that started the design drafts. I added some comments here:

![image](https://github.com/awslabs/aws-dataall/assets/71252798/cf07d6c9-9b05-4d36-855e-74a1fbc1f4e7)

**Additional context**
It is designed to be part of the modularization repo structure

dlpzx:
@ironspur5 I just rebased the branch and did some changes to it that will help you in implementing the above design

zsaltys:
@dlpzx could we add more information here what is Omics, what problem it helps solve etc?

ironspur5:
Hello @zsaltys Omics, now called HealthOmics, is an AWS service that helps users such as bioinformaticians and/or researchers to process and generate insights from genomics and other biological data stored in the cloud. Raw genomic data can be sent to a HealthOmics managed workflow and then stored in a manageable format for querying/visualizing/sharing. For health/life sciences Data.all users, this would a welcome feature, similar to the existing integration with SageMaker Jupyter notebooks.

PR #1177, opened 2024-04-15
run db migrations in a custom resource

petrkalos:
### Feature or Bugfix
- Bugfix

### Detail
Currently the DB is being initialised in the DBMigrations pipeline stage (using CodeBuild) which runs after the BackendStage.
The SavePermissions TriggerFunction runs during the deployment of Backend, just after deployment of the DBCluster.
As a result on clean deployments the SavePermissions step will fail because the DB is uninitialized.

To resolve that in this PR we do the following
* remove DBMigrations stage based on CodeBuild
* run DBMigrations as part of a TriggerFunction/CustomResource after the DB deployment
* run SavePermissions TriggerFunction after the DBMigrations TriggerFunction  

### Security

  - Is the input sanitized?


backend/trigger_handler.py
```diff
@@ -2,16 +2,20 @@
...
 from dataall.base.loader import load_modules, ImportMode
 from dataall.core.permissions.services.tenant_policy_service import TenantPolicyService
 
-logger = logging.getLogger()
-logger.setLevel(os.environ.get('LOG_LEVEL', 'INFO'))
-log = logging.getLogger(__name__)
+
+def db_migrations(event, context) -> None:
+    alembic_cfg = Config('alembic.ini')
+    alembic_cfg.set_main_option('script_location', './migrations')
```
SofiaSazonova, line 17: Why `script_location` is set programatically here and not loaded from .ini?

dlpzx: If I understand this correctly, we are replacing the CLI commands by alembic classes (Config) and packages (Command). So lines 16 and 18 should be equivalent to `'alembic -c backend/alembic.ini upgrade head',` but then line 17 that Sofia pointed out... is it there the same way `'export PYTHONPATH=backend',` was in the CodeBuild stage?

dlpzx (approved):
Looks good. There is still one comment to review. And we should verify that the CDK execution role policy that we provide is able to deploy data.all with this new piece of infra

backend/trigger_handler.py
petrkalos: The script location is loaded from the ini I overwrite it here.

That's because in the programmatic invocation we call alembic from within the Docker container. When we creating the image we copy the `./backend/*` into `/app` so the path is now `/app/migrations/...` not `/app/backend/migrations/...`.

I didn't change the `.ini` because I wanted to allow devs to invoke alembic manually as they currently do today (via the `Makefile`)

petrkalos:
> And we should verify that the CDK execution role policy that we provide is able to deploy data.all with this new piece of infra

My dev pipeline managed to deploy everything successfully is that enough evidence?

dlpzx:
While developing #563 in our dev environments in AWS, we run into the scenario of what would happen if the migration scripts fails. If there is anything wrong with the migration scripts, the custom resource Lambda fails, which results in a `UPDATE_FAILED` for the whole backend stack. We verified that the stack goes to `UPDATE_ROLLBACK_COMPLETE` without issues and that the failed stage in CodePipeline can be retried.

PR #1181, opened 2024-04-16
Refactoring env stack part4

SofiaSazonova:
### Feature or Bugfix
- Refactoring

### Detail
- detangle StackRepository & StackService & EnvironmentService
- Move logic from resolvers and repositories to services
- Remove unused params

### Relates
- #741 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/core/stacks/api/resolvers.py
+++ b/backend/dataall/core/stacks/api/resolvers.py
@@ -3,6 +3,7 @@ import logging
 
 from dataall.base.api.context import Context
 from dataall.core.environment.services.environment_service import EnvironmentService
+from dataall.core.stacks.services.keyvaluetag_service import KeyValueTagService
 from dataall.core.stacks.services.stack_service import StackService
 from dataall.core.stacks.db.stack_models import Stack
 
@@ -60,7 +61,7 @@ def update_stack(context: Context, source, targetUri: str = None, targetType: st
 
 
 def list_key_value_tags(context: Context, source, targetUri: str = None, targetType: str = None):
-    return StackService.list_stack_tags(targetUri, targetType)
+    return KeyValueTagService.list_key_value_tags(targetUri, targetType)
 
 
 def update_key_value_tags(context: Context, source, input=None):
--- a/backend/dataall/core/stacks/api/resolvers.py
+++ b/backend/dataall/core/stacks/api/resolvers.py
@@ -1,29 +1,17 @@
 import json
 import logging
-import os
 
 from dataall.base.api.context import Context
-from dataall.core.environment.db.environment_models import Environment
 from dataall.core.environment.services.environment_service import EnvironmentService
 from dataall.core.stacks.services.stack_service import StackService
-from dataall.core.stacks.aws.cloudformation import CloudFormation
-from dataall.core.stacks.aws.cloudwatch import CloudWatch
 from dataall.core.stacks.db.stack_models import Stack
-from dataall.core.stacks.db.keyvaluetag_repositories import KeyValueTag
-from dataall.core.stacks.db.stack_repositories import StackRepository
-from dataall.base.db import exceptions
-from dataall.base.utils import Parameter
 
 log = logging.getLogger(__name__)
 
 
 def get_stack(context: Context, source, environmentUri: str = None, stackUri: str = None):
-    with context.engine.scoped_session() as session:
-        env: Environment = EnvironmentService.get_environment_by_uri(session, environmentUri)
-        stack: Stack = StackRepository.get_stack_by_uri(session, stackUri)
-        cfn_task = StackService.save_describe_stack_task(session, env, stack, None)
-        CloudFormation.describe_stack_resources(engine=context.engine, task=cfn_task)
-        return StackService.get_environmental_stack_by_uri(uri=environmentUri, stack_uri=stackUri)
+    env = EnvironmentService.find_environment_by_uri(environmentUri)
+    return StackService.get_and_describe_stack_in_env(env, stackUri)
 
 
 def resolve_link(context, source, **kwargs):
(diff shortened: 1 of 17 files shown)


backend/dataall/core/stacks/services/stack_service.py
```diff
@@ -1,25 +1,55 @@
...
+        if not target_uri:
+            raise RequiredParameter('targetUri')
+        if not target_type:
+            raise RequiredParameter('targetType')
 
 
 class StackService:
     @staticmethod
-    def get_stack_with_cfn_resources(targetUri: str, environmentUri: str):
+    def get_stack_with_cfn_resources(targetUri: str, env):
```
dlpzx, line 50: This is a pretty important function, very minor, can we add type hints for env:Environment?

backend/dataall/core/stacks/services/stack_service.py
```diff
@@ -99,3 +129,77 @@ def delete_stack(target_uri, accountid, cdk_role_arn, region):
 
         Worker.queue(context.db_engine, [task.taskUri])
         return True
+
+    @staticmethod
+    def get_stack_by_uri(stack_uri):
+        with get_context().db_engine.scoped_session() as session:
+            return StackRepository.get_stack_by_uri(session, stack_uri)
+
```
dlpzx, line 137: I think this method, `get_environmental_stack_by_uri` should go back to the environment, and should be called `get_environment_stack_by_uri`

backend/dataall/core/stacks/services/stack_service.py
SofiaSazonova: Done: [5ddf7b3](https://github.com/data-dot-all/dataall/pull/1181/commits/5ddf7b3340e1540b445a53a8052de37291b22483)

backend/dataall/core/stacks/services/stack_service.py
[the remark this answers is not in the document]
dlpzx: But if we think about it, checking GET_ENVIRONMENT permissions when we get ANY type of stack does not make much sense. We should check GET_ENVIRONMENT when we get the Environment stack, GET_DATASET for the Dataset stack and so on. But instead we check GET_ENVIRONMENT when we get the dataset, which means that a user could execute an API call to get other teams' infrastructure stack details if they are in the same environment.

To fix this we could use something similar to what is implemented in `update_stack_by_target_uri` or in `list_key_value_tags`
```
ResourcePolicyService.check_user_resource_permission(
                session=session,
                username=context.username,
                groups=context.groups,
                resource_uri=target_uri,
                permission_name=TargetType.get_resource_read_permission_name(target_type),
            )
```

backend/dataall/core/stacks/db/keyvaluetag_repositories.py
```diff
@@ -1,96 +1,41 @@
...
-            username=context.username,
-            groups=context.groups,
-            resource_uri=uri,
-            permission_name=TargetType.get_resource_read_permission_name(target_type),
-        )
-        return KeyValueTag.find_key_value_tags(session, uri, target_type)
-
-    @staticmethod
-    def find_key_value_tags(session, target_uri, target_type) -> [models.KeyValueTag]:
+    def find_key_value_tags(session, target_uri, target_type) -> [KeyValueTag]:
```
petrkalos, line 10: Is that a valid hint? I think you need `List[KeyValueTag]`

backend/dataall/core/stacks/api/resolvers.py
```diff
@@ -68,50 +53,16 @@ def resolve_task_id(context, source: Stack, **kwargs):
...
-        return results
+    StackService.get_stack_logs(environmentUri, stackUri)
 
 
 def update_stack(context: Context, source, targetUri: str = None, targetType: str = None):
-    with context.engine.scoped_session() as session:
-        stack = StackRepository.update_stack(session=session, uri=targetUri, target_type=targetType)
-    StackService.deploy_stack(stack.targetUri)
-    return stack
+    return StackService.update_stack_by_target_uri(targetUri, targetType)
```
petrkalos, line 60: General comment about the refactoring...

* doesn't the existence of the `resolvers.py` become redundant?  They are now just proxy calls so can't we just point the resolver (where the GraphQL query is defined) directly to the service?
* by creating the session within the business logic it's going to be more difficult to unit-test those methods because it will by harder to mock the results from the db. Right?

SofiaSazonova: Well, it's open question yet. At least it gives us asyncronous calls (like in get_logs)

backend/dataall/core/stacks/db/keyvaluetag_repositories.py
noah-paige: Let's import List from typing and add this for `find_key_value_tags` and `find_environment_cascade_key_value_tags`

backend/dataall/core/stacks/services/stack_service.py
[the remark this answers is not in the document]
noah-paige: 1 - can we remove `get_environmental_stack_by_uri` since it is no longer used?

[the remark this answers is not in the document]
noah-paige: This #2 above is a bigger change than originally anticipated - I have been looking into a bit and can open an enhancement PR after this one and can push this through if we clean up `get_environmental_stack_by_uri` and address type hints comment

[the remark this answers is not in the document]
noah-paige: Link to draft PR - https://github.com/data-dot-all/dataall/pull/1182

noah-paige:
Tested creating Environments, describing env stacks, describing dataset stacks, etc.

All looks good!! (have one related PR regarding some additional stack clean up parts [here](https://github.com/data-dot-all/dataall/pull/1182) but I think is not of immediate priority and to come after this PR)

noah-paige (approved):
approving!

backend/dataall/core/stacks/api/resolvers.py
dlpzx: The principle was that in resolvers we check the api input and in service we just execute business logic. In the PRs for refactoring we are adding validation in the service layer but not in the resolvers. I did not add any comment in these PRs, because we already talked about implementing a better validation mechanism. I opened [an issue ](https://github.com/data-dot-all/dataall/issues/1183)so that we do not forget about it

PR #1182, opened 2024-04-16
[...After v2.4] Refactoring getStack API

noah-paige:
### Feature or Bugfix

- Feature
- Bugfix
- Refactoring

### Detail
- Add additional permission checks to ensure read permissions on specific target resource in order to call `getStack`

- Removed `StackStatus` component from frontend 

Additional Description in Comments Below

### Relates
N/A

### Security

  - Is the input sanitized?


SofiaSazonova:
About first point 
>> Add additional permission checks to ensure read permissions on specific target resource in order to call getStack

Do we need it? Is there any way for user to getStack without entering the resource page first?
Security is cool, but may be it's overkill.

dlpzx:
As a third option I would rescue the EventBridge notifications from environment to central account design by @SofiaSazonova in https://github.com/data-dot-all/dataall/issues/922

noah-paige:
This PR is ready for review @dlpzx @SofiaSazonova @petrkalos


The goal is this PR is to:

- (1) Add permissions checks for a user to read associated stack information
  - I would argue these permissions are definitely useful as a user who does not have access to a give dataset or environment or similar should not be able to learn anything about the underlying infra in CFN of that resource

- (2) Remove some unneeded return values from child stack types when listing data.all resources w/ associated infra
  -  Ex: `ListDataPipelines` or `ListEnvironment`

- (3) Remove StackStatus component for Data.all resources w/ associated infra (i.e. env, mlstudio, notebook, dataset, pipeline)
  - Stack status already updated via
    - Stack Tab of the resource (via `getStack` API) 
    - CDKProxy before and after `cdk deploy` command executes
  - StackStatus was used to run `getStack` API on a given interval **indefinitely** if on page for that resource
    - This seemed overkill and makes debugging logs very unreadable / hard to navigate
    - Was only used to pass snack bar notification

---
Additional Thoughts:
- If there is some need to periodically check the status of stacks - maybe introduce a background ECS task to handle stack status updates? Or find a better way to sync stack status in RDS when changes occur to status in the env account?
  - I would argue that even if this is needed it only is needed for environments b/c of `listValidEnvironments` API filtering on particular statuses

dlpzx:
Hi @noah-paige I just tested the PR. I agree with 1. permissions on targetType and uri; there is a loophole in which a user could get details of a dataset without permissions. Not a big deal, but still, better to restrict it.

I checked that removing StackStatus in the FE did not have any influence and that the remaining references use the StackStatus component in `design`

The only thing that I would remark is that in the stack_service we have 2 methods doing almost the same `get_stack_with_cfn_resources` and `get_and_describe_stack_in_env`. `get_stack_with_cfn_resources` is used in the resolver functions of the getXObject functions; so the permissions are already checked so we could keep both; the only thing is that I would do some renaming to make clear what does what

noah-paige:
> Hi @noah-paige I just tested the PR. I agree with 1. permissions on targetType and uri; there is a loophole in which a user could get details of a dataset without permissions. Not a big deal, but still, better to restrict it.
> 
> I checked that removing StackStatus in the FE did not have any influence and that the remaining references use the StackStatus component in `design`
> 
> The only thing that I would remark is that in the stack_service we have 2 methods doing almost the same `get_stack_with_cfn_resources` and `get_and_describe_stack_in_env`. `get_stack_with_cfn_resources` is used in the resolver functions of the getXObject functions; so the permissions are already checked so we could keep both; the only thing is that I would do some renaming to make clear what does what

The other big difference I see is `get_stack_with_cfn_resources` resolves the stack of the parent object and then **queues** a CFN describe task while `get_and_describe_stack_in_env` directly calls a CFN describe task and whats for a return

From a user side I think it would be best to directly call describe stack amd get the latest status for both calls however I worry about users with large numbers of stacks that all require synchronous calls to Cloudformation in their respective envs. 

Maybe we could think of a more lightweight `describe_stack_resources` that only looks for and updates `stackStatus` but still the heavy lifting of assuming a role and describing stack needs to happen for each. This makes me circle back to the Eventbridge idea and pushing notifications back to data.all backend on any changes to stacks in env accounts - but that I would deem out of scope for the above.

For now I have renamed `get_stack_with_cfn_resources` -> `resolve_parent_obj_stack` - I will merge the above but will give some time if you want to give one final look @dlpzx

dlpzx:
We keep coming back to the push notifications from environments to central data.all. I think at some point we will implement it. Polling repeated info is just a pain. Good for me

PR #1179, opened 2024-04-16
[AFTER 2.4] Refactor: uncouple datasets and dataset_sharing modules 

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail
`datasets` and `dataset_sharing` were tightly coupled, for that reason the `datasets_base` module was created. However if we want to expand the `dataset` or the `dataset_sharing` modules, managing the relationships gets more complex. 

Instead what we want is a `datasets` module that is an standalone module of S3 datasets. The `dataset_sharing` module depends on `datasets` and `datasets` does NOT depend on `datasets_sharing`.

This PR (check the commits, I tried to follow some logical order)
- [ ] Moves the `datasets_base` logic into the `datasets` module
- [ ] Updates the __init__ files to define the target dependencies (dataset_sharing depends on datasets)
- [ ] Deletes the `datasets_base` module
- [X] Splits the DatasetAlarm service into the alarms dedicated to datasets and those dedicated to dataset_sharing --> in #1184 
- [X] Creates an interface to execute checks and clean-ups of data sharing objects when dataset objects are deleted (initially it was going to be an db interface, but I think it is better in the service) in https://github.com/data-dot-all/dataall/pull/1213
- [X] Moves the verify dataset shares mutation to the datasets_sharing module in https://github.com/data-dot-all/dataall/pull/1185
- [X] Split the KMS client and create one for datasets in https://github.com/data-dot-all/dataall/pull/1186
- [X] Move dataset_subscription task to dataset_sharing in https://github.com/data-dot-all/dataall/pull/1185
- [X] Use interface to add listSharedDatasets to listDatasets and create a listSharedDatasets in the dataset_sharing module in https://github.com/data-dot-all/dataall/pull/1213
-  [X] Use interface to resolve dataset roles related to datasets shared and implement logic in the dataset_sharing module in #1214 
- [X] Split the getDatasetAssumeRole API into 2 APIs, one for dataset owners role (in datasets module) and another one for share requester roles (in datasets_sharing module) in https://github.com/data-dot-all/dataall/pull/1187
- [X] Extend and clean-up stewards share permissions through interface #1214 
- [X] Move listDatasetShares query to dataset_sharing module in https://github.com/data-dot-all/dataall/pull/1185
[...]

PR #1186, opened 2024-04-17
Refactor: uncouple datasets and dataset_sharing modules - part 3

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail
Remove and move logic from dataset to datasets_sharing module. This is needed as explained in full PR [AFTER 2.4] Refactor: uncouple datasets and dataset_sharing modules https://github.com/data-dot-all/dataall/pull/1179
- [X] Split the KMS client and create one dedicated KMS client for datasets

### Relates
https://github.com/data-dot-all/dataall/pull/1179

### Security
[OWASP 10](https://owasp.org/Top10/en/). `N/A`

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/datasets/aws/kms_dataset_client.py
+++ b/backend/dataall/modules/datasets/aws/kms_dataset_client.py
@@ -1,0 +1,52 @@
+import logging
+
+from dataall.base.aws.sts import SessionHelper
+from botocore.exceptions import ClientError
+
+
+log = logging.getLogger(__name__)
+
+
+class KmsClient:
+    _DEFAULT_POLICY_NAME = 'default'
+
+    def __init__(self, account_id: str, region: str):
+        session = SessionHelper.remote_session(accountid=account_id, region=region)
+        self._client = session.client('kms', region_name=region)
+        self._account_id = account_id
+        self.region = region
+
+    def get_key_id(self, key_alias: str):
+        # The same client function is defined in the data_sharing module. Duplication is allowed to avoid coupling.
+        try:
+            response = self._client.describe_key(
+                KeyId=key_alias,
+            )
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'AccessDenied':
+                raise Exception(
+                    f'Data.all Environment Pivot Role does not have kms:DescribeKey Permission for key {key_alias}: {e}'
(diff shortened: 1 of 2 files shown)


SofiaSazonova:
Where do you remove this logic from? In changelog I can only see the inserted function

backend/dataall/modules/datasets/aws/kms_dataset_client.py
```diff
@@ -0,0 +1,52 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class KmsClient:
+    _DEFAULT_POLICY_NAME = 'default'
+
+    def __init__(self, account_id: str, region: str):
+        session = SessionHelper.remote_session(accountid=account_id, region=region)
+        self._client = session.client('kms', region_name=region)
```
SofiaSazonova, line 15: since it's a new part, may be we can start using 'dependency injection patter' as we discussed before?
Smth like
```
  def __init__(self, account_id: str, region: str, client = None):
        self._account_id = account_id
        self.region = region
        if client is not None:
             self._client = client
        else:
             session = SessionHelper.remote_session(accountid=account_id, region=region)
             self._client = session.client('kms', region_name=region)
```

SofiaSazonova (changes requested):

backend/dataall/modules/datasets/aws/kms_dataset_client.py
dlpzx: I will do it on a separate PR, to study how to extend it to other boto3 clients. I think we can look for a way of injecting without doing the distinction between tests and no-tests. 

Before we call the KmsClient we would create the session and then pass it directly, that way we do not need the if-else. 
```
session = SessionHelper.remote_session(accountid=account_id, region=region)
boto_client = session.client('kms', region_name=region)
kms_client = KmsClient(client=boto_client)
```

```
  def __init__(self, account_id: str, region: str, client: boto3.session.Session):
        self._account_id = account_id
        self.region = region
        self._client = client
```

But creating the boto_client every time we initialise the KmsClient seems a bit painful, so I need to look for a better way of doing that. That's why I want a separate PR :)

SofiaSazonova (approved):

PR #1185, opened 2024-04-17
Refactor: uncouple datasets and dataset_sharing modules - part 2

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail

Remove and move logic from dataset to datasets_sharing module. This is needed as explained in full PR [AFTER 2.4] Refactor: uncouple datasets and dataset_sharing modules  #1179
- [X] Moves the verify dataset shares mutation to the datasets_sharing module
- [X] Move dataset_subscription task to dataset_sharing
- [X] Move listDatasetShares query to dataset_sharing module
- [X] Remove unused `shares` field from the Dataset graphql type as it was not used in the frontend: listDatasets, listOwnedDatasets, listDatasetsOwnedByEnvGroup, listDatasetsCreatedInEnvironment and getDataset
- [x] Move getSharedDatasetTables to data_sharing module and fix reference to DatasetService

I am aware that some of the queries and mutations that this PR moves look a bit odd in the dataset_sharing module, but this will be solved once data sharing is divided into dataset_sharing_base and s3_dataset_sharing.

### Relates
#1179

### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)?
  - Is the input sanitized?
  - What precautions are you taking before deserializing the data you consume?
  - Is injection prevented by parametrizing queries?
  - Have you ensured no `eval` or similar functions are used?
- Does this PR introduce any functionality or component that requires authorization?
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms?
  - Are you logging failed auth attempts?
- Are you using or adding any cryptographic features?
  - Do you use a standard proven implementations?
  - Are the used keys controlled by the customer? Where are they stored?
- Are you introducing any new policies/roles/users?
  - Have you used the least-privilege principle? How?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1187, opened 2024-04-17
Refactor: uncouple datasets and dataset_sharing modules - part 2-2

dlpzx:
### Feature or Bugfix
- Refactoring
⚠️ MERGE AFTER https://github.com/data-dot-all/dataall/pull/1185

### Detail
- Split the getDatasetAssumeRole API into 2 APIs, one for dataset owners role (in datasets module) and another one for share requester roles (in datasets_sharing module)

### Relates
-  #1179

### Security

  - Is the input sanitized?


SofiaSazonova (approved):

backend/dataall/modules/dataset_sharing/api/queries.py
```diff
@@ -61,3 +62,11 @@
     type=gql.ArrayType(gql.Ref('SharedDatasetTableItem')),
     resolver=list_shared_tables_by_env_dataset,
 )
+
+getDatasetSharedAssumeRoleUrl = gql.QueryField(
```
petrkalos, line 66: This is a generic comment about how the conditional module loading works and the feature flag works.

I see that this Query relies on the `get_dataset_shared_assume_role_url` resolver but this is enabled conditionally. Does it mean that the user can call this query but will receive `Disabled by config ...`? 
Wouldn't it better if we don't define the query at all?

frontend/src/modules/Folders/views/FolderView.js
```diff
@@ -51,13 +52,24 @@ function FolderPageHeader(props) {
...
     } else {
-      dispatch({ type: SET_ERROR, error: response.errors[0].message });
+      const response = await client.query(
+        getDatasetSharedAssumeRoleUrl(folder.dataset.datasetUri)
+      );
+      if (!response.errors) {
+        window.open(response.data.getDatasetSharedAssumeRoleUrl, '_blank');
+      } else {
+        dispatch({ type: SET_ERROR, error: response.errors[0].message });
+      }
```
petrkalos, line 72: nit: I'd DRY this part with a small lambda

petrkalos (approved):

backend/dataall/modules/dataset_sharing/services/dataset_sharing_service.py
```diff
@@ -49,3 +53,37 @@ def list_shared_tables_by_env_dataset(dataset_uri: str, env_uri: str):
                     session, env_uri, dataset_uri, context.username, context.groups
                 )
             ]
+
+    @staticmethod
+    @ResourcePolicyService.has_resource_permission(CREDENTIALS_DATASET)
+    def get_dataset_shared_assume_role_url(uri):
```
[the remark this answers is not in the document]
dlpzx: That would be nice, not only for this API call but to the original get_dataset_assume_role_url. It involves renaming quite a number of files so I'd do it in a separate PR

backend/dataall/modules/dataset_sharing/api/queries.py
dlpzx: I see, so having the same `@is_feature_enabled('modules.datasets.features.aws_actions')` logic on the Query creation instead of in the resolver call right? Yes, it does make sense and it might be possible with AppSync I guess as we define API calls as infra as code independently right?

frontend/src/modules/Folders/views/FolderView.js
dlpzx: what do you mean?

petrkalos: it's the same code in if and else branch, with some different parameters, right? DRY means Don't Repeat Yourself.

backend/dataall/modules/dataset_sharing/api/queries.py
petrkalos: Yes that's what I mean, I think it's possible here, instead of putting the decorator in the resolver you can put an `if` in the definition of the QueryField.

Issue #1190, opened 2024-04-18
Deployment from scratch fails with latest changes in backend Lambda log group creation


dlpzx:
### Describe the bug

When deploying data.all in a fresh account using the latest code in `main` the CICD fails in the deployment of the backend stack. The failure comes from the creation of CloudWatch alarms for the log-groups of the backend lambdas.
In this PR #1134 the behavior on how the log groups for the Lambda functions changes, instead of letting Lambda create the groups automatically, it explicitly checks for the log group name and then performs the check. Unfortunately, the output of `from_log_group_name` is not None even when the log group does not exist.

```
       esproxy_loggroup = logs.LogGroup.from_log_group_name(
            self, 'esproxyloggroup', f'/aws/lambda/{resource_prefix}-{envname}-esproxy'
        )
....

        self.elasticsearch_proxy_handler = _lambda.DockerImageFunction(
            self,
            'ElasticSearchProxyHandler',
            function_name=f'{resource_prefix}-{envname}-esproxy',
            log_group=esproxy_loggroup
            if esproxy_loggroup is not None
            else logs.LogGroup(
                self, 'esproxyloggroup', log_group_name=f'/aws/lambda/{resource_prefix}-{envname}-esproxy'
            ),
```

### How to Reproduce

Deploy data.all with the current changes including #1134

### Expected behavior

Smooth behavior, log groups created and cloudwatch alarms created referencing the created log groups

### OS

n/a

### Python version

n/a

### AWS data.all version

2.3+additional PRs

dlpzx:
From testing we can conclude:
1. If we do not specify the `log_group` in the Lambda creation, it will automatically create a LogGroup for the Lambda using the name of the function, which is exactly what we create using `logs.LogGroup`
2. If we create the  `logs.LogGroup` but it already exists, the whole stack fails creation
3. if we create the `logs.LogGroup` and it did not exist before, it succeeds

I think because of number 2, the check was added, however the check does not effectively check the existence of a previous stack. 

I am inclined to revert the changes and go for the previous setup in which we did not define the log-group and use the automatically created ones. But maybe there is something that I am missing @mourya-33

PR #1134, opened 2024-03-30
Implement least privilege principle for cloudfront, lambda and db migration stacks

mourya-33:
…front permissions

### Feature or Bugfix
- Bugfix

### Relates
[- <URL or Ticket>](https://github.com/data-dot-all/dataall/issues/877)

### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)? N/A
  - Is the input sanitized? N/A 
  - What precautions are you taking before deserializing the data you consume? N/A
  - Is injection prevented by parametrizing queries? N/A
  - Have you ensured no `eval` or similar functions are used? N/A
- Does this PR introduce any functionality or component that requires authorization? N/A
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms? N/A
  - Are you logging failed auth attempts? N/A
- Are you using or adding any cryptographic features? N/A
  - Do you use a standard proven implementations? yes
  - Are the used keys controlled by the customer? Where are they stored? N/A
- Are you introducing any new policies/roles/users? N/A
  - Have you used the least-privilege principle? How? Yes, by removing the * for cloudfront permissions and explicitly specifying the distribution id arn.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1192, opened 2024-04-18
Fix: Remove creation of log-groups in Lambdas

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
In #1134 log groups are created explicitly for each Lambda in the backend, which results in errors when the log-groups do not exist.  (see #1190 )

This PR removes the `log_group` parameter from the definition of the Lambdas. It reverts back to the previous (2.3) settings. 

Tested: on a fresh deployment, making sure that no log-groups were previously created, the stack creates the log-groups with the Lambda name. For example `/aws/lambda/<resource_prefix>-<env_name>-esproxy`.

Update: we need to explicitly define the log groups to avoid the log retention lambdas and their roles and policies that CDK automatically creates. So to cope with existing groups and non-existing groups this PR adds a CfnCondition to the creation of the log-group

### Relates
- #1134

### Security
[OWASP 10](https://owasp.org/Top10/en/). `n/a`

  - Is the input sanitized?

### DIFF
--- a/deploy/stacks/lambda_api.py
+++ b/deploy/stacks/lambda_api.py
@@ -15,6 +15,7 @@ from aws_cdk import (
     aws_sqs as sqs,
     aws_logs as logs,
     Duration,
+    CfnCondition,
     CfnOutput,
     Fn,
     RemovalPolicy,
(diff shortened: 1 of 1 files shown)


noah-paige:
pretty sure the change to explicitly create log groups was primarily to avoid CDK from constructing a policy with PutLog* permissions on resource * 

I think no specifying a log group and letting CDK/AWS create defaults does behave that way but maybe there is another way around (i.e. `.without_role_policy_updates()` as we do with CodeBuild and giving log permissions in `create_function_role()`)

dlpzx:
Hi @mourya-33 @petrkalos. 

**We can discard using `CfnConditions` with log groups imported**
I have tried the implementation in the latest [commit](https://github.com/data-dot-all/dataall/pull/1192/commits/a89aa6576c28e51d36a5949f95f20be260e5c52c) and some other local alternatives using CfnConditions. It turns out that [ILogGroup ](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_logs/ILogGroup.html) which is what we get when we import the log_group, does not have any method that we can successfully evaluate in a condition. I have tested `log_group_physical_name()` which is the only one that gives a Token as string, but it returns an empty string as the log group does not have a physical name even when it exists. All other parameters from the IlogGroup class do not check if the log group actually exists (it makes sense, that is why it failed so late in the Cfn template)


**Other alternatives**
1. Create new log-groups (with slightly different names). It does not impose any issues in fresh deployments or in existing deployments. The only issue is that for existing deployments users will have 2 log-groups, the old one and the new one. I will investigate ways of migrating the log-streams from one to the other.
2. Use boto3 to check if the log group exists. It would not be an issue because log groups are not deleted as we do not set expiration date for them. But it is a bit hacky.
3. Use version to decide whether to create a new log-group

deploy/stacks/lambda_api.py
```diff
@@ -69,24 +69,12 @@ def __init__(
...
         self.elasticsearch_proxy_handler = _lambda.DockerImageFunction(
             self,
             'ElasticSearchProxyHandler',
             function_name=f'{resource_prefix}-{envname}-esproxy',
-            log_group=esproxy_loggroup
-            if esproxy_loggroup is not None
-            else logs.LogGroup(
-                self, 'esproxyloggroup', log_group_name=f'/aws/lambda/{resource_prefix}-{envname}-esproxy'
+            log_group=logs.LogGroup(
+                self, 'esproxyloggroup', log_group_name=f'/aws/lambda/{resource_prefix}-{envname}-backend-esproxy'
```
petrkalos, line 77: We discussed offline that just by renaming we will lose the historical metrics? are we ok with that?

dlpzx: That is what we need to decide, I have been looking for ways of migrating logs from one log group to another and it is not completely straightforward

petrkalos (approved):
As discussed offline the already emitted metrics will not go away from CloudWatch so I am approving

dlpzx:
@petrkalos I double check and a CW metric keeps the history even after changing the log-group. Good to know :)

PR #1213, opened 2024-04-23
Refactor: uncouple datasets and dataset_sharing modules - part 2-3

alopezlopez4:
### Feature or Bugfix
- Refactoring
⚠️ MERGE AFTER https://github.com/data-dot-all/dataall/pull/1187

### Detail

- [X] Creates an interface to execute checks and clean-ups of data sharing objects when dataset objects are deleted (initially it was going to be an db interface, but I think it is better in the service)
- [X] Move listDatasetShares query to dataset_sharing module in https://github.com/data-dot-all/dataall/pull/1185

### Relates
-  #1179

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/datasets/services/dataset_service.py
+++ b/backend/dataall/modules/datasets/services/dataset_service.py
@@ -56,19 +56,19 @@ class DatasetServiceInterface(ABC):
     @abstractmethod
     def check_before_delete(session, uri, **kwargs) -> bool:
         """Abstract method to be implemented by dependent modules that want to add checks before deletion for dataset objects"""
-        raise NotImplementedError('Method check_before_delete is not implemented')
+        ...
 
     @staticmethod
     @abstractmethod
     def execute_on_delete(session, uri, **kwargs) -> bool:
         """Abstract method to be implemented by dependent modules that want to add clean-up actions when a dataset object is deleted"""
-        raise NotImplementedError('Method execute_on_delete is not implemented')
+        ...
 
     @staticmethod
     @abstractmethod
     def append_to_list_user_datasets(session, username, groups):
         """Abstract method to be implemented by dependent modules that want to add datasets to the list_datasets that list all datasets that the user has access to"""
-        raise NotImplementedError('Method append_to_list_user_datasets is not implemented')
+        ...
 
 
 class DatasetService:


backend/dataall/modules/datasets/services/dataset_service.py
```diff
@@ -50,7 +51,67 @@
...
+
+    @classmethod
+    def register(cls, interface: DatasetServiceInterface):
+        cls._interfaces.append(interface)
+
+    @classmethod
+    def check_before_delete(cls, session, uri, **kwargs) -> bool:
+        """All actions from other modules that need to be executed before deletion"""
+        can_be_deleted = [interface.check_before_delete(session, uri, **kwargs) for interface in cls._interfaces]
+        return False not in set(can_be_deleted)
```
SofiaSazonova, line 96: You can just use `all(can_be_deleted)`

SofiaSazonova (changes requested):

backend/dataall/modules/datasets/api/dataset/resolvers.py
```diff
@@ -57,10 +57,10 @@ def get_file_upload_presigned_url(context, source, datasetUri: str = None, input
     return DatasetService.get_file_upload_presigned_url(uri=datasetUri, data=input)
 
 
-def list_owned_shared_datasets(context: Context, source, filter: dict = None):
+def list_all_user_datasets(context: Context, source, filter: dict = None):
     if not filter:
         filter = {'page': 1, 'pageSize': 5}
```
SofiaSazonova, line 62: This values are the default ones inside 
```
@staticmethod
 def paginated_dataset_shares(session, uri, data=None) -> [ShareObject]:
        query = ShareObjectRepository.query_dataset_shares(session, uri)
        return paginate(query=query, page=data.get('page', 1), page_size=data.get('pageSize', 5)).to_dict()
```

Maybe it's better to have default value as `{}` then?

SofiaSazonova: Same for ` list_owned_datasets`

backend/dataall/modules/datasets/services/dataset_location_service.py
```diff
@@ -87,14 +87,8 @@ def remove_storage_location(uri: str = None):
...
             dataset = DatasetRepository.get_dataset_by_uri(session, location.datasetUri)
-            has_shares = ShareObjectRepository.has_shared_items(session, location.locationUri)
-            if has_shares:
-                raise ResourceShared(
-                    action=DELETE_DATASET_FOLDER,
-                    message='Revoke all folder shares before deletion',
-                )
-
-            ShareObjectRepository.delete_shares(session, location.locationUri)
+            DatasetService.check_before_delete(session, location.locationUri, action=DELETE_DATASET_FOLDER)
```
SofiaSazonova, line 90: We can check this on the line 89 before getting dataset. Search for dataset is unnecessary, if the check fails.

backend/dataall/modules/datasets/services/dataset_service.py
dlpzx: way more elegant indeed

backend/dataall/modules/datasets/api/dataset/resolvers.py
dlpzx: I agree on not setting it as default, but in this case the default is 10 not 5 `dataall.modules.datasets_base.db.dataset_repositories.DatasetRepository.paginated_all_user_datasets` I need to verify what is needed

backend/dataall/modules/datasets/services/dataset_service.py
```diff
@@ -50,7 +51,67 @@
...
+        return None
+
+    @staticmethod
+    def extend_attach_steward_permissions(session, dataset, new_stewards) -> bool:
+        """Abstract method to be implemented by dependent modules that want to attach additional permissions to Dataset stewards"""
+        return True
+
+    def extend_delete_steward_permissions(session, dataset, new_stewards) -> bool:
+        """Abstract method to be implemented by dependent modules that want to attach additional permissions to Dataset stewards"""
+        return True
```
petrkalos, line 82: Should we use [@abstractmethod](https://docs.python.org/3/library/abc.html#abc.abstractmethod) and `...` for unfinished code?

backend/dataall/modules/datasets/services/dataset_service.py
```diff
@@ -283,24 +347,16 @@ def get_dataset_assume_role_url(uri):
...
-            else:
+            if dataset.SamlAdminGroupName in context.groups:
                 role_arn = dataset.IAMDatasetAdminRoleArn
                 account_id = dataset.AwsAccountId
                 region = dataset.region
 
+            else:
+                raise exceptions.UnauthorizedOperation(
+                    action=CREDENTIALS_DATASET,
+                    message=f'User: {context.username} is not a member of the group {dataset.SamlAdminGroupName}',
```
petrkalos, line 358: nit: Instead of `f'User: {context.username}...` you can use `f'{context.username=}...` and it will also print the variable name.  [docs](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging)

backend/dataall/modules/datasets/services/dataset_service.py
```diff
@@ -50,7 +51,67 @@
...
+
+    @classmethod
+    def _list_all_user_interface_datasets(cls, session, username, groups) -> List:
+        """All list_datasets from other modules that need to be appended to the list of datasets"""
+        all_subqueries = []
+        for interface in cls._interfaces:
+            interface_subquery = interface.append_to_list_user_datasets(session, username, groups)
+            if interface_subquery.first() is not None:
+                all_subqueries.append(interface_subquery)
+        return all_subqueries
```
petrkalos, line 113: can this become...
```python
return [
        query
        for interface in cls._interfaces
        for query in interface.append_to_list_user_datasets(session, username, groups)
        if query is not None
    ]
```

backend/dataall/modules/datasets/services/dataset_service.py
SofiaSazonova: wow

SofiaSazonova (changes requested):
When the changes, requested by Petros are adressed, good to go

backend/dataall/modules/datasets/services/dataset_location_service.py
dlpzx: Resolved in https://github.com/data-dot-all/dataall/pull/1213/commits/ebdcdb058b961c2de32837de52268554e16f8680

backend/dataall/modules/datasets/services/dataset_service.py
dlpzx: My concern is that not all implementations of DatasetServiceInterface might use all methods. But for that the solution would be to break down the interface into smaller interfaces following (Interface Segregation principle)

backend/dataall/modules/datasets/services/dataset_service.py
dlpzx: I think that loop is not correct because `interface.append_to_list_user_datasets(session, username, groups)` does not return a list of queries, it returns a single query. Also, the query itself won't be none, we need to check that it does not return any item. Maybe something like:


```
        return [
            interface.append_to_list_user_datasets(session, username, groups)
            for interface in cls._interfaces
            if interface.append_to_list_user_datasets(session, username, groups).first() is not None
        ]
```

backend/dataall/modules/datasets/services/dataset_service.py
dlpzx: Very cool! Implemented

backend/dataall/modules/datasets/services/dataset_service.py
dlpzx: Shouldn't it be:
```python
return [
        query
        for interface in cls._interfaces
        for query in [interface.append_to_list_user_datasets(session, username, groups)]
        if query.first() is not None
    ]
```

dlpzx: Then yes!

backend/dataall/modules/datasets/services/dataset_service.py
```diff
@@ -53,33 +53,22 @@
 
 class DatasetServiceInterface(ABC):
     @staticmethod
+    @abstractmethod
     def check_before_delete(session, uri, **kwargs) -> bool:
         """Abstract method to be implemented by dependent modules that want to add checks before deletion for dataset objects"""
-        return True
+        raise NotImplementedError('Method check_before_delete is not implemented')
```
petrkalos, line 59: just an ellipsis here will do to avoid the visual clutter, python will throw because of the decorator

check this out...
```python
from abc import abstractmethod, ABC

import pytest


class FooABC(ABC):
    @staticmethod
    @abstractmethod
    def abc():
        ...


class FooImplMissing(FooABC):
    ...


def test_fooabc_insta_error():
    with pytest.raises(TypeError, match="Can't instantiate abstract class FooABC with abstract method abc"):
        FooABC()


def test_fooimpl_missing_error():
    with pytest.raises(TypeError, match="Can't instantiate abstract class FooImplMissing with abstract method abc"):
        FooImplMissing()
```

petrkalos (approved):
:ship:

SofiaSazonova (approved):

PR #1214, opened 2024-04-23
Refactor: uncouple datasets and dataset_sharing modules - part 2-4

dlpzx:
### Feature or Bugfix
- Refactoring
⚠️ MERGE AFTER https://github.com/data-dot-all/dataall/pull/1213

### Detail
-  [X] Use interface to resolve dataset roles related to datasets shared and implement logic in the dataset_sharing module
- [X] Extend and clean-up stewards share permissions through interface

### Relates
- #1179 

### Security

  - Is the input sanitized?


backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
```diff
@@ -368,11 +368,11 @@ def get_share_by_uri(session, uri):
         return share
 
     @staticmethod
-    def get_share_by_dataset_attributes(session, dataset_uri, dataset_owner):
+    def get_share_by_dataset_attributes(session, dataset_uri, dataset_owner, groups=[]):
```
noah-paige, line 371: Related to L160 in `DatasetSharingService.get_dataset_shared_assume_role_url()` where this function is called - should we also pass `groups=context.groups` for `get_dataset_shared_assume_role_url()` so not only filtering based on user is share owner

noah-paige (approved):
Local testing looks good - left one comment on `get_dataset_shared_assume_role_url()` (take a look but it is out of scope for this PR), approving

backend/dataall/modules/dataset_sharing/db/share_object_repositories.py
dlpzx: `get_dataset_shared_assume_role_url` already checks the context.groups of the user and returns a different role depending on the groups. So they can navigate to the S3 bucket, but assuming their own role. I am also not 100% convinced about the current behavior, but I don't think limiting the groups is going to help

```
    def get_dataset_shared_assume_role_url(uri):
        context = get_context()
            dataset = DatasetRepository.get_dataset_by_uri(session, uri)

            if dataset.SamlAdminGroupName in context.groups:
                role_arn = dataset.IAMDatasetAdminRoleArn
                account_id = dataset.AwsAccountId
                region = dataset.region
            else:
                share = ShareObjectRepository.get_share_by_dataset_attributes(
                    session=session, dataset_uri=uri, dataset_owner=context.username
                )
                shared_environment = EnvironmentService.get_environment_by_uri(
                    session=session, uri=share.environmentUri
                )
                env_group = EnvironmentService.get_environment_group(
                    session=session, group_uri=share.principalId, environment_uri=share.environmentUri
                )
                role_arn = env_group.environmentIAMRoleArn
                account_id = shared_environment.AwsAccountId
                region = shared_environment.region
```

noah-paige: So `get_dataset_shared_assume_role_url()` only gets called if the user is not an "Admin" (i.e. Creator, Owner, Admin) of the dataset

The issues I am seeing:
- (1) `get_dataset_shared_assume_role_url()` requires `CREDENTIALS_DATASET` permissions which is a part of `DATASET_READ` which I do not think we grant to share requestors
- (2) The initial `if` condition will never be true
- (3) Would getting share by ShareObject.owner == context.username even work? What if UserA of GroupA creates share object and UserA2 of GroupA tries to get this role?
- (4) Even assuming you are able to get credentials to the requestor env group role would this redirect to AWS page work if share is cross-account?

Because of all the above reasons maybe we remove this API call :) @dlpzx

PR #1219, opened 2024-04-25
Add integration tests on a real API client and integrate the tests in CICD

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Add integration tests that use a real Client to execute different validation actions. 

- Define the Client and the way API calls are posted to API Gateway in the conftest
- Define the Cognito users and the different fixtures needed for all tests
- Write tests for the Organization core module as example
- Add feature flag in `cdk.json` called `with_approval_tests` that can be defined at the deployment environment level. If set to True, a CodeBuild stage running the tests is created.

### Relates
- https://github.com/data-dot-all/dataall/issues/1220

### Security

  - Is the input sanitized?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,9 @@ install-cdkproxy:
 install-tests:
 	pip install -r tests/requirements.txt
 
+install-integration-tests:
+	pip install -r tests_new/integration_tests/requirements.txt
+
 lint:
 	pip install ruff
 	ruff check --fix
@@ -51,6 +54,10 @@ check-security: upgrade-pip install-backend install-cdkproxy
 test:
 	export PYTHONPATH=./backend:/./tests && \
 	python -m pytest -v -ra tests/
+
+integration-test: install-integration-tests
+	export PYTHONPATH=./backend:/./tests_new && \
+	python -m pytest -v -ra tests_new/integration_tests/
 
 coverage: upgrade-pip install-backend install-cdkproxy install-tests
 	export PYTHONPATH=./backend:/./tests && \
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ test:
 	export PYTHONPATH=./backend:/./tests && \
 	python -m pytest -v -ra tests/
 
-integration-tests: upgrade-pip install-backend install-cdkproxy install-integration-tests
+integration-tests: upgrade-pip install-integration-tests
 	export PYTHONPATH=./backend:/./tests_new && \
 	python -m pytest -v -ra tests_new/integration_tests/ \
 		--junitxml=reports/integration_tests.xml
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ test:
 	export PYTHONPATH=./backend:/./tests && \
 	python -m pytest -v -ra tests/
 
-integration-test: install-integration-tests
+integration-tests: upgrade-pip install-backend install-cdkproxy install-integration-tests
 	export PYTHONPATH=./backend:/./tests_new && \
 	python -m pytest -v -ra tests_new/integration_tests/
 
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,8 @@ test:
 
 integration-tests: upgrade-pip install-backend install-cdkproxy install-integration-tests
 	export PYTHONPATH=./backend:/./tests_new && \
-	python -m pytest -v -ra tests_new/integration_tests/
+	python -m pytest -v -ra tests_new/integration_tests/ \
+		--junitxml=reports/integration_tests.xml
 
 coverage: upgrade-pip install-backend install-cdkproxy install-tests
 	export PYTHONPATH=./backend:/./tests && \
--- a/tests_new/integration_tests/client.py
+++ b/tests_new/integration_tests/client.py
@@ -1,0 +1,43 @@
+import requests
+import boto3
+import os
+from munch import DefaultMunch
+from dataall.base.aws.parameter_store import ParameterStoreManager
+
+
+ENVNAME = os.getenv('ENVNAME', 'dev')
+
+
+class Client:
+    def __init__(self, username, password):
+        self.username = username
+        self.password = password
+        self.token = self._get_jwt_token()
+
+    def query(self, query: str):
+        endpoint = ParameterStoreManager.get_parameter_value(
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -52,6 +55,10 @@ test:
 	export PYTHONPATH=./backend:/./tests && \
 	python -m pytest -v -ra tests/
 
+integration-test: install-integration-tests
+	export PYTHONPATH=./backend:/./tests_new && \
+	python -m pytest -v -ra tests_new/integration_tests/
```
petrkalos, line 60: let's export the XML to make it available in CodeBuild, see [here](https://docs.aws.amazon.com/codebuild/latest/userguide/test-report-pytest.html)

dlpzx: Done!

dlpzx:
I tested the PR by merging it to a real AWS deployment. 

I verified that the results report was being correctly generated:
image

Makefile
```diff
@@ -52,6 +55,11 @@ test:
 	export PYTHONPATH=./backend:/./tests && \
 	python -m pytest -v -ra tests/
 
+integration-tests: upgrade-pip install-backend install-cdkproxy install-integration-tests
```
petrkalos, line 58: why do we need install-backend and cdkproxy?

tests_new/integration_tests/client.py
```diff
@@ -0,0 +1,51 @@
...
+        self.token = self._get_jwt_token()
+
+    def query(self, query: str):
+        endpoint = (
+            os.getenv('API_ENDPOINT', False)
+            if os.getenv('API_ENDPOINT', False)
+            else ParameterStoreManager.get_parameter_value(
+                region=os.getenv('AWS_REGION', 'eu-west-1'), parameter_path=f'/dataall/{ENVNAME}/apiGateway/backendUrl'
+            )
+        )
```
petrkalos, line 24: I think the `API_ENDPOINT` variable is enough here. If devs want to pull the endpoint from ParameterStore they can use AWS CLI and a .envrc file to automate it.

tests_new/integration_tests/conftest.py
```diff
@@ -0,0 +1,133 @@
...
+
+from tests_new.integration_tests.client import Client
+
+load_modules(modes=ImportMode.all())
+
+
+collect_ignore_glob = []
+
+
+def ignore_module_tests_if_not_active():  # TODO: maybe move to commons between tests
```
petrkalos, line 16: Isn't there a way to query which modules are active from the API directly?
Don't we get different response when the query is missing vs when something goes wrong?

Makefile
dlpzx: It is related to one of your comments below. When we use the `ignore_module_tests_if_not_active` method to get what is. loaded and what's not we initialize the dataall package and we import all subpackages some of which depend on the packages installed. We would not have this issue if we check if modules are enabled/disabled in a different way.

What I can do is to remove `ignore_module_tests_if_not_active` because at the moment it is not needed (there are tests only for organization) and think of a different way of implementing module test switch. wdyt?

tests_new/integration_tests/conftest.py
dlpzx: If a module is not enabled the API call won't exist, so the response will be something like (I just ran a query with a typo): `"Cannot query field 'listTenantPermission' on type 'Query'. Did you mean 'listTenantPermissions' or 'listTenantGroups'?"`

It is a bit hacky, I have been looking for ways to define the test scope of an API call using graphql, with no luck. Another option is to list the available Queries and Mutations (no with API GW but maybe with AppSync introspection?). But we still need to map them to modules or we run tests based on each API call....

tests_new/integration_tests/client.py
```diff
@@ -0,0 +1,36 @@
...
+class Client:
+    def __init__(self, username, password):
+        self.username = username
+        self.password = password
+        self.token = self._get_jwt_token()
+
+    def query(self, query: str):
+        graphql_endpoint = f'{os.getenv("API_ENDPOINT", False)}graphql/api'
+        headers = {'AccessKeyId': 'none', 'SecretKey': 'none', 'authorization': self.token}
+        r = requests.post(graphql_endpoint, json=query, headers=headers)
```
petrkalos, line 19: nit: I know this is going to be replaced with SDK but does it make sense to use [GQL's client](https://github.com/graphql-python/gql)? Maybe it does have some retry mechanisms and might also provide better logging.

tests_new/integration_tests/client.py
```diff
@@ -0,0 +1,36 @@
...
+        r.raise_for_status()
+
+        return DefaultMunch.fromDict(r.json())
+
+    def _get_jwt_token(self):
+        cognito_client = boto3.client('cognito-idp', region_name=os.getenv('AWS_REGION', 'eu-west-1'))
```
petrkalos, line 25: Can't we authenticate same way as we do with the browser? 
Why do we need the boto3 calls? 
Also this ties the authentication to Cognito, what about customers that they use a custom IdP?
Maybe something like [this](https://gist.github.com/miguelmota/8b519212aca47210d529532b3d8e5b2f)?

I guess this one will be solved by the SDK as well.

tests_new/integration_tests/conftest.py
petrkalos: I don't think it's hacky, you can create a decorator and check which APIs must be available, otherwise test will be disabled. Something like `disable_if(api='Query.listTenantGroup')`

Another option is to provide a real query that returns all the enabled modules and enable it only on dev stages 

Also introspection is not a bad idea, you can [enable it in Apollo server as well](https://www.apollographql.com/docs/apollo-server/api/apollo-server/#introspection), is not an AppSync specific feature.

tests_new/integration_tests/client.py
dlpzx: Exactly I did not want to spend much time on something that will be replaced by the SDK/CLI authentication, because it is going to mimic it

tests_new/integration_tests/client.py
dlpzx: I have been working a bit  with it and tried to use the httpx trasnport of the gql client, at the end I ended up writing the same code as the one we have in the CLI/SDK. I don't want to make it difficult to replace when the time comes, I just want to have the minimum functionality needed for the tests to work

petrkalos (approved):

petrkalos:
This looks awesome.

Issue #1212, opened 2024-04-23
Data.all UPDATE/DELETE team in environment referencing the environment and not the Team


dlpzx:
### Describe the bug

At the moment for update_group_permissions or remove_group from environment we check the permissions of the user in relation with the environment. This is wrong, we should check the permissions of the user with respect to the invited_group. The same way we do with consumption roles.

### How to Reproduce

Given that userA belings to teamA, userB to teamB and userC to teamC.
1. UserA creates environmentA with teamA
2. UserA invites teamB to environmentA
3. UserA invites teamC to environmentA
4. UserB CANNOT update permissions of any team in environment A
5. UserB CAN remove teamC from the environment

### Expected behavior

Teams that should be able to update permissions and remove groups:
- the environment Admin
- the invited group itself
- the group that invited the group

Situations like the one in step 5 should not happen.

We will need to backfill permissions :( 

### OS

n/a

### Python version

n/a

### AWS data.all version

2.3+additional PRs

SofiaSazonova:
There are three steps to properly solve this issue:

- [ ]  Quick bugfix. Right now the user can remove group, if they can invite group. Quick solution: couple remove permission with update permission, rather then invite permission.

- [ ] Introduce ResourcePolicy with the resource_type = 'EnvironmentGroup'. In invitation form the user will have to input 'Supervisors' -- group, that have permissions to update and remove EnvironmentGroup.

- [ ] Implement 'permission boundaries' mechanism. Each group can grant other group with limited permissions.

SofiaSazonova:
The described bug is fixed by the PR.
All other enhancements -- in separate issue #1235

PR #1234, opened 2024-04-30
bugfix: EnvironmentGroup can remove other groups

SofiaSazonova:
### Feature or Bugfix

- Bugfix

### Detail
- Now, if the group can't update other group, it also can not remove them.
- 
### Relates
- #1212 

### Security

  - Is the input sanitized?


noah-paige:
So if I understand correctly...

Before we had permissions s.t. - If the invited team can invite other env groups than they can remove others from env

After this PR we have permissions s.t. - If the invited team can update others env group permissions then they can remove others from env
- Currently in dataall `UPDATE_ENVIRONMENT_GROUP` is only added to EnvAdmin as part of `ENVIRONMENT_ALL` permissions set
- Meaning this is the same logic as saying on Env Admin can update permissions or remove teams from the Environment (however invited teams can invite other teams as well if they are given that permission)


2 Pending Qs:

1. Should we allow other non-admin env groups to invite additional groups on the data.all (if they cannot update / delete)? Is there an easy way to disable that permissions temporarily until we introduce "Supervisors" concept mentioned in [issue](https://github.com/data-dot-all/dataall/issues/1235) 
2. In order to close the loop on this bug, should we add a migration script to remove all `REMOVE_ENVIRONMENT_GROUP` permissions from non EnvAdmins with respect to each environment?

SofiaSazonova:
@dlpzx what do you think about restrict access to group management?

dlpzx:
So we have 3 types of permissions:
- INVITE_GROUP that is currently applied to Environment objects as it should 👍 
- UPDATE_GROUP and REMOVE_GROUP, currently applied to Environments, but they should protect EnvironmentGroups 

Besides the object they "protect" we need to verify who is getting access:

|Permission|Should|Has|
|--|--|--|
| INVITE_GROUP| EnvAdmin and other invited groups with permissions | EnvAdmin and other invited groups with permissions 👍 |
| UPDATE_GROUP| EnvAdmin and invited group that invited this group| EnvAdmin only|
| REMOVE_GROUP| EnvAdmin and invited group that invited this group| EnvAdmin and other invited groups with permissions|


We can solve the problems in multiple steps:
1. Remove the REMOVE_GROUP permission from other invited groups with permissions, to mimic UPDATE_GROUP (this PR)
2. Migrate from checking `Environment` to checking `EnvironmentGroup` ---> that means that we would need to update the resourceUri of the UPDATE_GROUP and REMOVE_GROUP permissions and the `uri` used in the checks to use the environment group uri
3. Grant the "inviter" group UPDATE_GROUP and REMOVE_GROUP permissions --> i am not sure how to make it retroactive, we would need to check the RDS table info, maybe there is already an "invitedBy" if we are lucky

noah-paige:
So ultimately we will get to a state where on invite_group:
- Env Group is created w/ assignd permissions
- EnvAdmin is granted UPDATE and REMOVE on the invited group
- Group who invited is granted UPDATE and REMOVE on the invited group
  - I imagine we will need to add a input param for who this group is (otherwise how do you determine for UserA a part of both Group 1 and Group2 both in EnvA?)

For this PR - we will restrict UPDATE and REMOVE to EnvAdmins and create migration script to remove improper permissions correct?

SofiaSazonova:
@noah-paige I also see it this way. 
In my POV we also need to implement permission boundary mechanism for groups: upon invitation, user enters which permissions can be granted by the invited group.

backend/migrations/versions/a991ac7a85a2_drop_remove_group_permissions.py
```diff
@@ -0,0 +1,63 @@
...
+                    ResourcePolicy.principalId != admin_group,
+                    ResourcePolicy.resourceType == Environment.__name__,
+                    ResourcePolicy.resourceUri == env.environmentUri,
+                )
+            )
+            .all()
+        )
+        for group in suspicious_permissions_principals:
+            permissions = ResourcePolicyService.get_resource_policy_permissions(session, group, env.environmentUri)
+            permissions.remove(REMOVE_ENVIRONMENT_GROUP)
```
noah-paige, line 49: can we add logic to remove only if `REMOVE_ENVIRONMENT_GROUP` is in the list of permissions

Otherwise we may get an Error when trying to remove this permission for a env group where it does not exist (i.e. they never were assigned `INVITE_ENVIRONMENT_GROUP` beforehand

noah-paige: UPDATE: After some testing I believe need to change the way we are removing `REMOVE_ENVIRONMENT_GROUP`, the below worked (or similar logic):

```
permissions = [permission.name for permission in permissions if permission.name != REMOVE_ENVIRONMENT_GROUP]
```

noah-paige (commented):
Small changes needed on migration script - then I think this PR is good and lets open a new issue for the other points we discussed as enhancements

backend/migrations/versions/a991ac7a85a2_drop_remove_group_permissions.py
SofiaSazonova: I always forget, that in JS remove is protected, but in python it's not! Argh)

backend/migrations/versions/a991ac7a85a2_drop_remove_group_permissions.py
```diff
@@ -0,0 +1,64 @@
...
+                    ResourcePolicy.resourceType == Environment.__name__,
+                    ResourcePolicy.resourceUri == env.environmentUri,
+                )
+            )
+            .all()
+        )
+        for group in suspicious_permissions_principals:
+            permissions = ResourcePolicyService.get_resource_policy_permissions(session, group, env.environmentUri)
+            if REMOVE_ENVIRONMENT_GROUP in permissions:
+                permissions.remove(REMOVE_ENVIRONMENT_GROUP)
```
noah-paige, line 50: I think you need to be checking `permission.name` for each permission rather than the entire `permission` object that is returned from `get_resource_policy_permissions()`

Can refer to the below
```
permissions = [permission.name for permission in permissions if permission.name != REMOVE_ENVIRONMENT_GROUP]
```

noah-paige (approved):
looks good and tested migration script locally - approving

Issue #998, opened 2024-01-24
Add Maintenance UI View / Window


noah-paige:
**Is your idea related to a problem? Please describe.**
Need a way to describe what new features had recently been introduced to data.all on latest releases and if there should be any warnings, deprecation notices, etc.

**Describe the solution you'd like**
To communicate new features or maintenance actions - add a maintenance UI view or window where we can post messages and info about the current status of the platform etc..

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

zsaltys:
My proposal to implement this:

1) add a settings page which is only visible to data all administrators
2) on this settings page add a button to put data.all into maintenance mode on/off.. store the state in RDS.
3) when maintenace mode is on all users except da admins see maintenance window
4) during maintenance mode all graphql calls are blocked except those needed to make maintenance mode work
5) no new cron jobs / periodic jobs should start while maintenance mode is on.. no stacks should be updated.
6) da admins only see the settings page while maintenance mode is on.. cannot navigate to anything else

TejasRGitHub:
Hi @zsaltys. Thanks for the insight. 

**WIP**
From the description by Noah on this issue .We also need a place to show what's in the new release 

Where do we get this information about new features ? 
One place could be from the release notes of the UI Or make another readme/ text file which will contain information about the new release 

Where should this UI be ? 

Do we wan to keep a time limit to how long we want to show this UI ? 

**Maintenance Window** 

The maintenance window UI which will be only visible to the data.all admin group could be added in the Admin settings page as another tab.
<img width="1479" alt="image" src="https://github.com/data-dot-all/dataall/assets/71188245/8fcda7a9-ff79-47fd-a8f2-889430885eac">

The UI will show a button to create a maintenance window and a table showing past maintenance windows. 
When clicked the button will show a pop-up to confirm the actions and then will start the maintenance window. The user will be shown if its okay to start the deployment ( For e.g. when clicked the maintenance window will show IN-PROGRESS and then once it is confirmed that no more pending graphQL calls, stack updates remain then the UI will  change to READY-FOR-DEPLOYMENT. Once the deployment is done the data.all admin can then remove the maintenance window ) 

A `maintenance` table will have to be created with attributes like 
User - Who started the window
Start time
End time 
Status of maintenance window 

List of tasks that need to be shut off for maintenance window 
[...]

PR #1236, opened 2024-04-30
[Gh 998] Maintenance window

TejasRGitHub:
### Feature or Bugfix
- Feature

### Detail
- GH issue with description - https://github.com/data-dot-all/dataall/issues/998

### Instructions on using maintenance window
1. Include yourself in the "DAAdministrators" group.
2. Goto Admin Section 
3. Switch to Maintenance Tab
4. You can put data.all in Read-only mode or No-access mode. Select either mode and click on "Start Maintenance" window button 
5. Once the status of the maintenance window is in ACTIVE state. You can be sure that the all ECS tasks have completed runnning and also the scheduled ECS tasks are turned OFF. ( Check this in AWS accounts for testing ) 
6. If in Read-only mode, then a red banner will appear whenever any user visits data.all UI . In Read-Only mode, all mutation calls ( are blocked ) 
7. If in No-Access mode, then after login a blank page displaying message about data.all maintenance window will appear. In No-Access mode, all graphQL calls are blocked ( Except getGroupForUser and getMaintenanceWindowStatus - which are required for the UI to load minimally). 
8. In order to disable maintenance window, goto the maintenance tab and click stop maintenance window. 

Note- In all the mode, data.all admin user is allowed to perform any gql call and access and modify anything on the UI.   

### Relates
- https://github.com/data-dot-all/dataall/issues/998

### Security

  - Is the input sanitized?
  - Have you used the least-privilege principle? How? Least Privelege


backend/api_handler.py
```diff
@@ -163,6 +148,41 @@ def handler(event, context):
             'schema': SCHEMA,
         }
 
+        query = json.loads(event.get('body'))
+
+        # Logic to block when in maintenance
+        # Check if in some maintenance mode
+        # Check if in maintenance status is not INACTIVE
+        # Check if the user belongs to a 'DAAdministrators' group
+        if config.get_property('modules.maintenance.active'):
```
rbernotas, line 157: Suggestion:  this whole if block might be simplified like the following, create vars for each of the calls, it is a bit more readable this way as well.

Screenshot 2024-05-03 at 4 06 50 PM

`
maintenance_active = config.get_property('modules.maintenance.active')
maintenance_window_mode = MaintenanceService._get_maintenance_window_mode(engine=ENGINE)
maintenance_window_status = MaintenanceService.get_maintenance_window_status(engine=ENGINE).status
is_tenant_admin = TenantPolicyValidationService.is_tenant_admin(groups)

if maintenance_active and maintenance_window_status != MaintenanceStatus.INACTIVE.value and not is_tenant_admin:
    if maintenance_window_mode == MaintenanceModes.NOACCESS.value and query.get('operationName', '') not in MAINTENANCE_ALLOWED_OPERATIONS:
        return send_unauthorized_response(
            query=query,
            message='Access Restricted: data.all is currently undergoing maintenance, and your actions are temporarily blocked.',
        )
    elif maintenance_window_mode == MaintenanceModes.READONLY.value and query.get('query', '').split()[0] == 'mutation':
        return send_unauthorized_response(
            query=query,
            message='Access Restricted: data.all is currently undergoing maintenance, and your actions are temporarily blocked.',
        )

`

backend/dataall/base/utils/api_handler_utils.py
```diff
@@ -0,0 +1,47 @@
...
+def get_cognito_groups(claims):
+    if not claims:
+        raise ValueError(
+            'Received empty claims. ' 'Please verify authorizer configuration',
+            claims,
+        )
+    groups = list()
+    saml_groups = claims.get('custom:saml.groups', '')
+    if len(saml_groups):
+        groups: list = saml_groups.replace('[', '').replace(']', '').replace(', ', ',').split(',')
```
rbernotas, line 15: Could also possibly use a translation table here instead of the successive replace statements?

Screenshot 2024-05-03 at 4 17 30 PM



`
translation_table = str.maketrans({'[': None, ']': None, ', ': ','})
groups = []
saml_groups = claims.get('custom:saml.groups', '')

if len(saml_groups) > 0:
    groups = saml_groups.translate(translation_table).split(',')
`

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,119 @@
...
+                        'Maintenance window already in PENDING or ACTIVE state. Cannot start maintenance window. Stop the maintenance window and start again'
+                    )
+                    return False
+                MaintenanceRepository(session).save_maintenance_status_and_mode(
+                    maintenance_status=MaintenanceStatus.PENDING.value, maintenance_mode=mode
+                )
+            # Disable scheduled ECS tasks
+            # Get all the SSM Params related to the scheduled tasks
+            ecs_scheduled_rules = ParameterStoreManager.get_parameters_by_path(
+                region=os.getenv('AWS_REGION', 'eu-west-1'),
```
rbernotas, line 41: Should we really have a default value of eu-west-1 here?  Seems like there should be no default and that we check region value and error if it is not a valid region.

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,119 @@
...
+                parameter_path=f"/dataall/{os.getenv('envname', 'local')}/ecs/ecs_scheduled_tasks/rule",
+            )
+            logger.debug(ecs_scheduled_rules)
+            ecs_scheduled_rules_list = [item['Value'] for item in ecs_scheduled_rules]
+            event_bridge_session = EventBridge(region=os.getenv('AWS_REGION', 'eu-west-1'))
```
rbernotas, line 46: Same as above, curious why we would default to eu-west-1 here.

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,119 @@
...
+                maintenance_record = MaintenanceRepository(session).get_maintenance_record()
+                if maintenance_record.status == MaintenanceStatus.INACTIVE.value:
+                    logger.error('Maintenance window already in INACTIVE state. Cannot stop maintenance window')
+                    return False
+                MaintenanceRepository(session).save_maintenance_status_and_mode(
+                    maintenance_status='INACTIVE', maintenance_mode=''
+                )
+            # Enable scheduled ECS tasks
+            ecs_scheduled_rules = ParameterStoreManager.get_parameters_by_path(
+                region=os.getenv('AWS_REGION', 'eu-west-1'),
```
rbernotas, line 70: Same question about default region as above.

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,119 @@
...
+                parameter_path=f"/dataall/{os.getenv('envname', 'local')}/ecs/ecs_scheduled_tasks/rule",
+            )
+            logger.debug(ecs_scheduled_rules)
+            ecs_scheduled_rules_list = [item['Value'] for item in ecs_scheduled_rules]
+            event_bridge_session = EventBridge(region=os.getenv('AWS_REGION', 'eu-west-1'))
```
rbernotas, line 75: Same question about default region as above.

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,119 @@
...
+    @staticmethod
+    def get_maintenance_window_status(engine):
+        logger.info('Checking maintenance window status')
+        try:
+            with engine.scoped_session() as session:
+                maintenance_record = MaintenanceRepository(session).get_maintenance_record()
+                if maintenance_record.status == MaintenanceStatus.PENDING.value:
+                    # Check if ECS tasks are running
+                    ecs_cluster_name = ParameterStoreManager.get_parameter_value(
+                        region=os.getenv('AWS_REGION', 'eu-west-1'),
```
rbernotas, line 93: Same question about default region as above.

frontend/src/authentication/components/AuthGuard.js
```diff
@@ -1,18 +1,58 @@
...
+        ) &&
+        response.data.getMaintenanceWindowStatus.mode === 'NO-ACCESS' &&
+        !groups.includes('DAAdministrators')
+      ) {
+        setNoAccessMaintenanceFlag(true);
+      } else {
+        setNoAccessMaintenanceFlag(false);
+      }
+    }
+  };
```
rbernotas, line 45: Suggestion, can we use a try/catch block here like the following:

Screenshot 2024-05-03 at 4 48 52 PM

`
const checkMaintenanceMode = async () => {
    try {
        const response = await client.query(getMaintenanceStatus());
        const maintenanceStatus = response?.data?.getMaintenanceWindowStatus;

        if (maintenanceStatus && [PENDING_STATUS, ACTIVE_STATUS].includes(maintenanceStatus.status) && maintenanceStatus.mode === 'NO-ACCESS' && !groups.includes('DAAdministrators')) {
            setNoAccessMaintenanceFlag(true);
        } else {
            setNoAccessMaintenanceFlag(false);
        }
    } catch (error) {
        // Handle error
    }
};

`

frontend/src/design/components/layout/DefaultNavbar.js
```diff
@@ -17,9 +26,55 @@ const useStyles = makeStyles((theme) => ({
...
+          setMaintenanceFlag(true);
+        }
+      } else {
+        const error = response.errors
+          ? response.errors[0].message
+          : 'Could not fetch status of maintenance window';
+        dispatch({ type: SET_ERROR, error });
+      }
+    }
+  }, [client]);
```
rbernotas, line 53: Can we use a try/catch block here like:

Screenshot 2024-05-03 at 4 52 12 PM

`
useEffect(() => {
    const fetchMaintenanceStatus = async () => {
        try {
            if (client) {
                const response = await client.query(getMaintenanceStatus());
                const maintenanceStatus = response?.data?.getMaintenanceWindowStatus;
                
                if (maintenanceStatus && [ACTIVE_STATUS, PENDING_STATUS].includes(maintenanceStatus.status)) {
                    setMaintenanceFlag(true);
                }
            }
        } catch (error) {
            const errorMessage = error?.response?.data?.errors?.[0]?.message || 'Could not fetch status of maintenance window';
            dispatch({ type: SET_ERROR, error: errorMessage });
        }
    };

    fetchMaintenanceStatus();
}, [client]);

`

frontend/src/modules/Administration/views/AdministrationView.js
```diff
@@ -13,12 +13,21 @@ import { useState } from 'react';
...
+  AdministrationTeams,
+  DashboardViewer,
+  MaintenanceViewer
+} from '../components';
 
 const tabs = [
   { label: 'Teams', value: 'teams' },
   { label: 'Monitoring', value: 'dashboard' }
 ];
+// Using 'config' as the isModulesEnabled needs Modules.MAINTENANCE which involves much bigger setup which is not needed for maintenance module
```
dlpzx, line 27: There is a utility in the frontend to check if a module is enabled :) In `frontend/src/utils/helpers/moduleUtils.js` 


You can do something like:
```
import {
  isModuleEnabled,
  ModuleNames
} from 'utils';


if isModuleEnabled(ModuleNames.MAINTENANCE) {
...
}
```

frontend/src/modules/Administration/components/MaintenanceViewer.js
```diff
@@ -0,0 +1,438 @@
...
+} from '@mui/material';
+import React, { useCallback, useEffect, useState } from 'react';
+import { Article, CancelRounded, SystemUpdate } from '@mui/icons-material';
+import { LoadingButton } from '@mui/lab';
+import { Label } from '../../../design';
+import {
+  getMaintenanceStatus,
+  stopMaintenanceWindow,
+  startMaintenanceWindow
+} from '../../../services/graphql/MaintenanceWindow';
```
dlpzx, line 23: In the frontend we agreed on using global imports for common resources and relative imports inside each module or component. 

Anything outside of the `modules` directory is considered common. Common submodules such as `services` (refers to services in `frontend/src/services`, `utils`, `design`... can be imported directly:


```
import { Label } from 'design';
import { useClient } from 'services';
```

In another comment I also talked about the layout, which is related to the imports.For the maintenance services we would import them from the module as relative imports (not sure about the amount of `../`)
```
import {
  getMaintenanceStatus,
  stopMaintenanceWindow,
  startMaintenanceWindow
} from '../../services';

```

backend/dataall/modules/maintenance/api/types.py
```diff
@@ -0,0 +1,8 @@
+"""Defines the object types of the Maintenance activity"""
+
+from dataall.base.api import gql
+
+
+Maintenance = gql.ObjectType(
+    name='Maintenance', fields=[gql.Field(name='status', type=gql.String), gql.Field(name='mode', type=gql.String)]
```
dlpzx, line 7: nit: `status` should be non-nullable

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,119 @@
...
+                        parameter_path=f"/dataall/{os.getenv('envname', 'local')}/ecs/cluster/name",
+                    )
+                    if Ecs.is_task_running(cluster_name=ecs_cluster_name):
```
dlpzx, line 96: If we run `is_task_running` without startedBy we take the risk of getting other tasks (scheduled tasks, deployment tasks..)

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,119 @@
...
+                        return maintenance_record
+                    else:
+                        maintenance_record.status = MaintenanceStatus.ACTIVE.value
+                        session.commit()
+                        return maintenance_record
+                else:
+                    logger.info('Maintenance window is not in PENDING state')
```
dlpzx, line 103: nit: can we show a better log for this? Maybe show the actual status

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,119 @@
...
+                    logger.error(
+                        'Maintenance window already in PENDING or ACTIVE state. Cannot start maintenance window. Stop the maintenance window and start again'
+                    )
+                    return False
+                MaintenanceRepository(session).save_maintenance_status_and_mode(
+                    maintenance_status=MaintenanceStatus.PENDING.value, maintenance_mode=mode
+                )
+            # Disable scheduled ECS tasks
+            # Get all the SSM Params related to the scheduled tasks
+            ecs_scheduled_rules = ParameterStoreManager.get_parameters_by_path(
```
dlpzx, line 40: nit: lines 40-46 are duplicated in 69-75, maybe we can create a _get_ecs_rules()

backend/api_handler.py
```diff
@@ -60,27 +66,6 @@ def adapted(obj, info, **kwargs):
...
-    return groups
-
-
-def get_custom_groups(user_id):
-    service_provider = ServiceProviderFactory.get_service_provider_instance()
-    return service_provider.get_groups_for_user(user_id)
-
-
 def handler(event, context):
     """Sample pure Lambda function
```
dlpzx, line 70: the handler function looks very messy, it is trying to do many things:
- setting the api context
- adding tenant permissions
- checking maintenance
- checking reauth
Can we divide it into smaller functions?

backend/search_handler.py
```diff
@@ -21,21 +28,53 @@ def handler(event, context):
...
+            return {
+                'statusCode': 200 if success else 400,
+                'headers': {
+                    'content-type': 'application/json',
+                    'Access-Control-Allow-Origin': '*',
+                    'Access-Control-Allow-Headers': '*',
+                    'Access-Control-Allow-Methods': '*',
+                },
+                'body': json.dumps(response),
+            }
```
dlpzx, line 80: It looks almost like the api handler, as next steps for the team: should we remove the search handler completely and integrate it into the api handler?

backend/dataall/modules/maintenance/services/maintenance_service.py
TejasRGitHub: Yes. That's right!

backend/dataall/modules/maintenance/services/maintenance_service.py
TejasRGitHub: Sure . Will do!

frontend/src/services/hooks/useGroups.js
```diff
@@ -16,12 +16,9 @@ export const useGroups = () => {
...
     } else if (process.env.REACT_APP_CUSTOM_AUTH) {
-      if (!auth.user) {
-        dispatch({
-          type: SET_ERROR,
-          error: 'Cannot Set User Groups as the User is not defined'
-        });
-      }
+      // Returning when auth.user is not present
+      // Not dispatching error as useGroups is triggered in auth guard when the user is not authenticated
+      if (!auth.user) return;
```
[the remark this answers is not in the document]
TejasRGitHub: When the user is just authenticated but the auth.user is not yet initialized this dispatch error is thrown on the screen due to code changes here - frontend/src/authentication/components/AuthGuard.js L27 
image

This would cause unnecessary confusion on the UI and thus I have instead added return

tests/modules/maintenance/test_mainenance_gql.py
```diff
@@ -0,0 +1,124 @@
...
+        maintenance_record = Maintenance(status='INACTIVE', mode='')
+        session.add(maintenance_record)
+        session.commit()
+    yield
+    with db.scoped_session() as session:
+        maintenance_record = session.query(Maintenance).one()
+        session.delete(maintenance_record)
+        session.commit()
+
+
```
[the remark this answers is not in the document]
TejasRGitHub: Ahh. Will fix this. Thanks!

tests/modules/maintenance/test_mainenance_gql.py
```diff
@@ -0,0 +1,124 @@
...
+                stopMaintenanceWindow
+            }
+            """,
+        username='alice',
+        groups=['DAAdministrators', 'Engineers'],
+    )
+
+    assert response
+    assert response.data.stopMaintenanceWindow is True
+
```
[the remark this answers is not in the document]
TejasRGitHub: Will add this test. Thanks!

backend/dataall/base/utils/api_handler_utils.py
TejasRGitHub: Yes. I can do this one! More cleaner

backend/dataall/modules/maintenance/services/maintenance_service.py
TejasRGitHub: I am taking the default value from the other similar boto3 wrappers which are defined in dataall/base/aws/ folder. I am in agreement that we should through error if the AWS_REGION env var is nt set. Maybe we should do it for other boto wrappers. 

@dlpzx , what are you views ? Is there any particular reason to keep the eu-west-1

frontend/src/authentication/components/AuthGuard.js
TejasRGitHub: @rbernotas, for this async method, the errors from client.query are attached in the `response`. This doesn't throw an exception. 

For catching unexpected exception , this function is catching exception in `catch` block defined in the useEffect hook. 
```js
if (config.modules.maintenance.active === true) {
      if (client) {
        checkMaintenanceMode().catch((e) => dispatch({ type: SET_ERROR, e }));
      }
}
```

frontend/src/design/components/layout/DefaultNavbar.js
TejasRGitHub: Thanks for pointing out that there is no try catch. I will decompose this into a function and call it in useEffect. I will use the `.catch` method to catch any unexpected exceptions. Similar to how it is done in `AuthGuard.js` -> `useEffect()`

backend/dataall/modules/maintenance/api/resolvers.py
```diff
@@ -0,0 +1,26 @@
...
+from dataall.modules.maintenance.api.types import Maintenance
+from dataall.modules.maintenance.services.maintenance_service import MaintenanceService
+
+
+def start_maintenance_window(context: Context, source: Maintenance, mode: str):
+    """Starts the maintenance window"""
+    if mode not in [item.value for item in list(MaintenanceModes)]:
+        raise Exception('Mode is not conforming to the MaintenanceModes enum')
+    # Check from the context if the groups contains the DAAAdminstrators group
+    if not TenantPolicyValidationService.is_tenant_admin(context.groups):
```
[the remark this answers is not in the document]
TejasRGitHub: Just curious, any particular reason this should be done in services .Is it because of the use of `TenantPolicyValidationService ` which inherently is a service and thus service calls should be in the service layer ? Asking this because we validate before calling the Service layers , like for e.g. in create_dataset method, we call the `RequestValidator.validate_creation_request(input)` before calling the services layer functions

frontend/src/services/graphql/MaintenanceWindow/index.js
```diff
@@ -0,0 +1,3 @@
+export * from './getMaintenanceStatus';
+export * from './stopMaintenanceWindow';
+export * from './startMaintenanceWindow';
```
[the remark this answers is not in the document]
TejasRGitHub: Sure. I will move it to the `modules` folder.

frontend/src/modules/Administration/components/MaintenanceViewer.js
TejasRGitHub: Addressed this in the next update! Thanks for the insight into the structure

backend/api_handler.py
TejasRGitHub: Thanks for the suggestion @rbernotas . Updated in the next commit

frontend/src/services/hooks/useGroups.js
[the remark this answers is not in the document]
TejasRGitHub: Another option would be that I can have a console.log and add a eslint ignore line so that eslint doesn't create problems while building

backend/dataall/modules/maintenance/api/types.py
TejasRGitHub: Thanks!!. Made updates in the next commit

backend/api_handler.py
TejasRGitHub: Yes. Updated in the next commit

backend/dataall/modules/maintenance/services/maintenance_service.py
TejasRGitHub: Updated in the next commit. Thanks!

TejasRGitHub:
Hi @dlpzx , Updated code. Please take a look and review whenever you get time. 

Testing -
Testing on local dev ( with and without maintenance module enabled ) ✅ 
Testing on aws data.all instance : TODO

backend/dataall/base/aws/parameter_store.py
```diff
@@ -37,6 +37,18 @@ def get_parameter_value(AwsAccountId=None, region=None, parameter_path=None):
             raise Exception(e)
         return parameter_value
 
+    @staticmethod
+    def get_parameters_by_path(AwsAccountId=None, region=None, parameter_path=None):
```
anushka-singh, line 41: What does this function do?

TejasRGitHub: This function is used to fetch all the parameters belonging to a hierachy of path name. For e.g. /path1/param1 and /path1/param2, this will fetch all the path under the /path1 

This is used to get the names of all the scheduled ECS tasks created in data.all

frontend/src/modules/Administration/views/AdministrationView.js
```diff
@@ -14,12 +14,18 @@
 import { Link as RouterLink } from 'react-router-dom';
 import { ChevronRightIcon, useSettings } from 'design';
 import { AdministrationTeams, DashboardViewer } from '../components';
+import { MaintenanceViewer } from '../../Maintenance/views/MaintenanceViewer';
+import { isModuleEnabled, ModuleNames } from '../../../utils';
```
dlpzx, line 18: Change this import to relative import to absolute import ( `from utils`)

frontend/src/authentication/components/AuthGuard.js
```diff
@@ -1,18 +1,59 @@
...
+import { useEffect, useState } from 'react';
 import { Navigate, useLocation } from 'react-router-dom';
 import { Login } from '../views/Login';
 import { useAuth } from '../hooks';
 import {
+  isModuleEnabled,
+  ModuleNames,
   RegexToValidateWindowPathName,
   WindowPathLengthThreshold
 } from '../../utils';
```
dlpzx, line 11: All these relative imports from line 11 to 19, both included, should be absolute imports (even the one in line 11 which was introduced in this PR).

frontend/src/modules/Administration/components/MaintenanceViewer.js
dlpzx: There are still some imports that need to be fixed (e.g. globalErrors, design) Basically, anything inside the top-level directories except for modules

frontend/src/services/hooks/useGroups.js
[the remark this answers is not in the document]
dlpzx: Let me confirm I understood the situation: when we load AuthGuard with or without maintenance enabled, we call useGroups (to get the groups and compare with DAAdmins). In other situations the user is already authenticated when we call useGroups, but in Authguard it might happen that it is not authenticated yet?

frontend/src/modules/Maintenance/views/MaintenanceViewer.js
```diff
@@ -0,0 +1,438 @@
+import {
```
dlpzx, line 1: MInor: MaintenanceViewer is not a view, it is a component of a view, so we should have something like `components/MaintenanceViewer.js` instead

backend/api_handler.py
```diff
@@ -131,56 +116,64 @@ def handler(event, context):
...
-        try:
-            reauth_apis = ParameterStoreManager.get_parameter_value(
-                region=os.getenv('AWS_REGION', 'eu-west-1'), parameter_path=f'/dataall/{ENVNAME}/reauth/apis'
-            ).split(',')
-        except Exception:
-            log.info('No ReAuth APIs Found in SSM')
-            reauth_apis = None
+        query = json.loads(event.get('body'))
+
+        isBlockedResponse = validate_and_block_if_maintenance_window(query=query, groups=groups)
```
dlpzx, line 133: I am not sure is Blocked returns anything

dlpzx: same for `check_reauth`

backend/dataall/modules/maintenance/api/enums.py
```diff
@@ -0,0 +1,18 @@
+"""Contains the enums used in maintenance module"""
+
```
dlpzx, line 2: For graphql we use GraphQLEnum class tha maps enums to what we get from graphql. 

Here is an example in the notebooks module
```python
from dataall.base.api.constants import GraphQLEnumMapper


class SagemakerNotebookRole(GraphQLEnumMapper):
    """Describes the SageMaker Notebook roles"""

    CREATOR = '950'
    ADMIN = '900'
    NO_PERMISSION = '000'
```

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        # Check from the context if the groups contains the DAAAdminstrators group
+        if groups is None:
+            groups = []
+        if not TenantPolicyValidationService.is_tenant_admin(groups):
+            raise Exception('Only data.all admin group members can start maintenance window')
+
+        logger.info('Putting data.all into maintenance')
+        try:
+            with engine.scoped_session() as session:
```
dlpzx, line 40: See comment in resolvers first. This should change to:

`with get_context().db_engine.scoped_session() as session`

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,153 @@
...
+            1. Perform validation to check if the user belongs to the DAAdministrators group
+            2. Put the maintenance window status to PENDING and update the maintenance mode
+            3. Get all the ECS Scheduled tasks and disable the schedule for them
+        @param engine: db engine
+        @param mode: mode to set for maintenance window
+        @param groups: user groups from context.groups
+        @return: returns True if successful or False
+        """
+        # Check from the context if the groups contains the DAAAdminstrators group
+        if groups is None:
```
dlpzx, line 33: See comment in resolvers first. This should change to:
`groups = get_context().groups if get_context().groups is not None else []`

dlpzx: Same for all other service calls

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,153 @@
...
+            with engine.scoped_session() as session:
+                maintenance_record = MaintenanceRepository(session).get_maintenance_record()
+                if maintenance_record.status == MaintenanceStatus.INACTIVE.value:
+                    logger.error('Maintenance window already in INACTIVE state. Cannot stop maintenance window')
+                    return False
+                MaintenanceRepository(session).save_maintenance_status_and_mode(
+                    maintenance_status='INACTIVE', maintenance_mode=''
+                )
+            # Enable scheduled ECS tasks
+            ecs_scheduled_rules_list = MaintenanceService._get_ecs_rules()
```
dlpzx, line 91: nit: rename event_bridge_session to event_bridge_client. I have missunderstood this line twice hahahahh because it says session I understood it returns a boto3 session and then the session is used to directly call the boto3 function

dlpzx (commented):
Did a more in depth review!

petrkalos:
I think this feature should be covered by the new [integration tests](https://github.com/data-dot-all/dataall/tree/main/tests_new/integration_tests)

backend/api_handler.py
```diff
@@ -207,20 +200,60 @@ def handler(event, context):
...
+            and not isAdmin
+        ):
+            if query.get('operationName', '') not in MAINTENANCE_ALLOWED_OPERATIONS:
+                return send_unauthorized_response(
+                    query=query,
+                    message='Access Restricted: data.all is currently undergoing maintenance, and your actions are temporarily blocked.',
+                )
+        elif (
+            (maintenance_mode == MaintenanceModes.READONLY.value)
+            and (maintenance_status is not MaintenanceStatus.INACTIVE.value)
```
anushka-singh, line 226: MaintenanceStatus.INACTIVE is an enum, so you should compare directly with the enum instead of its value.

backend/api_handler.py
```diff
@@ -207,20 +200,60 @@ def handler(event, context):
...
+            and not isAdmin
+        ):
+            # If its mutation then block and return
+            if query.get('query', '').split()[0] == 'mutation':
```
anushka-singh, line 230: Lots of issues with using this code. 
I dont like that we are doing a string comparison in the query to match the first word with 'mutation'. It seems super easy to break. For one, if the query string is empty or contains only white spaces, the call to split() will result in an index error because there won't be any elements to access.

The comparison is case-sensitive, so if the query string contains 'Mutation' or 'MUTATION', the condition will not be satisfied.

It assumes that the first word of the query string represents the operation type. However, GraphQL queries can be structured in various ways, and the operation type might not always be the first word.

Use graphql offering OperationDefinitionNode to determine the query operation properly. There are usages of it already in the code.

backend/api_handler.py
```diff
@@ -207,20 +200,60 @@ def handler(event, context):
...
+        maintenance_mode = MaintenanceService._get_maintenance_window_mode(engine=ENGINE)
+        maintenance_status = MaintenanceService.get_maintenance_window_status(engine=ENGINE).status
+        isAdmin = TenantPolicyValidationService.is_tenant_admin(groups)
+
+        if (
+            (maintenance_mode == MaintenanceModes.NOACCESS.value)
+            and (maintenance_status is not MaintenanceStatus.INACTIVE.value)
+            and not isAdmin
+        ):
+            if query.get('operationName', '') not in MAINTENANCE_ALLOWED_OPERATIONS:
```
anushka-singh, line 219: Read my comment about using OperationDefinitionNode to fetch operations from graphql query below.

frontend/src/services/hooks/useGroups.js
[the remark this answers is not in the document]
TejasRGitHub: Yes. That's right. So if the user is not authenticated then this useGroups is called and at that time the `auth.user` is not set so it calls the dispatch showing up on the UI. This occurs not at the login screen but when the user tries to login and then after authentication when the user is sent to the default route (https://github.com/data-dot-all/dataall/blob/bab2f3e6cec36e0de4c9229a25108395910ac2bc/frontend/src/routes.js#L409) there is a fraction of time when this AuthGuard calls the useGroups and at that time the `auth.user` is not set ( as it is still loading the user context after auth ) and this useGroups is triggered , sending a dispatch call which shows up on the UI.

backend/search_handler.py
TejasRGitHub: Yes, the API handler and Search Handler are very similar. I agree we should integrate them into one as next steps. 
I can create a GH issue for this change. 

Not including this change in this PR as it might involve multiple infra changes in the code.

frontend/src/modules/Administration/views/AdministrationView.js
TejasRGitHub: My bad. Addressed in next update

frontend/src/authentication/components/AuthGuard.js
TejasRGitHub: Updated imports for `utils`, `design`, `globalErrors`. For others like `../../modules/Maintenance/services` & `../../modules/Maintenance/views/MaintenanceViewer` I don't know if these can be absolute imports . Can you tell me if you were asking for these to also be absolute imports ?

frontend/src/modules/Maintenance/views/MaintenanceViewer.js
TejasRGitHub: So the reason I kept it as a view is because we are saying `Maintenance` as a module and this module provides as view - which works as a component - for the `AdministrativeView.js` . If MaintenanceViewer.js had a component inside of it then I was planning to add them in `component/` folder. 

I don't have much preference with the two. but I can refactor it to say as a component

backend/api_handler.py
TejasRGitHub: Sure. I can do that . I see the value of reusing them. Thanks @petrkalos!

backend/api_handler.py
TejasRGitHub: Hi @dlpzx , `validate_and_block_if_maintenance_window` and `check_reauth` both return responses. 

`validate_and_block_if_maintenance_window` returns response if its unauthorized for that type of maintenance window


`check_reauth` -> returns response if there is re_auth operation.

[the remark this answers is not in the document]
TejasRGitHub: These seem to be newer lambda's which are added. Thanks for pointing this out @petrkalos !! 

If these are only called at the time of deployment then I think we don't need to add the maintenance logic to block changes. 

The logic to block actions was for actions which a user could take from UI OR make an API request to modify data.all constructs ( datasets, environments, etc ) 

For deployments , we assume that the one deploying will have some tribal information about deployment and its impact. With the maintenance window, even if the maintenance window is not ready ( in ACTIVE state ) , deployment's could be started as we do not block anything on CodePipeline. Maintenance window is to make a safe environment but not enforce it yet ( maybe in the future we can also enforce it ). Maintenance window is not just to be used for deployments but to create a window in which data.all admins could make any other hot fixes ( like updating RDS tables, etc )  

@petrkalos , Are these lambda's behind API Gateway ?

backend/api_handler.py
```diff
@@ -131,56 +116,64 @@ def handler(event, context):
...
-            ).split(',')
-        except Exception:
-            log.info('No ReAuth APIs Found in SSM')
-            reauth_apis = None
+        query = json.loads(event.get('body'))
+
+        isBlockedResponse = validate_and_block_if_maintenance_window(query=query, groups=groups)
+        if isBlockedResponse is not None:
+            return isBlockedResponse
+        isReauthResponse = check_reauth(query=query, auth_time=claims['auth_time'], username=username)
```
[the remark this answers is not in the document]
TejasRGitHub: Thanks . For of habit. Will make those changes and update the names as well.

frontend/src/modules/Administration/views/AdministrationView.js
TejasRGitHub: Fixed. Thanks! Missed these 😞

backend/dataall/modules/maintenance/services/maintenance_service.py
TejasRGitHub: Updated and fixed

backend/api_handler.py
TejasRGitHub: maintenance_status contains the value of the MaintenanceStatus.INACTIVE enum. Let me check if I can get an enum there instead

backend/api_handler.py
TejasRGitHub: Thanks a lot @anushka-singh  for pointing this and explanation about the potential issue. 

Let me check what is available in the ariadne lib for fetching this information.

Can you please point me where is this already used in code ?

backend/api_handler.py
```diff
@@ -207,20 +200,60 @@ def handler(event, context):
...
+                    print(f'No policy found for Team {group}. Attaching TENANT_ALL permissions')
+                    TenantPolicyService.attach_group_tenant_policy(
+                        session=session,
+                        group=group,
+                        permissions=TENANT_ALL,
+                        tenant_name=TenantPolicyService.TENANT_NAME,
+                    )
+        return groups
+    except Exception as e:
+        print(f'Error managing groups due to: {e}')
```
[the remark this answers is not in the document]
TejasRGitHub: updated in the next git push . Thanks!

backend/api_handler.py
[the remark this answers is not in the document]
petrkalos: No, they are invoked as [TriggerFunctions](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.triggers.TriggerFunction.html) on every deployment

backend/dataall/modules/maintenance/api/resolvers.py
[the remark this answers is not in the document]
TejasRGitHub: @dlpzx , I have moved this to the services

backend/dataall/modules/maintenance/api/enums.py
TejasRGitHub: Used the `GraphQLEnumMapper ` instead of Enum

backend/api_handler.py
```diff
@@ -38,6 +41,9 @@
 Worker.queue = SqsQueue.send
 
 
+MAINTENANCE_ALLOWED_OPERATIONS = ['getGroupsForUser', 'getMaintenanceWindowStatus']
```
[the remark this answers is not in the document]
TejasRGitHub: so this is used to allow `'getGroupsForUser', 'getMaintenanceWindowStatus'` operations when the maintenance window mode is NO-ACCESS ( in which all graphql - query and mutations - will be blocked ) so that the UI works. These are the minimally needed functions. I don't knw if there would be a case in someone wants to keep this configurable, but please let me know if you see any use case of this as such.

TejasRGitHub:
> I think this feature should be covered by the new [integration tests](https://github.com/data-dot-all/dataall/tree/main/tests_new/integration_tests)

I have added the tests in the tests/modules/maintenance. 

For the integration tests, i think they are using the cognito idp to authenticate. As we use our custom auth, I don't think I can add tests easily. I think this could be a part of another PR where-in the same integration tests can be run with a custom IDP user

backend/api_handler.py
TejasRGitHub: Implemented a parser and used graphQL utilities method to fetch the type of graphQL

TejasRGitHub:
Updated code 

Tested -
Local dev UI and sanity checks - ✅ 
AWS deployed data.all - Tested for READ-ONLY and NO-ACCESS modes - ✅

backend/api_handler.py
```diff
@@ -131,82 +111,31 @@ def handler(event, context):
...
-
-        except Exception as e:
-            print(f'Error managing groups due to: {e}')
-            groups = []
 
-        set_context(RequestContext(ENGINE, username, groups, user_id))
+        groups: list = extract_groups(user_id=user_id, claims=claims)
+        attach_tenant_policy_for_groups(groups=groups)
 
+        # Set Context
```
anushka-singh, line 118: nit: Dont really need this comment

backend/dataall/base/utils/api_handler_utils.py
```diff
@@ -0,0 +1,181 @@
...
+from dataall.base.services.service_provider_factory import ServiceProviderFactory
+from dataall.core.permissions.services.tenant_permissions import TENANT_ALL
+from dataall.core.permissions.services.tenant_policy_service import TenantPolicyService
+from dataall.modules.maintenance.api.enums import MaintenanceModes, MaintenanceStatus
+from dataall.modules.maintenance.services.maintenance_service import MaintenanceService
+from dataall.base.config import config
+from dataall.core.permissions.services.tenant_policy_service import TenantPolicyValidationService
+
+logger = logging.getLogger()
+logger.setLevel(os.environ.get('LOG_LEVEL', 'INFO'))
```
anushka-singh, line 18: What is this used for? We are using "log" everywhere else.

TejasRGitHub: Do you mean why are we using the setLevel command ?

backend/dataall/modules/maintenance/api/resolvers.py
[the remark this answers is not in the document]
dlpzx: It is just an application agreement. All business logic, which includes data.all custom permissions are part of the service layer

backend/dataall/modules/maintenance/services/maintenance_service.py
dlpzx: We have both types, I don't mind event_bridge is fine as well

backend/dataall/base/aws/event_bridge.py
dlpzx, line 1: I think you placed the event_bridge class in base because it enables and disables the ECS tasks. But this class is only. used by the maintenance module. We can place it directly in the maintenance/aws package

backend/dataall/modules/maintenance/api/resolvers.py
```diff
@@ -0,0 +1,20 @@
+from dataall.base.api.context import Context
+from dataall.core.permissions.services.tenant_policy_service import TenantPolicyValidationService
```
dlpzx, line 2: Unused import

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,149 @@
...
+            raise Exception('Only data.all admin group members can stop maintenance window')
+        logger.info('Stopping maintenance mode')
+        try:
+            with get_context().db_engine.scoped_session() as session:
+                maintenance_record = MaintenanceRepository(session).get_maintenance_record()
+                if maintenance_record.status == MaintenanceStatus.INACTIVE.value:
+                    logger.error('Maintenance window already in INACTIVE state. Cannot stop maintenance window')
+                    return False
+                MaintenanceRepository(session).save_maintenance_status_and_mode(
+                    maintenance_status='INACTIVE', maintenance_mode=''
```
dlpzx, line 84: Use Maintenence Enum instead of hardcoded INACTIVE

backend/dataall/modules/maintenance/services/maintenance_service.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        Start maintenance window by performing following actions
+            1. Perform validation to check if the user belongs to the DAAdministrators group
+            2. Put the maintenance window status to PENDING and update the maintenance mode
+            3. Get all the ECS Scheduled tasks and disable the schedule for them
+        @param mode: mode to set for maintenance window
+        @return: returns True if successful or False
+        """
+        # Check from the context if the groups contains the DAAAdminstrators group
+        groups = get_context().groups if get_context().groups is not None else []
```
dlpzx, line 32: I thought get_context().groups returned an empty list when the user belonged to no group.... good finding

backend/migrations/versions/b833ad41db68_maintenance_window_schema.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+
+import os
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy import Column, String, orm
+
+from dataall.base.db import get_engine, has_table, Base
+
```
dlpzx, line 16: Can you merge `main` into this branch? I think we will face some migration conflicts

backend/dataall/modules/maintenance/services/maintenance_service.py
dlpzx: `groups: list = extract_groups(user_id, claims)`?

backend/search_handler.py
dlpzx: Yes, but let's create an issue

frontend/src/authentication/components/AuthGuard.js
dlpzx: This is a very particular case, because in general frontend modules do not influence shared components, but this is a banner on the whole screen. 

**OPTION 1:** maintenance as part of the shared elements in FE
- move whatever is in `modules/Maintenance/services' to `services`
- move `modules/Maintenance/components/MaintenanceViewer` to `design/components`

We import services and viewer with absolute imports in
- frontend/src/authentication/components/AuthGuard.js
- frontend/src/modules/Administration/views/AdministrationView.js


**OPTION 2**: maintenance as module in FE modules
We keep it as it is now.


Downside of OPTION 1, for a new developer it is more difficult to find Maintenance code. But there is a reason why in the FE we enforce absolute imports all over the place, it has to do with the way the FE modularization is intended. 
The idea is that all packages except for `modules` expose some methods that all the other packages can use, the methods exposed are chosen in the index.js. Code that is not in the index should not be used by other packages. This way working inside each of the packages sets a boundary of the responsibilities "this are the functions in the index.js anything else I can touch as I please".

`modules` use the global packages and develop additional code inside the module. **Removing one module should not break anything!** Right now if we remove the Maintenance module there will be errors, that's why I think we should go for OPTION 1

backend/dataall/base/utils/api_handler_utils.py
TejasRGitHub: the setLevel method is used mostly at the top of logger initialization. I am adding it here to set the log level to whatever is set for each lambda in their environment variable - which is available via the `os.environ.get('LOG_LEVEL', 'INFO')`.

backend/dataall/base/aws/event_bridge.py
TejasRGitHub: Okah. I can refactor it to keep it in maintenance/aws package

backend/dataall/modules/maintenance/services/maintenance_service.py
TejasRGitHub: This is been used only for initializing the groups list when the API handler is called. This takes in the claims which come from the cognito/custom authorizer.

backend/migrations/versions/b833ad41db68_maintenance_window_schema.py
TejasRGitHub: Will do sure!!

backend/dataall/modules/maintenance/services/maintenance_service.py
TejasRGitHub: Addressed in the next update

backend/search_handler.py
TejasRGitHub: Created GH issue - https://github.com/data-dot-all/dataall/issues/1302

dlpzx (approved):
Looks good! I was a bit unsure about how the maintenance mode disabled was going to behave in the frontend. My concern was that the import `import { MaintenanceViewer } from '../../Maintenance/components/MaintenanceViewer';` in `frontend/src/modules/Administration/views/AdministrationView.js` was going to result in errors, but I tested it locally and because it is not used in the tabs it works normal.

If we wanted to be ultra-clean we would check `if (isModuleEnabled(ModuleNames.MAINTENANCE))` before importing and using the MaintenanceViewer, but this is not required.

Approving!

petrkalos (approved):

Issue #1165, opened 2024-04-11
Alembic migrations miss drop_table statement


SofiaSazonova:
### Describe the bug

After this [PR](https://github.com/data-dot-all/dataall/pull/1084) we encounter a problem for fresh deployments.
It seems, that alembic prepares migrations in batch, so, for fresh deployments with empty database all `has_table` functions will return false. 
So, if we apply all migrations to fresh deployment, we will have all tables from this [PR](https://github.com/data-dot-all/dataall/pull/1084) still present in database. And all further migrations, autogenerated with alembic, will try to delete them.

### How to Reproduce

locally:
```
export PYTHONPATH=backend
export envname=local 
make drop-tables
make upgrade-db 
make generate-migrations
```
generated file will have delete statements for tables:
```
redshiftcluster
redshiftcluster_datasettable
redshiftcluster_dataset
dataset_quality_rule
tag
item_tags
dataset_table_profiling_job
group_member
```

### Expected behavior

After we perform all migrations on empty DB, this tables should not be present in database

### OS

macOs Sonoma 14.2

### Python version

3.11

### AWS data.all version

2.3.0

PR #1240, opened 2024-05-01
Fix Alembic Migration: has table checks

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Fix `has_table()` check to ensure dropping the tables if the exists as part of alembic migration upgrade 

### Relates
- https://github.com/data-dot-all/dataall/issues/1165

### Security

  - Is the input sanitized? N/A
- Are you introducing any new policies/roles/users? No


SofiaSazonova:
1. **Check passed:** All 'extra' tables are not created after 'make upgrade-db' of the clean DB
2. **Problem**. `make generate-migrations` creates a non empty migration. It means, smth was not tracked by our migrations, so maybe, we have to create another migration script.
 @dlpzx  have a look please

```
def upgrade():
    op.alter_column('dataset_lock', 'isLocked',
               existing_type=sa.BOOLEAN(),
               nullable=True)
    # ### end Alembic commands ###


def downgrade():
    op.alter_column('dataset_lock', 'isLocked',
               existing_type=sa.BOOLEAN(),
               nullable=False)
    # ### end Alembic commands ###
```

SofiaSazonova:
@noah-paige good with me

noah-paige:
> @noah-paige good with me

Should be fixed now

SofiaSazonova:
Could you please add it as a new migration? 
E.g. customer may want to upgrade their current DB (without dropping) and then create their own migrations. 
Changes in past migrations won't help here(

noah-paige:
> Could you please add it as a new migration? E.g. customer may want to upgrade their current DB (without dropping) and then create their own migrations. Changes in past migrations won't help here(

For the has table fix it is only an issue on fresh deployments right? For existing data.all deployments that then run that script it should be fine

For the dataset lock we aren't changing the DB or the way we create the `dataset_locks` table - instead we are changing the model class object for `DatasetLock()` so I think it should be fine here right?

SofiaSazonova:
Hm, I see. Ok, then it should be fine.

SofiaSazonova (approved):

Issue #1241, opened 2024-05-01
Unable to view all environments in the environment list view


TejasRGitHub:
### Describe the bug

After upgrading to 2.4 , user is unable to see environment more than 8 environments on the UI. The user can see that the RDS records have 9 environments and also can navigate to the environment which is not showing up by visiting the org->environment tab -> goto the environment. 

The environment which is now showing up is in proper condition and the user has access to all the teams belonging to each environment

### How to Reproduce

1. Create 9 - 10 environments in data.all
2. Check environment list UI for if the environment is showing 
3. 8 environments will be shown and only 1 page is shown 

### Expected behavior

All environments should be displayed

### OS

Mac

### Python version

3.9

### AWS data.all version

2.4

noah-paige:
Look into this issue and able to recreate the same behavior where the list Env is missing an environment, more specifically:

- 8 Envs --> All show on 1 page (Good)
- 9 Envs --> All show on 1 page (Good)
- 10 Envs --> 9 on 1 page and No Second Page (Error)
- 11 Envs --> 9 on 1 page and 2 on second page (Fine --> but page size should be 10)

Additionally with 10 Envs I looked at the `listEnv` Query and I see the count returned in 10 but the number of nodes (i.e. # of envs) is only 9 -- to do some additionally digging here

noah-paige:
After a bit more digging it looks like the origin of this issue is well summarized in this [post](https://stackoverflow.com/questions/6933478/sqlalchemy-only-one-result-being-returned-when-count-says-there-are-more) 

For the 10 Env Error case above:
After running the query as  `query.limit(page_size).offset((page - 1) * page_size).all()` in `def paginate()` we are receiving back only 9 items and somehow losing 1 record in the process

One way to fix is to add a `.distinct()` option to the Query which then seems to properly return all 10 items

However, a more general practice approach I think would be to add `.order_by()` for any query that we want to paginate which also fixes the issue and ensures that the offset call when we get the next page does not return the same items as the previous page

noah-paige:
Below is a comprehensive list of each query that we paginate reponses to and by what value we are now introducing `order_by()` condition for the query:

By EnvironmentGroup label
- query_user_environments (has outer join)

By EnvironmentGroup groupUri
- query_environment_invited_groups (has inner join)
- query_all_environment_groups
- query_user_environment_groups
- query_user_groups

By ConsumptionRole Name:
- query_all_environment_consumption_roles
[...]

PR #1249, opened 2024-05-03
Add order_by for paginated queries 

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- This PR aims to solve the following

    - (1) for particular queries (identified as ones that perform `.outerjoin()` operations and have results paginated with `paginate()` function - sometimes the returned query results is *less than* the limit set by the pageSize of the paginate function even when the total count is greater than the pageSize 
        -  Ex 1: 11 envs total, `query_user_environments()` returning 9 envs on 1st page + 2 on 2nd page
        -  Ex 2: 10 envs total, `query_user_environments()` returning 9 envs on 1st page + no 2nd page

        -  Believe this is to be happening due to the way SQLAlchemy is "uniquing" the records resulted from an outerjoin and then returning that result back to the frontend

        - Adding a `.distinct()` check on the query ensures each distinct record is returned (tested successfully)

    - (2) Currently we often times do not implement an `.order_by()` condition for the query used in `paginate()` and do not have a stable way of preserving order of the items returned from a query (i.e. when navigating through pages of response)
        - A generally good practice seems to include an `order_by()` on a column or set of columns
        - For each query used in `paginate()` this PR adds an `order_by()` condition (full list in comments below)

Can read a bit more context from related issue linked below

### Relates
- https://github.com/data-dot-all/dataall/issues/1241

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/datapipelines/db/datapipelines_repositories.py
+++ b/backend/dataall/modules/datapipelines/db/datapipelines_repositories.py
@@ -135,7 +135,7 @@ class DatapipelinesRepository(EnvironmentResource):
         if filter and filter.get('type'):
             if len(filter.get('type')) > 0:
                 query = query.filter(DataPipeline.devStrategy.in_(filter.get('type')))
-        return query
+        return query.order_by(DataPipeline.label).distinct()
 
     @staticmethod
     def paginated_user_pipelines(session, username, groups, data=None) -> dict:
@@ -150,7 +150,7 @@ class DatapipelinesRepository(EnvironmentResource):
         query = session.query(DataPipelineEnvironment).filter(
             DataPipelineEnvironment.pipelineUri.ilike(uri + '%%'),
         )
-        return query
+        return query.order_by(DataPipelineEnvironment.envPipelineUri)
 
     @staticmethod
     def paginated_pipeline_environments(session, uri, data=None) -> dict:
--- a/backend/dataall/modules/datapipelines/db/datapipelines_repositories.py
+++ b/backend/dataall/modules/datapipelines/db/datapipelines_repositories.py
@@ -135,7 +135,7 @@ class DatapipelinesRepository(EnvironmentResource):
         if filter and filter.get('type'):
             if len(filter.get('type')) > 0:
                 query = query.filter(DataPipeline.devStrategy.in_(filter.get('type')))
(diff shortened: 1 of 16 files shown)


noah-paige:
I checked that none of the above queries throw an error when invoking and ensured that the original issue with listing environments is resolved

backend/dataall/modules/datapipelines/db/datapipelines_repositories.py
```diff
@@ -150,7 +150,7 @@ def query_pipeline_environments(session, uri) -> Query:
         query = session.query(DataPipelineEnvironment).filter(
             DataPipelineEnvironment.pipelineUri.ilike(uri + '%%'),
         )
-        return query
+        return query.order_by(DataPipelineEnvironment.envPipelineUri)
```
dlpzx, line 153: if the label is available maybe better to order by label or name

backend/dataall/modules/datasets/db/dataset_repositories.py
```diff
@@ -375,7 +375,7 @@ def _query_user_datasets(session, username, groups, filter) -> Query:
                     Dataset.label.ilike(filter.get('term') + '%%'),
                 )
             )
-        return query.distinct(Dataset.datasetUri)
+        return query.order_by(Dataset.label).distinct(Dataset.datasetUri, Dataset.label)
```
dlpzx, line 378: same as above

dlpzx (approved):
Looks good, left small comments

backend/dataall/modules/datapipelines/db/datapipelines_repositories.py
noah-paige: rather than label - do you mean by `stage` name (i.e. dev, test, prod, etc.)? I can make that change

backend/dataall/modules/datasets/db/dataset_repositories.py
noah-paige: same response as above - order by + distinct requirements

backend/dataall/modules/datapipelines/db/datapipelines_repositories.py
noah-paige: switched

PR #1250, opened 2024-05-06
Generic dataset module and specific s3_datasets module - part 1 (Rename datasets as s3_datasets)

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail
- Rename `datasets` module to `s3_datasets` module

This PR is the first step to extract a generic datasets_base module that implements the undifferentiated concepts of Dataset in data.all. s3_datasets will use this base module to implement the specific implementation for S3 datatasets.

### Relates
- #1123 
- #955 

### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)?
  - Is the input sanitized?
  - What precautions are you taking before deserializing the data you consume?
  - Is injection prevented by parametrizing queries?
  - Have you ensured no `eval` or similar functions are used?
- Does this PR introduce any functionality or component that requires authorization?
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms?
  - Are you logging failed auth attempts?
- Are you using or adding any cryptographic features?
  - Do you use a standard proven implementations?
  - Are the used keys controlled by the customer? Where are they stored?
- Are you introducing any new policies/roles/users?
  - Have you used the least-privilege principle? How?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1123, opened 2024-03-25
Create generic dataset_base and s3_dataset modules from current datasets


dlpzx:
**Is your feature request related to a problem? Please describe.**
It is difficult to add new dataset types with the current implementation of the `datasets` module.

**Describe the solution you'd like**
If we want to add more types of dataset (e.g. Redshift dataset) a first step will be the creation of a generic class that defines a data.all Dataset abstraction that can be used by each of the particular dataset implementations.

**Describe alternatives you've considered**
This are rough design considerations that might change during implementation:

-------------------
### 1. Avoid circular dependencies and remove the current `datasets_base` - [COMPLETED ✅ ]
There is some work to do to avoid circular dependencies between the initial `datasets` and `dataset_sharing`. There are some API calls that use each others' graphql models. From a code logic perspective, `datasets` should not depend on `dataset_sharing` code, so as a first step we will remove this dependency by defining the corresponding code related to dataset_sharing in dataset_sharing always.

### 2. Define basic code layout

![image](https://github.com/data-dot-all/dataall/assets/71252798/b34c87af-6629-4379-baf8-cf68f08d49f4)

- `datasets_base` --> includes `db` base models for Dataset + list APIs that are generic
       - DEPENDS ON: nothing - we need to remove dependencies with sharing. Anything related to sharing should be moved to the sharing module.
- `dataset_sharing_base` --> includes `db` base models for ShareObject, base permissions + list methods
       - DEPENDS ON: `datasets_base` 
- `s3_datasets` --> includes all `db` models of S3 datasets + implementation S3 APIs
       - DEPENDS ON: `datasets_base` and `dataset_sharing_base` (maybe)
- `s3_dataset_sharing` ---> business logic for S3 sharing
       - DEPENDS ON: `s3_dataset`, `dataset_sharing_base`
- in the future `redshift_datasets` and `redshift_dataset_sharing`

### 3. `datasets_base` module

1. contains the API calls that list Datasets that are generic (without details to S3/Glue)
2. contains the base code for other modules to implement a dataset

I thought of splitting the module into 2 modules: `datasets_list` module and `datasets_base` module. Although `datasets_base` would not be a module, it is some base code that will be re-used by other modules. Because of this and because the dataset_list uses dataset_base models I decided to leave it in the same module, but we could change this behavior, it is not a one way door.

#### Base code for Datasets
- NO API CALLS
[...]

Issue #955, opened 2024-01-09
Redshift Data Sharing


anmolsgandhi:
### **Description:**
Enable seamless data integration with Redshift as a new data source in ‘data.all’. This feature enhances collaboration by allowing users to easily publish, discover and share Redshift data within the data.all platform. Users can securely configure Redshift instance, streamlining the process of making Redshift datasets accessible.

### **Details:**
**Adding Redshift Instance and Publishing Tables**

- Users initiate the process by selecting “Create Dataset” and choosing Redshift from the dropdown menu.
- The interface guides users through a secure credential input, ensuring a streamlined and secure configuration process.
- Once configured, the dataset owners can select specific tables to publish to the ‘data.all’ catalog, ensuring a controlled inclusion of Redshift data.

**Tables Available for Discovery**

- Cataloged Redshift tables automatically become part of the ‘data.all’ catalog, visible to users exploring datasets within the platform.
- The catalog provides detailed metadata for each table, facilitating a comprehensive understanding of available data.
- Users can navigate the ‘data.all’ UI to effortlessly discover and explore Redshift tables
- Dataset owners can edit metadata for each table such as description, tags.

**Self-service Share Process for Redshift Data Sharing**

- Consumers interested in specific Redshift tables initiate the share process by selecting the desired dataset.
- Owners of the shared Redshift tables within data.all Datasets receive access requests, with an easy-to-use interface for managing permissions and approvals.
- Upon approval, the shared Redshift data becomes dynamically accessible to consumers, maintaining a consistent and user-friendly experience.

### **Benefits:**

- Additional Data Source Integration: The added capability of Redshift as a new data source enhances flexibility, enabling users to integrate diverse data sources beyond S3, expanding the platform’s utility.
- User-Friendly Configuration: A guided process ensures a connection of Redshift instances with secure credentials.
- Efficient Discovery: Automated cataloging promotes effortless exploration of Redshift tables within ‘data.all’ catalog.
- Streamlined Sharing Workflow: The self-service share process maintains simplicity and consistency across different types of data, allowing users to request and access Redshift data seamlessly as they do with S3 data.

@dlpzx

dlpzx:
# Design

## Assumptions
* Redshift clusters/namespaces are created and maintained by DevOps teams outside of data.all
* Database admin teams manage users in their clusters/namespaces outside of data.all
* Data producers and consumers can access their clusters/namespaces with the access provided by the database admin teams.
[...]

PR #1257, opened 2024-05-07
Generic dataset module and specific s3_datasets module - part 2 (Create datasets_base and move enums)

dlpzx:
### Feature or Bugfix
⚠️ This PR should be merged after #1250. 
- Refactoring

### Detail

This PR:
- Creates the skeleton of the `datasets_base` module consisting of 3 packages: `db`, `api`, `services`. And adds the `__init__` file.
- Adds the dependency of `s3_datasets` to `datasets_base` in the `__init__` file of the `s3_datasets` module
- Moves datasets_enums to datasets_base

### Relates
- #1123 
- #955 

### Security

  - Is the input sanitized?


dlpzx:
Tested in AWS:
- Loading of modules works as expected (e.g. we can see the Datasets module)
- All FE views show enums as normal (e.g. DatasetRoles, Topics, classification)
- Edit dataset and add topics works as expected
- Edit dataset confidentiality works as expected

backend/dataall/modules/datasets_base/__init__.py
```diff
@@ -0,0 +1,18 @@
+from typing import Set
+from dataall.base.loader import ModuleInterface, ImportMode
+
+
+class DatasetBaseModuleInterface(ModuleInterface):
+    @staticmethod
+    def is_supported(modes: Set[ImportMode]) -> bool:
+        supported_modes = {
```
noah-paige, line 8: can we have this as so no need to hardcode?
```
supported_modes = ImportMode.all()
```

noah-paige, line 8: could supported_modes just be `ImportMode.all()` ?

config.json
```diff
@@ -9,6 +9,9 @@
         "datapipelines": {
             "active": true
         },
+        "datasets_base": {
```
noah-paige, line 12: I think ideally we would have `dataset_base` be automatically activated if any of the child dataset modules (i.e. `s3_datasets`) is active and not have to expose it here on the config.json

Configuring the `dataset_base` module as active or not does not really mean anything as it is really the child dataset modules that are meaningful to activate or not

I am not sure how easy it is to do such a thing or if it requires a big shift in the loader logic but if an easy fix

backend/dataall/modules/datasets_base/__init__.py
noah-paige: I see it is not used for `CDK_CLI_EXTENSION`

config.json
noah-paige: I saw your comment that it is required here because of how we load ImportModes of CDK and CDK_CLI_EXTENSION - but can you explain why that is / is there an easy way to resolve?

dlpzx: First of all, `datasets_base` will be added in the config.json at some point (check the full design steps in https://github.com/data-dot-all/dataall/issues/1123) because some parameters are relevant for all datasets no only for s3_datasets or redshift_datasets (e.g. confidentiality). That is why I did not give it more thought.

But if you are curious, the issue is a bit hidden, the `loader` has some methods to check that it initialized the modules correctly. With the current implementation I think that **a module cannot have a module interface CDK and not have a CDK_CLI_EXTENSION if it is not a config.json module**.  I think the problem is not that much on the loader but on the way we are using it in the cdk proxy. I will have a closer look

backend/dataall/modules/datasets_base/__init__.py
dlpzx: I did not want to just give a wildcard of import; but as I explained in my comment, the result is not very pretty

noah-paige (approved):
Doing some digging on the loader class and exposing `datasets_base` in config.json but tested this PR and changes look good

config.json
noah-paige: I think I found the issue as to why there is an error when we remove `datasets_base` from `config.json`:

- When we `load_modules(modes={ImportMode.CDK_CLI_EXTENSION})`
    - We first import all modules `in_config` as part of `def _load_modules(...)`
    - This would theoretically include `s3_datasets` and `dataset_sharing` at this point in time
    - When we import `dataset_sharing` we have the following line in top level of `__init__.py` that subsequently imports `s3_datasets` and `datasets_base` modules
```
from dataall.modules.dataset_sharing.db.share_object_repositories import ShareEnvironmentResource
```

- Then when we `_check_loading_correct(...)` in loader.py
     - it fails on L252 with `ImportError` since `datasets_base` is in `sys.modules.keys()` but is not in `checked_modules_names` (since `datasets_base` is not in config / expected_load)

Previously we never exposed `datasets_base` or `datasets_sharing` as configurable modules in `config.json` so we avoided this problem. Moving the line 
```
from dataall.modules.dataset_sharing.db.share_object_repositories import ShareEnvironmentResource
```

in `backend/dataall/modules/dataset_sharing/__init__.py` to within the the `def __init__(self):` in `class SharingApiModuleInterface()` resolves this issue on loading modules but may need some additional testing to ensure share APIs still working as expected

dlpzx:
@noah-paige I will implement the import change in another PR, probably with other modules that also have global imports in their __init__ files

PR #1258, opened 2024-05-07
Generic dataset module and specific s3_datasets module - part 3 (Create DatasetBase db model and S3Dataset model) 

dlpzx:
### Feature or Bugfix
⚠️ This PR should be merged after #1257.
- Feature
- Refactoring

### Detail

**This PR does**:
- Adds a generic `DatasetBase` model in datasets_base.db that is used in s3_datasets.db to build the `S3Dataset` model using joined table inheritance in [sqlalchemy](https://docs.sqlalchemy.org/en/20/orm/inheritance.html)
- Rename all usages of Dataset to S3Dataset (in the future some will be returned to DatasetBase, but for the moment we will keep them as S3Dataset)
- Add migration script that backfills `datasets` table and renames `s3_datasets` ---> ⚠️ In the process of migrating we are doing some "scary" operations on the dataset table, if for any reason the migration encounters any issue it could result in catastrophic loss of information --> for this reason this [PR](https://github.com/data-dot-all/dataall/pull/1267) implements RDS snapshots on migrations.

**This PR does not**:
- Feed registration stays as: `FeedRegistry.register(FeedDefinition('Dataset', S3Dataset))` using `Dataset` with the `S3Dataset` resource type. It is out of the scope of this PR to migrate the Feed definition.
- Exactly the same for the GlossaryRegistry registration. We keep `object_type='Dataset'` to avoid backwards compatibility issues.
- It does not change the resourceType for permissions. We keep using a generic `Dataset` as target for S3 permissions. If we are to split permissions into DatasetBase permissions and S3Dataset permissions we would do it on a different PR

#### Remarks
Inserting new items of S3Dataset does not require any changes. SQL Alchemy joined inheritance automatically inserts data in the parent table and then another one to the child table as explained in this stackoverflow [link](https://stackoverflow.com/questions/39926937/sqlalchemy-how-to-insert-a-joined-table-inherited-class-instance-when-the-pare) (I was not able to find it in the official docs)

### Relates
- #1123 
- #955 
- https://github.com/data-dot-all/dataall/pull/1267

### Security

  - Is the input sanitized?


dlpzx:
Testing:
- [X] locally run all migration scripts from scratch - without data
- [X] locally downgrade each of the migration scripts one by one - without data
- [X] locally downgrade both migration scrips at once - without data

backend/dataall/modules/datasets_base/db/dataset_models.py
```diff
@@ -0,0 +1,42 @@
...
+    inProject = query_expression()
+
+    businessOwnerEmail = Column(String, nullable=True)
+    businessOwnerDelegationEmails = Column(ARRAY(String), nullable=True)
+    stewards = Column(String, nullable=True)
+
+    SamlAdminGroupName = Column(String, nullable=True)
+    autoApprovalEnabled = Column(Boolean, default=False)
+
+    datasetType = Column(String, nullable=False)
```
noah-paige, line 32: any need for `datasetType` to be an enum or have a default of S3>

backend/migrations/versions/458572580709_remove_dataset_table_read_permissions_from_env_admins.py
```diff
@@ -30,6 +35,53 @@ def get_session():
     return session
 
 
+Base = declarative_base()
+
+
+class Dataset(Resource, Base):
```
noah-paige, line 41: Do we need all of the below columns / do they even exist on base `dataset` table anymore?

Can we just query `dataset` table for 
`environmentUri` 
`SamlAdminGroupName`
`stewards`
`datasetUri`

backend/migrations/versions/c6d01930179d_add_backfill_read_folder_permissions.py
```diff
@@ -30,6 +36,53 @@ def get_session():
     return session
 
 
+Base = declarative_base()
+
+
+class Dataset(Resource, Base):
```
noah-paige, line 42: Do we need all of the below columns / do they even exist on base dataset table anymore?

Can we just query `dataset` table for
`SamlAdminGroupName`
`stewards`
`datasetUri`

backend/migrations/versions/d059eead99c2_rename_dataset_table_as_s3_dataset.py
```diff
@@ -0,0 +1,51 @@
...
+        local_cols=['environmentUri'],
+        remote_cols=['environmentUri'],
+    )
+    op.add_column(
+        's3_dataset',
+        sa.Column('datasetType', sa.String(), nullable=True),
+    )
+    op.execute('UPDATE s3_dataset SET "datasetType" = \'S3\'')
+    op.alter_column('s3_dataset', 'datasetType', nullable=False)
+
```
noah-paige, line 38: Would it be easier to add datasetType with a defaultValue and `nullabel=False` initiallly (... and then can then remove the defaultValue if no longer need before finishing upgrade)

backend/migrations/versions/d059eead99c2_rename_dataset_table_as_s3_dataset.py
```diff
@@ -0,0 +1,51 @@
...
+
+def downgrade():
+    print('Renaming s3_dataset as dataset...')
+    op.drop_column('s3_dataset', 'datasetType')
```
noah-paige, line 42: should we drop constraint `s3_dataset_environmentUri_fkey` in downgrade

noah-paige:
Overall left some minor comments - additionally tested out the migrations scripts back-filling a single dataset and all works locally for me as well

Will do one last look through first thing tomorrow once you are finished with you check list of testing as well

backend/dataall/modules/s3_datasets/__init__.py
noah-paige, line 1: Let's change `log.info(...)` statements here to refer to S3Datasets rather than Datasets now?

backend/migrations/versions/5cdcf6cc1d73_create_and_backfill_dataset_table.py
```diff
@@ -0,0 +1,374 @@
...
+        dataset.businessOwnerDelegationEmails = generic_datasets[dataset.datasetUri].businessOwnerDelegationEmails
+        dataset.stewards = generic_datasets[dataset.datasetUri].stewards
+        dataset.SamlAdminGroupName = generic_datasets[dataset.datasetUri].SamlAdminGroupName
+        dataset.autoApprovalEnabled = generic_datasets[dataset.datasetUri].autoApprovalEnabled
+        dataset.datasetType = generic_datasets[dataset.datasetUri].datasetType
+        dataset.imported = generic_datasets[dataset.datasetUri].imported
+    session.commit()
+    session.close()
+
+    # Update non-nullable columns
```
noah-paige, line 333: On downgrade I am getting an error `column "label" of relation "s3_dataset" contains null values` even after we set the `label` value in the above for loop

backend/migrations/versions/458572580709_remove_dataset_table_read_permissions_from_env_admins.py
dlpzx: No we don't need them all, I just added the Dataset that was defined at the time this migration script was added, instead of importing the Dataset model that might be newer. This is the table that exists at this point of the migration. Although we should never go and modify alembic scripts (we create new ones), having the Dataser model prevents users to add things on columns that don't exist at this stage

backend/migrations/versions/c6d01930179d_add_backfill_read_folder_permissions.py
dlpzx: Same as above; I think it makes it clear that importing the current DatasetBase or S3Dataset is not the same as using Dataset at that moment in time

backend/dataall/modules/datasets_base/db/dataset_models.py
dlpzx: I opt not to use sa.Enums as type for the column because I remember running into a similar situation in the past; I checked and I found that when I refactored notifications, I transformed one column from enum to string. To be honest I just had a soft memory of enums in columns = problems; but I think that it only applies to cases in which the list in the Enum varies often or if we get the values from another module, because that results in circular depdendencies. In short, I will add it

backend/migrations/versions/d059eead99c2_rename_dataset_table_as_s3_dataset.py
dlpzx: Let me test it out, because it is indeed easier :)

backend/dataall/modules/s3_datasets/__init__.py
dlpzx: Done

backend/migrations/versions/5cdcf6cc1d73_create_and_backfill_dataset_table.py
dlpzx: I will re-test with the changes from PR review, see the results below

dlpzx:
- [ ] locally create datasets with `main` code, update schema `dkrcompose` with alembic migrations (running only the ones that we introduce in this PR) check that data is correctly backfilled.
- [ ] Same for downgrade

backend/migrations/versions/d059eead99c2_rename_dataset_table_as_s3_dataset.py
dlpzx: good catch

backend/migrations/versions/d059eead99c2_rename_dataset_table_as_s3_dataset.py
dlpzx: So, setting a default does not write it  as default when we create a new column. I think the default only applies to new inserts to the db not to column creation

dlpzx: but it is implemented here: https://github.com/data-dot-all/dataall/pull/1258/commits/ffe372e455c49047cf7d17d45137ea68e9806735

backend/migrations/versions/5cdcf6cc1d73_create_and_backfill_dataset_table.py
```diff
@@ -0,0 +1,380 @@
...
+        's3_dataset',
+        sa.Column('businessOwnerEmail', sa.String(), nullable=True),
+    )
+    op.add_column(
+        's3_dataset',
+        sa.Column('businessOwnerDelegationEmails', postgresql.ARRAY(sa.String()), nullable=True),
+    )
+    op.add_column(
+        's3_dataset',
+        sa.Column('stewards', postgresql.ARRAY(sa.String()), nullable=True),
```
noah-paige, line 298: We add stewards as string column in upgrade to `dataset` but add it back to `s3_dataset` as Array

backend/dataall/modules/s3_datasets/db/dataset_models.py
```diff
@@ -120,26 +108,14 @@ class Dataset(Resource, Base):
...
     importedAdminRole = Column(Boolean, default=False)
-    imported = Column(Boolean, default=False)
-
-    autoApprovalEnabled = Column(Boolean, default=False)
 
-    @classmethod
-    def uri(cls):
-        return cls.datasetUri
+    __mapper_args__ = {
+        'polymorphic_identity': DatasetType.S3.value,
```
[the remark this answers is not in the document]
dlpzx: It is strange because when filling the column we always use the enum values. I'll work on it

noah-paige:
Last last things:

WHen running alembic autogenerate migration I get the following

```
def upgrade():
    op.alter_column('dataset', 'datasetType',
               existing_type=postgresql.ENUM('S3', name='datasettypes'),
               type_=sa.Enum('S3', name='datasettype'),
               existing_nullable=False)
    op.drop_constraint('s3_dataset_bucket_datasetUri_fkey', 'dataset_bucket', type_='foreignkey')
    op.create_foreign_key(None, 'dataset_bucket', 'dataset', ['datasetUri'], ['datasetUri'], ondelete='CASCADE')
    # ### end Alembic commands ###
```

Our models and migration scripts may not be in sync?

noah-paige:
Changing datasetUri in `dataset_bucket` table to 
```
    datasetUri = Column(String, ForeignKey('s3_dataset.datasetUri', ondelete='CASCADE'), nullable=False)
```
fixes the foreign key change generated by alembic

dlpzx:
Hello hello, thanks @noah-paige for such a deep review :) 
- foreign_key in dataset_bucket - I can make the change, looks reasonable
- postgresql.ENUM vs sa.Enum - there is a problem with sa.Enum and that is that it forces the creation of the enum object itself. I used postgresql.ENUM because it has the option to `create_type=False` which uses the existing datasettypes object

noah-paige (approved):
tested and looks good - approving

PR #1267, opened 2024-05-14
Create RDS database snapshot before executing alembic migrations

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Alembic migrations can get complex and in some cases we are using alembic for not only schema migrations but also data migrations. When moving columns with data from one table to another we might accidentally make a mistake in a migration script. We strive to test all migration scripts and avoid bugs in such sensitive operations, but to protect users from the catastrophic situation in which there is a bug, a service issue or any other exceptional situation this PR introduces the creation of manual database snapshots before running alembic migration scripts.

This PR modifies the db_migration handler that is triggered with every backendStack update. It checks if there are new migration scripts (if the current head in the database is different from the new head in the code). If True, it will create a cluster snapshot.

Remarks:
- Snapshots cannot be created when the cluster in not `available`, the PR introduces a check to wait for this condition. If the Lambda timeout is reached waiting for the cluster, then the CICD pipeline will fail and will need to be retried
- During the creation of an snapshot we can still run alembic migration scripts
- Snapshots are incremental, the first time will take a long time, but new snapshots will be faster

### Relates
- #1258 - This PR is a good example of complex data migration operations.

### Security

  - Is the input sanitized?


dlpzx:
⚠️ I need to test it in AWS!

dlpzx:
Tested in AWS!
- [X] backend stack gets updated with new IAM policies for the custom resource execution role
- [X] If no new migration script is found, then no snapshot is created
- [ ] If new migration scripts need to be run, an snapshot is created

backend/dbmigrations_handler.py
```diff
@@ -4,15 +4,67 @@
...
+            f'Creating RDS snapshot for cluster {cluster_id}, head revision {head_rev} is ahead of {current_rev}...'
+        )
+        try:
+            rds_client = boto3.client('rds', region_name=os.getenv('AWS_REGION'))
+            cluster_id = 'res-fr-db'
+            cluster_status = ''
+            while cluster_status != 'available':
+                # Edge case in which the cluster is performing backup and/or maintenance operations.
+                # If it times out the CICD pipeline fails and needs to be retried.
+                logger.info(f'Waiting while the cluster is available, status={cluster_status}')
```
petrkalos, line 53: is this `Waiting for the cluster to become "available"`? also should we use `{cluster_status=}`?

backend/dbmigrations_handler.py
```diff
@@ -4,15 +4,67 @@
...
+        snapshot_id = f'dataall-migration-{head_rev}-{datetime.datetime.now().strftime("%Y-%m-%d-%H-%M-%S")}'
+        cluster_id = engine.dbconfig.host.split('.')[0]
+        logger.info(
+            f'Creating RDS snapshot for cluster {cluster_id}, head revision {head_rev} is ahead of {current_rev}...'
+        )
+        try:
+            rds_client = boto3.client('rds', region_name=os.getenv('AWS_REGION'))
+            cluster_id = 'res-fr-db'
+            cluster_status = ''
+            while cluster_status != 'available':
```
petrkalos, line 50: nit: I am not saying it should be done like that but take a look at the[ walrus operator](https://nickdrozd.github.io/2022/12/14/walrus-while.html)

Essentially this can be simplified to

```
while (cluster_status := rds_client.describe_db_clusters(DBClusterIdentifier=cluster_id)['DbClusters'][0]['Status'])) is not 'available':
  logger.log(...)
  time.sleep(30)
```

backend/dbmigrations_handler.py
```diff
@@ -4,15 +4,67 @@
...
+
+    if head_rev != current_rev:
+        snapshot_id = f'dataall-migration-{head_rev}-{datetime.datetime.now().strftime("%Y-%m-%d-%H-%M-%S")}'
+        cluster_id = engine.dbconfig.host.split('.')[0]
+        logger.info(
+            f'Creating RDS snapshot for cluster {cluster_id}, head revision {head_rev} is ahead of {current_rev}...'
+        )
+        try:
+            rds_client = boto3.client('rds', region_name=os.getenv('AWS_REGION'))
+            cluster_id = 'res-fr-db'
```
petrkalos, line 48: Let's move this in a separate function maybe even a different file

petrkalos (changes requested):

backend/dbmigrations_handler.py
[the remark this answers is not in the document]
dlpzx: It is easier to get it from the envine.dbconfig (that gets it from SSM) My bad left this hardcoding reference

backend/dbmigrations_handler.py
dlpzx: Yes, for the rare case that the database is running an scheduled backup creating an snapshot would fail

backend/dbmigrations_handler.py
dlpzx: Solved in https://github.com/data-dot-all/dataall/pull/1267/commits/e72c76b0c7dd1b95d909e06974995054858defd3

backend/dbmigrations_handler.py
dlpzx: Solved in https://github.com/data-dot-all/dataall/pull/1267/commits/e72c76b0c7dd1b95d909e06974995054858defd3

backend/dbmigrations_handler.py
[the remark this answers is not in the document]
dlpzx: Solved in https://github.com/data-dot-all/dataall/pull/1267/commits/e72c76b0c7dd1b95d909e06974995054858defd3

dlpzx: Solved in latest changes. I tested the solution in AWS

petrkalos (approved):

PR #1270, opened 2024-05-15
Remove global imports in modules

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
The loader class that loads modules dependending on the module interface loads all module init files and then uses the interfaces to load specific parts of the module or to register/perform other operations. The issue is that in some modules we are importing some parts of the package at the top level of the `__init__` file, which means that no matter what the module interface is, those imports will be executed.

Let's see it with an example. I am loading modules using the `CDK_CLI_EXTENSION` module in the `cdk_cli_wrapper`. The loader loads all init files of the modules in the config.json, and then uses the interface to load the pipelines module only. The issue is that in the init files of the modules it has also imported things like the `DashboardRepository` or `SagemakerStudioEnvironmentResource` for example. Pieces of code that are not needed in the `CDK_CLI_EXTENSION`. 

This would be a minor issue, some extra pieces of code are loaded, okey. The bigger problem comes from "child-modules". The loader after loading everything checks if what is has been loaded has its correct interface or if it is a config.json module. Modules such as `feed` that are not directly specified in the config.json could be imported indirectly through some of this top-level imports in the init files and result in failure of the loading.

We have 2 options:
1. Make our loader aware of the source of each file it imports
2. Be strict with how we import in the init files

In this PR I went for 2 because our loader is complex enough and is already name-sensitive. I think adding more logic would make it very difficult to handle and it might have unexpected results.

### Relates
- #1257 - in this PR we realized about the issue

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/dashboards/__init__.py
+++ b/backend/dataall/modules/dashboards/__init__.py
@@ -3,9 +3,6 @@
 import logging
 from typing import Set, List, Type
 
-from dataall.core.environment.services.environment_resource_manager import EnvironmentResourceManager
-from dataall.modules.dashboards.db.dashboard_repositories import DashboardRepository
-from dataall.modules.dashboards.db.dashboard_models import Dashboard
 from dataall.base.loader import ImportMode, ModuleInterface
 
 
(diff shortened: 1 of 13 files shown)


dlpzx:
Testing locally:
- [X] modules load as normal when everything in config.json is enabled
- [X] creation of organization, environment
- [X] dataset, dataset indexed in catalog
- [ ] creation of play tools mlstudio, notebook, datapipeline, dashboard work as expected

petrkalos (approved):

Issue #1147, opened 2024-04-05
environment and dataset teams missing GET_ORGANIZATION permission


zsaltys:
There was recently a fix for 2.3 related to how GET_ORGANIZATION permission works: https://github.com/data-dot-all/dataall/pull/1139

The summary of the issue is that if you are a member of an environment team or dataset team then you will not have GET_ORGANIZATION permission and that will cause multiple issues in data.all UI:

- You will not be able to list environments in the environment UI and get an error saying that you do not have a GET_ORGANIZATION permission
- You will see organizations in the Organizations list but if you try to view an organization you will not be able to view it (there won't even be a UI error and graphql response will contain an error. **This should also be fixed**).

The PR above partially fixes this issue by adding a new organization resolver which does not require a GET_ORGANIZATION permission. This means that anyone can view some basic information about any organization.

I argue that this fix is ultimately not the best one and the way it should work is this:

If you are a member of an environment team or any dataset team on that environment then you should be granted GET_ORGANIZATION permission on that organization implicitly even if your team is not directly invited into an organization. Logically this makes sense:

1) You can only create environments with a specific environment team if that environment is invited to the organization (if that is not the case today then this should be fixed)
2) We do not invite dataset teams into organizations because the only reason to invite teams into organizations is to let them create environments.

To summarize:

1) Require that creating an environment in an organization with a team requires that that team is invited into the organization. This will ensure that this team will always have the GET_ORGANIZATION permission.
2) Any dataset team when it is invited to an environment should also be at the same time given the GET_ORGANIZATION permission on the organization. It should be removed if the dataset team is removed from the environment.

Additionally we should add tests to check that dataset teams can view organizations or that they are granted GET_ORGANIZATION permission upon invitation. Also to make sure we test this for environment teams when they are invited to organizations.

zsaltys:
@dlpzx was this addressed:

`Any dataset team when it is invited to an environment should also be at the same time given the GET_ORGANIZATION permission on the organization. It should be removed if the dataset team is removed from the environment.`

And was the code cleaned up to remove the fix added to bypass the get_organization permission error?

PR #1273, opened 2024-05-15
GET_ORGANIZATION permission for Env teams

SofiaSazonova:
### Feature or Bugfix
- Bugfix

### Detail
-  When Environment is linked, check that SamlGroup is invited to Organization
- When group is invited to Env, add GET_ORGANIZATION permissions, if it doesn't exist yet
- When group is deleted from Env Teams, check 1)it's not invited to Org, 2)it's not invited to another Env of this Org -> remove GET_ORGANIZATION permissions
- Related tests
- list_organisations now is based on GET_ORGANIZATION permissions, so non-mebers with GET_ORGANIZATION permissios can see Orgs 
- Inside OrgView all buttons (link, Invite, update, delete) are hidden basing on permissions
- migration script for permission backfill

### Relates
- #1147

### Security

  - Is the input sanitized?


SofiaSazonova:
Checked in AWS:
Org, Env1, Env2, Group (not in Org)

- [ ] Can't create an Env with Group
- [ ] Can create Env with group that is invited into Org

- [ ] When Group is invited to the Env1, it gets GET_ORGANIZATION permissions for Org
- [ ] When Group is invited to the Env2, GET_ORGANIZATION permission for Org is not duplicated

- [ ] When Group is deleted from Env2,  GET_ORGANIZATION permission for Org is NOT deleted
- [ ] When Group is deleted from Env1,  GET_ORGANIZATION permission for Org is deleted

- [ ] Group is invited into Org, then invited into Env1, GET_ORGANIZATION permission for Org is not duplicated
- [ ] When Group is deleted from Env1,  GET_ORGANIZATION permission for Org is NOT deleted
- [ ] When Group is deleted from Org,  GET_ORGANIZATION permission for Org is deleted

backend/dataall/core/environment/services/environment_service.py
```diff
@@ -213,6 +215,12 @@ def create_environment(uri, data=None):
         context = get_context()
         with context.db_engine.scoped_session() as session:
             EnvironmentRequestValidationService.validate_creation_params(data, uri, session)
+            if not OrganizationRepository.is_group_invited(session, uri, data['SamlGroupName']):
+                raise Exception(
+                    f'Group {data["SamlGroupName"]} is not a member of the organization {data.get("organizationUri")}. '
+                    f'Invite this group to the organisation before linking the environment.'
+                )
+
```
noah-paige, line 223: Can we move to to part of `validate_user_groups()` in `EnvironmentRequestValidationService`

backend/dataall/core/organizations/db/organization_repositories.py
```diff
@@ -106,6 +106,19 @@ def query_organization_groups(session, uri, filter) -> Query:
...
+        return (
+            session.query(models.OrganizationGroup)
+            .filter(
+                and_(
+                    models.OrganizationGroup.organizationUri == uri,
+                    models.OrganizationGroup.groupUri == group,
+                )
+            )
+            .count()
+        ) > 0
```
noah-paige, line 120: nit: Don't need to specify `>0` - if returns a value greather than 0 will evaluate as True in the if condition either way

backend/dataall/core/environment/db/environment_repositories.py
```diff
@@ -303,6 +303,22 @@ def is_user_invited_to_environment(session, groups, uri):
...
+            .join(Environment, EnvironmentGroup.environmentUri == Environment.environmentUri)
+            .filter(Environment.deleted.is_(None))
+            .filter(
+                and_(
+                    EnvironmentGroup.groupUri == group,
+                    Environment.organizationUri == organizationUri,
+                )
+            )
+            .count()
+            > 0
```
noah-paige, line 319: nit: Don't need to specify >0 - if returns a value greather than 0 will evaluate as True in the if condition either way

backend/dataall/core/environment/services/environment_service.py
```diff
@@ -463,6 +487,20 @@ def remove_group(uri, group):
                 resource_uri=environment.environmentUri,
                 resource_type=Environment.__name__,
             )
+
+            # if group is no longer invited to organization or environments inside it, remove GET_ORGAISATION policy
```
noah-paige, line 491: nit: type in comment `remove GET_ORGANIZATION policy`

backend/dataall/core/environment/services/environment_service.py
```diff
@@ -409,6 +417,22 @@ def invite_group(uri, data=None) -> (Environment, EnvironmentGroup):
...
+                resource_type=Organization.__name__,
+                permission_name=GET_ORGANIZATION,
+            ):
+                ResourcePolicyService.attach_resource_policy(
+                    session=session,
+                    group=group,
+                    resource_uri=environment.organizationUri,
+                    permissions=[GET_ORGANIZATION],
+                    resource_type=Organization.__name__,
+                )
```
noah-paige, line 434: The way `def attach_resource_policy()` is written is that the policy is only created and permissions associated if and only if they do not already exist (like a "create if not exists" expressions)

Open ended Q - is there added benefit to having the check `if check_group_resource_permission()`?

noah-paige:
Also a note - I believe there is a bug with the confirmation pop ups when removing an invited team or consumption role from an environment

Let's say I have 3 groups in an env (GroupA, B, and C)... When removing GroupA - I receive the pop up and need to type `permanently delete` for the group
- Once I do so GroupA is removed but the pop up remains and the team it is asking me to confirm to remove now is GroupB 

In `EnvironmentTeams.js` we have one variable `isDeleteTeamModalOpen` to control if the mondals for N groups is open - some bug in the logic is being introduced because of this

backend/dataall/core/environment/services/environment_service.py
dlpzx: True, attach_resource_policy does not create duplicates if the policy already exists. I don't think we need the check at creation, only at deletion

backend/dataall/core/environment/services/environment_service.py
SofiaSazonova: done

backend/dataall/core/environment/services/environment_service.py
SofiaSazonova: done

backend/dataall/core/environment/services/environment_service.py
SofiaSazonova: done

SofiaSazonova:
@noah-paige 
(1) I think, it's for separate PR
(2) Done. I didn't remove REMOVE_ORGANIZATION_GROUP yet, but backfilled GET_ORGANISATION
(3) I changed OrganisationView. Now readers are displayed. Also, I changed list_organisation logic to show Orgs for "readers"
(4) I will fix this bug in separate PR

SofiaSazonova:
image

backend/dataall/core/groups/services/group_service.py
```diff
@@ -72,6 +72,6 @@ def list_groups_without_invited(filter: dict = None):
             groups = []
             for group in cognito_groups:
                 if group not in invited_groups:
-                    groups.append({'groupName': group})
+                    groups.append({'groupUri': group})
```
noah-paige, line 75: why this change?

backend/dataall/core/groups/api/types.py
```diff
@@ -45,6 +45,18 @@
...
+
+OrganizationReadersSearchResult = gql.ObjectType(
+    name='OrganizationReaders',
+    fields=[
+        gql.Field(name='count', type=gql.Integer),
+        gql.Field(name='page', type=gql.Integer),
+        gql.Field(name='pages', type=gql.Integer),
+        gql.Field(name='hasNext', type=gql.Boolean),
+        gql.Field(name='hasPrevious', type=gql.Boolean),
+        gql.Field(name='nodes', type=gql.ArrayType(GroupsInfo)),
```
noah-paige, line 60: Is this type used anywhere?

backend/dataall/core/groups/api/types.py
```diff
@@ -45,6 +45,18 @@
 GroupsInfo = gql.ObjectType(
     name='GroupsInfo',
     fields=[
-        gql.Field(name='groupName', type=gql.String),
+        gql.Field(name='groupUri', type=gql.String),
```
noah-paige, line 48: why change this from groupName to groupUri?

backend/dataall/core/organizations/db/organization_repositories.py
```diff
@@ -106,6 +122,52 @@ def query_organization_groups(session, uri, filter) -> Query:
...
+            .join(
+                ResourcePolicyPermission,
+                ResourcePolicy.sid == ResourcePolicyPermission.sid,
+            )
+            .join(
+                Permission,
+                Permission.permissionUri == ResourcePolicyPermission.permissionUri,
+            )
+            .filter(
+                or_(
```
noah-paige, line 140: don't need the `or_(...)` condition

frontend/src/modules/Organizations/components/OrganizationTeamInviteForm.js
```diff
@@ -48,8 +48,8 @@ export const OrganizationTeamInviteForm = (props) => {
         setGroupOptions(
           response.data.listGroups.map((g) => ({
             ...g,
-            value: g.groupName,
-            label: g.groupName
+            value: g.groupUri,
```
noah-paige, line 51: similar to above - any need to change `groupName` to `groupUri`?

backend/migrations/versions/11b716ddb036_backfill_get_org_permissions.py
```diff
@@ -0,0 +1,50 @@
...
+from sqlalchemy import orm
+from dataall.core.organizations.db.organization_models import Organization
+from dataall.core.environment.db.environment_models import EnvironmentGroup, Environment
+from dataall.core.permissions.services.resource_policy_service import ResourcePolicyService
+from dataall.core.permissions.services.organization_permissions import GET_ORGANIZATION
+
+
+# revision identifiers, used by Alembic.
+revision = '11b716ddb036'
+down_revision = '458572580709'
```
noah-paige, line 19: I see the migration script is failing because of multiple heads in alembic ... I believe you will need to merge the latest from `main` to this branch and then update the `down_revision` here to be pointing from the latest revision currently in main

noah-paige:
The new frontend UI looks great! 1 general question - does the query still run in manageable time for either `getOrganization` or `listOrganizationReadOnlyGroups` with the nested resolvers / added logic?

I was considering the idea that we could add a new column to `OrganizationGroup` called `userType` and backfiling for existing Groups. Then for each invite or remove from envs we also create/remove a new OrganizationGroup record with `userType` of value `Reader` 
- Could make it easier to determine what should be loaded for a given user on UI / when listing reader vs admin users 

... but this is not really needed unless we see some big performance differences I would argue

SofiaSazonova:
Well, right now I can't spot any big difference. We have queries like this in many places. But of course, it's better to test with some loaded infrastructure.

As for OrganisationGroups. Well, we can do this, but it means, that we also need to redo the UI and backend of inviting groups: we want then to add opportunity to give Readonly permissions. 
May be, it's a good thing to do. I'm not sure right now. Let's discuss

noah-paige:
Last required change I think - the resolved user role for an invite EnvGroup (i.e. an Org Reader but not an Administrator) shows as `NotMember`

Screenshot 2024-05-24 at 10 59 34 AM

Would need to update the way `find_organization_membership` identifies who is part of org group or not

backend/dataall/core/permissions/db/resource_policy/resource_policy_repositories.py
```diff
@@ -108,9 +108,9 @@ def find_resource_policy(session, group_uri: str, resource_uri: str, resource_ty
 
     @staticmethod
     def find_all_resource_policies(
-        session, group_uri: str, resource_uri: str, resource_type: str = None, permissions: List[str] = None
+        session, groups: [str], resource_uri: str, resource_type: str = None, permissions: List[str] = None
```
noah-paige, line 111: nit: `List[str]`

noah-paige (approved):
Left 2 minor comments but overall tested and looks good

Issue #1009, opened 2024-01-26
Refactor access request UI flow to avoid draft shares


zsaltys:
From our experience users who are using data.all for the first time find it very confusing/unintuitive that when they request access to a dataset that this creates a draft share which needs to be further filled out and then submitted. There are a few text notices for this but most folks just don't see them.

I propose that we completely get rid of access request modal and create a new page form where a user has to additionally specify what he wants to be shared with them and submitting this form actually submits the share to the dataset owner. This will ensure that it is impossible to think that a share was created and sent to another team while in fact it is still in draft mode.

I would also use this to make another improvement. From my experience a lot of new users get confused with the optional field of consumer role. They often submit an access request to their team role rather than consumer role. They then think that somehow their environment now has access and becuse they registered the roles on their environment they have access too which is obviously not the case. People don't understand what requesting access for a team means. We should make this explicit and make people choose what they want to share for - a team role or their own iam role... The change is basically making people make a conscious decision vs having an optional field which can be ignored. We want users to actively think - what should I be choosing here?

zsaltys:
@noah-paige @dlpzx I would put a lot of priority on this one.. This one is currently causing a lot of issues where people reach out to us because they think sharing does not work when in fact the issue is because they don't understand the UI...

dlpzx:
Hi @zsaltys thanks for the issue. There is definitely room for potential in the sharing UX. I will tag this item ans high priority and we will try to work on it as soon as possible.

dlpzx:
Completed

PR #1277, opened 2024-05-16
ShareView remake

SofiaSazonova:
### Feature or Bugfix
- Feature

### Detail
- Share View page is reorganised. Now all important info is on one screen.
- Red reminder to submit the draft
- Edit and Submission buttons call the Modal window for share editing. User can save share as a draft or submit it.
- When request is created from Catalog, the modal window give user the form for share editing and submission (or draft). 

### Relates
- #1009 

### Security

  - Is the input sanitized?


SofiaSazonova:
image

image

SofiaSazonova:
@dlpzx @zsaltys have a look

zsaltys:
@SofiaSazonova @dlpzx @noah-paige I think perhaps this is an improvement but I think we could do better...

**I think draft shares should not exist at all**. A share either exists as submitted or doesn't exist at all. I don't see any value in draft shares and I think they they just serve to add confusion. I think we should rework the graphql endpoints to allow to submit a share together with all the items selected. This would be more work and you couldn't reuse likely the existing UI components but it solves a lot of problems we're trying to work around: putting reminders in red text, adding more modals and buttons to click. I see this modal + red text as bandaging over symptoms vs solving the core problem.

if you really insist to keep drafts:

1) Please make role name LONG so you can see how it affects the UI.

2) Revoke/Verify/Re-Apply should not show on draft shares.

3) It would be nice to remap Your Role to Approver and Requester and not ApproverAndRequester

4) Put on the title **Draft share object...** as one more cue visual reminder this is a draft.

5) Rename the Submit Share in the modal to Review Share... I really don't like this modal because it feels an annoyance to the user who knows what they are doing and now has to click more buttons.. It's not a nice experience.

7) Showing status the submit modal only makes sense for drafts.. but it makes no sense to show status on the modal because the status can only be PENDINGAPPROVAL... we can show it but there's little value for it.

8) Change the button on the Request Access modal to "Create Draft Share"

9) Remove the request purpose on the request access modal and then it makes more sense to set it on the submit modal and it will be a bit more clear why the submit modal exists.. 

10) When creating a share from a TABLE item in the catalog do not automatically preselect the table but leave no items selected to force users to think through what they need to select on the share items..

anushka-singh:
I agree with Zi. It would be preferable not to encounter an additional page after navigating through the request access modal. I suggest that upon clicking the lock button on a dataset, the user should be directed either to a request access modal or a share page. Here, they should be prompted to fill out all the details from the original request access modal along with the current share page options, such as selecting share items. Having two pages solely for submitting a share request seems unnecessary.

Additionally, if we opt to retain the draft stage page, we should implement the steps outlined by Zi. I particularly emphasize step 7, where instead of a submit button, we should display a "Create Draft Share" button. Subsequently, upon someone clicking on this button, a red pop-up notification should appear at the top right, indicating that the draft request has been submitted and prompting the user to complete the rest of the form to finalize their share request."

TejasRGitHub:
Hi @SofiaSazonova , the sleek UI looks nicer than the earlier one. 

Adding to the comments from @anushka-singh and @zsaltys , 

1. I would suggest having the submit button at the end of the UI ( like how it is when we create an IAM role, etc in AWS) . This way the user goes from top to bottom and also ( hopefully 😄 ) sees the share items section. 
2. Just above the submit button, there can be a small banner which provides a note about how the user should proceed with with this share. Maybe here we can mention that the user needs to select atleast one share item and then submit the share request. 
3. If possible ( as a stretch ) if the user is a requestor , we can show another banner telling the user about how the share request will be approved by someone from the approvers team. If the user is an approver, then we can display something like , Next Steps : by clicking on approve, you would create a share for items {} , if you do not wish to share then you can reject the share request 
4. I really liked the way which Anushka described of how after clicking the Catalog's lock icon , the user is taken to the shares page. Additionally , I would suggest that we should replace the lock icon with some text maybe or a  better way to let the user know about how the share request should be created. 
5. During the share creation process, as the consumption role is an optional field, it creates confusion amongst first time users. Maybe we could design that modal such that after selecting the environment, they could take action as to if they want to create a share on the team role or want to select a consumption role ( if its available in that environment )

SofiaSazonova:
@zsaltys @anushka-singh @TejasRGitHub Thanks for the feedback!
I will review the workflow once again and will try to fulfil you requirements without dramatic changes to the current procedures.

dlpzx:
In addition to the above comments:
- The main downside to keep an eye on when simplifying the workflow and avoiding the draft state is the triggering of the shares tasks. Requester submits item1, then item2, then item3. Approver approves the share request right after item2 --> we need to ensure that item3 is not shared. I have not looked at the code but we do queries based on the item status so I guess we could run into that situation
- Small thing: when submitting a share already SHARE_SUCCEEDED items do not matter. In the case of having many many tables showing already shared items can be confusing

SofiaSazonova:
- Share View page is reorganised. Now all important info is on one screen.
- Red reminder to submit the draft
- Edit and Submission buttons call the Modal window for share editing. User can save share as a draft or submit it.
- When request is created from Catalog, the modal window give user the form for share editing and submission (or draft).

SofiaSazonova:
Screenshot 2024-06-06 at 13 21 02
Screenshot 2024-06-06 at 13 21 37

SofiaSazonova:
@zsaltys 

>> Please make role name LONG so you can see how it affects the UI. 
Long role names will be croped by the width of the line and "..." will be added. Full name will be available in the tooltip.

>> Revoke/Verify/Re-Apply should not show on draft shares.
"Revoke" is now shown only in Edit window. Verify and Reapply are still useful in drafts: share could be previously approved and processed. Then, it was drafted again, but previously shared items still can be diagnosed and reapplied if needed.  

>> It would be nice to remap Your Role to Approver and Requester and not ApproverAndRequester
There are some 'backend strings' that I would love to rename as well, but I will keep for separate PR

>> Put on the title Draft share object... as one more cue visual reminder this is a draft.
Good Idea

>>Rename the Submit Share in the modal to Review Share... I really don't like this modal because it feels an annoyance to the user who knows what they are doing and now has to click more buttons.. It's not a nice experience.
Now the share can be submitted as soon as it's edited, so usually the user won't have to use this modal. I keep it as it is for now.

>> Showing status the submit modal only makes sense for drafts.. but it makes no sense to show status on the modal because the status can only be PENDINGAPPROVAL... we can show it but there's little value for it.
Now the status is shown, because we have one editing table for all items

>> Change the button on the Request Access modal to "Create Draft Share"
Good Idea

>> Remove the request purpose on the request access modal and then it makes more sense to set it on the submit modal and it will be a bit more clear why the submit modal exists..
I keep it now in both places, for i don't want to break the already established ways users can be accustomed to.

>> When creating a share from a TABLE item in the catalog do not automatically preselect the table but leave no items selected to force users to think through what they need to select on the share items..
In current workflow, user will see right away, what is selected.

SofiaSazonova:
@TejasRGitHub 
>>I would suggest having the submit button at the end of the UI ( like how it is when we create an IAM role, etc in AWS) . This way the user goes from top to bottom and also ( hopefully 😄 ) sees the share items section.

Now, it's there

>>If possible ( as a stretch ) if the user is a requestor , we can show another banner telling the user about how the share request will be approved by someone from the approvers team. If the user is an approver, then we can display something like , Next Steps : by clicking on approve, you would create a share for items {} , if you do not wish to share then you can reject the share request

It's not really a UI change, but more the workflow change as well. We can discuss it separately and create a separate issue

>>I really liked the way which Anushka described of how after clicking the Catalog's lock icon , the user is taken to the shares page. Additionally , I would suggest that we should replace the lock icon with some text maybe or a better way to let 
the user know about how the share request should be created.

Now we can do the whole share submission from catalog page in modal window.

>> During the share creation process, as the consumption role is an optional field, it creates confusion amongst first time users. Maybe we could design that modal such that after selecting the environment, they could take action as to if they want to create a share on the team role or want to select a consumption role ( if its available in that environment )

Good suggestion. I don't want to overcomplicate this PR, but this option definitely needs some clarification in UI

frontend/src/modules/Shared/Shares/ShareEditForm.js
```diff
@@ -0,0 +1,452 @@
...
+      dispatch({ type: SET_ERROR, error: response.errors[0].message });
+    }
+  };
+
+  const revokeShareItem = async () => {
+    const response = await client.mutate(
+      revokeItemsShareObject({
+        input: {
+          shareUri: share.shareUri,
+          itemUris: [item.itemUri]
```
noah-paige, line 67: this needs to be `[item.shareItemUri]` rather than `[itemUri]` for this to work

frontend/src/modules/Shared/Shares/ShareEditForm.js
```diff
@@ -0,0 +1,452 @@
...
+        }
+      })
+    );
+
+    if (!response.errors) {
+      enqueueSnackbar('Item added', {
```
noah-paige, line 73: Message should be `Item Revoked` not Added

noah-paige:
Hi @SofiaSazonova - think we are getting there but found a couple issues when playing around with the UI and testing locally:

[1] Does this work for Dashboard Shares via requestAccessModal?

[2] If already have a share and open a share request from catalog for another item via RequestAccessModal
- Issues 
    - Updating request purpose of first step (i.e. step 0) does nothing (maybe remove?)
    - If Share Status is Submitted and then I “Include” a new Item
        - The Share status remains “Submitted” after I add the item
        - But when I click “View Share” it is actually of state “Draft” (a similar finding is below with editing directly via ShareView)

[3] When I have a share in submitted state with 1+ items
- I then click on edit from ShareView and add another item
- Issues
    - When I add a new item my options are then only to “Cancel” (no functional issue but a bit confusing that no other button)
    - When I click out via “Cancel” - ShareView is not updated with latest status (would be changed from Submitted to Draft because I added a new item)
        - Additionally should we have a way to add items and immediately re-submit (i.e. no draft) from the ShareView Edit Modal?
    - I can not edit purpose because the only button shown for this share edit view is “Cancel”

[4] From ShareEditForm opened via ShareView
- If status Share_Rejected - it says “Wait until this item is processed”
- But I can delete from Shared Items table in Share View which I believe is correct since this is separate from a revoke of an existing share


[5] From ShareEditForm opened via ShareView
- When Revoke there is no loading view from UI after first click of Revoke
    - I can keep spamming Revoke button and likely break something (no way to know something is happening) 

[6] From ShareEditForm opened via ShareView
- Pagination does not work if over 10 items

frontend/src/modules/Shared/Shares/ShareEditForm.js
SofiaSazonova: true

frontend/src/modules/Shared/Shares/ShareEditForm.js
SofiaSazonova: thanks

SofiaSazonova:
@noah-paige 

1. About Dashboard Shares. It should be the same as it was before. Isn't it?
2, 3. I must have missed something and share is not reloaded. 

About this 
>>But when I click “View Share” it is actually of state “Draft” (a similar finding is below with editing directly via ShareView)

It actually the way it was before. If share is already exist, we don't change the request, just load the old one. I'll think. how to let user know, why is that so.

Also, I will rename  “Cancel” -> "Close"

4. Hm, I was overwhelmed with the state machine))) I'll fix it.
5. Acknowledged. I'll fix
6. Hm...Does it work in Share View? I'll  create so many items to test it's better.

SofiaSazonova:
I fixed the problems @noah-paige mentioned.

The only thing left is to notify user, that the share is already exist before redirecting them to share page. I'm on it)

SofiaSazonova:
I added message for user that share already exists. So, they won't be surprised

frontend/src/modules/Shared/Shares/ShareEditForm.js
```diff
@@ -0,0 +1,485 @@
...
+    dispatch,
+    client
+  } = props;
+
+  const whatToDo = () => {
+    if (!item.status) return 'Request';
+    if (
+      item.status === 'Revoke_Succeeded' ||
+      item.status === 'PendingApproval' ||
+      item.status === 'Share_Rejected'
```
noah-paige, line 48: looking at SM - also should include `Share_Failed` which one should be able to delete from right?

noah-paige:
Thank you @SofiaSazonova for the quick changes on the above - just a few more issues I am finding.

Pending General Issues I still am facing:

*[1] Still request purpose does nothing on first view in RequestAccessModal with existing share object*
- I guess it is fine because notification saying redirecting to existing share
    - But for a user if they write a lengthy description they then lose it all (annoying UX)
- Additionally no way to edit request purpose from Edit button on ShareView because the only button shown for this share edit view is “Close”

*[2] From ShareEditForm opened via ShareView*
- If I delete an item that is in PendingApproval the ShareEditModal updates but when I click “Close” the ShareView is not updated until I “Refresh” on ShareView Page
    - Maybe Close should just be a Refresh always?

*[3] MINOR - Submit with no included share item throws error and keeps share Draft state (may be confusing for user from RequestAccessModal at least)*
- Disable Submit button if none included?

---

**Additional Considerations (Looking further?):**
Maybe as a future issue for 2.7 or later given the amount of work @SofiaSazonova  has already invested here to make these views far better…
- But now that we have separate modal views to add/remove items - should we have only dataset objects in catalog
    - Opening shares for child Tables and Folders already redirect to dataset related shares
    - As we expand cataloged items (i.e. redshift datasets, s3 datasets, dashboards, etc etc) things may begin to get cluttered in catalog
- As long as the user can fully submit a share from the Catalog page I think it is a much cleaner experience but open to thoughts!

SofiaSazonova:
1. Purpose texfield moved to second step of edit form
2. Request purpose can't be edited, when request is already processed or approved. This button also disabled in ShareView
3. Reject Purpose can be edited only in statuses Submitted or Draft
4. Share_Failed added to possible statuses for deletion of item
5. Submit button is disabled if there is no items
6. Button to Save&Close  when it's not draft. When the request is saved, the share is reloaded, edits are saved

frontend/src/modules/Shared/Shares/ShareEditForm.js
```diff
@@ -346,7 +359,8 @@ export const ShareEditForm = (props) => {
   return (
     <Box sx={{ p: 3, minHeight: 800 }}>
       <Typography align="center" color="textPrimary" gutterBottom variant="h4">
-        Share status: {shareStatus}
+        Share status: {shareStatus}{' '}
+        {smthChanged || requestPurpose !== share.requestPurpose ? 'yes' : 'no'}
```
noah-paige, line 363: remove `yes` / `no` print out

noah-paige:
Thank you @SofiaSazonova for all of your work iterating on this PR

---
Overall - the process of opening up a new share request is much approved with these changes and will become a more streamlined process.

An Important Remark(!):
With this PR it will not be possible to revoke a list of items in a share object but rather revokes will happen one at a time per share item
- Enhancements we can look into for future releases include - (1) adding search / filters to make finding objects to add or to revoke simpler, (2) enabling revoke on a list of items, or other ideas as well
---

Please take a look at the latest and if no additional concerns I will work with Sofia to get this PR approved and merged in the next day or two
cc: @dlpzx @petrkalos @zsaltys @TejasRGitHub @anushka-singh

anushka-singh:
Thanks @SofiaSazonova for all the work on this PR! 
The changes already look much more user friendly. I found no issues in the code, but I was just wondering if it would be possible to get newer screenshots after the final changes have been made so we can review it easily?

frontend/src/modules/Shared/Shares/ShareEditForm.js
SofiaSazonova: oops

SofiaSazonova:
@anushka-singh here you go!
Screenshot 2024-06-11 at 07 30 19
Screenshot 2024-06-11 at 07 30 54
Screenshot 2024-06-11 at 07 31 22

anushka-singh:
Originally, the Request Purpose box used to be on the "Request Access" page. I see now it has moved to Share status page. Is there a reason for that? I think it made sense to have it on the "Request Access" Page along with the other fields that needed to be filled out.

noah-paige:
> Thank you @SofiaSazonova for all of your work iterating on this PR
> 
> Overall - the process of opening up a new share request is much approved with these changes and will become a more streamlined process.
> 
> An Important Remark(!): With this PR it will not be possible to revoke a list of items in a share object but rather revokes will happen one at a time per share item
> 
>     * Enhancements we can look into for future releases include - (1) adding search / filters to make finding objects to add or to revoke simpler, (2) enabling revoke on a list of items, or other ideas as well
> 
> 
> Please take a look at the latest and if no additional concerns I will work with Sofia to get this PR approved and merged in the next day or two cc: @dlpzx @petrkalos @zsaltys @TejasRGitHub @anushka-singh

@SofiaSazonova - can you just pull the latest from main to fix the issue with `npm-audit` GH workflow check

Afterwards I think we are good on this PR and I will approve

noah-paige:
> Originally, the Request Purpose box used to be on the "Request Access" page. I see now it has moved to Share status page. Is there a reason for that? I think it made sense to have it on the "Request Access" Page along with the other fields that needed to be filled out.

Before we had the Request Purpose box on both the Request Access initial page (opened from Catalog) and the Share Edit page (opened as second page from Catalog or from `Edit` button on ShareView)

IMO - It felt less intuitive to have the edit on both pages and I thought it would be best to keep the text box in on `ShareEditForm` because that UI view is used in both places making it far more accessible for a user to edit this field

Also the Share Object gets created on the first "Request Access" page (opened from Catalog) but if the share object already exists it just gets redirected to re-use the existing share object. Currently the logic we have in backend would just return the existing share object as is and not update the request purpose if one provided as well. Instead of adding more one-off logic to backend and additional conditions,  it seemed best to clean up frontend views in this case

Let me know if I missed anything in the above rationale @SofiaSazonova

SofiaSazonova:
@noah-paige done)

anushka-singh:
@noah-paige 
Thanks for addressing this. I like that we modified the frontend instead of adding more clunky code in the backend. 
IMO I would still like having request purpose on Request Access page. 
As a user, when I create a share request, I can put in my reasoning for requesting resources and the approver will be able to see my reason. However, with the latest change it sounds like the user will have to first create a share request and then go to edit the share request and only then put a reason for requesting the resources. IMO, it should be in request access page. But its not a blocker, we can discuss it further or slate the change for later if we decide to go ahead with it. But otherwise you have my approval to merge.

SofiaSazonova:
@anushka-singh 
>>user will have to first create a share request and then go to edit the share request and only then put a reason for requesting the resources.

It's not quite right. The edit page (screenshot 3) appears immediately after the share created. So, just consider Request Access as step 1 and then Share fill in/edit form as step 2. No additional actions needed

anushka-singh:
Okay! That makes sense. 
Thanks for clarifying that for me @SofiaSazonova. 
Let's merge it in!

TejasRGitHub:
Hi @SofiaSazonova , The changes look really good. I was able to pull in your changes and checked the UI. The UI experience is really nice and I like that the user can create a share request on the same request modal itself.

noah-paige (approved):
approving and will merge - thanks again @SofiaSazonova

Issue #926, opened 2023-12-18
Search returns results with very weak matches


zsaltys:
1. Let's imagine I have 3 datasets published in my data.all:

* test-iam-policy
* test-data
* best-data

2. If I search for test-iam-policy then data.all will return ALL 3 datasets in the search. This is because test-iam-policy matches completely, test-data has word test, and best-data has ending *est

3. I think returning first test-iam-policy and test-data is probably fine though I would argue that when I'm using - it should find the exact thing. Especially because the UI autocompletes the dataset for me so there's no need for it to return test-data too. And it should never return matches on partial words like best-data as that is simply confusing. Perhaps partial word search would be fine if it was searching if any of the words i searched appear fully in the text.. but if I search test it should not return anything that matches with "est".

dlpzx:
Hi @zsaltys thanks for raising an issue, we will look into it. I have taken the liberty to edit the title and add tags to the issue.

PR #1279, opened 2024-05-16
set DataSearch fuzziness to 0 -- strict search

SofiaSazonova:
### Feature or Bugfix

- Bugfix

### Detail
We use DataSearch component, that has the parameter [fuzziness](https://www.elastic.co/guide/en/elasticsearch/guide/current/fuzziness.html) that describes [Damerau-Levenshtein edit distance](https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance). To make search more strict I change this parameter to 0. Lest's see, if this is enough.

### Relates
- #926 

### Security

  - Is the input sanitized?


frontend/src/modules/Catalog/views/Catalog.js
noah-paige, line 1: Changing fuzziness to `{0}` works exactly as expected now requiring a strict prefix search of data object (i.e. dataset, table, or folder) name or associated metadata to return results in catalog

The one additional thing I would suggest is that we remove filtering for metadata categories on the free-form string entered in the search bar for Catalog. For example, when I type `useast1` in the search bar of Catalog I should not be returned data objects that have no reference of us-east-1 other than the `region` metadata field of the data object

More generally, I think as a rule of thumb we can say any `enum` type associated to a data object can be a filter Multi-list filter but should not also be a string match filter (i.e. `type`, `topic`, `classification`, `region`) - it is a bit redundant and confuses the search a bit

If we do pick up this change it would just be editing L175 to remove `region` and deleting out L177-180 which add these enums to the `dataFieldList` on `Catalog.js`

Open to hear others thoughts @SofiaSazonova @zsaltys

SofiaSazonova: I agree, let's not mix up search and filters

noah-paige (approved):
lgtm

PR #1282, opened 2024-05-17
Generic dataset module and specific s3_datasets module - part 5.2 (Move list queries to dataset_base or rename them)

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail

In this PR we:
- Restructure listDatasetsOwnedByEnvGroup as listS3DatasetsOwnedByEnvGroup and move it into Worksheets in FE: the reason why it is moved to Worksheets is that it is the only place where it is used in the FE. One could argue that in the BE listS3DatasetsOwnedByEnvGroup is part of the S3_Dataset module. The way I see it, FE and BE are independent and their modularization strategies fit the type of programming, what makes sense in FE might not make it in BE. In BE queries belong to the module whose services/models they are performing actions on, in this case s3_datasets. In FE queries belong to the module where they are used and if a query is used by more than one module then it can be placed in the generic `services` directory. What is important is that we define the dependencies. In this case it is important to make Worksheets dependent of S3_Datasets (as we do in the index in `frontend/src/modules/Worksheets/index.js` and in `backend/dataall/modules/worksheets/__init__.py`
- Move listDatasetsCreatedInEnvironment to datasets_base

### Relates
- #1123 
- #955 

### Security

  - Is the input sanitized?


noah-paige (approved):
Looks good - tested worksheet listing datasets, env dataset tab, and share filters all still work

Approving

PR #1289, opened 2024-05-20
automate bootstrapping of integrations tests

petrkalos:
### Feature or Bugfix
- Feature

### Detail
* introduce an SSM param that holds the test userdata
* migrate ConfigureCognito step from a frontend codebuild step to a TriggerFunction that runs after the deployment of the CognitoClient
* add logic in ConfigureCognito to create the test users and assigned them to groups if `with_approval_tests=True` (reads info from SSM param)
   * minor: change prints to logging
* tests are reading the userdata from the SSM param
* if `with_approval_tests=True` Cognito allows `ALLOW_USER_PASSWORD_AUTH` auth flow

### Security

  - Is the input sanitized?

### DIFF
--- a/deploy/configs/cognito_urls_config.py
+++ b/deploy/configs/cognito_urls_config.py
@@ -20,6 +20,7 @@ def setup_cognito(
     internet_facing='True',
     custom_domain='False',
     enable_cw_canaries='False',
+    with_approval_tests='False',
 ):
     ssm = boto3.client('ssm', region_name=region)
     user_pool_id = ssm.get_parameter(Name=f'/dataall/{envname}/cognito/userpool')['Parameter']['Value']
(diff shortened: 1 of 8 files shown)


tests_new/integration_tests/README.md
```diff
@@ -3,7 +3,18 @@ The purpose of these tests is to automatically validate functionalities of data.
 
 ## Pre-requisites
 - A real deployment of data.all in AWS
-- 4 Cognito users (at the moment only Cognito is supported) like the ones defined in `conftest`(e.g. `testUser1` with password `Pass1Word!`)
+- A SecretManager secret (`'{self.resource_prefix}-{target_env["envname"]}-cognito-test-users'`) with the followings contents
```
SofiaSazonova, line 6: Do we need SecretManager here? We don't need auto rotation or so, I think SSM Parameter store is enough, also it's free.

tests_new/integration_tests/README.md
```diff
@@ -3,7 +3,18 @@ The purpose of these tests is to automatically validate functionalities of data.
...
+    ```
```
SofiaSazonova, line 7: I know, that's for test purposes, but why expose it in code (especially opensource)? May be it's better just to leave comment like: `userdata with username, password and groups`.

SofiaSazonova (changes requested):

tests_new/integration_tests/README.md
petrkalos: I just gave an example password, real deployments should use a different one. Is your concern that people will just copy and paste the example and hence make themselves vulnerable?

tests_new/integration_tests/README.md
petrkalos: That's a good point, I am not an expert but my go-to is that when a password is involved I use SecretsManager (even if it's just for tests). Apart from the pricing aspect I don't see a downside.

dlpzx: We had issues in the past with checkov flagging some secrets in SecretsManager because some best practices dictate that encrypted Secrets in secret manager should be rotated at least annually and encrypted with a KMS key. So if we add a secret, it needs to be encrypted with KMS and with rotation enabled. So as @SofiaSazonova was saying if it does not rotate, we might go for SSM. The question is, do we need to rotate them? what can happen if the password gets exposed?

In the scenario where a malicious user:
- knows that a company uses data.all. Tries to guess the cloudfront domain that might be something like <company-name>.dataall.com
- tries to hit the domain with the default test user password (company is careless and did not change it)
- can overflood a company CICD pipeline with requests?

tests_new/integration_tests/README.md
```diff
@@ -3,7 +3,18 @@ The purpose of these tests is to automatically validate functionalities of data.
...
+    {
+        "testUserTenant": {"password": "yourPassword", "groups": ["DAAdministrators"]},
+        "testUser1": {"password": "yourPassword", "groups": ["testGroup1"]},
+        "testUser2": {"password": "yourPassword", "groups": ["testGroup2"]},
+        "testUser3": {"password": "yourPassword", "groups": ["testGroup3"]},
+        "testUser4": {"password": "yourPassword", "groups": ["testGroup4"]}
+    }
+    ```
+- If you are not using Cognito then you must manually create the users/groups
```
dlpzx, line 16: If not using Cognito, they still need to create the secret to store the passwords right?

deploy/configs/cognito_urls_config.py
```diff
@@ -94,61 +95,68 @@ def setup_cognito(
...
+            MessageAction='SUPPRESS',
+        )
+        print(f'User Created Successfully...: {response}')
+    except ClientError as e:
+        if 'UsernameExistsException' in str(e):
+            print('User already exists')
+        else:
+            raise e
+
+    print('Updating Canaries user password...')
```
dlpzx, line 134: nit: prints logs referencing CW canaries. Can we replace prints by logs?

tests_new/integration_tests/README.md
petrkalos: After discussing with @dlpzx offline we decide to use SSM params for simplicity. Thanks for the spot on commend Sofia :)

deploy/configs/cognito_urls_config.py
```diff
@@ -44,73 +49,68 @@ def setup_cognito(
...
 
-    print(f'UI: {signin_singout_link}')
-    print(f'USERGUIDE: {user_guide_link}')
+    log.info(f'UI: {signin_singout_link}')
+    log.info(f'USERGUIDE: {user_guide_link}')
 
     cognito = boto3.client('cognito-idp', region_name=region)
-    try:
-        user_pool = cognito.describe_user_pool_client(UserPoolId=user_pool_id, ClientId=app_client)
+    user_pool = cognito.describe_user_pool_client(UserPoolId=user_pool_id, ClientId=app_client)
```
dlpzx: You removed the general try/except because if it fails then the stack will rollback right?

petrkalos: Nope, I removed it because it was only catching the exception printing and throwing it again which isn't required as the exception will currently bubble up and print the exact same information...

```
 except ClientError as e:
        print(f'Failed to setup cognito due to: {e}')
        raise e
```

SofiaSazonova (approved):

Issue #1283, opened 2024-05-17
Create generic shares_base and s3_datasets_shares modules from current dataset_sharing


dlpzx:
**Is your feature request related to a problem? Please describe.**
It is difficult to add new sharing item types with the current implementation of the `datasets_sharing` module.

**Describe the solution you'd like**
If we want to add more types of sharing(e.g. Redshift dataset sharing, notebook sharing) a first step will be the creation of a generic class that defines a data.all Shares abstraction that can be used by each of the particular sharing implementations.

**Describe alternatives you've considered**
This are rough design considerations that might change during implementation:

-------------------
⚠️ [WIP]
This item is a continuation of https://github.com/data-dot-all/dataall/issues/1123 

## Backend

![image](https://github.com/data-dot-all/dataall/assets/71252798/b34c87af-6629-4379-baf8-cf68f08d49f4)

### 1. `shares_base` module
With the generic code to create share objects and share items.
- api
- aws - nothing, it should be technology agnostic
- db - generic ShareObject and ShareItem models. Operations on these models 
       - in ShareObject: rename datasetUri as targetUri
       - in ShareItem: make it a base class and remove the S3/Glue details from it
       - repositories: all operations that do not involve details on S3/Glue
- cdk - nothing, it does not involve any particular AWS IAM permissions or infra
- handlers - nothing
- services
    - shares_enums - they are generic - ShareItemTypes will be used to know the items that can be shared
    - shares_permissions - they are generic
    - TBD
- tasks
   - TBD

### 2. `s3_datasets_shares` module
With the specific code to share S3 dataset items.
- api
- aws
- db
- cdk
- handlers
- services
- tasks

dlpzx:
# Step-by-step implementation plan
- [ ] Rename `dataset_sharing` to `s3_datasets_shares` and create `shares_base` module ([PART1](https://github.com/data-dot-all/dataall/pull/1284))
- [ ] Move enums and permissions to `shares_base`([PART1](https://github.com/data-dot-all/dataall/pull/1284))
- [ ] Move ShareObject db model and make base ShareItem model in shares_base. Use ShareItem to create S3ShareItems or GlueShareItems, tbd
- [ ] .....

dlpzx:
The core functionality has been completed. There will be follow-up enhancements related to shares tackled in separate github issues

PR #1292, opened 2024-05-21
Generic dataset module and specific s3_datasets module - part 6 (Frontend)

dlpzx:
### Feature or Bugfix
- Feature

### Detail

In this PR we:
- Create DatasetsBase module in frontend. Depends on S3_Datasets module
- Move DatasetsList view and DatasetListItem component to DatasetsBase
- Add CreateDataset modal that allows multiple types of datasets creation
- Fix routes and redirects to point at /datasets/ or any other /X-dataset/
- Move dataset_base services. In backend/datasets_base/api we define the following queries that are good candidates to become part of the DatasetsBase module
    - listDatasets - only used in DatasetsBase/DatasetList view - it should be in DatasetsBase/services
    - listOwnedDatasets - only used in Shares/SharesBoxList view - it should be in Shares/services
    - listDatasetsCreatedInEnvironment - only used in Environments/EnvDataset tab - it should be in Environment/services

If we want to keep everything clean we could rename all "datasets" as "s3_datasets" or equivalent in the S3_Datasets module. Because it is a cosmetic change that would pollute the PR a lot I have decided not to include it.

⚠️ UPDATE: Next steps
The Data Shared With You table in Environments>Datasets tab needs a remake. It contains references to each type of item and it is very coupled with s3-dataset-shares. For the moment I just made it work for the changes of s3-datasets, but when completing the work in #1283 we should fix this. Maybe in favor of DataGrid

### Relates
- #1123 
- #955 

### Security

  - Is the input sanitized?


frontend/src/routes.js
noah-paige, line 1: If we are changing all of these routes then I believe we also need to change everywhere we navigate `/console/datasets/...` to `/console/s3-datasets/...` 

The above is true for all paths except for `/console/datasets` (which shows top level `<DatasetList>` UI View - otherwise we get a bunch of 404 page not found errors when navigating in data.all

dlpzx: I thought that I changed them but I missed a good number of them. On it!

dlpzx:
Testing locally:
- [X] listDatasets from the menu side bar
- [X] click on a specific Dataset > go to DatasetView
- [X] createFolder, click Folder, edit Folder
- [X] click Table, edit table
- [X] hit NewDataset button from listDatasets > create new dataset > at the end it redirects to DatasetView
- [X] hit NewDataset button from listDatasets > import new dataset
- [X] Go to Catalog, click any Dataset/Table/Folder and get redirected to DatasetView/FodlerView/TableView
- [X] from Catalog NewDataset button, window opens
- [X] in Environments Dataset tab lists owned and shared items

frontend/src/modules/Catalog/index.js
```diff
@@ -7,7 +7,8 @@ export const CatalogsModule = {
   resolve_dependency: () => {
     return (
       getModuleActiveStatus(ModuleNames.S3_DATASETS) ||
-      getModuleActiveStatus(ModuleNames.DASHBOARDS)
+      getModuleActiveStatus(ModuleNames.DASHBOARDS) ||
+      getModuleActiveStatus(ModuleNames.DATASETS_BASE)
```
noah-paige, line 11: minor: `getModuleActiveStatus(ModuleNames.DATASETS_BASE)` will always return false in `datasets_base` not in `config.json` 

does not affect logic though

noah-paige (approved):
looks good! Approving

frontend/src/modules/Catalog/index.js
dlpzx: I need to review that, because maybe not now but I am counting on being able to define all datasets enabled like this.... I'll have a second look because by the way ModuleNames and dependencies are defined it should work

PR #1294, opened 2024-05-22
Generic shares_base module and specific s3_datasets_shares module - part 2 (db objects to shares_base)

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail
As explained in the design for #1123 and #1283 we are trying to implement generic `datasets_base` and `shares_base` modules that can be used by any type of datasets and by any type of shareable object in a generic way. 

In this PR:
- Move shares state machines into its own file and move them to the `shares_base` module.
- Create new ShareObjectRepository and copy some generic methods to it
- Move db share objects to shares_base:
     - ShareObject has a field called datasetUri. For the scope of S3 and Redshift datasets we can leave it as dataseturi, but if we want to implement other kinds of sharing we should rethink it: we need to store the "approvers" of the share in some way. For the moment I am not going to go down that path until shares and s3 are uncoupled, then we can see how we would implement a complete generic sharing.
     - ShareObjectItem includes 3 fields related to the glue tables... we need to get rid of them in the backend code and look up the info with the itemType and itemUri. Left for part3 to keep the PR clean.

### Relates
- #1283 
- #1123 
- #955 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/shares_base/db/share_object_state_machines.py
+++ b/backend/dataall/modules/shares_base/db/share_object_state_machines.py
@@ -1,0 +1,296 @@
+import logging
+
+from dataall.base.db import exceptions
+from dataall.modules.shares_base.services.shares_enums import (
+    ShareObjectActions,
+    ShareObjectStatus,
+    ShareItemActions,
+    ShareItemStatus,
+)
+from dataall.modules.shares_base.db.share_object_repositories import ShareObjectRepository
+
+logger = logging.getLogger(__name__)
+
+
+class Transition:
+    def __init__(self, name, transitions):
+        self._name = name
+        self._transitions = transitions
+        self._all_source_states = [*set([item for sublist in transitions.values() for item in sublist])]
+        self._all_target_states = [item for item in transitions.keys()]
+
+    def validate_transition(self, prev_state):
+        if prev_state in self._all_target_states:
+            logger.info(f'Resource is already in target state ({prev_state}) in {self._all_target_states}')
+            return False
+        elif prev_state not in self._all_source_states:
+            raise exceptions.UnauthorizedOperation(
+                action=self._name,
+                message=f'This transition is not possible, {prev_state} cannot go to {self._all_target_states}. '
+                f'If there is a sharing or revoking in progress wait until it is complete and try again.',
+            )
+        else:
+            return True
+
+    def get_transition_target(self, prev_state):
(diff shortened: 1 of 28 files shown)


backend/dataall/modules/shares_base/db/share_object_state_machines.py
```diff
@@ -0,0 +1,296 @@
...
+                message=f'This transition is not possible, {prev_state} cannot go to {self._all_target_states}. '
+                f'If there is a sharing or revoking in progress wait until it is complete and try again.',
+            )
+        else:
+            return True
+
+    def get_transition_target(self, prev_state):
+        if self.validate_transition(prev_state):
+            for target_state, list_prev_states in self._transitions.items():
+                if prev_state in list_prev_states:
```
petrkalos, line 38: nit: `not in` and drop the else branch

backend/dataall/modules/shares_base/db/share_object_state_machines.py
```diff
@@ -0,0 +1,296 @@
...
+from dataall.modules.shares_base.db.share_object_repositories import ShareObjectRepository
+
+logger = logging.getLogger(__name__)
+
+
+class Transition:
+    def __init__(self, name, transitions):
+        self._name = name
+        self._transitions = transitions
+        self._all_source_states = [*set([item for sublist in transitions.values() for item in sublist])]
```
petrkalos, line 19: nits
* `itertools.chain.from_iterable(transitions.values())` should flatten your values
*  why convert the set to list? I guess you want to dedup but why convert back to list?
*  by converting to set you assume that the values are hashable, I can see that are enum values so they must be but give it a thought

backend/dataall/modules/shares_base/db/share_object_state_machines.py
```diff
@@ -0,0 +1,296 @@
...
+        self._all_target_states = [item for item in transitions.keys()]
```
petrkalos, line 20: nit: why is this needed? what's wrong with `self._all_target_states = transitions.keys()`?

backend/dataall/modules/shares_base/db/share_object_state_machines.py
dlpzx: I just copy/pasted the code. I was very tempted to do changes on the implementation of the state machines, but I think we should do a review of them in a separate PR

PR #1305, opened 2024-05-30
Allow descriptions schema

noah-paige:
### Feature or Bugfix

- Enhancement

### Detail
- Allow dataset admin role glue permissions for dataset crawler 

- Allow GQL Types, Arguments, Inputs, and Fields to take in `description` instance variable and write GQL schema including the description docstring 

### Relates
- Data.all SDK/CLI

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/base/api/gql/graphql_argument.py
+++ b/backend/dataall/base/api/gql/graphql_argument.py
@@ -18,9 +18,9 @@ class Argument:
 
     def gql(self):
         description_str = ''
-        n = "\n"
+        n = '\n'
         if self.description:
-            description_str = f" \"\"\"{self.description}\"\"\" {n}"
+            description_str = f' """{self.description}""" {n}'
 
         if isinstance(self.type, Enum):
             return f'{description_str} {self.name} : {self.type.name}'
--- a/backend/dataall/base/api/gql/graphql_argument.py
+++ b/backend/dataall/base/api/gql/graphql_argument.py
@@ -17,10 +17,7 @@ class Argument:
             raise Exception('Invalid Argument Type')
 
     def gql(self):
-        description_str = ''
-        n = '\n'
-        if self.description:
-            description_str = f' """{self.description}""" {n}'
+        description_str = f'"""{self.description}"""\n' if self.description else ''
 
         if isinstance(self.type, Enum):
             return f'{description_str}{self.name} : {self.type.name}'
--- a/backend/dataall/base/api/gql/graphql_argument.py
+++ b/backend/dataall/base/api/gql/graphql_argument.py
@@ -8,23 +8,29 @@ from .utils import get_named_type
 
 
 class Argument:
-    def __init__(self, name, type):
+    def __init__(self, name, type, description=''):
         self.name = name
+        self.description = description
         if isinstance(get_named_type(type), (Scalar, InputType, Ref, Enum)):
             self.type = type  # get_named_type(type)
         else:
             raise Exception('Invalid Argument Type')
 
     def gql(self):
+        description_str = ''
+        n = "\n"
+        if self.description:
+            description_str = f" \"\"\"{self.description}\"\"\" {n}"
+
         if isinstance(self.type, Enum):
-            return f'{self.name} : {self.type.name}'
+            return f'{description_str} {self.name} : {self.type.name}'
         elif isinstance(self.type, Ref):
-            return f'{self.name} : {self.type.name}'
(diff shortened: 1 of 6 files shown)


backend/dataall/base/api/gql/graphql_argument.py
```diff
@@ -8,23 +8,29 @@
...
         self.name = name
+        self.description = description
         if isinstance(get_named_type(type), (Scalar, InputType, Ref, Enum)):
             self.type = type  # get_named_type(type)
         else:
             raise Exception('Invalid Argument Type')
 
     def gql(self):
+        description_str = ''
+        n = '\n'
```
SofiaSazonova, line 21: Do we really need '\n' as a variable?
May we just make it
`description_str = f' """{self.description}""" \n' if self.description else ''`
I think it's more python-like

backend/dataall/base/api/gql/graphql_argument.py
```diff
@@ -8,23 +8,29 @@
...
+        if self.description:
+            description_str = f' """{self.description}""" {n}'
+
         if isinstance(self.type, Enum):
-            return f'{self.name} : {self.type.name}'
+            return f'{description_str}{self.name} : {self.type.name}'
```
SofiaSazonova, line 26: Also, do we want to store \n in description itself? 
IMHO, it's better to format it in place
'{description_str}\n{self.name} : {self.type.name}'

backend/dataall/base/api/gql/graphql_field.py
```diff
@@ -48,13 +49,19 @@ def gql(self, with_directives=True) -> str:
             t = self.type.name
         else:
             raise Exception(f'Invalid type for field `{self.name}`: {type(self.type)}')
+
+        gql = ''
+        n = '\n'
+        if self.description:
+            gql = f' """{self.description}""" {n}'
```
SofiaSazonova, line 56: Same comment about  f' """{self.description}""" \n' if self.description else ''

backend/dataall/base/api/gql/graphql_input.py
```diff
@@ -5,12 +5,17 @@
...
+    def __init__(self, name, arguments, description=''):
         self.name = name
         self.arguments = arguments
+        self.description = description
 
     def gql(self):
+        description_str = ''
         n = '\n'
+        if self.description:
+            description_str = f' """{self.description}""" {n}'
```
SofiaSazonova, line 17: same

backend/dataall/modules/catalog/api/queries.py
```diff
@@ -21,7 +21,7 @@
 
 SearchGlossary = gql.QueryField(
     name='searchGlossary',
-    doc='Search glossary ',
+    description='Search glossary ',
```
SofiaSazonova, line 24: Here, after the transformation described above, we will have both space and \n.
Is it ok?

SofiaSazonova (changes requested):

backend/dataall/base/api/gql/graphql_argument.py
noah-paige: sure I can change that if you prefer - more so just following the way it is already defined in other GQL class modules such as `backend/dataall/base/api/gql/graphql_enum.py` where `n='\n'` is defined as such

since we are only calling on `\n` once here I will just make the change you described above

backend/dataall/base/api/gql/graphql_argument.py
noah-paige: well we do not need `\n` if there is a description - so the new line is tied to the description a bit

I have not played around with how descriptions can be interpreted - I can play around a bit to see if can keep all on same line but if not I will keep it as is

noah-paige: From my research - the standard seems to be adding comments on a new line above the type or argument you are describing (ex:  https://medium.com/@krishnaregmi/how-to-write-comments-in-a-graphql-schema-to-enhance-auto-generated-documentation-c0047125ea24)

I think best to follow the same

backend/dataall/modules/catalog/api/queries.py
noah-paige: not sure I follow but the extra space is not a problem - will have a later PR that adds descriptions to all of our GQL queries, mutations, etc. and can clean up this description then

SofiaSazonova (approved):

Issue #1230, opened 2024-04-29
Integration tests executed on a real deployment as part of the CICD  - Environments


dlpzx:
Description in #1220.

This issue is to track the progress for the `Environments` module. 
It has its complications because we need to work with real AWS accounts that need to be bootstrapped and trusting the real deployment account. 

Design will follow

petrkalos:
In #1334 some basic tests that create/delete/modify an environment were added but more should be added to  cover the most commonly used functionality around environments

dlpzx:
Completed!

Issue #1136, opened 2024-04-02
Feature Flags for different environments like Dev, UAT, Prod


shivalkarrahul:
Hello,

We have a requirement as follows.

1. We have 3 different Environments, viz Dev, Staging and Prod.
2. We should be able to have Dev, Staging & Prod Environment on different versions of Data.all (Dev on 2.3.0, Staging on 2.2.0, Prod on 2.1.0)
3. There should be a way to Enable or Disable features based on Environments.
e.g. Dev can have modules.dashboards.active enabled, and Staging/Prod can have it disabled. (We believe different config.json would come into picture here) 
e.g Dev environment would be having few functionalities which are under development and not present in staging or Prod, Staging Environment would be having few functionalities getting tested but will not be moved to Prod until UAT is done

5. Do we need to have config-dev.json, config-staging.json, config-prod.json files to enable to disable features?
6. How do we enable or disable UI for the features?

Any guidance would be appreciated.

Thanx in advance.

dlpzx:
Hi @shivalkarrahul thanks for the issue! Definitely an interesting one. 

For the issue on having **multiple code versions in different deployment environments** we can either:
1. go for a trunk-based approach and define a ManualApproval stage between the development stages. In data.all there is a parameter called `"with_approval": "boolean_ADD_CODEPIPELINE_APPROVAL_STEP|DEFAULT=false",` that creates a manual approval stage between the deployment environments. 
2. or go for a gitflow approach and deploy multiple CICD pipelines, each one would be like an standalone data.all deployment reading from a branch in the repository. You would control promotion between stages by branch merging rules.

The second issue is related to the config.json files. For gitflow approach, every deployment would have its own config.json, so it is already solved. For the trunk-based approach which is the standard in data.all we would need to introduce multiple config.json files as you pointed out in point4.

Last thing, I am not sure I understand your point number 5, could you give us some more details?

Bests

petrkalos:
@shivalkarrahul currently our config system is a bit convoluted. 

We have 2 config files `cdkjson` and `configjson`  the first is responsible for configuring only the cdk code.
The second though is configuring the backend (lambda/ecs tasks, migration scripts etc), the frontend but also the cdk.
As suggested one solution could be introducing multiple files but that will add even more mental overhead.
So we want to take some time and design a better solution.

In the meantime in #1318 I introduced a way to override config flags using an SSM parameter (i.e `/dataall/dev/configjson`).
[...]

PR #1318, opened 2024-06-06
add support for full or partially updating Config params from SSM

petrkalos:
### Feature or Bugfix
Feature

### Detail
Currently data.all reads the config that is commited to the root directory of the repo.
In this PR we allow customers to customize the params as defined in the config.json (or completely overwrite it) using an SSM parameter (`/datall/{ENV}/configjson`).
The new workflow will be as follows
1. Read the `config.json` (as always)
2. Read the configjson SSM param and merge it with the contents of config.json (essentially overwriting the specified params)

This is useful for the following cases...
* Running a CI/CD pipeline directly from GitHub repo and want to customize the behaviour
* Running multiple stages/envs (dev, uat, prod etc) and want to customize them separately

### Relates
- #1230 
- #1136 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/base/config.py
+++ b/backend/dataall/base/config.py
@@ -1,18 +1,36 @@
 """Reads and encapsulates the configuration provided in config.json"""
 
-import json
 import copy
-from typing import Any, Dict
+import json
+import logging
 import os
 from pathlib import Path
+from typing import Any, Dict
+
+from dataall.base.utils import Parameter
+
+ENVNAME = os.getenv('envname', 'local')
+
+log = logging.getLogger(__name__)
 
 
 class _Config:
-    """A container of properties in the configuration file
-    and any other that can be specified/overwritten later in the application"""
+    """
+    A container of properties that configure data.all.
+    * Reads from config.json
+    * Optionally updates/overwrite from SSM /dataall/{ENVNAME}/configjson if it exists
+    * Updates dynamically at runtime via set_property
+    """
 
     def __init__(self):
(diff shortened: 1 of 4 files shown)


backend/dataall/base/config.py
```diff
@@ -1,18 +1,36 @@
...
+    * Reads from config.json
+    * Optionally updates/overwrite from SSM /dataall/{ENVNAME}/configjson if it exists
+    * Updates dynamically at runtime via set_property
+    """
 
     def __init__(self):
+        log.info('reading data.all config from local file')
         self._config = _Config._read_config_file()
+        try:
+            if ssm_config := _Config._read_config_parameter():
```
SofiaSazonova, line 29: Where does `ssm_config` come from? I can't see the declaration. Same about `configjson` (line 79).

SofiaSazonova: There is double work happening here: first, try to read file and then ssm_config. Can we somehow do only one option? Or is it unnecessary and it's quick enough not to worry about double loading?

petrkalos: My thought here is that we use the SSM parameter to overwrite the config.json so we need to read both.
The reason for overwriting instead of providing the entire thing is for brevity and for keeping compatibility with upstream changes.

noah-paige (approved):

petrkalos:
Closing because @dlpzx pointed out that CDK also reads this config file hence changing it dynamically is not a very good idea. We need to come up with a better plan.

Issue #1319, opened 2024-06-07
Enable encryption for environment variables in lambdas - cont


mourya-33:
### Describe the bug

The lambda environment variables are not encrypted. This is flagged by checkov as failures

Check: CKV_AWS_173: "Check encryption settings for Lambda environmental variable"
	FAILED for resource: AWS::Lambda::Function.TriggerFunctiondbmigrationshandlerhandler8A64572A
	File: /assembly-dataall-main-cicd-stack-dataall-sandbox-backend-stage/dataallmaincicdstackdataallsandboxbackendstagebackendstackDbMigrations80B1C3E5.nested.template.json:378-452
	Guide: https://docs.prismacloud.io/en/enterprise-edition/policy-reference/aws-policies/aws-serverless-policies/bc-aws-serverless-5

Check: CKV_AWS_173: "Check encryption settings for Lambda environmental variable"
	FAILED for resource: AWS::Lambda::Function.CognitoParamsSyncHandlersandbox22A17F25
	File: /assembly-dataall-main-cicd-stack-dataall-sandbox-backend-stage/dataallmaincicdstackdataallsandboxbackendstagebackendstackCognito0421C128.nested.template.json:589-648
	Guide: https://docs.prismacloud.io/en/enterprise-edition/policy-reference/aws-policies/aws-serverless-policies/bc-aws-serverless-5

Check: CKV_AWS_173: "Check encryption settings for Lambda environmental variable"
	FAILED for resource: AWS::Lambda::Function.CognitoProvidersandboxframeworkonEventE89AB8F9
	File: /assembly-dataall-main-cicd-stack-dataall-sandbox-backend-stage/dataallmaincicdstackdataallsandboxbackendstagebackendstackCognito0421C128.nested.template.json:734-778
	Guide: https://docs.prismacloud.io/en/enterprise-edition/policy-reference/aws-policies/aws-serverless-policies/bc-aws-serverless-5
		

Check: CKV_AWS_173: "Check encryption settings for Lambda environmental variable"
	FAILED for resource: AWS::Lambda::Function.TriggerFunctionsavepermshandlerhandlerDA90B406
	File: /assembly-dataall-main-cicd-stack-dataall-sandbox-backend-stage/dataallmaincicdstackdataallsandboxbackendstagebackendstackSavePermsBAF6E160.nested.template.json:378-452
[...]

PR #1322, opened 2024-06-08
Updating encryption for lambda env vars - cont

mourya-33:
### Feature or Bugfix
- Bugfix

### Detail
Adding encryption with KMS for lambda environment variables

### Relates
- https://github.com/data-dot-all/dataall/issues/1319

### Security

  - Is the input sanitized? N/A
- Are you using or adding any cryptographic features? Adding KMS key for encrypting lambda env vars for trigger functions and cognito functions


noah-paige:
Hi @mourya-33 - these changes look good but is there any reason you did not also pass `lambda_env_key` to the `SavePerms` TriggerFunctionStack?

deploy/stacks/trigger_function_stack.py
```diff
@@ -52,6 +54,7 @@ def __init__(
             memory_size=256,
             timeout=Duration.minutes(15),
             environment=env,
+            environment_encryption=env_var_encryption_key,
```
petrkalos, line 57: Why not create the KMS key in a method within this stack? Are you planning to reuse a single key across all the lambdas? 

Since this is a security requirement (which I obviously missed) I don't think there will be a scenario that we want a TriggerFunction without env_encryption.

mourya-33: @noah-paige @petrkalos I added the encryption for both the functions as well. 
With regards to creating the key in each method, this will create too many KMS keys just for encrypting the lambda env variables. I am planning to unify this key to be a single key that will encrypt all lambda functions as well in the next iteration. Currently we have different keys for lambda_api , TriggerFunctions and Cognito.

petrkalos (approved):
lgtm!

qq, how did you realise that there were unencrypted env vars? did you use some automation?

zsaltys:
@petrkalos checkov scans... Which we need to add into data.all and to scan synthesized CDK stacks .. It's one of the logged tasks.

Issue #374, opened 2023-03-15
Error in Monitoring tab in Admin Settings


kasturmp:
### Describe the bug

It is not a functional error but happens every time when switching between monitoring and teams tab:
```
An error occurred (AccessDenied) when calling the AssumeRole operation: User: arn:aws:sts::1111111111111:assumed-role/dataall-dev-graphql-role/dataall-dev-graphql is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::1111111111111:role/dataallPivotRole

```

### How to Reproduce

It is not a functional error but happens every time when switching between monitoring and teams tab:

### Screenshots

```
An error occurred (AccessDenied) when calling the AssumeRole operation: User: arn:aws:sts::1111111111111:assumed-role/dataall-dev-graphql-role/dataall-dev-graphql is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::1111111111111:role/dataallPivotRole

```

### OS

All

### Python version

3.1

### AWS data.all version

1.4,1.3

dlpzx:
Hi @kasturmp, thanks for raising the issue. I renamed it and replaced your images as they contained AWS Account IDs. 

From the log I think that the issue is on the Monitoring tab that is trying to fetch some parameter in the infra account assuming the pivotRole, which does not exist. It only affects users using the Monitoring feature, I'll have a look to confirm the hypothesis and I'll try to work on the fix

dlpzx:
@SofiaSazonova @noah-paige @petrkalos @anmolsgandhi ---> Instead of fixing this issue, why don't we implement a better monitoring experience as part of 2.5?

PR #1328, opened 2024-06-12
Move quicksight monitoring to config.json and disable it in FE

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
Quicksight monitoring was enabled or disabled in the cdk.json, however it is more of a functional feature than an infrastructure configuration. Moving it to the config.json allows us to disable the monitoring tab in the frontend. This solves the bug reported in #374 

### Relates
- #374 

### Security

  - Is the input sanitized?


dlpzx:
I tested the frontend locally, but I have not tested the deployment. That's why it is still in draft

noah-paige:
I like the way you changed Monitoring tab to fix this error...

But a general question - do we have any rule of thumb for what should be included in `cdk.json`  vs. `config.json` ?

frontend/src/modules/Administration/views/AdministrationView.js
```diff
@@ -16,11 +16,13 @@ import { ChevronRightIcon, useSettings } from 'design';
...
 import { isModuleEnabled, ModuleNames } from 'utils';
+import config from '../../../generated/config.json';
 
-const tabs = [
-  { label: 'Teams', value: 'teams' },
-  { label: 'Monitoring', value: 'dashboard' }
-];
+const tabs = [{ label: 'Teams', value: 'teams' }];
+
+if (config.core.enable_quicksight_monitoring) {
```
noah-paige, line 23: should it be `config.core.features.enable_quicksight_monitoring`

deploy/stacks/param_store_stack.py
```diff
@@ -146,6 +135,21 @@ def __init__(
                 ),
             )
 
+    @run_if(['core.enable_quicksight_monitoring'])
```
noah-paige, line 138: Should it be `core.features.enable_quicksight_monitoring` ?

deploy/stacks/backend_stack.py
```diff
@@ -436,6 +363,80 @@ def __init__(
                 internet_facing=internet_facing,
             )
 
+    @run_if(['core.enable_quicksight_monitoring'])
```
noah-paige, line 366: Should it be `core.features.enable_quicksight_monitoring` ?

noah-paige (approved):
lgtm - approving

dlpzx:
> I like the way you changed Monitoring tab to fix this error...
> 
> But a general question - do we have any rule of thumb for what should be included in `cdk.json` vs. `config.json` ?

That's the elephant in the room. In principle I would define `cdk.json` parameters as infrastructure parameters used in the deployment of data.all application infra (CICD+FE+BE).  `config.json` parameters are application features' parameters used at runtime in the FE and BE. Most parameters fall into one of those 2 definitions, but we have some exceptions of config.json runtime parameters that affect the infra deployed. We cannot access cdk.json at runtime, (and it is very weird) that's why so far if a parameter needs to be in the config.json but also affects the infra (cdk.json) I have defined them in the config.json. In short:
- parameter that affects infra = cdk.json
- parameter that affects a feature at runtime = config.json
- parameter that affects both = config.json

PR #1330, opened 2024-06-13
Gql schema cleanup sdkcli

noah-paige:
### Feature or Bugfix

- Enhancement / Bugfix

### Detail
- GraphQL Class Improvements
  - Allow Descriptions to be passed to Arguments of GQL Field
  - Handle Field Args for Fields Args of Type Non Null, Array, and Thunk
    - After doing so had to make a bunch of changes on Frontend + Test GQL query strings to ensure we are specifying non null type if required on the backend

-  GQL Return Type Improvements
   - (For Glossary) Remove `assetLink` nested field for types Glossary, Category, & Term -> Not Used
   - (For Glossary) Remove `name` nested field for types Glossary, Category, & Term -> `name` does not exist in RDS (only `label` since not inheriting `Resource`)
   -  (For Glossary) Remove `status` nested field for type Glossary -> Avoid running into errors when returning entire Glossary tree (i.e. Categories and Terms)
       - Error Message Found - `"graphql.error.graphql_error.GraphQLError: Fields 'status' conflict because they return conflicting types 'String' and 'String!'"`
       - Also `status` does not make sense for high level glossary - only for associated Glossary Terms associated/linked with particular data objects

   -  (For Dataset) Remove `isPublishedInEnvironment` and `projectPermissions` fields from Dataset Type
       - Not used by frontend

### Relates
N/A

### Security

  - Is the input sanitized?


petrkalos (approved):
lgtm but I just have a few questions!

* How did you figure out that we had unused fields? Did you automate it?
* Is there any chance that any user is using them in their automation?

noah-paige:
> lgtm but I just have a few questions!
> 
>     * How did you figure out that we had unused fields? Did you automate it?
> 
>     * Is there any chance that any user is using them in their automation?

I found the issues doing some manual tests building out GQL query strings with all possible return fields specified and seeing discrepancies / errors in some cases

For the unused fields `isPublishedInEnvironment` and `projectPermissions` for dataset these columns were used for an older iteration of dataset model but there is no concept of published to env or projects in current data.all + they values are set as `query_expression()` but never resovled with some `with_expression` SQL statement on backend

For `assetLink` for GlossaryNode - there are other queries to directly list term associations/links to glossary nodes so highly unlikely this `assetLink` nested field is used when getting glossaries

Issue #1093, opened 2024-03-07
Extend locking mechanism to different resources


anushka-singh:
**Is your idea related to a problem? Please describe.**
We implemented locking for dataset level. However, we are now facing issues when 2 or more shares are getting approved for the same requestor at the same time. This gives rise to a race condition where the IAM policy of requestor is modified at the same time. 
We observed an error like: 
```
Failed to update policy SHARE_POLICY_ARN : An error occurred (NoSuchEntity) when calling the DeletePolicyVersion operation: Policy SHARE_POLICY_ARN version v3 does not exist or is not attachable.
```

Example:
- UserA in GroupA requests shares to bucket in Dataset B_1 and folder in Dataset B_2
- GroupB (who owns both datasets) approves the share at the same time
- Both shares will be processed at the same time and update the IAM share policy of UserA to give S3 and KMS permissions to the resources (here causes the error)

**Describe the solution you'd like**
Think about a way to extend locking mechanism beyond just dataset locking to different resource types. What if we need to lock resource X during activity Y (i.e. lock datasets from other actions when updating, or similar)

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

dlpzx:
@anushka-singh thanks for opening an issue. This feature came up during validation of 2.3. Do you think you have capacity to take on the extension? In any case, we have added it to our 2.5 project and try to pick it up as soon as possible.

anushka-singh:
@dlpzx I will be working on expiration of shares feature next. But if I have bandwidth for 2.5 then can pick this up shortly after.

noah-paige:
Thoughts on the design for this issue below...

**Changes to DB Schema**
- Rename `dataset_lock` to `resource_lock`
- Rename column `datasetUri` to `resourceUri`
- Add column `acquiredByResourceType` to represent a string/enum value for what resource it is acquired by (i.e. `share`, etc.)
  - This will be additional information alongside the `acquiredBy` Column to store the `shareUri`
  - Do we think this would be useful in long-run (i.e. think if we extend locking to not only be a part of the sharing mechanism) ?

**Changes to Share Logic**
- Check if either datasetUri is locked **_OR_** Requester Principal is locked
  - OPEN ENDED Q: The primary key for Env Group is composite key of groupUri and environmentUri
    - This issue is that GroupA can request access from many dif Envs (need to check lock for groupUri-envUri pair!)
[...]

PR #1338, opened 2024-06-18
Convert Dataset Lock Mechanism to Generic Resource Lock

noah-paige:
### Feature or Bugfix

- Feature
- Bugfix
- Refactoring

### Detail
- Convert Dataset Lock Mechanism to Generic Resource Lock
  - Extend locking to Share principals (i.e. EnvironmentGroup and Consumption Roles)

- Making locking a generic component not tied to datasets

### Relates
- #1093 

### Security

  - Is the input sanitized?


noah-paige:
Local Testing

- [ ] Invite Env Group - Create Resource Lock
- [ ] Add CR - Create Resource Lock
- [ ] Create Dataset - Create Resource Lock
- [ ] Approve Share w/ Env Group
    - 2 Locks Acquired (dataset + Env Group)
    - 2 Locks Released when sharing complete
- [ ] Approve Share w/ ECR
    - 2 Locks Acquired (dataset +  CR)
    - 2 Locks Released when sharing complete
- [ ] Revoke Share w/ Env Group
    - 2 Locks Acquired (dataset + Env Group)
    - 2 Locks Released when revoke complete
- [ ] Revoke Share w/ ECR
    -  2 Locks Acquired (dataset +  CR)
    -  2 Locks Released when revoke complete
- [ ] Remove Env Group - Delete Resource Lock
- [ ] Remove CR - Delete Resource Lock
- [ ] Delete Dataset - Delete Resource Lock

TODO
- [ ] Testing in AWS with multiple concurrent shares

backend/dataall/core/environment/services/environment_service.py
```diff
@@ -295,6 +296,12 @@ def create_environment(uri, data=None):
             )
             session.commit()
 
+            ResourceLockRepository.create_resource_lock(
+                session=session,
+                resource_uri=f'{env_group.groupUri}-{env_group.environmentUri}',
```
petrkalos, line 301: don't we assume that groupUris are unique? why do you need to append the envUri?

backend/dataall/core/resource_lock/db/resource_lock_repositories.py
petrkalos, line 1: Can't we use `acquire` to create an entry in the ResourceLock table and `release` to drop this entry. That way we don't need the create/delete and also the `isLocked` column. All uris that exist in the table we assume that are locked. Wdyt?

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -373,103 +424,26 @@ def _get_share_data_and_items(session, share_uri, status, healthStatus=None):
...
-                return False
-        except Exception as e:
-            session.expunge_all()
-            session.rollback()
-            log.error('Error occurred while acquiring lock:', e)
-            return False
-
-    @staticmethod
-    def acquire_lock_with_retry(dataset_uri, session, share_uri):
+    def acquire_lock_with_retry(
```
petrkalos, line 427: this looks like a useful method, shouldn't we move it in the ResourceLockRepo?

backend/dataall/core/resource_lock/db/resource_lock_repositories.py
```diff
@@ -0,0 +1,122 @@
...
+        session.commit()
+
+    @staticmethod
+    def delete_resource_lock(session, resource_uri):
+        resource_lock = session.query(ResourceLock).filter(ResourceLock.resourceUri == resource_uri).first()
+        session.delete(resource_lock)
+        session.commit()
+
+    @staticmethod
+    def acquire_locks(resources, session, acquired_by_uri, acquired_by_type):
```
petrkalos, line 31: should this be a [ContextManager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager)?

petrkalos (changes requested):

backend/dataall/core/environment/services/environment_service.py
noah-paige: The same group can be a memeber of many different environments

In practice primary key for `environment_group_permission` (aka EnvironmentGroup) is groupUri + environmentUri

We similarly need a way to distinguish each Env + Group Pair for resource lock since the underlying resource we are "locking" is the consumer IAM role for the group in that particular env

backend/dataall/modules/shares_base/services/sharing_service.py
noah-paige: moved

backend/dataall/core/resource_lock/db/resource_lock_repositories.py
noah-paige: I changed logic to make `acquire_lock_with_retry(...)` as a context manager which will acquire on enter and release on exit

Let me know if this is what you were thinking as well

backend/dataall/core/resource_lock/db/resource_lock_models.py
```diff
@@ -0,0 +1,25 @@
...
+        self,
+        resourceUri: str,
+        resourceType: str,
+        acquiredByUri: Optional[str] = None,
+        acquiredByType: Optional[str] = None,
+    ):
+        self.resourceUri = resourceUri
+        self.resourceType = resourceType
+        self.acquiredByUri = acquiredByUri
+        self.acquiredByType = acquiredByType
```
petrkalos, line 25: could this be a `@dataclass` and avoid the boilerplate ctor?

backend/dataall/core/resource_lock/db/resource_lock_repositories.py
petrkalos: I think `acquire_locks` and `release_lock` should be private methods and as such must be prefixed with `_` and optionally moved to the bottom of the file.
If we want to expose acquire/release without a retry we should write another contextmanager.

petrkalos (approved):

Issue #1168, opened 2024-04-12
Show Errors if Share Fails


SofiaSazonova:
As a user I want to see the reason why share request is processed with errors.
Right now we have only health status, but it doesn't show the errors occurred during share processing.

SofiaSazonova:
Suggested places 
<img width="1476" alt="Screenshot 2024-04-12 at 12 44 00" src="https://github.com/data-dot-all/dataall/assets/1146495/5e8cc69e-f6c5-4853-b798-63053a04fd36">
<img width="1443" alt="Screenshot 2024-04-12 at 12 44 10" src="https://github.com/data-dot-all/dataall/assets/1146495/d2caee33-2b44-44e1-8c3d-4d0e4480396c">

PR #1340, opened 2024-06-18
Generic shares_base module and specific s3_datasets_shares module - part 7 (share_object_service)

dlpzx:
### Feature or Bugfix
- Refactoring

### Detail
As explained in the design for #1123 and #1283 we are trying to implement generic `datasets_base` and `shares_base` modules that can be used by any type of datasets and by any type of shareable object in a generic way. 

The goal of this PR is to move the `share_object_service` to `shares_base` and refactor any dependency to S3 in the service.

- Move file and fix imports of ShareObjectService
- Use DatasetsBase and DatasetsBaseRepository instead of the S3 equivalents
- ⚠️ Avoid Dashboard check logic in `ShareObjectService.submit_share_object` see below
- ⚠️ Avoid SharePolicyService logic in `ShareObjectService.create_share_object` see below
- Create ShareLogsService for logs
- Remove unused methods
- I also copied share_item_service to shares_base (it will be used in next PR)

#### Avoid Dashboard check logic in `ShareObjectService.submit_share_object`
Currently, whenever a share request is submitted, we check if the REQUESTER environment has dashboards enabled and if there are shared tables we verify that the Quicksight subscription is active.

Alternative: perform this check in the share processor of tables. It solves the issue, but it gives a poorer user experience as it is difficult to figure out for the requester why the share failed. This can be solved holistically as requested in https://github.com/data-dot-all/dataall/issues/1168.

Decision: move the logic to the processor and make the table share fail.

#### Avoid SharePolicyService logic in `ShareObjectService.create_share_object`

When a share request is first created, we perform a series of operations to ensure that an IAM policy for the share requester principal IAM role is created.

Alternative 1: move this logic inside the share processor. Not sure if it is possible. It would be the ideal solution, but the SharePolicyService throws errors in the create share object API if the policy is not attached. 

Alternative 2: implement interface to define share_policies (similar to the dataset-actions that use share logic in `backend/dataall/modules/s3_datasets_shares/services/dataset_sharing_service.py`).

Decision: we want to preserve the user experience of having the IAM policy created before the share request is processed. Plus, it is not an uncommon pattern that could get extended by other dataset types, for example redshift sharing might need additional share policies. For this reason this PR implements alternative 2

### Relates
- #1283 
- #1123 
- #955 

### Security

  - Is the input sanitized?


dlpzx:
## Testing
- [ ] Open share request modal and view that Consumption roles policies are checked and attached
- [ ] Share request gets created
- [ ] Add items to share request
- [ ] Submit share request WITH TABLES!
- [ ] Approve share request with tables QS subscription enabled --> succeeds
- [ ] same but with QS subcription disabled ---> share failed

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -42,6 +42,14 @@ def generate_empty_policy(self) -> dict:
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def create_managed_policy_from_inline_and_delete_inline(self) -> str:
+        """
+        Returns policy arn and needs to be implemented in the ManagedPolicies inherited classes
+        It is used for backwards compatibility. It should be deprecated and removed in future releases.
+        """
+        raise NotImplementedError
```
petrkalos, line 51: nit: `raise NotImplementedError` isn't required, when there is at least one abstract method python will throw `TypeError: Can't instantiate abstract class T with abstract method ...` when you try to instantiate a derived class. So I suggest that since you are touching this file you might as well remove it from everywhere.

backend/dataall/modules/shares_base/services/share_item_service.py
```diff
@@ -0,0 +1,334 @@
...
+        with get_context().db_engine.scoped_session() as session:
+            return ShareObjectRepository.list_shareable_items(session, share, states, filter)
+
+    @staticmethod
+    @ResourcePolicyService.has_resource_permission(LIST_ENVIRONMENT_SHARED_WITH_OBJECTS)
+    def paginated_shared_with_environment_datasets(session, uri, data) -> dict:
+        return ShareObjectRepository.paginate_shared_datasets(session, uri, data)
+
+    @staticmethod
+    def _get_glue_database_for_share(glueDatabase, account_id, region):
```
petrkalos, line 217: nit: this is a "private" method that I don't think is being used anywhere

petrkalos (approved):

backend/dataall/core/environment/services/managed_iam_policies.py
dlpzx: True, I can leave the `...` !

backend/dataall/modules/shares_base/services/share_item_service.py
dlpzx: I'll handle it in part8 where I have to refactor the share_item_service

PR #1335, opened 2024-06-17
Add search (Autocomplete) in environment/group dropdown in Share request modal

dlpzx:
### Feature or Bugfix
- Feature

### Detail
In the following view, instead of using a fixed list in the environment, team and consumption roles dropdowns; this PR introduces search capabilities as requested in #1012. There are still many other dropdowns to change, where we will be able to extract a common frontend component. This one however, is a bit particular so I implemented it separately.

image

### Relates
- #1012 

### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)?
  - Is the input sanitized?
  - What precautions are you taking before deserializing the data you consume?
  - Is injection prevented by parametrizing queries?
  - Have you ensured no `eval` or similar functions are used?
- Does this PR introduce any functionality or component that requires authorization?
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms?
  - Are you logging failed auth attempts?
- Are you using or adding any cryptographic features?
  - Do you use a standard proven implementations?
  - Are the used keys controlled by the customer? Where are they stored?
- Are you introducing any new policies/roles/users?
  - Have you used the least-privilege principle? How?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1012, opened 2024-01-26
Add possibility to search to all team and environment dropdowns


zsaltys:
I have access to a lot of environments and teams and these team and environment dropdowns are very long. There are multiple places where they are not searchable and are not sorted either. On the team invite modal the team search is sorted and searchable. I propose we use those components everywhere where we need to list teams or environments.

dlpzx:
Hi @zsaltys, thanks for the issue. This feature is implemented in some of the select dropdowns, so we could make it the standards approach for all field selectors. We will pick it up based on bandwidth

dlpzx:
There are multiple places where we list Environments, Teams and Consumption roles. Some can be implemented using reusable frontend components, while others are more peculiar and will be implemented as standalone frontend code snippets.

Environment list:
- [ ] RequestAccessModal.js  ([PR](https://github.com/data-dot-all/dataall/pull/1335))
- [ ] DashboardImportForm.js  
- [ ] MLStudioCreateForm.js 
- [ ] NotebookCreateForm.js 
- [ ] OmicsRunCreateForm.js 
- [ ] PipelineEnvironmentCreateForm.js 
- [ ] PipelineEnvironmentEditForm.js
- [ ] PipelineCreateForm.js 
- [ ] PipelineEditForm.js
- [ ] DatasetCreateForm.js 
- [ ] DatasetImportForm.js
- [ ] WorksheetView.js 


....

noah-paige:
Closing this issue as it is implemented in a series of PRs:
- #1367 
- #1368 
- #1356 
- #1347 

@dlpzx - if there still remain any follow up tasks or enhancements that are "nice to have", let's open a separate issue for them

PR #1347, opened 2024-06-19
Add search (Autocomplete) in environment/group dropdown in create/import forms

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Autocomplete for environments and teams in the following frontend views as requested in 1012. 

This PR introduces a 2 new Shared frontend components: EnvironmentTeamDropdown and EnvironmentTeamDatasetsDropdown. They are generic formik fields that can be reused in Formik forms in several modules. Some creation forms include custom parameters that are fetched alongside the environments and teams (e.g. Notebooks sets VPCs). For those cases, we implement the Autocomplete logic in the view itself

- [X] DashboardImportForm.js - using EnvironmentTeamDropdown
- [X] MLStudioCreateForm.js - using EnvironmentTeamDropdown
- [X] NotebookCreateForm.js - custom
- [X] OmicsRunCreateForm.js - using EnvironmentTeamDatasetsDropdown
- [X] DatasetCreateForm.js - custom
- [X] DatasetImportForm.js - custom

Pipelines in a separate PR

In all forms we use `SamlAdminGroupName` instead of `SamlGroupName` as a standard that allows us to use Shared components

### Relates
- https://github.com/data-dot-all/dataall/pull/1335
- #1012 

### Security

  - Is the input sanitized?


frontend/src/modules/Dashboards/views/DashboardImportForm.js
```diff
@@ -136,7 +81,7 @@ const DashboardImportForm = (props) => {
             dashboardId: values.dashboardId,
             environmentUri: values.environment.environmentUri,
             description: values.description,
-            SamlGroupName: values.SamlGroupName,
+            SamlAdminGroupName: values.SamlAdminGroupName,
```
noah-paige, line 84: Can we keep the key value `SamlGroupName` ? Otherwise - wouldn't we also have to change the Backend importDashboard input type and resolver logic?

frontend/src/modules/Omics/views/OmicsRunCreateForm.js
```diff
@@ -151,7 +64,7 @@ const OmicsRunCreateForm = (props) => {
           workflowUri: omicsWorkflow.workflowUri,
           parameterTemplate: values.parameterTemplate,
           SamlAdminGroupName: values.SamlAdminGroupName,
-          destination: values.destination
```
noah-paige, line 154: similar to importDashboard - can we keep the key value of `destination` for `createOmicsRun` the same

frontend/src/modules/S3_Datasets/views/DatasetCreateForm.js
```diff
@@ -429,38 +429,45 @@ const DatasetCreateForm = (props) => {
...
+                                setFieldValue('environment', value);
+                                fetchGroups(value.environmentUri).catch((e) =>
+                                  dispatch({
+                                    type: SET_ERROR,
+                                    error: e.message
+                                  })
+                                );
+                              } else {
+                                setFieldValue('environment', '');
+                                setGroupOptions([]);
```
noah-paige, line 449: Calling this out here but I think it will be the same to everywhere where we have `<Autocomplete>` of environment which is also setting `GroupOptions`

In my current testing when clicking `X` to delete `environment` value on form - the groupOptions delete but the `SamlAdminGroupName` value remains the same

May have to wrap `SamlAdminGroupName` in a similar way to how we do with `RequestAccessModal.js` as such:

```
                            <Box>
                              {groupOptions.length > 0 ? (
                                ... // CODE WE HAVE NOW
                              ) : (
                                <TextField
                                  error={Boolean(
                                    touched.groupUri && errors.groupUri
                                  )}
                                  helperText={
                                    touched.groupUri && errors.groupUri
                                  }
                                  fullWidth
                                  disabled
                                  label="Team"
                                  value="No teams found for this environment"
                                  variant="outlined"
                                />
                              )}
                            </Box>

Comment (dlpzx) [reply frontend/src/modules/S3_Datasets/views/DatasetCreateForm.js:449]: The PR comments are addressed in https://github.com/data-dot-all/dataall/pull/1347/commits/2298fa7f9a488edcd831b8da37be371e0e01f276. Although , I don't think the issue is completely fixed.

If instead of clicking in the `x` I select another environment the value shown in the form persists. The actual value passed to the API call and the value evaluated in the form is correct, it is just the value shown in the field.

In the next screenshot the group Scientist was selected for env1; then I selected the env wnv2 that does not have that group. I tried to submit the form, but the value for SamlAdminGroup is null. So the value is correct, but whatever is plotted int he form is not updated.

image

Comment (dlpzx) [reply frontend/src/modules/S3_Datasets/views/DatasetCreateForm.js:449]: It is fixed now with the commit: https://github.com/data-dot-all/dataall/pull/1347/commits/a094fbca8d643cf63ea1d18020e8630bbf07de15

File: frontend/src/modules/S3_Datasets/views/DatasetImportForm.js [hunk 4]
```diff
@@ -442,38 +442,45 @@ const DatasetImportForm = (props) => {
...
-                                dispatch({ type: SET_ERROR, error: e.message })
-                              );
-                              setFieldValue('environment', event.target.value);
+                          <Autocomplete
+                            id="environment"
+                            disablePortal
+                            options={environmentOptions.map((option) => option)}
+                            onChange={(event, value) => {
+                              setFieldValue('SamlAdminGroupName', '');
+                              setFieldValue('steward', '');
```
noah-paige, line 451: Change to `stewards` - otherwise text does not clear appropriately

frontend/src/modules/S3_Datasets/views/DatasetCreateForm.js
```diff
@@ -429,38 +429,45 @@ const DatasetCreateForm = (props) => {
...
-                                dispatch({ type: SET_ERROR, error: e.message })
-                              );
-                              setFieldValue('environment', event.target.value);
+                          <Autocomplete
+                            id="environment"
+                            disablePortal
+                            options={environmentOptions.map((option) => option)}
+                            onChange={(event, value) => {
+                              setFieldValue('SamlAdminGroupName', '');
+                              setFieldValue('steward', '');
```
noah-paige, line 438: Change to `stewards` - otherwise text does not clear appropriately

frontend/src/modules/S3_Datasets/views/DatasetImportForm.js
```diff
@@ -558,45 +566,103 @@ const DatasetImportForm = (props) => {
...
-                            renderInput={(renderParams) => (
-                              <TextField
-                                {...renderParams}
-                                label="Stewards"
-                                margin="normal"
-                                onChange={handleChange}
-                                value={values.stewards}
-                                variant="outlined"
-                              />
+                            inputValue={values.steward}
```
noah-paige, line 634: Change to `stewards`

frontend/src/modules/S3_Datasets/views/DatasetCreateForm.js
```diff
@@ -494,45 +502,103 @@ const DatasetCreateForm = (props) => {
...
+                                    error={Boolean(
+                                      touched.SamlAdminGroupName &&
+                                        errors.SamlAdminGroupName
+                                    )}
+                                    helperText={
+                                      touched.SamlAdminGroupName &&
+                                      errors.SamlAdminGroupName
+                                    }
+                                    label="Team"
+                                    variant="outlined"
```
noah-paige, line 535: nit: for here and all other related TextField componenets changed in commit a094fbca8d643cf63ea1d18020e8630bbf07de15 - removing the `onChange={handleChange}` option now no longer allows typing in the textfield 

Would add back unless that option was causing some issue with updating groupOptions and such

frontend/src/modules/S3_Datasets/views/DatasetCreateForm.js
```diff
@@ -494,45 +502,103 @@ const DatasetCreateForm = (props) => {
...
+                                      errors.SamlAdminGroupName
+                                    }
+                                    fullWidth
+                                    disabled
+                                    label="Team"
+                                    value="No teams found for this environment"
+                                    variant="outlined"
+                                  />
+                                )}
+                              </Box>
```
noah-paige, line 554: nit: This TextField component (and all other ones for when groupOptions is empty) has a different sizing then the other TextFields on the form and the other TextField when groupOptions are present

dlpzx: I noticed, not very sure why

frontend/src/modules/S3_Datasets/views/DatasetCreateForm.js
dlpzx: Reverted back, great catch

noah-paige (approved):
lgtm - approving

PR #1306, opened 2024-05-31
Organization Group Permissions Add|Edit

SofiaSazonova:
### Feature or Bugfix
- Feature

### Detail
- Flexible permissions appointment to Ogr Groups
- User now can edit permissions of Ogr Groups
- Can't link env with group that is not in Org
- Can't invite into env groups, that are not in Org

### Relates
- #1147 
- https://github.com/data-dot-all/dataall/pull/1273
### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)?
  - Is the input sanitized?
  - What precautions are you taking before deserializing the data you consume?
  - Is injection prevented by parametrizing queries?
  - Have you ensured no `eval` or similar functions are used?
- Does this PR introduce any functionality or component that requires authorization?
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms?
  - Are you logging failed auth attempts?
- Are you using or adding any cryptographic features?
  - Do you use a standard proven implementations?
  - Are the used keys controlled by the customer? Where are they stored?
- Are you introducing any new policies/roles/users?
  - Have you used the least-privilege principle? How?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1356, opened 2024-06-21
Add search (Autocomplete) in group dropdown for useGroups

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Autocomplete for environments and teams in the following frontend views as requested in 1012.  In this PR the changes apply to those views that use `useGroups` to define the `groupOptions`
- [x] WorksheetCreateForm.js
- [x] EnvironmentCreateForm.js
- [x] OrganizationCreateForm.js
- [x] GlossaryCreateForm.js

❗ In the EnvironmentCreateForm we might need to filter out the list of groups provided. At the moment users can select any group that they belong to. With the changes done by @SofiaSazonova in #1306 I think we need to review if that is still appropriate. Do you think we should open a new PR for this change?

### Relates
- #1012 

### Security

  - Is the input sanitized?

### DIFF
--- a/frontend/src/modules/Environments/views/EnvironmentCreateForm.js
+++ b/frontend/src/modules/Environments/views/EnvironmentCreateForm.js
@@ -17,7 +17,6 @@ import {
   Grid,
   IconButton,
   Link,
-  MenuItem,
   Switch,
   TextField,
   Typography
@@ -73,9 +72,7 @@ const EnvironmentCreateForm = (props) => {
   const [trustedAccount, setTrustedAccount] = useState(null);
   const [pivotRoleName, setPivotRoleName] = useState(null);
   const [loading, setLoading] = useState(true);
-  const groupOptions = groups
-    ? groups.map((g) => ({ value: g, label: g }))
-    : [];
+
   const fetchItem = useCallback(async () => {
     setLoading(true);
     const response = await client.query(getOrganization(params.uri));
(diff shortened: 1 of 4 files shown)


dlpzx:
## Testing:
- [X] Create Worksheet, correct input is passed
- [X] Create Worksheet, list groups, remove group and select another group
- [X] Create Environment, correct input is passed
- [X] Create Organization, correct input is passed
- [X] Create Glossary, correct input is passed


https://github.com/data-dot-all/dataall/assets/71252798/a2f099c0-c1ca-48b6-b15d-13f6c6610c2e

SofiaSazonova:
@dlpzx 
About filter for groups. I thinks, It's more appropriate to list only groups, that are valid for this case: 
1. user belongs to this group
2. group is invited into org

Same holds for the invitation into environment: only groups invited into Org, can be invited into Env.

But on the other hand in can lead to a confusion, when user can't understand, why the target group is not listed. 
That's why I didn't touch this question in #1147

noah-paige (approved):
Changes look good and tested locally - approving

Agree we probably want to filter groups for Env Create Form but think this PR focus is only on Autocomplete component and meets that requirement

Feel free to create new PR or can re-request review on this after adding group filter if you want - but I think we are good to merge this as is and pick up that task separately

dlpzx:
To track this issue on the filtered listGroups queries I opened a separate [issue](https://github.com/data-dot-all/dataall/issues/1370)

PR #1361, opened 2024-06-25
fix migration to not rely on OrganizationService or RequestContext

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Ensure migration script does not need RequestContext - otherwise fails in migration trigger lambda as context info not set / available

### Relates
- https://github.com/data-dot-all/dataall/pull/1306

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/migrations/versions/328e35e39e1e_invite_env_groups_as_readers.py
+++ b/backend/migrations/versions/328e35e39e1e_invite_env_groups_as_readers.py
@@ -11,9 +11,6 @@ from sqlalchemy import orm
 from dataall.core.environment.db.environment_models import EnvironmentGroup, Environment
 from dataall.core.organizations.db.organization_repositories import OrganizationRepository
 from dataall.core.permissions.services.organization_permissions import GET_ORGANIZATION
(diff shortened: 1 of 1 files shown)


noah-paige:
cc @TejasRGitHub - fix for Organization Group migration

backend/migrations/versions/328e35e39e1e_invite_env_groups_as_readers.py
petrkalos, line 1: Is this specific to the customresource/lambda context? it seems to me that even when we had it with a CodeBuild step we'd have the same problem. 

Would it be better to instantiate/set_context in env.py? That way we will be able to use the high-level functions (i.e invite_group) in the migrations scripts.

noah-paige: maybe... we would then need to determine which username and group to set and how to appropriately check permissions for all of the high level functions decorated with `has_resource_permission(...)` and similar

There may be an elegant way to do it and maybe should take some time to think through - for this PR just focused on fixing a bug found

petrkalos (approved):

PR #1367, opened 2024-06-26
Add search (Autocomplete) in miscellaneous dropdowns

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Autocomplete for environments and teams in the following frontend views as requested in #1012. In this case the views required custom dropdowns. 

❗ I used `noOptionsText` whenever it was necessary instead of checking groupOptions lenght >0
- [x] DatasetEditForm.js -> ❗ I kept the stewards field as `freesolo` - what that means is that users CAN specify options that are not on the list. I would like the reviewer to confirm this is what we want. At the end stewardship is a delegation of permissions, it makes sense that delegation happens to other teams. Also changed DatasetCreateForm
- [X] RequestDashboardAccessModal.js - already implemented, minor changes
- [X] EnvironmentTeamInviteForm.js - already implemented, minor changes. -> Kept `freesolo` because invited teams might not be the user teams. Same reason why there is no check for groupOptions == 0, if there are no options there is still the free text option.
- [X] EnvironmentRoleAddForm.js
- [X] NetworkCreateModal.js 

### Relates
- #1012 

### Security

  - Is the input sanitized?

### DIFF
--- a/frontend/src/modules/Catalog/components/RequestDashboardAccessModal.js
+++ b/frontend/src/modules/Catalog/components/RequestDashboardAccessModal.js
@@ -23,9 +23,6 @@ export const RequestDashboardAccessModal = (props) => {
   const dispatch = useDispatch();
   const client = useClient();
   const groups = useGroups();
-  const idpGroupOptions = groups
-    ? groups.map((g) => ({ value: g, label: g }))
-    : [];
 
   async function submit(values, setStatus, setSubmitting, setErrors) {
     try {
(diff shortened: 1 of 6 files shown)


noah-paige:
for `EnvironmentRoleAddForm` and `NetworkCreateModal` I can type out any group name and submit the form even if it is not in the options

- removing the lines `// onChange={handleChange}` and changing `inputValue={values.groupUri}` to `value={values.groupUri}` seemed to resolve

- this may be occuring in a number of places may want to check
    - update: tested with DatasetCreateForm and also could type in random string for Team without it being discarded when clicking out
    - commenting out `inputValue={values.SamlAdminGroupName}` seemed to resolve (I believe we added this to help with text clearing when clear env but it seems like text is still clearing appropriately)

noah-paige:
Instead of `freeSolo` would it make more sense for the groupOptions for `stewards` to be the groups in the Environment

and for `EnvironmentTeamInvite` the options for group be the OrganizationGroup members?  

`freeSolo` can be fine too but should likely have some validation whether a group is valid on the backend in that case (if not already there)

dlpzx:
To track this issue on the filtered listGroups queries I opened a separate [issue](https://github.com/data-dot-all/dataall/issues/1370)

dlpzx:
Hi @noah-paige, I have fixed the glitch in the NetworkCreate view and in the inviteRole form. For the views where we have an environment selector and a group selector is a little more difficult.
- if we use inputValue, the group current text does not get cleared out
- if we use value, the group selected does not get cleared out after another environment is selected
I think we could consider it as a feature out of the scope of this PR

noah-paige (approved):
lgtm

Issue #1220, opened 2024-04-25
Integration tests executed on a real deployment as part of the CICD


dlpzx:
**Is your feature request related to a problem? Please describe.**
I would like data.all to be tested "in reality" once it is deployed. That allows us to verify code changes on real infrastructure. It speeds up development because it increases the confidence that developers and maintainers have that a feature does not introduce bugs.

**Describe the solution you'd like**
A way to execute locally and in a CICD pipeline tests that run against a real API of data.all.

**Describe alternatives you've considered**
Defining the tests using pytest. Once #950 is complete we will be able to use the SDK directly to configure the test clients and the API call definitions.

**Additional context**

*P.S. Please Don't attach files. Add code snippets directly in the message body instead.*

PR #1371, opened 2024-06-27
add environment tests

petrkalos:
### Feature or Bugfix
Feature

### Detail
* add list_environment tests
* add test for updating an environment (via update_stack)
* generalise the polling functions for stacks

### Relates
#1220 

### Security

  - Is the input sanitized?

### DIFF
--- a/tests_new/integration_tests/core/environment/global_conftest.py
+++ b/tests_new/integration_tests/core/environment/global_conftest.py
@@ -1,33 +1,32 @@
 import logging
-import re
 
 import pytest
 
 from integration_tests.client import GqlError
-from integration_tests.core.environment.queries import create_environment, get_environment, delete_environment
-from integration_tests.utils import poller
+from integration_tests.core.environment.queries import (
+    create_environment,
+    get_environment,
+    delete_environment,
+    list_environments,
+)
+from integration_tests.core.organizations.queries import create_organization
+from integration_tests.core.stack.utils import check_stack_ready
 
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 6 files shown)


tests_new/integration_tests/core/environment/global_conftest.py
```diff
@@ -87,5 +86,21 @@ def temp_env1(client1, group1, org1, testdata):
...
-def persistent_env1(client1, group1, org1, testdata): ...  # TODO
+def get_or_create_persistent_env(env_name, client, group, testdata):
+    envs = list_environments(client, term=env_name).nodes
+    if envs:
+        return envs[0]
+    else:
+        envdata = testdata.envs[env_name]
+        org = create_organization(client, f'org_{env_name}', group)
+        env = create_env(client, group, org['organizationUri'], envdata.accountId, envdata.region, tags=[env_name])
+        if env.stack.status == 'CREATE_COMPLETE':
```
dlpzx, line 97: What about UPDATE_COMPLETE?

dlpzx: Every day there is an scheduled task that updates environment and dataset stacks. It is an edge case that exactly at the same time we run our tests, but it could happen. Should we consider it in the design?

dlpzx (commented):
Left a small comment, but it looks great

tests_new/integration_tests/core/environment/global_conftest.py
petrkalos: Although this will be pretty unlikely, polling is 10seconds, but it won't harm adding it. Will also have to add it to the assertions.

dlpzx (approved):

PR #1280, opened 2024-05-16
Scope down dataset sharing requester IAM role managed IAM policy S3 permissions

mourya-33:
### Feature or Bugfix
- Bugfix

### Detail
This PR will address the issue - 1226

### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)? N/A
  - Is the input sanitized? N/A
  - What precautions are you taking before deserializing the data you consume? N/A
  - Is injection prevented by parametrizing queries? N/A
  - Have you ensured no `eval` or similar functions are used? N/A
- Does this PR introduce any functionality or component that requires authorization? N/A
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms? N/A
  - Are you logging failed auth attempts? N/A
- Are you using or adding any cryptographic features? N/A
  - Do you use a standard proven implementations? N/A
  - Are the used keys controlled by the customer? Where are they stored? N/A
- Are you introducing any new policies/roles/users? No
  - Have you used the least-privilege principle? How?Yes, restricted the s3:* permissions to s3 readonly permissions

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1374, opened 2024-06-28
Dataall code/infrastructure AWS migrations

SofiaSazonova:
### Feature or Bugfix
- Feature

### Detail
During the work on #1280 we encountered a problem: we need to revise all share-policies in already existed AWS deployments (remove s3:* permissions and add appropriate ones).   It sounds like "migration", but has nothing to do with database, so it doesn't belong to alembic migrations module.

As  a general solution to such problems, we introduced 'dataall migrations' -- a module, that is called in lambda function, triggered by update of backend stack, after DB-migrations. This module has the same logic of migrations, which are executed one by one. Current revision uid is stored in SSM Parameter store.

Module contains:|
 - BaseDataAllMigration -- python class  -- a base for future migrations
 - Herder -- python class, that handles migrations
 - folder "versions", where python files are stored with subclasses of BaseDataAllMigration.

Each subclass has parameters
   - uid of the revision `key = '51132fed-c36d-470c-9946-5164581856cb'
   - name ` name = 'Remove Wildcard from Sharing Policy'
   - short description `description = 'Remove Wildcard from Sharing Policy'
   - uid of previous migration, to specify the order `previous_migration = '0'  # initial migration`
   - possible methods `up` and `down` for upgrade and downgrade respectively
   
In order to add new migration, developer needs to create a python class in folder "versions" (folder name is hardcoded by now) and specify the parameters described above. Right now it can be done only manually.

### Security

  - Is the input sanitized?


backend/deployment_triggers/dataall_migrate_handler.py
```diff
@@ -0,0 +1,44 @@
...
+        print(f'Error: {e}')
+        return -1
+
+
+def put_parameter_to_parameter_store(value):
+    try:
+        client.put_parameter(Name=PARAM_KEY, Value=value, Type='String', Overwrite=True)
+    except Exception as e:
+        # Handle other exceptions
+        print(f'Error: {e}')
```
petrkalos, line 34: Should you reuse [ParameterStoreManager](https://github.com/data-dot-all/dataall/blob/main/backend/dataall/base/aws/parameter_store.py#L14)?

backend/deployment_triggers/dataall_migrate_handler.py
```diff
@@ -0,0 +1,44 @@
...
+logger = logging.getLogger()
+logger.setLevel(os.environ.get('LOG_LEVEL', 'INFO'))
+
+client = boto3.client('ssm')
+
+envname = os.environ.get('ENVNAME', 'local')
+PARAM_KEY = (f'/dataall/{envname}/dataall-migration/revision',)
+
+
+def get_parameter_from_parameter_store():
```
petrkalos, line 15: Did you consider storing the revision in the db?

backend/deployment_triggers/dataall_migrate_handler.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def handler(event, context) -> None:
+    revision = get_parameter_from_parameter_store()
+    if revision == -1:
+        logger.error('Failed to retrieve revision from parameter store')
+        return
+    H = Herder()
+    H.upgrade(start_key=revision)
+    put_parameter_to_parameter_store(H.last_key)
```
petrkalos, line 44: `H` -> `herder`

backend/migrations/dataall_migrations/base_migration.py
```diff
@@ -0,0 +1,58 @@
...
+    name = 'Base Migration'
+    description = 'Base Migration'
+
+    @classmethod
+    def up(cls):
+        print(f'Upgrade is not defined for migration {cls.name}')
+
+    @classmethod
+    def down(cls):
+        print(f'Downgrade is not defined for migration {cls.name}')
```
petrkalos, line 30: I think this class is mixing management and execution/business logic. 

I think this should be an `ABC` with abstact props/methods `name` `description` `up` `down`  which will force all derived classes to implement them. Similarly to alembic instead of key I'd have a `previous_revision` and `next_revision`.

Then the herder/management should deal with the execution order much easier than it currently does.


Wdyt?

backend/deployment_triggers/dataall_migrate_handler.py
```diff
@@ -0,0 +1,44 @@
...
+PARAM_KEY = (f'/dataall/{envname}/dataall-migration/revision',)
+
+
+def get_parameter_from_parameter_store():
+    try:
+        parameter = client.get_parameter(Name=PARAM_KEY, WithDecryption=True)
+        return parameter['Parameter']['Value']
+    except client.exceptions.ParameterNotFound:
+        # Handle the case where the parameter is not found
+        print(f"Error: Parameter '{PARAM_KEY}' not found.")
```
petrkalos, line 21: use loggers

zsaltys:
@SofiaSazonova I could not find a ticket for this PR. Few things that are interesting to me:

1) what happens if a migration fails? How does the owner of data.all find out that some migrations failed?
2) what path on CloudWatch we can see migration logs?
3) What is the purpose of the key? Can we generate it dynamically for example just hashing the migration class name with self reflection?

SofiaSazonova:
@zsaltys 
It's a PR without a ticket, it was born from #1280 
Please, note, that it's still a draft.

1. If migration fails, pipeline fails. The same as for alembic migrations
2.  CloudWatch log is named smth like `/aws/lambda/dataall-dev-deployment_triggers_dataall_migrate_handler_handler` but with respect of deployment resource-prefix
3. Key is UID (except for the initial one, its key is '0'). Right now it's not automated, but in future (not in 2.6) I will add autogeneration of keys (as, e.g. alembic does).

backend/deployment_triggers/dataall_migrate_handler.py
SofiaSazonova: already) I'll submit changes soon. So, this PR is a draft

deploy/stacks/trigger_function_stack.py
```diff
@@ -42,12 +43,27 @@ def __init__(
 
         function_sgs = self.create_lambda_sgs(envname, handler, resource_prefix, vpc)
         statements = self.get_policy_statements(resource_prefix) + (additional_policy_statements or [])
+
+        self.role_name = role_name
```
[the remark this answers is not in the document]
SofiaSazonova: I can go for a default option.
If role name is not specified -- autogenerate it with  suffix  = self.id

In this case, I wanted to grant assume role permissions for pivotRole, and followed the naming, that was used there
`                'aws:PrincipalArn': [
                  !Sub "arn:aws:iam::${AwsAccountId}:role/*graphql-role",
                  !Sub "arn:aws:iam::${AwsAccountId}:role/*awsworker-role",
                  !Sub "arn:aws:iam::${AwsAccountId}:role/*ecs-tasks-role",
                  !Sub "arn:aws:iam::${AwsAccountId}:role/*dataall-migration-role",`

backend/migrations/dataall_migrations/versions/remove_wildcard_share_policy.py
```diff
@@ -0,0 +1,66 @@
...
+    name = 'Remove Wildcard from Sharing Policy'
+    description = 'Remove Wildcard from Sharing Policy'
+
+    previous_migration = '0'
+
+    @classmethod
+    def up(cls):
+        print('removing wildcard from sharing policy')
+        ENVNAME = os.environ.get('envname', 'local')
+        ENGINE = get_engine(envname=ENVNAME)
```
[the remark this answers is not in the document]
SofiaSazonova: ok

backend/deployment_triggers/dataall_migrate_handler.py
SofiaSazonova: Good idea. Thanks, missed it

backend/deployment_triggers/dataall_migrate_handler.py
SofiaSazonova: Yes, I was thinking about it. But then I thought: 
1. It's only one parameter and it's related to AWS resources
2. Dataall migrations are not related to DB in general, so, in general it shouldn't access DB
3. SSP Parameter store is free

So, I went for SSM

backend/deployment_triggers/dataall_migrate_handler.py
SofiaSazonova: ok

backend/migrations/dataall_migrations/base_migration.py
SofiaSazonova: I thought about bi-directional links to previous and next revisions.
I postponed it, because now the keys are appointed manually, so I don't want developer to do more actions, than necessary. I'm not sure about our timeline, so I don't know, if I can come back to this any time soon. 
Since, there are only one migration now and I don't think we will have many more, I would leave it for further enhancements. 
Of course, if you think, that the current solution is not good enough for main, I can fix it)) But I don't want to postpone this PR, because we are close to 2.6.

SofiaSazonova:
Tests:

Pre-reqs:
Dataset is shared with Consumer Role.

1. Edit share policy of the consumer role with Sid "BucketStatementS3" to have only "s3:*" permission. 
2. Update pipeline

- [ ] Pipeline finished successfully
- [ ] Share policy edited. `"s3:*"` permission removed, 
- [ ] `'s3:List*', 's3:Describe*', 's3:GetObject' `permissions are added
- [ ] SSM parameter `dataall/{envname}/dataall-migration/revision` is created with value `'51132fed-c36d-470c-9946-5164581856cb'`

backend/deployment_triggers/dataall_migrate_handler.py
petrkalos: would having it in SSM make local development trickier?

deploy/stacks/trigger_function_stack.py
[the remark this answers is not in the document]
petrkalos: if the default naming convention is enough to distinguish the role then why add the extra complication of setting it explicitly?

backend/migrations/dataall_migrations/base_migration.py
petrkalos: Thinking about it we don't even need bidirectional since we probably don't have a scenario that we will want to go to the previous node. 
I think such an implementation (simple single linked list traversal) is quite easy to implement, do you think you can do it in the current timeframe?

backend/deployment_triggers/dataall_migrate_handler.py
petrkalos: also I am thinking that RDS is the only place we keep the app's state but with this change we will also use SSM. I am trying to over think it because the decision we make now might be difficult to be changed in the future.

backend/dataall/base/aws/parameter_store.py
```diff
@@ -29,37 +29,29 @@ def client(AwsAccountId=None, region=None, role=None):
...
             raise Exception('Parameter name is None')
-        try:
-            parameter_value = ParameterStoreManager.client(AwsAccountId, region).get_parameter(Name=parameter_path)[
-                'Parameter'
-            ]['Value']
-        except ClientError as e:
-            raise Exception(e)
+        parameter_value = ParameterStoreManager.client(AwsAccountId, region).get_parameter(Name=parameter_path)[
+            'Parameter'
+        ]['Value']
```
dlpzx, line 34: what are the implications of removing the try-except for other parts of the code?

dlpzx: I see that it is used by the pivot_role utils, maintenance_service, other API calls

dlpzx (commented):
We need to add a README file the same way we do with db migrations. Users need to understand how to add a new migration, how to test locally...

backend/deployment_triggers/dataall_migrate_handler.py
dlpzx: They are very similar options, for me the "untie" factor is that if we store it in db we can run the migrations locally without a manually-created SSM parameter. On the other hand, I think most migrations will have to look-up information in RDS

backend/migrations/dataall_migrations/herder.py
```diff
@@ -0,0 +1,79 @@
...
+        self.migration_path = {}
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        self.folder_path = os.path.join(dir_path, 'versions')
+        self.initial_key = '0'
+        self.last_key = None
+
+        logger.info('Loading migrations...')
+        logger.info(f'Folder path: {self.folder_path}')
+
+        for py_file in Path(self.folder_path).glob('*.py'):
```
dlpzx, line 25: will this get the __init__ file?

backend/migrations/dataall_migrations/base_migration.py
```diff
@@ -0,0 +1,65 @@
...
+    @classmethod
+    def set_next(cls, next_migration_key):
+        cls.next_migration = next_migration_key
+
+    @classmethod
+    def next(cls):
+        if 'next_migration' in cls.__dict__:
+            return cls.next_migration
+        return None
+
```
dlpzx, line 52: don't we need to ensure that every new migration contains the class variable `'previous_migration'`? If we return None it might break the herder and go unnoticed

dlpzx:
What happens when a migration fails? do we revert changes, update the version? Are we pushing this responsibility to the test-author? meaning, do tests need to be robust enough that they can be run several times?

backend/deployment_triggers/dataall_migrate_handler.py
SofiaSazonova: @petrkalos 
Basically, this "dataall-migrations" are designed for software changes, that are not related to database -- so, they are AWS-related. It means, that anyhow we need to have aws-access to run them, can't run them locally.

backend/dataall/base/aws/parameter_store.py
SofiaSazonova: @dlpzx these try-except blocks make no sense. Even more, they make obstacles for error handling.
1. All of them simply catch error and do nothing but throw it further
2. Instead of ClientError (which inherits Exception) there blocks just throw basic Exception making it hard to distinguish different types of errors.
We had conversation with @petrkalos and decided to drop these statements

backend/migrations/dataall_migrations/herder.py
SofiaSazonova: @dlpzx It will, but there is nothing there. It is only for module init purposes.

backend/migrations/dataall_migrations/base_migration.py
SofiaSazonova: No, It won't break anything for sure. The worst case scenario the migration without parent will simply not be run.

SofiaSazonova:
@dlpzx  Good point about reverting the changes.
If migrations fails now, the stage will fail, but yes, previous migrations won't be reverted. I should add this.

backend/deployment_triggers/dataall_migrate_handler.py
dlpzx: Kind of, they are AWS operations on data.all deployed infra or other SDK calls (e.g. shares) that can be deployed/executed locally. I would like to have a running deployment of data.all locally run the migrations and see if the changes are correct.

backend/migrations/dataall_migrations/herder.py
dlpzx, line 1: Maybe we can have a more descriptive name for herder. Like MigrationManager

backend/migrations/dataall_migrations/base_migration.py
dlpzx: The point is that alembic migrations are only valid when we have a single thread of migrations: A -> B -> C -> D. If no previous migration is specified the herder won't notice it which means that we might have a broken thread of migrations: A -> B, C (without previous), C->D. In this case  the herder will just run A and B and stop there. I think we need to ensure that migrations are valid and that there is one single unbroken thread

backend/migrations/dataall_migrations/herder.py
SofiaSazonova: done

backend/migrations/dataall_migrations/base_migration.py
SofiaSazonova: I made the BaseMigration ABC

SofiaSazonova:
@dlpzx  changes in my last commit protect 'single thread' migration structure

deploy/stacks/trigger_function_stack.py
```diff
@@ -42,12 +43,27 @@ def __init__(
 
         function_sgs = self.create_lambda_sgs(envname, handler, resource_prefix, vpc)
         statements = self.get_policy_statements(resource_prefix) + (additional_policy_statements or [])
+
+        self.role_name = role_name if role_name else f'{id}Role'
```
petrkalos, line 47: nit: `role_name or f'{id}Role'`

backend/migrations/dataall_migrations/migrationmanager.py
```diff
@@ -0,0 +1,67 @@
...
+            except Exception as e:
+                # Something went wrong revert
+                logger.info(f'An error occurred while applying the migration.{e}.')
+                while migration := self.executed_upgrades.pop():
+                    migration.down()
+                return False
+        logger.info('Upgrade completed')
+        return self.executed_upgrades.pop().key()
+
+    def downgrade(self, target_key='0'):
```
petrkalos, line 47: this is unused, thus I don't see why having it in the codebase

backend/migrations/dataall_migrations/migrationmanager.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self, key='0'):
+        self.current_migration = InitMigration
+        self.previous_migrations: Deque[BaseDataAllMigration] = deque()
+        while self.current_migration.key() != key:
+            self.previous_migrations.append(self.current_migration)
+            self.current_migration = self.current_migration.next_migration()
+        self.executed_upgrades: Deque[BaseDataAllMigration] = deque()
+        self.executed_downgrades: Deque[BaseDataAllMigration] = deque()
+
+    def upgrade(self, target_key=None):
```
petrkalos, line 23: I see that that we don't invoke upgrade with `target_key`, should we remove it along with all the extra checks/logic around it?

backend/migrations/dataall_migrations/migrationmanager.py
```diff
@@ -0,0 +1,67 @@
...
+            try:
+                self.executed_upgrades.append(self.current_migration)
+                logger.info(f'Applying migration {self.current_migration.name()}')
+                self.current_migration.up()
+                logger.info(f'Migration {self.current_migration.name()} completed')
+                self.current_migration = self.current_migration.next_migration()
+            except Exception as e:
+                # Something went wrong revert
+                logger.info(f'An error occurred while applying the migration.{e}.')
+                while migration := self.executed_upgrades.pop():
```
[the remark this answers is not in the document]
SofiaSazonova: Changed it to safe cycle.

Test
`from collections import deque

q = deque()
for i in range(10):
    q.append(i)

while q:
    print(q.pop())`

backend/migrations/dataall_migrations/migrationmanager.py
SofiaSazonova: I would save it for local testing purposes.

backend/migrations/dataall_migrations/migrationmanager.py
SofiaSazonova: I would save it for local testing purposes.

backend/migrations/dataall_migrations/base_migration.py
```diff
@@ -0,0 +1,49 @@
...
+    @abstractmethod
+    def description() -> str:
+        """
+        Short description of migration logic and purpose. To be implemented in the  inherited classes
+        """
+        ...
+
+    @staticmethod
+    @abstractmethod
+    def next_migration() -> Type['MigrationBase'] | None:
```
petrkalos, line 29: I think you need `Union` instead of `|` for `python<3.12`

backend/migrations/dataall_migrations/base_migration.py
```diff
@@ -0,0 +1,49 @@
...
+logger = logging.getLogger()
+logger.setLevel(os.environ.get('LOG_LEVEL', 'INFO'))
+
+
+class MigrationBase(ABC):
+    @staticmethod
+    @abstractmethod
+    def revision_id() -> str:
+        """
+        Uniq revision identifier.  To be implemented in the  inherited classes
```
petrkalos, line 15: nit: I'd remove the "To be implemented.." comment, it's obvious from the `@abstractmethod`

backend/migrations/dataall_migrations/base_migration.py
```diff
@@ -0,0 +1,49 @@
...
+import os
+from abc import ABC, abstractmethod
+from typing import Type
+
+logger = logging.getLogger()
+logger.setLevel(os.environ.get('LOG_LEVEL', 'INFO'))
+
+
+class MigrationBase(ABC):
+    @staticmethod
```
petrkalos, line 11: should all the methods be @classmethods in case we ever need the cls info?

backend/migrations/dataall_migrations/migrationmanager.py
```diff
@@ -0,0 +1,65 @@
...
+
+logger = logging.getLogger()
+logger.setLevel(os.environ.get('LOG_LEVEL', 'INFO'))
+
+
+class MigrationManager:
+    def __init__(self, key='0', initial_migration=InitMigration):
+        self.current_migration = initial_migration
+        self.previous_migrations: Deque[MigrationBase] = deque()
+        while self.current_migration.revision_id() != key:
```
petrkalos, line 17: if `key` doesn't exist this will blow up ungracefully I think `AttributeError: 'NoneType' object has no attribute 'revision_id'`

backend/migrations/dataall_migrations/migrationmanager.py
petrkalos: Let's assume that we have the migrations `0->1->2->3`, our current revision is `2` (hence ctor key=`2`).
Problems...
* If we call upgrade `target_key=None`
* If we call upgrade `target_key=1` it will execute all the migations to the end
* If we call upgrade 2 times then we will get an exception because `self.current_migration = None`

I suspect there are more combinations of init/up/down that will make this blow up as well.

backend/migrations/dataall_migrations/versions/initial.py
```diff
@@ -0,0 +1,25 @@
+from migrations.dataall_migrations.base_migration import MigrationBase
+from migrations.dataall_migrations.versions.remove_wildcard_share_policy import RemoveWildCard
+from typing import Type
+
+
+class InitMigration(MigrationBase):
```
petrkalos, line 6: I am thinking it over again and I don't think the initial migration (dummy node) is required. We can just do `mm = MigrationManager(RemoveWildcard, current_key)`

petrkalos (changes requested):

petrkalos (changes requested):

backend/migrations/dataall_migrations/migrationmanager.py
```diff
@@ -0,0 +1,124 @@
...
+            return self.current_migration.revision_id()
+
+        logger.info(f"Upgrade from {self.current_migration.revision_id()} to {target_revision_id or 'latest'}")
+        executed_upgrades: Deque[Union[Type[MigrationBase], None]] = deque()
+        saved_start_migration = self.current_migration
+        while self.current_migration is not None:
+            try:
+                logger.info(f'Applying migration {self.current_migration.__name__}')
+                self.current_migration.up()
+                executed_upgrades.append(self.current_migration)
```
petrkalos, line 75: If we throw on `up` that means that the current_migration won't be added to executed_upgrades. As a result in the `Exception` block `down` will not be executed for it and might leave the environment in limbo. I am not sure what's the correct behaviourm, should we call `down` on the migration that caused the exception on `up`?

backend/migrations/dataall_migrations/migrationmanager.py
```diff
@@ -0,0 +1,124 @@
...
+                logger.info(f'Migration {self.current_migration.__name__} completed')
+                if (
+                    self.current_migration.revision_id() == target_revision_id
+                    or self.current_migration.next_migration() is None
+                ):
+                    break
+                self.current_migration = self.current_migration.next_migration()
+            except Exception as e:
+                # Something went wrong revert
+                logger.info(f'An error occurred while applying the migration.{e}.')
```
petrkalos, line 85: This will only print the exception message, I think it would be nice if we can see the stack trace here. Try `logger.exception('An error occurred while applying the migration.')` [docs](https://docs.python.org/3/library/logging.html#logging.Logger.exception) which does that by default.

backend/migrations/dataall_migrations/migrationmanager.py
```diff
@@ -0,0 +1,124 @@
...
+                while executed_upgrades:
```
petrkalos, line 86: would it make sense to call `self.downgrade` here?

backend/migrations/dataall_migrations/migrationmanager.py
```diff
@@ -0,0 +1,124 @@
...
+        while self.current_migration:
+            if self.current_migration.revision_id() == target_revision_id:
+                break
+            try:
+                logger.info(f'Reverting migration {self.current_migration.__name__}')
+                self.current_migration.down()
+                executed_downgrades.append(self.current_migration)
+                logger.info(f'Migration {self.current_migration.__name__} completed')
+            except Exception as e:
+                logger.info(f'An error occurred while reverting the migration.{e}.')
```
petrkalos, line 111: I wonder if that's something we want to do (upgrade again in case downgrade fail). Also similar to my upgrade function should we call `self.upgrade` here (but avoid inf loops)?

tests/migrations/test_dataall_migrations.py
```diff
@@ -1,71 +1,190 @@
...
-    try:
-        manager = MigrationManager(current_revision, TestInitMigration)
-    except Exception as e:
-        assert f'{e}' == f'Revision {current_revision} is not found.'
+    incorrect_key = 'incorrect'
+
+    m3 = get_migration('3', None)
+    m2 = get_migration('2', m3)
+    m1 = get_migration('1', m2)
+    m0 = get_migration('0', m1)
```
petrkalos, line 67: nit: you can DRY it tests by creating 2 fixtures one that returns m0->m3 and one m0->m3 with m2 throwing.

backend/migrations/dataall_migrations/migrationmanager.py
```diff
@@ -68,6 +68,7 @@ def upgrade(self, target_revision_id=None):
         logger.info(f"Upgrade from {self.current_migration.revision_id()} to {target_revision_id or 'latest'}")
         executed_upgrades: Deque[Union[Type[MigrationBase], None]] = deque()
         saved_start_migration = self.current_migration
+        self.current_migration = self.current_migration.next_migration()
```
petrkalos, line 71: I think this will make the following fail although it's valid
```
mm = Manager('0', Initial)
mm.upgrade()
mm.downgrade()
```
That's because `previous_migrations` won't contain the InitialMigration

Also I think for the same reason...
```
    manager = MigrationManager('1', m0)
    new_version = manager.upgrade()
    m0.up.assert_not_called()
    m1.up.assert_not_called()
    m2.up.assert_called_once()
    m3.up.assert_called_once()
    manager.downgrade()
    m1.down.assert_not_called()
    m2.down.assert_called_once() <--called twice
    m3.down.assert_called_once() <--called twice
```

petrkalos (changes requested):

backend/migrations/dataall_migrations/migrationmanager.py
SofiaSazonova: I think, if it's not executed, then we don't need to downgrade.
If it fails and leaves garbage behind -- I think, it's a bad migration.
If we execute here downgrade, it may as well be unsafe, if poorly written

SofiaSazonova: @dlpzx what do you think?

dlpzx: If I understand correctly, we are trying to minimize the danger of garbage or halfway done upgrades right? We can either 1) make the migrations atomic - if anything fails in the upgrade it should revert back the changes. Or we can 2) make the migrations retry-able - if anything fails we execute downgrade.Although in theory I think 1) makes more sense, in practice I think it is easier to implement 2), plus it has other advantages (e.g. retry for any other reason). At this moment, the remove_wildcard script fulfills 2 but not 1

PR #1382, opened 2024-07-02
Integration Tests Notebooks

noah-paige:
### Feature or Bugfix

- Feature

### Detail
- <feature1 or bug1>
- <feature2 or bug2>

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?

### DIFF
--- a/tests_new/integration_tests/conftest.py
+++ b/tests_new/integration_tests/conftest.py
@@ -14,6 +14,7 @@ logging.basicConfig(level=logging.INFO, stream=sys.stdout)
 pytest_plugins = [
     'integration_tests.core.organizations.global_conftest',
     'integration_tests.core.environment.global_conftest',
+    'integration_tests.modules.notebooks.global_conftest',
 ]
 
 
--- a/tests_new/integration_tests/modules/notebooks/global_conftest.py
+++ b/tests_new/integration_tests/modules/notebooks/global_conftest.py
@@ -1,0 +1,154 @@
+import logging
+
+import pytest
+
+from integration_tests.client import GqlError
+from integration_tests.modules.notebooks.queries import (
+    create_sagemaker_notebook,
+    get_sagemaker_notebook,
+    delete_sagemaker_notebook,
+    list_sagemaker_notebooks,
+)
+from integration_tests.core.stack.utils import check_stack_ready
+
+from integration_tests.modules.notebooks.aws_clients import VpcClient
+
+log = logging.getLogger(__name__)
+
+
+def create_notebook(client, group, org_uri, env_uri, vpc_id, subnet_id, tags=[]):
+    notebook = create_sagemaker_notebook(
+        client=client,
+        name='TestNotebook',
+        group=group,
+        organizationUri=org_uri,
+        environmentUri=env_uri,
+        VpcId=vpc_id,
+        SubnetId=subnet_id,
+        tags=tags,
+    )
+    check_stack_ready(client, env_uri, notebook.stack.stackUri)
+    return get_sagemaker_notebook(client, notebook.environmentUri)
+
+
+def delete_notebook(client, env_uri, notebook):
+    check_stack_ready(client, env_uri, notebook.stack.stackUri)
+    try:
+        return delete_sagemaker_notebook(client, notebook.environmentUri)
+    except GqlError:
+        log.exception('unexpected error when deleting environment')
+        return False
+
+
+"""
+Session envs persist accross the duration of the whole integ test suite and are meant to make the test suite run faster (env creation takes ~2 mins).
+For this reason they must stay immutable as changes to them will affect the rest of the tests.
+"""
+
+
+@pytest.fixture(scope='session')
+def session_notebook1(client1, group1, org1, session_env1, session_id, session_env1_aws_client):
+    resource_name = 'sessionnotebook1'
+    notebook = None
+    try:
+        vpc_client = VpcClient(session=session_env1_aws_client, region=session_env1['region'])
+        vpc_id = vpc_client.create_vpc(vpc_name=resource_name, cidr='172.31.0.0/26')
+        subnet_id = vpc_client.create_subnet(vpc_id=vpc_id, subnet_name=resource_name, cidr='172.31.0.0/28')
+
+        notebook = create_notebook(
+            client1,
+            group=group1,
+            org_uri=org1['organizationUri'],
+            env_uri=session_env1['environmentUri'],
+            tags=[session_id],
+            vpc_id=vpc_id,
+            subnet_id=subnet_id,
+        )
+        yield notebook
+    finally:
+        if notebook:
+            delete_notebook(client1, session_env1['environmentUri'], notebook)
+        vpc_client = VpcClient(session=session_env1_aws_client, region=session_env1['region'])
+        vpc_client.delete_subnet_by_name(resource_name)
+        vpc_client.delete_vpc_by_name(resource_name)
+
+
+"""
+Temp envs will be created and deleted per test, use with caution as they might increase the runtime of the test suite.
+They are suitable to test env mutations.
+"""
+
+
+@pytest.fixture(scope='function')
+def temp_notebook1(client1, group1, org1, session_env1, session_id, session_env1_aws_client):
+    resource_name = 'tempnotebook1'
+    notebook = None
+    try:
+        vpc_client = VpcClient(session=session_env1_aws_client, region=session_env1['region'])
+        vpc_id = vpc_client.create_vpc(vpc_name=resource_name, cidr='172.31.0.0/26')
+        subnet_id = vpc_client.create_subnet(vpc_id=vpc_id, subnet_name=resource_name, cidr='172.31.0.0/28')
+
+        notebook = create_notebook(
+            client1,
+            group=group1,
+            org_uri=org1['organizationUri'],
+            env_uri=session_env1['environmentUri'],
+            tags=[session_id],
+            vpc_id=vpc_id,
+            subnet_id=subnet_id,
+        )
+        yield notebook
+    finally:
+        if notebook:
+            delete_notebook(client1, session_env1['environmentUri'], notebook)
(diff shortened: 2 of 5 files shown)


noah-paige:
Need to do final testing end to end - then will open a PR

deploy/stacks/pipeline.py
```diff
@@ -687,6 +687,8 @@ def set_approval_tests_stage(
                                     f'export TESTDATA=$(aws ssm get-parameter --name /dataall/{target_env["envname"]}/testdata --profile buildprofile --output text --query "Parameter.Value")',
                                     f'export ENVNAME={target_env["envname"]}',
                                     f'export AWS_REGION={target_env["region"]}',
+                                    'export INTEGRATION_TESTS=True',
+                                    f'export TOOLING_ACCOUNT={self.account}',
```
petrkalos, line 691: * INTEGRATION_TESTS: we can infer wether they are enabled or not by using the presence of the `TESTDATA`
* TOOLING_ACCOUNT: isn't it better if we use an SSM param to figure it out?

backend/dataall/core/environment/cdk/environment_stack.py
```diff
@@ -138,6 +138,9 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
         self.environment_admins_group: EnvironmentGroup = self.get_environment_admins_group(
             self.engine, self._environment
         )
+        # Create test role for integration tests
+        if os.getenv('INTEGRATION_TESTS', None) == 'True':
```
petrkalos, line 142: You can infer this by checking if the `/{resource_prefix/{env_name}/testdata` SSM param is populated

dlpzx: or of working locally check for the testdata file right?

dlpzx: For the same deployment with autoapproval enabled we will deploy "test environments" as part of the testing and "real environments". The SSM param will exist no matter what no?

dlpzx: Summarizing, we would recommend users to:
- have a dev environment with approval_tests enabled - all environments in the actual deployment and in the codebuild  test session include an integration test role (very permissive role)
- have a prod environment with approval_tests disabled 

My concern is that users enable approval_tests in prod and start deploying a very open IAM role in all environments. I would only deploy the test role when we are doing tests (if possible)

dlpzx: The more I think about it the more I agree with you. Having an environment variable makes it quite complex, because we need to pass the env variable from the testing environment to API to ECS. I am good with having guardrails as you suggested and clear warnings and recommendations

dlpzx: There is a problem with checking in SSM and that is that the SSM parameter is stored in the tooling account right? the ECS task that deploys the environment is in the infra account, so it would need to read the env from the tooling account. We might need to create a different SSM parameter in the infra account

dlpzx: Recapping alternatives:
- Env variables - very complex because we are passing them from env to env 
- SSM - very complex because of the accounts isolation, plus we cannot distinguish which environments are test environments and which ones are prod
- parameter in API call - seems to be the best option, we can differentiate between test and no-test environments and we pass the info directly to the API call

noah-paige:
This PR is ready for review but has been merged to include changes from `feat/integration-tests-notebooks` in order to leverage some of the functionality there with IAM role creation for environments and similar

For review of this PR please focus on tests on notebook primarily - review of any dataset related tests can be a part of the separate PR #1379 and this can be merged AFTER the dataset tests is merged

petrkalos (commented):
lgtm, just some small suggestions

tests_new/integration_tests/modules/notebooks/global_conftest.py
```diff
@@ -0,0 +1,177 @@
...
+            env_uri=session_env1['environmentUri'],
+            tags=[session_id],
+            vpc_id=vpc_id,
+            subnet_id=subnet_id,
+        )
+        yield notebook
+    finally:
+        if notebook:
+            delete_notebook(client1, session_env1['environmentUri'], notebook)
+            stack_name = NamingConventionService(
```
petrkalos, line 85: can't we get the stack_name from `notebook.stack.name`?

tests_new/integration_tests/modules/notebooks/test_notebooks.py
```diff
@@ -0,0 +1,91 @@
...
+    notebook_uri = persistent_notebook1.notebookUri
+    target_type = 'notebook'
+    check_stack_ready(
+        client=client1, env_uri=env_uri, stack_uri=stack_uri, target_uri=notebook_uri, target_type=target_type
+    )
+    update_stack(client1, notebook_uri, target_type)
+    # wait for stack to move to "in_progress" state
+    # TODO: Come up with better way to handle wait in progress if applicable
+    # Use time.sleep() instead of poller b/c of case where no changes founds  (i.e. no update required)
+    # check_stack_in_progress(client1, env_uri, stack_uri)
```
petrkalos, line 85: I had exactly the same issue in envs, I think we need an API for that. All the async tasks must return a taskid and then we must create an api to poll for the status of it.

tests_new/integration_tests/core/environment/queries.py
```diff
@@ -61,6 +61,10 @@ def create_environment(client, name, group, organizationUri, awsAccountId, regio
                 'region': region,
                 'description': 'Created for integration testing',
                 'tags': tags,
+                'parameters': [
+                    {'key': 'notebooksEnabled', 'value': 'true'},
```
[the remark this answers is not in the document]
noah-paige: It will not - but enabling sagemaker studio will and may just want to make that common practice for all tests

tests_new/integration_tests/core/environment/global_conftest.py
```diff
@@ -50,6 +50,24 @@ def session_env1(client1, group1, org1, session_id, testdata):
             delete_env(client1, env)
 
 
+@pytest.fixture(scope='session')
+def session_env1_aws_client(session_env1, session_id):
```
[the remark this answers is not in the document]
noah-paige: To address in #1379 - this change is pulled from there

tests_new/integration_tests/modules/notebooks/global_conftest.py
```diff
@@ -0,0 +1,177 @@
...
+"""
+
+
+@pytest.fixture(scope='session')
+def session_notebook1(client1, group1, session_env1, session_id, session_env1_aws_client):
+    resource_name = 'sessionnotebook1'
+    notebook = None
+    try:
+        vpc_client = VpcClient(session=session_env1_aws_client, region=session_env1['region'])
+        vpc_id = vpc_client.create_vpc(vpc_name=resource_name, cidr='172.31.0.0/26')
```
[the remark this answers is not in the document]
noah-paige: For notebooks in particular it would be fine to use default VPC if just spinning up and down the notebook and making sure it is accessible

I think these helper functions to create and clean up VPC may be useful for SM Studio and/or Network Modules

tests_new/integration_tests/modules/notebooks/global_conftest.py
noah-paige: I wanted to ensure we clean up VPC but need to wait until notebook is deleted 

We likely will have to ensure clean up is orchesteated in some sequential order as we add more modules for testing

backend/dataall/core/environment/cdk/environment_stack.py
dlpzx: Solved and implemented in PR #1379

tests_new/integration_tests/modules/notebooks/global_conftest.py
petrkalos: Clean up order is controlled by the fixture creation order so we probably don't need to do anything explicitly. Did you maybe look into using the `notebook.stack.name`?

deploy/stacks/param_store_stack.py
```diff
@@ -118,6 +119,15 @@ def __init__(
             string_value=str(json.dumps(deploy_config.get_dataall_version())),
             description='Deployed data all version',
         )
+
+        aws_ssm.StringParameter(
+            self,
+            f'toolingAccountParam{envname}',
+            parameter_name=f'/dataall/{envname}/toolingAccount',
+            string_value=str(tooling_account_id),
+            description=f'Stores AWS account if for the tooling account that hosts the code for environment {envname}',
```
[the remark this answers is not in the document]
noah-paige: fixed in latest

tests_new/integration_tests/conftest.py
```diff
@@ -14,6 +14,8 @@
 pytest_plugins = [
     'integration_tests.core.organizations.global_conftest',
     'integration_tests.core.environment.global_conftest',
+    'integration_tests.modules.notebooks.global_conftest',
```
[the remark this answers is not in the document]
noah-paige: good call - changed in latest

tests_new/integration_tests/core/environment/global_conftest.py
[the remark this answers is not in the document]
noah-paige: removed and made get env session a re-usable function

tests_new/integration_tests/modules/notebooks/global_conftest.py
noah-paige: yes i just changed in latest and removed NamingConventionService

tests_new/integration_tests/modules/notebooks/test_notebooks.py
noah-paige: For now I just changed time.sleep(120) similar to dataset tests

Agreed - need a API to get task

noah-paige:
Tested with latest updates and all checks passing

petrkalos (approved):

PR #1386, opened 2024-07-03
Add check in delete environment for create_failed stacks

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
When an environment stack fails in its creation (the CloudFormation stack is not deployed), the cdk-pivot role won't be created. Then, when a user tries to delete the environment they will get an error `Data.all Environment Pivot Role does not have permissions to to get the list of attached policies to the role dataall-testenva-jv54380n: An error occurred (AccessDenied) when calling the AssumeRole operation: User: arn:aws:sts::XXXXXXX:assumed-role/tool-tst-graphql-role/tool-tst-graphql is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::1111111111:role/dataallPivotRole-cdk',`

Those managed permissions do not even exist if the environment stack was never deployed successfully. This PR adds a check that verifies that the environment stack was deployed before removing env-managed policies.

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/core/environment/services/environment_service.py
+++ b/backend/dataall/core/environment/services/environment_service.py
@@ -868,13 +868,19 @@ class EnvironmentService:
                     f'related objects before proceeding',
                 )
             else:
-                PolicyManager(
-                    role_name=environment.EnvironmentDefaultIAMRoleName,
-                    environmentUri=environment.environmentUri,
-                    account=environment.AwsAccountId,
-                    region=environment.region,
-                    resource_prefix=environment.resourcePrefix,
-                ).delete_all_policies()
+                if StackRepository.find_stack_by_target_uri(session, environment.environmentUri) not in [
+                    StackStatus.ROLLBACK_COMPLETE.value,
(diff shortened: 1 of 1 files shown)


backend/dataall/core/environment/services/environment_service.py
```diff
@@ -868,13 +868,19 @@ def delete_environment(uri):
...
                 )
             else:
-                PolicyManager(
-                    role_name=environment.EnvironmentDefaultIAMRoleName,
-                    environmentUri=environment.environmentUri,
-                    account=environment.AwsAccountId,
-                    region=environment.region,
-                    resource_prefix=environment.resourcePrefix,
-                ).delete_all_policies()
+                if StackRepository.find_stack_by_target_uri(session, environment.environmentUri) not in [
```
petrkalos, line 871: In case auto pivotRole creation the pivot role won't exist as well (I've come across this issue a few times) so I think that this check will fail regardless. Wdyt?

petrkalos (approved):

Issue #1376, opened 2024-07-01
  Integration tests executed on a real deployment as part of the CICD - Shares


dlpzx:
Same as for https://github.com/data-dot-all/dataall/issues/1220.

This issue is to track the progress for the Shares modules.
It has its own dedicated issue because of the challenge of validating the shares

dlpzx:
## Required tests for basic coverage

### For fresh deployments
For each of the following API calls we need to test authorized and unauthorized scenarios as well as all possible configurations (e.g. autoapproval...)
- [ ] Create Dataset - includes testing if dataset is indexed in Catalog
- [ ] Import Dataset - includes testing if dataset is indexed in Catalog
- [ ] List Datasets
- [ ] Edit Dataset
- [ ] Delete dataset
- [ ] Start crawler
- [ ] Sync tables
- [ ] Preview table
- [ ] Delete table
- [ ] Create folder
- [ ] Delete folder

### For backwards compatibility
- [ ] Update created Dataset stack
- [ ] Update imported Dataset stack
- [ ] Create Dataset in updated Environment
- [ ] Import Dataset in updated Environment

## Full coverage [WIP]

### For fresh deployments
- [ ] AWS access to Dataset - Credentials 
- [ ] AWS access to Dataset - S3 redirect
- [ ] Upload data
- [ ] Profile table

### For backwards compatibility
For the updated Dataset stacks:
- [ ] AWS access to Dataset - Credentials 
- [ ] AWS access to Dataset - S3 redirect
- [ ] Start crawler
- [ ] Sync tables

SofiaSazonova:
Current progress
![shares](https://github.com/user-attachments/assets/ba5e2f8b-cabf-47c6-8f5f-dce9df27bf95)

SofiaSazonova:
**Happy** path tests:

- [ ] Create Share request
- [ ] Submit request without Items (not allowed)
- [ ] Add items to request
- [ ] Submit request without autoapproval (status -- Submitted)
- [ ] Submit request with autoapproval (status -- Approved)
- [ ] Reject request
- [ ] Approve request
- [ ] Request succeeded
- [ ] Item health verification
- [ ] Folder sharing validation (cal list objects via s3 accesspoint , for same account check write|modify)
- [ ] Bucket sharing validation (cal list objects in s3 bucket , for same account check write|modify)
- [ ] Table sharing validation (cal perform athena query: select * from db.table , for same account check write|modify)
- [ ] Revoke items
[...]

PR #1389, opened 2024-07-05
Add Dataset integration tests - S3 Share requests

SofiaSazonova:
### Feature or Bugfix
- Tests

### Detail
- module `share_base`
- bugfix `delete_env` requires `env_object` not `envUri`
- TEMPORARY: hardcoded dataset_uri --> I wait for dataset module

### Relates
- #1376 

### Security

  - Is the input sanitized?


tests_new/integration_tests/modules/share_base/fresh_deployment.py
dlpzx, line 1: The naming does not match any of the other tests, we could call them something closer to the other modules `test_shares` and `test_shares_backwards_compatibility`

tests_new/integration_tests/modules/share_base/fresh_deployment.py
```diff
@@ -0,0 +1,106 @@
...
+    get_share_object,
+    reject_share_object,
+    approve_share_object,
+    revoke_share_items,
+)
+from tests_new.integration_tests.modules.share_base.utils import check_share_ready
+
+
+def test_create_share_object(share1):
+    assert_that(share1.status).is_equal_to(ShareObjectStatus.Draft.value)
```
dlpzx, line 18: We could also add assertions on the principal of the share and a new test for create_share_object for consumption roles that makes sure that the groupUri and the principal are correct

tests_new/integration_tests/modules/share_base/fresh_deployment.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def test_submit_empty_object(client2, share1):
+    # here Exception is not recognized as GqlError, so we use base class
+    # toDo: back to GqlError
+    assert_that(submit_share_object).raises(Exception).when_called_with(client2, share1.shareUri).contains(
+        'ShareItemsFound', 'The request is empty'
+    )
+
+
```
dlpzx, line 28: Once the dataset PRs are merged we should add tests on getShareObject with the filters used to list shared items. Making sure it lists the tables, folders and bucket defining them explicitly instead of depending on a list of items

tests_new/integration_tests/modules/share_base/fresh_deployment.py
```diff
@@ -0,0 +1,106 @@
...
+    item_to_add = items[0]
+    share_item_uri = add_share_item(client2, share1.shareUri, item_to_add.itemUri, item_to_add.itemType)
+    assert_that(share_item_uri).is_not_none()
+
+    updated_share = get_share_object(client2, share1.shareUri, {'isShared': True})
+    items = updated_share['items'].nodes
+    assert_that(items).is_length(1)
+    assert_that(items[0].shareItemUri).is_equal_to(share_item_uri)
+    assert_that(items[0].status).is_equal_to(ShareItemStatus.PendingApproval.value)
+
```
dlpzx, line 43: My main concern is that we cannot test `test_reject_share` without running `test_submit_object_no_auto_approval` before. The tests are coupled, which in this case it might be alright, we are testing test_share_workflow succeed in parts. I would just clearly indicate that as a comment and group them all together (all tests on share1 together and all tests on share3 together to show the path). @petrkalos wdyt?

tests_new/integration_tests/modules/share_base/fresh_deployment.py
```diff
@@ -0,0 +1,106 @@
...
+    assert_that(updated_share.status).is_equal_to(ShareObjectStatus.Approved.value)
+
+
+def test_share_succeeded(client1, share3):
+    check_share_ready(client1, share3.shareUri)
+    updated_share = get_share_object(client1, share3.shareUri, {'isShared': True})
+    items = updated_share['items'].nodes
+
+    assert_that(updated_share.status).is_equal_to(ShareObjectStatus.Processed.value)
+    for item in items:
```
dlpzx, line 81: Concern: how do we ensure there are items and this is not an empty list? As a next step after datasets PR, for the validation we would 1) s3 get object to the bucket and access point shares 2) run an athena query on the glue tables

tests_new/integration_tests/modules/share_base/input_types.py
dlpzx, line 1: Not sure having input_types is needed, we need a good reason to add more code

dlpzx:
It would be good to have a testing plan on the different scenarios that we want to cover:
- Happy path approved
    - without Auto-approval
    - with Auto-approval
- Happy path revoked
- Unauthorized cases
- Error handling
For the different types of shareable objects:
- S3 Bucket
- S3 Prefix (folder)
- Glue table
For the different types of principals:
- group
- consumption role
For the different types of dataset:
- created
- imported with SSE-S3 encyrption
- imported with KMS encryption
- imported and Glue database imported

tests_new/integration_tests/conftest.py
```diff
@@ -14,6 +14,7 @@
 pytest_plugins = [
     'integration_tests.core.organizations.global_conftest',
     'integration_tests.core.environment.global_conftest',
+    'integration_tests.modules.share_base.global_conftest',
```
dlpzx, line 17: global conftests are meant to be used by the other modules. I am not sure we will want all the configs that we are creating in the other modules. Maybe we can have 2 conftest, one `global_conftest` (exportable to other modules) and a plain `conftest` (used internally in the shares tests). This is important because the share1, share3 states are very unpredictable

dlpzx:
Issue to track PR: https://github.com/data-dot-all/dataall/issues/1409

tests_new/integration_tests/README.md
```diff
@@ -12,59 +12,73 @@ Currently **we support only Cognito based deployments** but support for any IdP
...
+                     "username": "testUser4",
+                     "password": "...",
+                     "groups": [
+                       "testGroup4"
+                     ]
+                   }
+                 },
+                 "aws_profiles": {
+                     "second": "second_int_test_profile"
+                 },
```
petrkalos, line 61: I think the use of `aws_profiles` will not play very well with CodeBuild, instead I propose to use the existing infrastracture  (see `session_env1_aws_client`).

As discussed offline you need this account to test consumption roles. By using the integration test account you can solve it using the two following patterns
1. (simpler) add the already created (during env deployment) integration-test role directly as a consumption role. Current (CodeBuild/Local) account have already permissions to assume this role so you can use STS to assume it and then run S3 queries to make sure that the share was succesful.
2. (more complex) use the integration-test role to create new roles in the target account that you will register as consumption roles. Then proceed with assuming those roles and testing for S3 access.

petrkalos (changes requested):

tests_new/integration_tests/modules/share_base/test_new_crossacc_s3_share.py
```diff
@@ -0,0 +1,393 @@
...
+)
+
+
+@pytest.mark.parametrize(
+    'principal_type',
+    ['Group', 'ConsumptionRole'],
+)
+def test_create_and_delete_share_object(
+    client5, persistent_cross_acc_env_1, session_s3_dataset1, consumption_role_1, group5, principal_type
+):
```
petrkalos, line 41: Instead of parametrizing the tests you can [parametrize a fixture](https://docs.pytest.org/en/latest/how-to/fixtures.html#parametrizing-fixtures) and then all tests that are using this fixture will run as many times as the parameters of the fixture.

For example in this case you can do something like...

```
@pytest.fixture(params=["Group", "ConsumptionRole"])
def principal1(request, group5, consumption_role_1):
    """
    :return: tuple with (principalUri, principalType)
    """
    if request.param is 'Group':
        yield (group5, request.param)
    else:
        yield (consumption_role_1.consumptionRoleUri, request.param)
```

tests_new/integration_tests/modules/share_base/test_new_crossacc_s3_share.py
```diff
@@ -0,0 +1,393 @@
...
+    updated_share = get_share_object(client5, share.shareUri)
+    assert_that(updated_share.rejectPurpose).is_equal_to('new purpose')
+
+    clean_up_share(client5, share.shareUri)
+
+
+@pytest.mark.parametrize(
+    'share_fixture_name',
+    ['session_share_1', 'session_share_consrole_1'],
+)
```
petrkalos, line 163: Similar to my previous comment do you think it's possible to use parametrized fixtures here as well to avoid the `getfixturevalue`?

SofiaSazonova: implemented

tests_new/integration_tests/modules/share_base/test_new_crossacc_s3_share.py
SofiaSazonova: implemented

tests_new/integration_tests/README.md
SofiaSazonova: I implemeted via AssumeRole

tests_new/integration_tests/aws_clients/iam.py
```diff
@@ -24,7 +24,16 @@ def get_role(self, role_name):
             log.info(f'Error occurred: {e}')
             return None
 
-    def create_role(self, account_id, role_name):
+    @staticmethod
+    def get_tooling_account_id():
+        session = boto3.Session()
+        param_client = session.client('ssm', os.environ.get('AWS_REGION', 'us-east-1'))
+        parameter_path = f"/dataall/{os.environ.get('ENVNAME', 'dev')}/toolingAccount"
+        print(parameter_path)
```
petrkalos, line 32: nit: logging

tests_new/integration_tests/aws_clients/iam.py
```diff
@@ -34,7 +43,9 @@ def create_role(self, account_id, role_name):
                 {{
                     "Effect": "Allow",
                     "Principal": {{
-                        "AWS": "arn:aws:iam::{account_id}:root"
+                        "AWS": ["arn:aws:iam::{account_id}:root",
+                        "arn:aws:iam::{IAMClient.get_tooling_account_id()}:root",
+                        "arn:aws:sts::{account_id}:assumed-role/{test_role_name}/{test_role_name}"]
```
petrkalos, line 48: I am no IAM expert but do we need this? I think the first principal (line 46) will allow all roles from `account_id` to assume this role. Check [this](https://niebardzo.github.io/2021-08-23-root-principal-in-aws/#:~:text=IAM%20role%20assigned.-,Root%20Principal,-Who%20is%20the)

tests_new/integration_tests/aws_clients/sts.py
```diff
@@ -2,12 +2,9 @@
...
 class StsClient:
-    def __init__(self, session, profile, region):
-        if session is None:
-            if profile is None:
-                session = boto3.Session()
-            else:
-                session = boto3.Session(profile_name=profile)
+    def __init__(self, session, region):
+        if not session:
+            session = boto3.Session()
```
petrkalos, line 7: nit: shorthand...
```python
    def __init__(self, region, session = boto3.Session()):
       ...
```

tests_new/integration_tests/modules/share_base/test_new_crossacc_s3_share.py
petrkalos, line 1: this looks awesome :+1:

tests_new/integration_tests/aws_clients/iam.py
```diff
@@ -34,7 +43,9 @@ def create_role(self, account_id, role_name):
                 {{
```
petrkalos, line 43: nit: I'd make this a dict and then do a `json.dumps`

petrkalos (changes requested):

tests_new/integration_tests/README.md
```diff
@@ -12,59 +12,69 @@ Currently **we support only Cognito based deployments** but support for any IdP
...
+              },
+              "session_env2": {
+                "accountId": "...",
+                "region": "eu-west-1"
+              },
+               "persistent_env1": {
+                       "accountId": "...",
+                       "region": "us-east-1"
+               },
+              "persistent_cross_acc_env_1": {
```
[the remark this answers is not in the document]
SofiaSazonova: 1. Speed up. 
2. Later we will need persistent shares as well

tests_new/integration_tests/aws_clients/iam.py
SofiaSazonova: No, it doesn't work. Assumed role is processed differently. I tried without it and got Access Denied, so I had to explicitly add this.

tests_new/integration_tests/README.md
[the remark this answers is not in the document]
petrkalos: I think @dlpzx made a very good point in another PR about using persistent env/shares.
If we use them without forcing an update (and wait for it to complete) then we might not be testing the latest changes but if we do that then we might as well create a new env every time.

I think we should be able to use persistent envs with the argument of speed only for not very significant features AND obviously to test backwards compatibility (but for that we should still force update and wait).

tests_new/integration_tests/aws_clients/iam.py
SofiaSazonova: done

tests_new/integration_tests/aws_clients/sts.py
SofiaSazonova: done

tests_new/integration_tests/aws_clients/iam.py
SofiaSazonova: removed

petrkalos (approved):

PR #1406, opened 2024-07-11
Introduce check for actions in share_verify bucket and access points

dlpzx:
### Feature or Bugfix
-Feature/ Bugfix

### Detail
When sharing access points or buckets, there is an IAM policy attached to the requester role that gets updated with every share request. This policy used to grant`s3:*` permissions, in https://github.com/data-dot-all/dataall/pull/1280/files we scoped down the S3 actions allowed in the IAM policy to `['s3:List*', 's3:Describe*', 's3:GetObject']`. 

The issue is that previous share requests will still contain the too permissive `s3:*`, which is incorrect. For this reason and for any additional action granted outside of data.all; in this PR we:
- add a verify step for s3 access points and buckets that checks the s3: actions granted
- if there are actions that do not belong to `['s3:List*', 's3:Describe*', 's3:GetObject']` the share verify tasks will report the error.

⚠️ TODO: fix and add tests

### Relates
- #1280
- #1374 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/s3_datasets_shares/services/s3_share_managed_policy_service.py
+++ b/backend/dataall/modules/s3_datasets_shares/services/s3_share_managed_policy_service.py
@@ -1,4 +1,5 @@
 import json
+from typing import List
 from dataall.base.aws.iam import IAM
 from dataall.base.utils.naming_convention import NamingConventionService, NamingConventionPattern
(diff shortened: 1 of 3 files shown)


noah-paige:
let's close in favor of #1407 ?

PR #1412, opened 2024-07-12
Fix/remove edit team modals

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Originally - we were opening up a modal per TeamRow for EnvironmentTeam and OrganizationTeam Tables whenever we wanted to edit or remove a singular team
  - For delete team in specific there was a bug when 1 team was removed another modal remained asking to remove a separate team from the environment (not the one user clicked on)

- This PR ensures that the Edit Team Modal or Remove Team Modal only opens for the particular team clicked on and not for any other for both EnvTeams and OrgTeams (prevents the above from happening)
  - This design is in line with how modal showing is handled with consumption Roles for Envs as well

### Relates
N/A

### Security

  - Is the input sanitized?

### DIFF
--- a/frontend/src/design/components/DeleteObjectWithFrictionModal.js
+++ b/frontend/src/design/components/DeleteObjectWithFrictionModal.js
@@ -95,6 +95,7 @@ export const DeleteObjectWithFrictionModal = (props) => {
               variant="contained"
               onClick={() => {
                 deleteFunction(deleteFromAWS);
+                setConfirmValue(null);
               }}
             >
               Delete
--- a/frontend/src/modules/Organizations/components/OrganizationTeams.js
+++ b/frontend/src/modules/Organizations/components/OrganizationTeams.js
@@ -45,29 +45,22 @@ import {
   OrganizationTeamInviteForm
 } from '../components';
 
-function TeamRow({ team, permissions, organization, fetchItems }) {
+function TeamRow({
+  team,
+  permissions,
+  organization,
+  fetchItems,
+  handleDeleteGroupModalOpen,
+  handleDeleteGroupModalClosed,
+  handlePermissionsModalOpen,
+  handlePermissionsModalClose,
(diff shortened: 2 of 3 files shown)


dlpzx:
Tested with 4 groups invited to an Environment and it works fine!

frontend/src/design/components/DeleteObjectWithFrictionModal.js
```diff
@@ -95,6 +95,7 @@ export const DeleteObjectWithFrictionModal = (props) => {
               variant="contained"
               onClick={() => {
                 deleteFunction(deleteFromAWS);
+                setConfirmValue(null);
```
dlpzx, line 98: I am not sure why this is needed; so it seems to be preventing that the next time the delete confirmation modal opens the confirm value is the target, but doesn't it get the null value from `const [confirmValue, setConfirmValue] = useState(null);` when it gets re-rendered?

frontend/src/modules/Organizations/components/OrganizationTeams.js
```diff
@@ -155,7 +155,7 @@ function TeamRow({ team, permissions, organization, fetchItems }) {
               objectName={team.groupUri}
               onApply={() => handleDeleteGroupModalClosed()}
               onClose={() => handleDeleteGroupModalClosed()}
-              open={isDeleteGroupModalOpen}
+              open={isDeleteGroupModalOpenId === team.groupUri}
```
dlpzx, line 158: very ingenious

dlpzx (approved):

frontend/src/design/components/DeleteObjectWithFrictionModal.js
noah-paige: I imagine the modal hides but is not re-rendered if you removed 2 groups in succession - so confirmValue never reset anywhere previously

PR #1424, opened 2024-07-22
Add Redshift datasets module

dlpzx:
### Feature or Bugfix
- Feature

### Detail
The design is up-to-date in the issue https://github.com/data-dot-all/dataall/issues/955 as well as next steps

### Relates
- #955 

### Security

  - Is the input sanitized?


dlpzx:
Redshift connection unit tests are running :) 
image

dlpzx:
## Testing locally
- Creation of invalid connections - wrong cluster id
- Creation of invalid connections - wrong namespace
- Creation of invalid connections - wrong workgroup
- Creation of invalid connections - wrong database
- Creation of invalid connections - database that cannot be accessed by this connection
- Creation of cluster connection with secret arn
- Creation of cluster connection with db user
- Creation of serverless connection with secret arn
- List all connections
- Delete a connection
- Import dataset with no tables (for each type of connection)
- Import dataset with some tables (for each type of connection)
- Add tables to a dataset (for each type)
- Delete a table
- Overview tab of dataset view
- Tables tab of dataset view
- Overview tab of tables view
- Columns tab of tables view
- Delete Dataset
- List Datasets
- List items in the Catalog: share request disabled
- Complimentary functions: glossary, terms, votes, feed

dlpzx:
How to test the PR @noah-paige @petrkalos 
1. Create cluster provisioned and/or serverless in a data.all environment account
2. Go to the Query editor
3. Create tables (check this [workshop](https://catalog.us-east-1.prod.workshops.aws/workshops/9f29cdba-66c0-445e-8cbb-28a092cb5ba7/en-US/lab1)  step 2
4. Create user, you can use the command: `CREATE USER "username" PASSWORD 'password';
5. `grant select on all tables in schema public to username;`
6. Store the user creds in a Secret in secrets manager - check this [ instructions ](https://docs.aws.amazon.com/redshift/latest/mgmt/redshift-secrets-manager-integration.html)- on top of the redshift tag, the secret needs to be tagged with `dataall` = `True`

dlpzx:
I have reviewed the design and there is one opportunity to simplify the user experience for the creation of Redshift Connections. @noah-paige @petrkalos 

### Current implementation
For Provisioned Redshift - input =  cluster-id
For Serverless cluster - input =  workgroup + namespace-id (we verify that the workgroup belongs to the namespace)

I verified and workgroup names must be unique in the account/region

### Target implementation
For Provisioned Redshift - input =  cluster-id
For Serverless cluster - input =  workgroup

For BOTH, data.all backend gets the namespace-id and/or arn and stores them in RDS. They are used later for datashares and whenever we connect to the clusters

backend/migrations/versions/852cdf6cf1e0_add_redshift_datasets.py
noah-paige, line 1: Are we handling backfilling of Redshift permissions for existing EnvGroups?

For `LIST_ENVIRONMENT_REDSHIFT_CONNECTIONS` think it needs to be added to all EnvGroups because it is default permission


For `CREATE_REDSHIFT_CONNECTION` and `IMPORT_REDSHIFT_DATASET`
-  For invited teams I think we can require Env Admins to edit the permissions of the invited groups to include those permissions after upgrade
- For Env Admins we likely want to backfill since they should have ALL permissions and is not editable

noah-paige:
I went through the entire create, update, delete and all of the associated tests above for Redshift Serverless (will do the same for provisioned cluster but I imagine very similar experience)

All in all the UI looks great and I think the user experience is as streamlined as we can make it - icons on datasets and the schema views are all amazing touches!

Some Minor UI Enhancements from my pov:

- Nit: boxes on create/import dataset are not equal size
Screenshot 2024-07-29 at 2 06 35 PM

- Nit: Columns Tab of Tables has a small height + no circular progress or bar to show it is loading columns/schema (if there is a away to lengthen the table or minimize the height of the card behind it)
Screenshot 2024-07-29 at 2 08 05 PM


- Nit: Status on Redshift Dataset is Blank but shows strangely on UI (can we remove - or populate with some default value since no IaC for redshift datasets? like "In Service" or similar?)

Screenshot 2024-07-29 at 2 07 20 PM

backend/migrations/versions/852cdf6cf1e0_add_redshift_datasets.py
dlpzx: You are right, good catch

dlpzx:
Hi @noah-paige; thanks for pushing the UI enhancements. Here are some screenshots of the final views:
- Alligned the cards and the buttons at the bottom in the create dataset selector window
image

- Removed the status completely from redshift datasets
image


- Columns Tab of Tables - added padding to the cells. We still use getRowHeight auto, that adapts the height to the content.
image

backend/migrations/versions/852cdf6cf1e0_add_redshift_datasets.py
dlpzx: Any of the options is a pretty big change, either in terms of code or in terms of design/feature design change. So I will go ahead and implement the migration for this case and we can discuss if this feature should be considered for 2.8/2.7

backend/migrations/versions/852cdf6cf1e0_add_redshift_datasets.py
```diff
@@ -0,0 +1,130 @@
...
+        sa.Column('deleted', sa.DateTime(), nullable=True),
+        sa.Column('description', sa.String(), nullable=True),
+        sa.Column('tags', postgresql.ARRAY(sa.String()), nullable=True),
+        sa.Column('datasetUri', sa.String(), nullable=False),
+        sa.Column('rsTableUri', sa.String(), nullable=False),
+        sa.Column('topics', postgresql.ARRAY(sa.String()), nullable=True),
+        sa.ForeignKeyConstraint(['datasetUri'], ['redshift_dataset.datasetUri'], ondelete='CASCADE'),
+        sa.PrimaryKeyConstraint('rsTableUri'),
+    )
+    op.execute("ALTER TYPE datasettypes ADD VALUE 'Redshift'")
```
noah-paige, line 92: did we look into if this fails if datasettype of 'Redshift' already exists?

dlpzx: I just checked and it fails

backend/dataall/modules/redshift_datasets/aws/redshift_serverless.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class RedshiftServerless:
+    def __init__(self, account_id: str, region: str, role=None) -> None:
+        session = SessionHelper.remote_session(accountid=account_id, region=region, role=role)
+        self.client = session.client(service_name='redshift-serverless', region_name=region)
+
+    def get_namespace_by_id(self, namespace_id: str):
+        response = self.client.list_namespaces()
+        namespaces = response['namespaces'] if 'namespaces' in response.keys() else []
```
petrkalos, line 18: nit:

```suggestion
        namespaces = response.get('namespaces', [])
```

backend/dataall/modules/redshift_datasets/db/redshift_dataset_repositories.py
```diff
@@ -0,0 +1,140 @@
...
+            topics=data.get('topics', []),
+            businessOwnerEmail=data.get('businessOwnerEmail', ''),
+            businessOwnerDelegationEmails=data.get('businessOwnerDelegationEmails', []),
+            stewards=data.get('stewards') if data.get('stewards') else data['SamlAdminGroupName'],
+            autoApprovalEnabled=data.get('autoApprovalEnabled', False),
+            connectionUri=data.get('connectionUri'),
+            schema=data.get('schema'),
+        )
+        session.add(dataset)
+        session.commit()
```
petrkalos, line 54: shouldn't we commit after the last add to make sure both succeed or get reverted?

frontend/src/modules/DatasetsBase/components/DatasetCreateWindow.js
```diff
@@ -17,27 +17,26 @@ import { isModuleEnabled, ModuleNames } from 'utils';
 
 export const DatasetCreateWindow = (props) => {
   const { onClose, open, ...other } = props;
-  const number_grid_items = isModuleEnabled(ModuleNames.S3_DATASETS) ? 2 : 0;
+  const number_grid_items =
```
petrkalos, line 20: not for this PR: 
Following the pattern set by @SofiaSazonova at #1435 I think we should start thinking how to detangle UI from the `config.json`. Here we could have a query that returns all the enabled modules.

frontend/src/modules/Environments/components/EnvironmentRedshiftConnectionAddForm.js
```diff
@@ -0,0 +1,365 @@
...
+
+export const EnvironmentRedshiftConnectionAddForm = (props) => {
+  const { environment, onClose, open, reload, ...other } = props;
+  const { enqueueSnackbar } = useSnackbar();
+  const dispatch = useDispatch();
+  const client = useClient();
+
+  let { groupOptions, loadingGroups } = useFetchGroups(environment);
+
+  const clusterOptions = [
```
petrkalos, line 32: maybe use the new query introduced in #1435?

tests/modules/redshift_datasets/conftest.py
```diff
@@ -0,0 +1,201 @@
...
+                'workgroupArn': 'arn:aws:redshift-serverless:eu-west-1:XXXXXXXXXXXXXX:workgroup/workgroup_name_1',
+            }
+        ]
+
+    def get_workgroup_arn(self, *args, **kwargs):
+        return 'arn:aws:redshift-serverless:eu-west-1:XXXXXXXXXXXXXX:workgroup/workgroup_name_1'
+
+
+@pytest.fixture(scope='function')
+def patch_redshift(module_mocker):
```
petrkalos, line 58: I think I disagree with creating Mock clients like that and I think there are practical dangers.
For example, let's say that you change your code and remove the `get_namespace_by_id` but you forget to remove the invocation from one line. Then you run the tests and they pass, you happily commit that to prod and then things start breaking because the function is missing in the real client.

I think the way to go here is...

```
fooClassMock = module_mocker.patch('path.to.foo.FooClass', autospec=True) # this returns a MagicMock bound to the FooClass 
fooClassMock.methoda.return_value = '123'
```

with `autospec=True` you make sure that calls are verified against the real class so if your code tries to access something that doesn't exist it will blow up.
Also this will allow you to assert that the expected methods were called with the correct arguments (if you have to validate that in any of your tests).

petrkalos, line 58: you create this fixture per function but you use `module_mocker` instead of `mocker`, does that mean that the objects will remain patched for the entire module? is that what you want to achieve?

tests/modules/redshift_datasets/test_unit_redshift_connection_service.py
```diff
@@ -0,0 +1,249 @@
...
+            'redshiftUser': None,
+            'secretArn': 'arn:aws:secretsmanager:*:111111111111:secret:secret-2',
+        },
+    ).contains('Redshift namespaceId not-existent-id does not exist')
+
+
+def test_create_redshift_connection_workgroup_not_in_namespace(env_fixture, api_context_1, group, module_mocker):
+    # Given a namespace but a workgroup that does not belong to it
+    mock_serverless = MagicMock()
+    module_mocker.patch(
```
petrkalos, line 51: what's the lifetime of this patch? I suspect (i haven't used mocker fixture before) that because you use `module_mocker` instead of `mocker` this will not be undone after the test is done resulting in cross-condamination

petrkalos (changes requested):
apart from some nits the prod code looks good, I am not approving yet because

1. dbmigrations are failing
2. there is an issue with mocking in the tests

backend/migrations/versions/852cdf6cf1e0_add_redshift_datasets.py
dlpzx: Added a check to pass the exception if the enum value already exists. That was tested in a local db

backend/dataall/modules/redshift_datasets/api/datasets/input_types.py
```diff
@@ -0,0 +1,46 @@
...
+    name='ImportRedshiftDatasetInput',
+    arguments=[
+        gql.Argument('label', gql.NonNullableType(gql.String)),
+        gql.Argument('organizationUri', gql.NonNullableType(gql.String)),
+        gql.Argument('environmentUri', gql.NonNullableType(gql.String)),
+        gql.Argument('description', gql.String),
+        gql.Argument('tags', gql.ArrayType(gql.String)),
+        gql.Argument('owner', gql.NonNullableType(gql.String)),
+        gql.Argument('topics', gql.ArrayType(gql.Ref('Topic'))),
+        gql.Argument(name='SamlAdminGroupName', type=gql.NonNullableType(gql.String)),
```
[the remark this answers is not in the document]
dlpzx: For the redshift module all arguments are non-keyword args now

backend/dataall/modules/redshift_datasets/__init__.py
```diff
@@ -0,0 +1,88 @@
+"""Contains the code related to datasets"""
+
+import logging
+from typing import List, Type, Set
+
+from dataall.base.loader import ModuleInterface, ImportMode
+
+log = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
dlpzx: To be honest I took another init file as reference, including the logging. When working locally or for example in the tests I can see the module loading logs, the typical `API of Redshift datasets has been imported`  which are INFO logs

backend/dataall/modules/redshift_datasets/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+        from dataall.modules.redshift_datasets.db.redshift_dataset_repositories import (
+            RedshiftDatasetEnvironmentResource,
+        )
+        from dataall.modules.redshift_datasets.db.redshift_connection_repositories import (
+            RedshiftConnectionEnvironmentResource,
+        )
+        from dataall.modules.redshift_datasets.db.redshift_models import RedshiftDataset, RedshiftTable
+        import dataall.modules.redshift_datasets.api
+
+        FeedRegistry.register(FeedDefinition('RedshiftDatasetTable', RedshiftTable))
```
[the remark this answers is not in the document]
dlpzx: I added a new file with constants. This is specially useful for the frontend development. As we do with the enums, reading constants would be nice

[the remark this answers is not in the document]
dlpzx: or transforming this constants into enums

frontend/src/modules/Environments/components/EnvironmentRedshiftConnectionAddForm.js
dlpzx: Added in # todo list in https://github.com/data-dot-all/dataall/issues/955

frontend/src/modules/DatasetsBase/components/DatasetCreateWindow.js
dlpzx: Added in todo list in https://github.com/data-dot-all/dataall/issues/955

backend/dataall/modules/redshift_datasets/aws/redshift_data.py
```diff
@@ -0,0 +1,150 @@
...
+                    if 'ColumnList' in response.keys():
+                        columns_list.extend(response['ColumnList'])
+                    next_token = response.get('NextToken', None)
+            for col in columns_list:
+                col['nullable'] = True if col['nullable'] == 1 else False
+            log.info(f'Returning {columns_list=}')
+            return columns_list
+        except ClientError as e:
+            log.error(e)
+            raise e
```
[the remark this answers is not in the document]
dlpzx: You are right, should we just remove all try/excepts that are just logging and raising the error?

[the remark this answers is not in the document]
petrkalos: yeah I think so, let the exception bubble-up and simplify the code :)

tests/modules/redshift_datasets/conftest.py
dlpzx: [[docs](https://pytest-mock.readthedocs.io/en/latest/usage.html)], It is also possible to use mocking functionality from fixtures of other scopes using the appropriate fixture (module_mocker, class_mocker...) What leads me to think that we can use module_mocker from the function scope, but as you pointed out it can be already pathing something that we are not aware is being patched

backend/dataall/modules/redshift_datasets/__init__.py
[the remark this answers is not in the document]
petrkalos: if you saw the logs then should be good

backend/dataall/modules/redshift_datasets/db/redshift_dataset_repositories.py
dlpzx: I implemented the change but the tests started failing. The issue is that if the object dataset is not commited the unique identifier is not created and then the activity creation gets a null value in `targetUri=dataset.datasetUri`

tests/modules/redshift_datasets/test_unit_redshift_connection_service.py
dlpzx: I have changed the mocks to be function-specific. And also for most fixtures for connections and datasets

tests/modules/redshift_datasets/conftest.py
```diff
@@ -54,25 +57,34 @@ def get_workgroup_arn(self, *args, **kwargs):
...
-        'dataall.modules.redshift_datasets.services.redshift_connection_service.redshift_client',
+        'dataall.base.aws.sts.SessionHelper.remote_session',
+        return_value=boto3.Session(),
+    )
+
+
+@pytest.fixture(scope='function')
+def patch_redshift(mocker):
+    # autospec=True ensures methods called in the MockClient correspond to real client methods
+    mocker.patch.object(
```
petrkalos, line 71: I don't think you need the Mock classes at all, take a look at the following example...

```python
from unittest.mock import NonCallableMagicMock

import pytest


class ARealClass:
    def get_str(self):
        return 'real_str'

    @staticmethod
    def static_func():
        return 'static_str'


def test_patch(mocker):
    mock_str = '123'
    mock_class = mocker.patch('scratch.ARealClass', autospec=True)
    mock_class.static_func.return_value = mock_str
    mock_class().get_str.return_value = mock_str

    # check that static function was mocked
    assert ARealClass.static_func() == mock_str
    # check that creating ARealClass() will now created a MagicMock
    cls_instance = ARealClass()
    assert isinstance(cls_instance, NonCallableMagicMock)
    # check that get_str return the mock_str and not the real_str
    assert cls_instance.get_str() == mock_str
    # check that autospec is working and calling arbitrary methods throws
    with pytest.raises(AttributeError):
        ARealClass().non_existend_method()
```

petrkalos (approved):

Issue #1324, opened 2024-06-08
Security configuration for Glue crawler


mourya-33:
### Describe the bug

Glue crawlers do not have any security configuration attached which is flagged by checkov as FAILURE with the error message below.

CheckID		: CKV_AWS_195
CheckName	: Ensure Glue component has a security configuration associated
File		: /dataall-dataset-bphz8d77.yaml:636-672
Resource	: AWS::Glue::Crawler.dataalld1bphz8d77crawler
Guideline	:<a href= "https://docs.prismacloud.io/en/enterprise-edition/policy-reference/aws-policies/aws-general-policies/ensure-aws-glue-component-is-associated-with-a-security-configuration"> CKV_AWS_195 </a>

### How to Reproduce

Run a checkov scan on the dataset template. The scan report will contain the entry mentioned above in the description.

### Expected behavior

Glue crawlers should have a security configuration.

### OS

Mac

### Python version

3.10

### AWS data.all version

2.5

dlpzx:
Hi @mourya-33 , thanks for the issue. Security enhancements are always welcome. Do you plan to work on this issue?

noah-paige:
Closing this issue as complete

PR #1434, opened 2024-07-26
Updated glue crawler security config

mourya-33:
### Feature or Bugfix
- Bugfix

### Detail
This PR is to add glue security configuration to glue crawlers and profiling jobs for passing checkov scans

### Relates
[- <URL or Ticket>](https://github.com/data-dot-all/dataall/issues/1324)

### Security

  - Is the input sanitized? N/A
- Are you using or adding any cryptographic features? Adding KMS key to encrypt crawler logs and Job bookmarks
  - Have you used the least-privilege principle? How? Restricted the permissions to only the required resources


mourya-33:
- Tested dataset creation, import and deletion
- Crawler runs

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
```diff
@@ -437,20 +449,68 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
             },
         )
 
+        glue_sec_conf_enc_key = kms.Key(
+            self,
+            f'crwlr_log_enc_key_{dataset.GlueCrawlerName}',
+            removal_policy=RemovalPolicy.DESTROY,
+            alias=f'crwlr_log_enc_key_{dataset.GlueCrawlerName}',
```
noah-paige, line 456: can we have both `alias` and `kms.Key` construct id be  `{dataset.GlueCrawlerName}_log_enc_key`

that way the naming convention is prefixed with `resource_prefix`

nit: we can also set it once as variable and pass that as both construct id and alias name

noah-paige, line 456: - can we have both `alias` and `kms.Key` construct id be  `{dataset.GlueCrawlerName}-log-enc-key` where dataset_basename is: 
```
        dataset_basename = NamingConventionService(
            target_uri=dataset.datasetUri,
            target_label=dataset.label,
            pattern=NamingConventionPattern.GLUE_ETL,
            resource_prefix=env.resourcePrefix,
        ).build_compliant_name()
```


    - that way the naming convention is prefixed with `resource_prefix`
    - also no use of `crawler` in the name since we are applying to both crawler and glue job


- nit: we can also set it once as variable and pass that as both construct id and alias name

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
```diff
@@ -437,20 +449,68 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
+                job_bookmarks_encryption=glue.CfnSecurityConfiguration.JobBookmarksEncryptionProperty(
+                    job_bookmarks_encryption_mode='CSE-KMS', kms_key_arn=glue_sec_conf_enc_key.key_arn
+                ),
+                s3_encryptions=[
+                    glue.CfnSecurityConfiguration.S3EncryptionProperty(
+                        s3_encryption_mode='SSE-KMS', kms_key_arn=glue_sec_conf_enc_key.key_arn
+                    )
+                ],
+            ),
+            name=f'crwlr_sec_config-{dataset.GlueCrawlerName}',
```
noah-paige, line 493: - similar to above - can we have both `name` and `glue.CfnSecurityConfiguration` construct id be `{dataset_basename}-security-config` (using basename from above)

   - that way the naming convention is prefixed with `resource_prefix`

- nit: we can also set it once as variable and pass that as both construct id and alias name

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
```diff
@@ -437,20 +449,68 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
             schedule={'scheduleExpression': f'{dataset.GlueCrawlerSchedule}'} if dataset.GlueCrawlerSchedule else None,
             role=dataset_admin_role.role_arn,
             targets=CfnCrawler.TargetsProperty(
                 s3_targets=[CfnCrawler.S3TargetProperty(path=f's3://{dataset.S3BucketName}')]
             ),
         )
         crawler.node.add_dependency(dataset_bucket)
+        crawler.node.add_dependency(dataset_admin_policy)
+        crawler.node.add_dependency(dataset_admin_role)
+        crawler.node.add_dependency(glue_crawler_security_config)
```
noah-paige, line 513: qq - do we need to add these `add_dependency(...)` statements? I would imagine there would be an implicit dependency since we are passing `crawler_security_configuration=glue_crawler_security_config.name,` as a parameter for the `CfnCrawler` but could be wrong

noah-paige (commented):
left some comments primarily around naming - to test this PR as well

noah-paige:
tested the following and all looks good!

- [x] Dataset Stack Updates (create KMS Key, Policy Change, + Security Config)
- [x] Crawler Run Successful
- [x] Glue Job Run Successful

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
```diff
@@ -437,20 +449,68 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
             },
         )
 
+        glue_sec_conf_enc_key = kms.Key(
```
noah-paige, line 452: I want to call out that this key comes with key policy which includes the block of code that delegates permissiosn to IAM, such as 

```
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::ACCOUNT_ID:root"
            },
            "Action": "kms:*",
            "Resource": "*"
        },
```

meaning along with the principals who have access in the key policy - a role/user with kms permissions via IAM policy will be able to use this key as well

For the sake of this key - which encrypts Glue Job and Crawler I do not see a major risk but if there is a need to change can set the `policy=` parameter directly in `kms.Key(...)` rather than using `add_to_resource_policy`

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
mourya-33: this was required @noah-paige . For some reason, during deployment, it was looking for permissions on the policy and without the dependency, the role was getting created first without the custom policy and crawler creation was failing all the time.

mourya-33:
@noah-paige I will work on the comments above and update the PR

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
mourya-33: done

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
mourya-33: done

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
mourya-33: i put the CDK admin for now, should i add this policy too Noah?

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
noah-paige: okay good data point to know - makes sense!

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
```diff
@@ -437,20 +460,68 @@ def __init__(self, scope, id, target_uri: str = None, **kwargs):
...
+            self,
+            f'{glue_sec_conf_enc_key_name}',
```
noah-paige, line 465: very minor --> can we just pass `glue_sec_conf_enc_key_name` instead of with f-string like `f'{glue_sec_conf_enc_key_name}',`

For both here in kms.Key using `glue_sec_conf_enc_key_name` and later on with `glue.CfnSecurityConfiguration` using `glue_sec_conf_name`

After that quick change I think this PR is good to merge!

mourya-33: Done. Made both changes Noah

backend/dataall/modules/s3_datasets/cdk/dataset_stack.py
```diff
@@ -11,10 +11,12 @@
     Duration,
     CfnResource,
     CustomResource,
+    RemovalPolicy,
     Tags,
 )
 from aws_cdk.aws_glue import CfnCrawler
 
+from backend.dataall.base.utils.naming_convention import NamingConventionPattern, NamingConventionService
```
noah-paige, line 19: @mourya-33 caught one last bug when testing need to import as 
```
from dataall.base.utils.naming_convention import NamingConventionPattern, NamingConventionService
```
 
removing `'backend.'` from import path

mourya-33: Done Noah

noah-paige (approved):
lgtm - thanks @mourya-33  approving

PR #1435, opened 2024-07-26
Utils for enum-queries

SofiaSazonova:
### Feature or Bugfix

- Refactoring

### Detail
- new api query `queryEnum` to query multiple enums at once by names
- shared frontend util to query several enums (fetchEnums`) by name. Output -- dictionary
```
{
 'enumName': [{name: '...', value: '..'}]
}
```
- example for Maintenance mode enum

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?


backend/dataall/base/utils/enum_utils.py
```diff
@@ -0,0 +1,26 @@
...
+        gql.Field(name='name', type=gql.String),
+        gql.Field(name='value', type=gql.String),
+    ],
+)
+
+
+def generate_enum_query(enum_class, test_scope):
+    return gql.QueryField(
+        name=enum_class.__name__,
+        type=gql.ArrayType(EnumItem),
```
petrkalos, line 23: IIRC we have some `int` enums, how are you thinking to handle this? do we have to or we should plan to migrate themt o strings?
should we have have dynamically generated `EnumItem`s something like the following?

```
if isinstance(`enum_class[0].value`, str)
  ...
elif  isinstance(`enum_class[0].value`, int)
  ...
```

backend/dataall/modules/maintenance/api/queries.py
```diff
@@ -1,9 +1,15 @@
...
 from dataall.modules.maintenance.api.resolvers import get_maintenance_window_status
 
 
 getMaintenanceWindowStatus = gql.QueryField(
     name='getMaintenanceWindowStatus', type=gql.Ref('Maintenance'), resolver=get_maintenance_window_status
 )
+
+queryMaintenanceModes = generate_enum_query(MaintenanceModes, 'Maintenance')
+
+queryMaintenanceStatus = generate_enum_query(MaintenanceStatus, 'Maintenance')
```
petrkalos, line 15: Instead of generating manually for each enum should we do it automatically [here](https://code.amazon.com/packages/dataall/blobs/1cec5fe5f3196634a709a11cfc9fe7dc87051d7a/--/backend/dataall/base/api/__init__.py#L32)?

backend/dataall/base/utils/enum_utils.py
```diff
@@ -0,0 +1,26 @@
...
+EnumItem = gql.ObjectType(
+    name='EnumItem',
+    fields=[
+        gql.Field(name='name', type=gql.String),
+        gql.Field(name='value', type=gql.String),
+    ],
+)
+
+
+def generate_enum_query(enum_class, test_scope):
```
petrkalos, line 20: what's the `test_scope`?

frontend/src/modules/Maintenance/components/MaintenanceViewer.js
```diff
@@ -314,11 +309,32 @@ export const MaintenanceViewer = () => {
     useState(START_MAINTENANCE);
   const [maintenanceWindowStatus, setMaintenanceWindowStatus] =
     useState(INACTIVE_STATUS);
+  const [maintenanceModes, setMaintenanceModes] = useState([]);
   const [dropDownStatus, setDropDownStatus] = useState(false);
   const [refreshingTimer, setRefreshingTimer] = useState('');
   const { enqueueSnackbar, closeSnackbar } = useSnackbar();
   const dispatch = useDispatch();
 
+  const fetchMaintenanceModes = useCallback(async () => {
```
petrkalos, line 318: where do you use the maintenanceModes state?

petrkalos, line 318: I am no FE expert but I am bit concerned that this might increase the latency and make the UI laggy (imagine waiting 100-200ms for a dropdown to populate). I think we might need caching or prepopulating the fields during packaging. I am happy to not do it now but I think we should generalise the enum fetching to a util `fetchEnum('MaintenanceModes', setMaintenanceModes)` that will allow us to customise the way we fetch later without changing the client code. wdyt?

backend/dataall/base/utils/enum_utils.py
```diff
@@ -0,0 +1,26 @@
...
+    return gql.QueryField(
+        name=enum_class.__name__,
```
petrkalos, line 22: * Should we get `args` as input?
* Do you think it makes sense to allow requesting multiple enums in one request to minimize RTT?

backend/dataall/base/utils/enum_utils.py
SofiaSazonova: I think we can just use str(value) since it is passed to frontend.

backend/dataall/modules/maintenance/api/queries.py
SofiaSazonova: >>You lack permission to view this package. 

I didn't want automatically expose all enums to frontend. And if we need them , we have to explicitly write it.

backend/dataall/base/utils/enum_utils.py
SofiaSazonova: QueryField has the parameter "test_scope". It means, the this part is loaded, when we test this module AFAIK

frontend/src/modules/Maintenance/components/MaintenanceViewer.js
SofiaSazonova: We should wait for client to be initialised before using this query.
I'm not really front end dev, so may be it's not necessary.

backend/dataall/base/utils/enum_utils.py
SofiaSazonova: Args? What for?
No, I don't think so. Usually, it's at most two enums per page with no heavy resolver. May be we can save it for later.

frontend/src/modules/Maintenance/components/MaintenanceViewer.js
SofiaSazonova: I like the idea of optimisation. I will try to do it, but I'm not sure I am good enough in react)

backend/dataall/base/utils/enum_utils.py
petrkalos: Ah wait my bad, I think I misunderstood your implementation. What you do is generate a query for each enum that's why you don't require an input argument.

Would it maybe be simpler to create a query with an input argument `enum_names` that points to a resolver that does the lookup for the enums in `GraphQLEnumMapper.__subclasses__()`? If you don't want to expose everything you can add an allow-list in the config.json but I don't see why.

SofiaSazonova: With this dynamic lookup we can get a circular dependencies issue. Every module depends on" base". In order to do that, we have to create another module, I'm not sure it worth it.

SofiaSazonova: It worked)

backend/dataall/base/utils/enum_utils.py
petrkalos, line 1: should we follow the generic pattern of the project? by that I mean

* move the `enum_resolver` to `dataall/base/api/resolvers.py`
* move the `EnumItem` and `EnumResult` in `dataall/base/api/types.py`
* move the `Query` in `dataall/base/api/queries.py` and drop the `generate_enum_query`, just `queryEnums = gql.QueryField(...)` should do the trick (as it does it for the rest)

backend/dataall/base/utils/enum_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    fields=[
+        gql.Field(name='name', type=gql.String),
+        gql.Field(name='items', type=gql.ArrayType(EnumItem)),
+    ],
+)
+
+
+def generate_enum_query():
+    return gql.QueryField(
+        name='queryEnum',
```
petrkalos, line 37: nit: plural

frontend/src/services/graphql/Shared/getEnumByName.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+};
+
+export const fetchEnums = async (client, enum_names) => {
+  const response = await client.query(
+    getEnumByName({ enum_names: enum_names })
+  );
+  let result = {};
+  if (!response.errors && response.data.queryEnum != null) {
+    for (let i = 0; i < response.data.queryEnum.length; i++) {
```
petrkalos, line 37: nit: instead of messing with `i` you could do either...
```
for (const element of theArray) {
    // ...use `element`...
}
```
or use `forEach`

frontend/src/services/graphql/Shared/getEnumByName.js
```diff
@@ -0,0 +1,44 @@
...
+    getEnumByName({ enum_names: [enum_name] })
+  );
+  if (!response.errors && response.data.queryEnum != null) {
+    return response.data.queryEnum[0].items;
+  } else {
+    return [];
+  }
+};
+
+export const fetchEnums = async (client, enum_names) => {
```
petrkalos, line 31: nit: I'd keep only this method and drop the fetchOneEnum and if I want to query one I'd pass `fetchEnums(..., ['EnumName'])`

petrkalos (changes requested):

frontend/src/services/graphql/Shared/getEnumByName.js
SofiaSazonova: Oh, yes. Honestly -- it's all Q's suggestions))

petrkalos (approved):

Issue #348, opened 2023-03-02
Granular sharing using Lake Formation


chamiles:
Customer has enabled granular (Row, Column, Cell) sharing using lake formation sharing and would like to see that capability in data.all sharing request, and in approval area. So that data owners and share the same datasets with restricted access to columns with out having to create duplicate another dataset and data.

Customer current solution is to use data filters and typing in a Manuel expression, this may be the best open a text area to add expression, but best user experience would show column names in a visual checkbox way with ability to put in expressions for rows and cells.

https://docs.aws.amazon.com/lake-formation/latest/dg/data-filters-about.html

dlpzx:
Hi @chamiles, for column level sharing we opened an issue a while ago. But there was no much interests (#84 ), instead we opted to implement Tag-based access control using Lake Formation, which also allows column level granularity. It is a feature in-progress at the moment developed in this issue #186 

For row/cell level filtering we can work together on implementing this sort of data filters in shared tables.

anmolsgandhi:
Bumping this up, This issue will be used to track the enhancement and extension of the current data.all sharing functionality to include [column-level and row-level access control](https://docs.aws.amazon.com/lake-formation/latest/dg/data-filters-about.html) in Lake Formation. This has been one of the top requests from customers based on our conversations. This will be included as a feature enhancement for completion in v2.7.

zsaltys:
@anmolsgandhi I think it is important we get the UI right to make it clear that this can only for table sharing and if you grant S3 access then you're effectively breaking this security measure because user will have access to the entire bucket and dataset behind the table. Let's discuss this.

noah-paige:
Hi @zsaltys - I am starting to work on design for this and can start sharing some mock ups of the UI here as I continue to progress this week

I agree some type of warning or call out from FE would be helpful to ensure user understands what they are doing when sharing data objects - but I think we can add warnings more so from the bucket sharing side because sharing an entire bucket is akin to sharing ALL folders and tables in data.all 

I will post additional design description on this issue today

noah-paige:
# Design (WIP)
---

## Assumptions
---
- Dataset Owners will be the team that is responsible for creating data filters and applying them to shares
- Data Filters will only be supported for S3 Dataset Table Share Items
  -  *Note: Data Filters are backed by LakeFormation Data Filters which is currently only used in S3 Dataset Table Sharing in data.all*
[...]

PR #1438, opened 2024-07-30
Row/Column Level Data Filters

noah-paige:
### Feature or Bugfix

- Feature

### Detail
- Add row/column level filters for table share items in data.all

### Relates
- https://github.com/data-dot-all/dataall/issues/348

### ⚠️ Important Call Outs for This PR ⚠️ 

- (1) data.all will NO LONGER support cross account sharing of table data to QuickSight Principals (i.e. QS Groups)
    - This PR introduces a more efficient sharing mechanism to directly share with foreign IAM principals (not support with QS Principals at the moment)
    - Access to data via QuickSight in same account is still supported as well as sharing of QuickSight dashboards
    - Please look at [comments in issue](https://github.com/data-dot-all/dataall/issues/348#issuecomment-2245400193) for more details on sharing mechanism, most importantly the changes involve:
        - Upgrade to using LF Data Catalog Settings Version 3
        - Direct shares to IAM Principals (for both cross or same account principals)

- (2) data.all will NO LONGER support "old_share" naming convention
    - Sharing of dataset tables was improved to reuse share DB and resource link resources as of [this PR](https://github.com/data-dot-all/dataall/pull/1016) released in [version v2.3.0](https://github.com/data-dot-all/dataall/releases/tag/v2.3.0) 
    - "old_share" naming convention can be identified as shared database resources that follow the naming convention `source_db_name + '_shared_' + shareUri` (new naming convention is ``source_db_name + '_shared`)
    - To update to new naming convention please revoke and recreate the table share on data.all >= v2.3.0
    - More details on migration path to be provided closer to release of v2.7.0!

- (3) As part of upgrade from v2.6 -> v2.7 data.all admins will have to verify + reapply all of the table shares in data.all
    - There are already built-in ECS Tasks to easily manage verify + reapply for you!
    - This will allow data.all to migrate all table shares to the new direct share to foreign IAM principals and clean up unused LakeFormation permissions (wherever applicable) 

- (4) With this PR there will be a change to the way preview-ability of table data + metadata works
    -  Please read [comment below](https://github.com/data-dot-all/dataall/pull/1438#issuecomment-2261536325) for more details

### Security

  - Is the input sanitized?


noah-paige:
### Testing
- [ ] Create Data Filter  (both COLUMN + ROW)
- [ ] List Data Filters
- [ ] Attach Filters to Share Item Table
- [ ] Update Filters to Share Item Table
- [ ] Remove Filters from Share Item Table
- [ ] Attach Filters to Share item Table on same table with same name --> violate uniqueness constraint!
- [x] Approve, Verify, Revoke on a Table Share w/ Default Full Access ( + Worksheet Query)
- [x] Approve, Verify, Revoke on a Table Share w/ Column Filter ( + Worksheet Query)
- [x] Approve, Verify, Revoke on a Table Share w/ Row Filter ( + Worksheet Query)
- [x] Approve, Verify, Revoke on a Table Share w/ Multiple Column + Row Filters ( + Worksheet Query)
- [ ] Delete Data Filter w/ Existing Share Item Filter(s) --> Error Thrown (must remove share item)
- [ ] Delete Data Filter
- [ ] Delete Dataset Table --> All Table Data Filters deleted

noah-paige:
## Comments on Table Preview-ability

Currently in data.all (prior to this PR) - the accessability of table information is a such:
- If a Table if `Unclassified`
    - Users can `Preview` the Table without requiring share with no restrictions (up to 50 rows)
    - Users can view `Columns` and `Metrics` for the Table without requiring share
- If a Table is `Secret` or `Official`
    - Users cannot `Preview` or see `Columns` or `Metrics` if there is no share
    - But once a user gets sahre access to a particular table
        - Users can `Preview` the Table without requiring share with no restrictions (up to 50 rows)
        - Users can view `Columns` and `Metrics` for the Table

This is problematic when it comes to granular shares because what if a user should not be able to see data for all columns in preview EVEN AFTER gaining share access (i.e. column filter to include specific subset ONLY)?


This PR Handles table information access in the following way:
- If a Table if `Unclassified`
    - Users can `Preview` the Table without requiring share with no restrictions (up to 50 rows)
    - Users can view `Columns` and `Metrics` for the Table without requiring share

- If a Table is `Secret` or `Official`
    - Users cannot `Preview` or see `Columns` or `Metrics` if there is no share
    - If there is a share for a user to this particular table
        - Users can view `Columns` and `Metrics` for the Table
        - If share has no filters --> Users can `Preview` the Table without requiring share with no restrictions (up to 50 rows)
        - If share has filters --> Users can `Preview` the Table with only the same level of access as they have on the on filtered table share item

There is an edge case that is important to call out on the above:
- If a given user is a part of 2 groups and each group has access to the same source table with different filters applied
    - How to resolve which level of data access is shown on the `Preview` Tab
    - Currently, it is at random which of the 2 (or more) types of previews are shown (i.e. whichever record is returned first when checking permissions in backend DB is used)


Because of the above complication - a more straightforward proposal would be to remove "Preview" altogether for `Secret` or `Official` tables
- At the point `Preview` is available to a user the user must already be given some level of share access anyway
- We would still keep `Columns` and Metrics` conditionally exposed once a user is given share access to these type of confidentials tables

Please let me know your thoughts on this - I am more inclined to switch to the model where `Preview` is only an option for `Unclassified` datasets: @petrkalos @dlpzx @SofiaSazonova

dlpzx:
Hi @noah-paige, reviewing the key decisions on this PR:
- Quicksight groups -> agree. In addition we could point out at next steps involving QS IAM IC instead of using QS groups and QS users as done now
- Deprecation of old/upgrade with ECS reapply needed - agreed on all. We just need to make sure to include it in the release notes
- Preview/Metrics on shares with filters: First of all, I don't like the random selection of role to read the preview hahahah. Secondly, previewing is very useful for the discovery of datasets (especially before share); but for confidential/secret datasets ensuring a very restricted data access is more important that some users being able to see actual real sample data. In my opinion, priority number one should be that there is no information disclosure. Again preview is useful for data discovery (before sharing), when data is already shared as you pointed out users already have some level of access. In short, I agree with option " a more straightforward proposal would be to remove "Preview" altogether for Secret or Official tables" because protecting secret/official data is a MUST, while the impact of not previewing already shared data is not that great for users

backend/dataall/modules/datasets_base/services/dataset_service_interface.py
```diff
@@ -50,3 +50,7 @@ def extend_attach_steward_permissions(session, dataset, new_stewards) -> bool:
     def extend_delete_steward_permissions(session, dataset, new_stewards) -> bool:
         """Abstract method to be implemented by dependent modules that want to attach additional permissions to Dataset stewards"""
         ...
+
```
dlpzx, line 53: What is this method used for?

backend/dataall/modules/s3_datasets/services/dataset_table_data_filter_enums.py
dlpzx, line 1: A general comment on enums; maybe out if this PR. We moved all enums to API package in #978; but there are enums that are used by the services, which often results in circular dependencies. From what I see there are 2 types of enums, those used in the services and those used in the API calls definitions. I we are not using DatFilterType as `gql.Argument('filterType', gql.NonNullableType(gql.String)),` in input_types I am not sure we need to use the graphqlEnumMapper

backend/dataall/modules/s3_datasets/db/dataset_table_data_filter_repositories.py
```diff
@@ -0,0 +1,52 @@
...
+from dataall.modules.s3_datasets.db.dataset_models import DatasetTableDataFilter
+from dataall.modules.s3_datasets.services.dataset_table_data_filter_enums import DataFilterType
+from dataall.base.db import paginate
+
+logger = logging.getLogger(__name__)
+
+
+class DatasetTableDataFilterRepository:
+    @staticmethod
+    def build_data_filter(session, context, table_uri, data):
```
dlpzx, line 13: nit: instead of passing the whole context, we could pass the username only

backend/dataall/modules/s3_datasets/services/dataset_table_service.py
```diff
@@ -16,16 +16,21 @@
...
+from dataall.modules.s3_datasets.aws.lf_data_filter_client import LakeFormationDataFilterClient
 from dataall.modules.s3_datasets.db.dataset_repositories import DatasetRepository
 from dataall.modules.datasets_base.services.datasets_enums import ConfidentialityClassification
 from dataall.modules.s3_datasets.db.dataset_models import DatasetTable, S3Dataset
 from dataall.modules.s3_datasets.services.dataset_permissions import (
     PREVIEW_DATASET_TABLE,
-    DATASET_TABLE_READ,
+    DATASET_TABLE_ALL,
     GET_DATASET_TABLE,
+    CREATE_TABLE_DATA_FILTER,
```
dlpzx, line 27: Remove unused imports:     CREATE_TABLE_DATA_FILTER,
    DELETE_TABLE_DATA_FILTER,
    LIST_TABLE_DATA_FILTERS,

backend/dataall/modules/s3_datasets/services/dataset_service.py
```diff
@@ -56,6 +56,16 @@ class DatasetService:
     def register(cls, interface: DatasetServiceInterface):
         cls._interfaces.append(interface)
 
+    @classmethod
+    def _resolve_consumption_details_resource_uri(cls, session, uri, group_uri) -> list:
+        """Find resources from other modules that need to be accessed"""
+        for interface in cls._interfaces:
+            consumption_detail = interface.find_consumption_details_resource_uri(session, uri, group_uri)
```
dlpzx, line 63: `find_consumption_details_resource_uri` is not part of the DatasetServiceInteface abstract classes, therefore when adding new interfaces it might fail because the method is not mandatory to be implemented

backend/dataall/modules/s3_datasets/services/dataset_table_service.py
```diff
@@ -85,19 +99,49 @@ def preview(uri: str):
...
+                    consumption_details = DatasetService._resolve_consumption_details_resource_uri(
+                        session, table.tableUri, resource_policy.principalId
+                    )
+                    if not consumption_details:
+                        raise exceptions.ResourceUnauthorized(
+                            username=context.username,
+                            action=PREVIEW_DATASET_TABLE,
+                            resource_uri=table.tableUri,
+                        )
+                    database_name = database_name + '_shared'
```
dlpzx, line 131: Some of this logic could be part of `_resolve_consumption_details_resource_uri` and the interface implementation. For example, the `_shared` suffix should not be here, it is a data_sharing parameter not a dataset parameter

backend/dataall/modules/datasets_base/services/dataset_service_interface.py
noah-paige: I switched the name to `find_consumption_details_resource_uri` and forgot to change here but only changed on `S3DatasetService` - good catch

It will likely go away anyway - was used for Preview on Confidential data

backend/dataall/modules/s3_datasets/services/dataset_service.py
noah-paige: yup - `find_resources_by_uri` from DatasetServiceInteface to be renamed `find_consumption_details_resource_uri` 

... but may be removed if removing Preview ability fro confidential datasets

backend/dataall/modules/s3_datasets/services/dataset_table_service.py
noah-paige: removed as part of removing Preview Tab for Confidential Data

backend/dataall/modules/s3_datasets/services/dataset_service.py
noah-paige: removed as part of removing Preview Tab for Confidential Data

backend/dataall/modules/datasets_base/services/dataset_service_interface.py
noah-paige: removed as part of removing Preview Tab for Confidential Data

backend/dataall/modules/s3_datasets/services/dataset_table_data_filter_enums.py
noah-paige: we do not need `graphqlEnumMapper` if do not need enum defined in GQL Schema - for the case of `DataFilterType` you are correct can just make `Enum` if that is the suggested pattern

we do use `DataFilterType` in services so I imagine will run into similar circular dependencies if move to API right?

tests/modules/s3_datasets/test_dataset_table.py
```diff
@@ -242,3 +256,152 @@ def test_delete_table(client, table, dataset_fixture, db, group):
         tableUri=table_to_delete.tableUri,
     )
```
dlpzx, line 257: nit: it would be nice to either add another test for `delete_table_with_filters` and assert that the filters are deleted in `lf_client.delete_table_data_filter`

backend/dataall/modules/s3_datasets/api/table/queries.py
```diff
@@ -36,3 +37,13 @@
     resolver=preview,
     type=gql.Ref('QueryPreviewResult'),
 )
+
+listTableDataFilters = gql.QueryField(
+    name='listTableDataFilters',
+    args=[
+        gql.Argument(name='tableUri', type=gql.NonNullableType(gql.String)),
+        gql.Argument('filter', DatasetTableFilter),
+    ],
```
dlpzx, line 46: picky nit: in a recent PR it was pointed out that we are mixing regular args with key-value args, let's try to be consistent

backend/dataall/modules/s3_datasets/db/dataset_table_repositories.py
```diff
@@ -39,6 +44,19 @@ def create_synced_table(session, dataset: S3Dataset, table: dict):
     def delete(session, table: DatasetTable):
         session.delete(table)
```
dlpzx, line 46: nit/suggestion: we could move this to the data filter repository. There is already a paginated_data_filters which is basically doing list_data_filters + pagination

backend/dataall/modules/s3_datasets/services/dataset_table_data_filter_service.py
```diff
@@ -0,0 +1,115 @@
...
+                'includedCols',
+                data.get('includedCols'),
+                f'must be provided for {data.get("filterType")} filter',
+            )
+
+    @staticmethod
+    def validate_creation_data_filter_params(uri, data):
+        DatasetTableDataFilterRequestValidationService._required_param(uri, 'tableUri')
+        DatasetTableDataFilterRequestValidationService._required_param(data, 'data')
+        DatasetTableDataFilterRequestValidationService._required_param(data.get('filterName'), 'filterName')
```
dlpzx, line 54: Does this constraint come from the naming convention of Lake Formation filters? In that case we could use the naming convention utilities no?

backend/dataall/modules/s3_datasets/services/dataset_table_data_filter_service.py
```diff
@@ -0,0 +1,115 @@
...
+
+    @staticmethod
+    @ResourcePolicyService.has_resource_permission(DELETE_TABLE_DATA_FILTER, parent_resource=_get_table_uri_from_filter)
+    def delete_table_data_filter(uri: str):
+        with get_context().db_engine.scoped_session() as session:
+            data_filter = DatasetTableDataFilterRepository.get_data_filter_by_uri(session, filter_uri=uri)
+
+            # Check if Share Items w Filter before Delete
+            DatasetService.check_before_delete(session, data_filter.filterUri, action=DELETE_TABLE_DATA_FILTER)
+
```
dlpzx, line 101: Nice! i had completely forgotten about this interface!

backend/dataall/modules/s3_datasets/services/dataset_table_service.py
```diff
@@ -70,6 +72,15 @@ def delete_table(uri: str):
             table = DatasetTableRepository.get_dataset_table_by_uri(session, uri)
             DatasetService.check_before_delete(session, table.tableUri, action=DELETE_DATASET_TABLE)
             DatasetService.execute_on_delete(session, table.tableUri, action=DELETE_DATASET_TABLE)
+
```
dlpzx, line 75: Why are we not defining (lines 76-83) in DatasetService.execute_on_delete(session, table.tableUri, action=DELETE_TABLE_DATA_FILTER)?

backend/dataall/modules/shares_base/db/share_object_models.py
```diff
@@ -57,3 +58,17 @@ class ShareObjectItem(Base):
     healthStatus = Column(String, nullable=True)
     healthMessage = Column(String, nullable=True)
     lastVerificationTime = Column(DateTime, nullable=True)
+    attachedDataFilterUri = Column(
```
dlpzx, line 61: We must be aware that this field is now affecting any type of share object (soon redshift shares, but also future notebook shares, X shares...) At first it does not seem to be very S3-Dataset-LF-filter specific, so I am okey introducing data filters to the share object; I'll review that there are no S3-LF specific assumptions in the shares_base module, but I wanted to bring it up because you know the code better

backend/dataall/modules/shares_base/db/share_object_models.py
```diff
@@ -57,3 +58,17 @@ class ShareObjectItem(Base):
...
+        String, ForeignKey('share_object_item_data_filter.attachedDataFilterUri'), nullable=True
+    )
+
+
```
dlpzx, line 65: Just curious. The implementation of ShareObjectItemDataFilter surprised me a bit, my first impulse would have been to leave ShareObjectItem without any additional columns and implement ShareObjectItemDataFilter with a reference to ShareObjectItem (same way ShareObject-ShareObjectItem do)

```
class ShareObjectItemDataFilter(Base):
    __tablename__ = 'share_object_item_data_filter'
    attachedDataFilterUri = Column(String, primary_key=True, default=utils.uuid('shareitemdatafilter'))
    label = Column(String, nullable=False)
    ++ shareObjectItemUri = Column(String, nullable=False)
    ++ dataFilterUri = Column(String, nullable=False)
    ++ dataFilterName = Column(String, nullable=False)
    -- dataFilterUris = Column(ARRAY(String), nullable=False)
    -- dataFilterNames = Column(ARRAY(String), nullable=False)
    itemUri = Column(String, nullable=False)
```

But looking at the PR solution, the queries might be faster, we don't need to resolve as many subqueries. I am just curious to understand why hahahhah

backend/migrations/versions/9efe5f7c69a1_add_data_filters_table.py
dlpzx, line 1: Despite having the correct file name, the internal alembic name for this migration is "describe_changes_shortly"
image

dlpzx: I think we can change that in the first line of this file

dlpzx, line 1: Tested! The migration script added the new tables and backfilled the filter permissions for an existing dataset!

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -216,14 +216,14 @@ def revoke_share(cls, engine: Engine, share_uri: str) -> bool:
...
-                                share_item_type=processor.type.value,
+                                share_item_type=processor.type,
                             )
                             ShareStatusRepository.update_share_item_status_batch(
                                 session,
                                 share_uri,
                                 old_status=ShareItemStatus.Revoke_In_Progress.value,
                                 new_status=ShareItemStatus.Revoke_Failed.value,
-                                share_item_type=processor.type.value,
+                                share_item_type=processor.type,
```
dlpzx, line 226: PR https://github.com/data-dot-all/dataall/pull/1452/files (already merged) solves this issue

dlpzx:
Tested:
- migration script
- create filter ----> ERROR: `An error occurred (AccessDeniedException) when calling the CreateDataCellsFilter operation: User: arn:aws:sts::XXXXX:assumed-role/dataall-aburrida-ia476u0x/dataall-aburrida-ia476u0x is not authorized to perform: lakeformation:CreateDataCellsFilter on resource: arn:aws:lakeformation:eu-west-1:XXXXXX:catalog:XXXXXXX because no identity-based policy allows the lakeformation:CreateDataCellsFilter action` ----> I think we are missing permissions in the dataset role that is used in the lf client no?

backend/dataall/modules/s3_datasets/db/dataset_table_repositories.py
noah-paige: good call out - moving

backend/dataall/modules/s3_datasets/services/dataset_table_data_filter_service.py
noah-paige: yes coming from LF / Glue naming convention - will look into using that instead of `re.search()`

backend/dataall/modules/s3_datasets/services/dataset_table_service.py
noah-paige: currently only interface registered in `DatasetService` is `S3ShareDatasetService` which was share item specific so this felt like the best place to leave in DatasetTableService since also specific only to Tables right now

backend/dataall/modules/shares_base/db/share_object_models.py
noah-paige: yeah I made data filter a generic `share_base` component thinking that the concept of share data filters is not necessarily table specific - only parts of the implementation are specific to handling table share items and LF

Think about for data filters maybe introducing a regex pattern to exclude could be an option on bucket or table shares and we could reuse this concept of attached filters?  Open to feedback if you think about it differently

backend/dataall/modules/shares_base/db/share_object_models.py
noah-paige: Honestly i just implemented how I thought made sense did not do much digging into other instances of the same but you are right it is a bit different

I do think it is helpful to quick determine if a share item has filters or not without having to do a separate query on `share_object_item_data_filter` to resolve if a data filter exists for that share item - what do you think?

backend/migrations/versions/9efe5f7c69a1_add_data_filters_table.py
noah-paige: will update

noah-paige:
>     * create filter ----> ERROR: `An error occurred (AccessDeniedException) when calling the CreateDataCellsFilter operation: User: arn:aws:sts::XXXXX:assumed-role/dataall-aburrida-ia476u0x/dataall-aburrida-ia476u0x is not authorized to perform: lakeformation:CreateDataCellsFilter on resource: arn:aws:lakeformation:eu-west-1:XXXXXX:catalog:XXXXXXX because no identity-based policy allows the lakeformation:CreateDataCellsFilter action` ----> I think we are missing permissions in the dataset role that is used in the lf client no?

Yup - you are exactly correct, had manually scoped down permissions on IAM Role but slipped my mind to include in dataset stack IaC for PR. Should be reflected now

tests/modules/s3_datasets/test_dataset_table.py
noah-paige: added another test for this case

backend/dataall/modules/s3_datasets/services/dataset_table_data_filter_service.py
noah-paige: Using Naming Convention now with a new `validate_name` method

backend/dataall/modules/shares_base/db/share_object_models.py
dlpzx: We would always work with dataset-specific filters, we will never do a full list or query filters from various datasets. That's why I think this implementation makes more sense in this particular case. I think I'll copy it for some cases with the same parent-child relationship

backend/dataall/modules/s3_datasets/services/dataset_table_service.py
dlpzx: Never mind I see it know. We don't need an interface since all code is in the s3_datasets_module. If we ever make filters generic and extract them out of s3_datasets_shares then we will need to introduce the interface. Resolving for the moment

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
dlpzx, line 1: nit:maybe we can rename it to ShareObjectFilterRepository

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
```diff
@@ -0,0 +1,92 @@
+import logging
+from sqlalchemy import and_, or_, func, case
+from sqlalchemy.orm import Query
+from typing import List
+
+from dataall.base.db import exceptions, paginate
```
dlpzx, line 6: Remove unused imports

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
```diff
@@ -0,0 +1,92 @@
...
+            )
+            .count()
+        )
+
+    @staticmethod
+    def get_share_item_by_item_filter_uri(session, uri):
+        return session.query(ShareObjectItem).filter(ShareObjectItem.attachedDataFilterUri == uri).first()
+
+    @staticmethod
+    def delete_share_item_filter(session, share_item_filter) -> bool:
```
dlpzx, line 48: nit: type annotations are returning a boolean but this function does not return anything

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
```diff
@@ -0,0 +1,92 @@
...
+        session.delete(share_item_filter)
+
+    @staticmethod
+    def update_share_item_filters(
+        session,
+        share_item_filter: ShareObjectItemDataFilter,
+        data: dict,
+    ) -> bool:
```
dlpzx, line 56: nit: type annotations returning a boolean when this function is returning a ShareObjectItemDataFilter

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
```diff
@@ -0,0 +1,92 @@
...
+        session.commit()
+        return share_item_filter
+
+    @staticmethod
+    def create_share_item_filters(
+        session,
+        share_item: str,
+        data: dict,
+    ) -> bool:
+        share_item_data_filter = ShareObjectItemDataFilter(
```
dlpzx, line 69: nit: type annotations returning a boolean when this function is returning a ShareObjectItemDataFilter

dlpzx: nit2: share_item is defined as str but it is a ShareObjectItem

backend/dataall/modules/shares_base/services/share_item_service.py
```diff
@@ -204,3 +216,53 @@ def list_shareable_objects(share, filter, is_revokable=False):
...
+            return ShareObjectItemRepository.get_share_item_filter_by_uri(session, uri)
+
+    @staticmethod
+    @ResourcePolicyService.has_resource_permission(
+        APPROVE_SHARE_OBJECT, parent_resource=_get_share_uri_from_item_filter_uri
+    )
+    def remove_share_item_data_filters(uri: str):
+        with get_context().db_engine.scoped_session() as session:
+            share_item = ShareObjectItemRepository.get_share_item_by_item_filter_uri(session, uri)
+            share_item.attachedDataFilterUri = None
```
dlpzx, line 262: I am not 100% sure, but will this line update the attachedDataFilters even when the exception in line 263 is raised?

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
dlpzx: nit3: rename it without being plural: `create_share_item_filter`

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
dlpzx: nit2: rename it without being plural: `update_share_item_filter`

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
```diff
@@ -0,0 +1,92 @@
...
+    @staticmethod
+    def update_share_item_filters(
+        session,
+        share_item_filter: ShareObjectItemDataFilter,
+        data: dict,
+    ) -> bool:
+        share_item_filter.label = data.get('label')
+        share_item_filter.dataFilterUris = data.get('filterUris')
+        share_item_filter.dataFilterNames = data.get('filterNames')
+        session.commit()
```
dlpzx, line 60: In the update, is it possible to not provide some of these keys? For example, let's say we do not update the label, would we update the label to None?

frontend/src/modules/Tables/components/TableFilters.js
```diff
@@ -0,0 +1,215 @@
...
+                  onChange={handleInputChange}
+                  onKeyUp={handleInputKeyup}
+                  placeholder="Search"
+                  value={inputValue}
+                  variant="outlined"
+                />
+              </Box>
+            </Grid>
+            <Grid item md={2} sm={6} xs={12}>
+              <Button
```
dlpzx, line 174: very very nit: disable this button if no rows are selected or no row expressions are saved

frontend/src/modules/Tables/components/TableFiltersDataGrid.js
```diff
@@ -0,0 +1,157 @@
...
+    return <CircularProgress />;
+  }
+
+  return (
+    <DataGrid
+      autoHeight
+      sx={{
+        wordWrap: 'break-word',
+        boxShadow: 2,
+        borderRadius: 1,
```
dlpzx, line 42: I ran into the same issue when implementing datagrids for Redshift. For our dark theme the borders look bad. I am guessing that's why you added the borders here, but i think there is something off with the grey shadows and the header vs normal rows.
image

I think that we should include the data grid style in the top-theme but for the time being here are the params that I used.

image




```
              sx={{
                wordWrap: 'break-word', //TODO: create a generic styled datagrid to be used across features
                '& .MuiDataGrid-row': {
                  borderBottom: '1px solid rgba(145, 158, 171, 0.24)'
                },
                '& .MuiDataGrid-columnHeaders': {
                  borderBottom: 0.5
                },
                '&.MuiDataGrid-root--densityStandard .MuiDataGrid-cell': {
                  py: '15px'
                }
              }}
```

frontend/src/modules/Shared/Shares/ShareEditForm.js
```diff
@@ -398,6 +399,14 @@ export const ShareEditForm = (props) => {
         {getExplanation(shareStatus)}
       </Typography>
       <Box>
+        <Box sx={{ p: 1 }}>
+          {sharedItems.nodes.find((item) => item.itemType === 'S3Bucket') && (
+            <Alert severity="warning" gutterBottom sx={{ mr: 1 }}>
+              Sharing S3Bucket gives Requestor read access to{' '}
+              <b>the entire S3 Bucket</b> superseding Table + Folder shares
```
dlpzx, line 406: For Folder shares yes, but for table shares not 100%; yes you can access the bucket but not the glue table

frontend/src/modules/Shares/components/ShareItemFilterModal.js
```diff
@@ -0,0 +1,386 @@
...
+  return (
+    <Dialog maxWidth="lg" fullWidth onClose={onClose} open={open} {...other}>
+      <Box sx={{ p: 3 }}>
+        <Typography
+          align="center"
+          color="textPrimary"
+          gutterBottom
+          variant="h4"
+        >
+          Assign data filters to a table share item {item.itemName}
```
dlpzx, line 246: nit: simplify title "Assign data filters to {item.itemName}"

dlpzx: nit2: I would remove the select Data filters text, I think it is pretty clear the data filters need to be selected. As for the create data filter text I would turn it into a button, it would be great to compact it and place it next to the search bar.

nit3: this data grid in unstyled, we could add the boarders as commented for the TableView

image

frontend/src/services/graphql/DatasetTable/listSharedDatasetTableColumns.js
dlpzx, line 1: the top-level services are reserved for those APIs used across multiple modules. In this case `listSharedDatasetTableColumns` is only used in Worksheets. I know it looks a bit weird

backend/dataall/modules/shares_base/api/types.py
```diff
@@ -31,9 +31,22 @@
...
+        gql.Field('attachedDataFilterUri', gql.String),
     ],
 )
 
+ShareObjectItemDataFilter = gql.ObjectType(
+    name='ShareObjectItemDataFilter',
+    fields=[
+        gql.Field(name='attachedDataFilterUri', type=gql.String),
+        gql.Field(name='label', type=gql.String),
+        gql.Field('dataFilterUris', type=gql.ArrayType(gql.String)),
```
dlpzx, line 43: very very nit: key-value vs value args

tests/modules/s3_datasets_shares/conftest.py
```diff
@@ -206,6 +211,29 @@ def factory(dataset: S3Dataset, name, username) -> DatasetTable:
...
+    ) -> DatasetTableDataFilter:
+        with db.scoped_session() as session:
+            data_filter = DatasetTableDataFilter(
+                tableUri=table.tableUri,
+                label=name,
+                filterType=filterType,
+                rowExpression=None,
+                includedCols=['id1', 'id2'],
+                owner='foo',
+            )
```
dlpzx, line 229: This can only be a COLUMN filter, as it is tested in the tests. Maybe we need make it more explicit to avoid unexpected errors down the line

tests/modules/s3_datasets_shares/conftest.py
```diff
@@ -237,13 +265,19 @@ def table_fixture(db, dataset_fixture, table, group, user):
         ResourcePolicyService.attach_resource_policy(
             session=session,
             group=group.groupUri,
-            permissions=DATASET_TABLE_READ,
+            permissions=DATASET_TABLE_ALL,
             resource_uri=table1.tableUri,
             resource_type=DatasetTable.__name__,
         )
     yield table1
```
dlpzx, line 274: Remove commented code

tests/modules/s3_datasets_shares/test_share.py
```diff
@@ -146,6 +146,12 @@ def table3(table: typing.Callable, dataset3: S3Dataset) -> DatasetTable:
     yield table(dataset=dataset3, name='table3', username='bob')
 
 
+@pytest.fixture(scope='module')
+def table_data_filter_fixture(db, table_fixture, table_data_filter, group, user):
+    table_data_filter = table_data_filter(table=table_fixture, name='datafilter1', filterType='COLUMN')
+    yield table_data_filter
+
```
dlpzx, line 153: nit: yield table_data_filter(table=table_fixture, name='datafilter1', filterType='COLUMN')

tests/modules/s3_datasets_shares/test_share.py
```diff
@@ -622,6 +653,37 @@ def add_share_item(client, user, group, shareUri, itemUri, itemType):
...
+        updateShareItemFilters(input: $input)
+      }
+        """
+
+    response = client.query(q, username=user.username, groups=[group.name], input=input)
+
+    print('Response from updateShareItemFilters: ', response)
+    return response
+
+
```
dlpzx, line 668: We are missing the test for `remove_share_item_filter` for `get_share_item_data_filters` i think it is fine with the test of the update filters function

dlpzx:
It is a great implementation of the feature, thanks @noah-paige! Some things that we should not forget to do:
- [ ] Add documentation!
- [ ] Add in the description of this PR some messages to warn users:
        - Deprecated old shared db names
        - force upgrade of lake formation to version 3 or more

backend/dataall/modules/s3_datasets_shares/db/s3_share_object_repositories.py
```diff
@@ -12,11 +12,12 @@
...
     PrincipalType,
 )
+from dataall.base.db import exceptions
 from dataall.modules.shares_base.db.share_state_machines_repositories import ShareStatusRepository
-from dataall.modules.shares_base.db.share_object_models import ShareObjectItem, ShareObject
+from dataall.modules.shares_base.db.share_object_models import ShareObjectItem, ShareObject, ShareObjectItemDataFilter
 from dataall.modules.shares_base.db.share_object_repositories import ShareObjectRepository
 from dataall.modules.s3_datasets.db.dataset_repositories import DatasetRepository
-from dataall.modules.s3_datasets.db.dataset_models import DatasetTable, S3Dataset
+from dataall.modules.s3_datasets.db.dataset_models import DatasetTable, S3Dataset, DatasetTableDataFilter
```
dlpzx, line 20: nit: unused import

backend/dataall/modules/s3_datasets_shares/db/s3_share_object_repositories.py
```diff
@@ -164,9 +165,11 @@ def get_share_by_dataset_attributes(session, dataset_uri, dataset_owner, groups=
         return share
 
     @staticmethod
-    def check_other_approved_share_item_table_exists(session, environment_uri, item_uri, share_item_uri):
+    def check_other_approved_share_item_table_exists(
+        session, environment_uri, item_uri, share_item_uri, data_filters=None
```
dlpzx, line 169: data_filters is not used

backend/dataall/modules/s3_datasets_shares/db/s3_share_object_repositories.py
```diff
@@ -268,10 +245,15 @@ def query_dataset_tables_shared_with_env(
                     ),
                 )
             )
-            .all()
         )
+        return env_tables_shared_query.all()
 
-        return env_tables_shared
+    @staticmethod
```
dlpzx, line 251: This method is kind of repeated in ShareObjectItemRepository.get_share_item_filter_by_uri (one with exception, one without).

backend/dataall/modules/s3_datasets_shares/services/s3_share_service.py
```diff
@@ -169,9 +178,14 @@ def reapply_share_items_for_dataset(uri: str):
...
         with context.db_engine.scoped_session() as session:
+            log.info(
+                S3ShareObjectRepository.query_dataset_tables_shared_with_env(
+                    session, env_uri, dataset_uri, context.username, context.groups
+                )
+            )
             return [
-                {'tableUri': t.tableUri, 'GlueTableName': t.GlueTableName}
-                for t in S3ShareObjectRepository.query_dataset_tables_shared_with_env(
+                {'tableUri': res[0], 'GlueTableName': res[1] + (f'_{res[2]}' if res[2] else '')}
```
dlpzx, line 187: why do we need to access fields by index, cannot we directly unpack the fields with dot notation?

backend/dataall/modules/s3_datasets_shares/services/share_managers/lf_share_manager.py
```diff
@@ -10,6 +10,7 @@
 from dataall.base.aws.sts import SessionHelper
 from dataall.base.db import exceptions
 from dataall.modules.shares_base.db.share_object_repositories import ShareObjectRepository
+from dataall.modules.s3_datasets_shares.db.s3_share_object_repositories import S3ShareObjectRepository
```
dlpzx, line 13: unused import

tests/modules/s3_datasets_shares/tasks/test_lf_share_manager.py
```diff
@@ -97,6 +97,35 @@ def share_item(share_item_table: Callable, share: ShareObject, table1: DatasetTa
     yield share_item_table(share=share, table=table1, status=ShareItemStatus.Share_Approved.value)
 
 
+@pytest.fixture(scope='module')
+def table_data_filter_fixture(db, table1, table_data_filter, group, user):
+    table_data_filter = table_data_filter(table=table1, name='datafilter1', filterType='COLUMN')
+    yield table_data_filter
```
dlpzx, line 103: very nit: yield table_data_filter(table=table1, name='datafilter1', filterType='COLUMN')

tests/modules/s3_datasets_shares/tasks/test_lf_share_manager.py
```diff
@@ -97,6 +97,35 @@ def share_item(share_item_table: Callable, share: ShareObject, table1: DatasetTa
...
+
+
+@pytest.fixture(scope='module')
+def share_item_data_filter(
+    share_item_table_data_filter: Callable, table1: DatasetTable, table_data_filter_fixture
+) -> ShareObjectItemDataFilter:
+    share_item_data_filter = share_item_table_data_filter(table=table1, table_data_filter=table_data_filter_fixture)
+    yield share_item_data_filter
+
```
dlpzx, line 112: same as above

backend/dataall/modules/shares_base/services/share_item_service.py
noah-paige: Hmmm - good catch I think it might, moving `share_item.attachedDataFilterUri = None` to after the check / raise of Exception

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
noah-paige: a user can try to leave one of those fields empty but label, dataFilterUris, and dataFilterNames are all non-nullable

From FE - we set the initial values for label, dataFilterUris, and dataFilterNames to be the existing values of ShareObjectItemDataFilter (if exist) so users can start from where they left off

If trying to invoke programmatically - one would have to specify all 3 of label, dataFilterUris, and dataFilterNames

frontend/src/modules/Tables/components/TableFilters.js
noah-paige: I think this would be for `Add Data Filter` Button on `TableDataFilterAddForm` and not for this button here right?

frontend/src/modules/Shares/components/ShareItemFilterModal.js
noah-paige: For `nit2` - what search bar?

noah-paige: Screenshot 2024-08-07 at 10 49 28 AM

is this more like what you envision?

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
noah-paige: fixed

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
noah-paige: update - return None

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
noah-paige: update - return ShareObjectItemDataFilter + no plural

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
noah-paige: updated for the above

backend/dataall/modules/shares_base/services/share_item_service.py
noah-paige: resolved above

frontend/src/modules/Tables/components/TableFilters.js
noah-paige: udpated TableDataFilterAddForm `Add Data Filter` to be disabled unless selected cols or row expression

frontend/src/modules/Tables/components/TableFiltersDataGrid.js
noah-paige: update DataGrid style to the above ++ hover piece because I think is a nice touch

frontend/src/modules/Shared/Shares/ShareEditForm.js
noah-paige: I updated the comment to be more verbose

While you cannot access the glue table you would have a workaround to access and download the underlying file in S3 when it comes to table shares

backend/dataall/modules/shares_base/api/types.py
noah-paige: using kwargs

tests/modules/s3_datasets_shares/conftest.py
noah-paige: updated naming to reflect col data filter

tests/modules/s3_datasets_shares/conftest.py
noah-paige: removed

tests/modules/s3_datasets_shares/test_share.py
noah-paige: fixed

backend/dataall/modules/s3_datasets_shares/db/s3_share_object_repositories.py
noah-paige: fixed

backend/dataall/modules/s3_datasets_shares/db/s3_share_object_repositories.py
noah-paige: removed param `data_filters`

backend/dataall/modules/s3_datasets_shares/db/s3_share_object_repositories.py
noah-paige: now using only `get_share_item_filter_by_uri` which raises ObjectNotFound if no filter found - checked where it is called and should be okay enforcing the object to exist when calling this method

backend/dataall/modules/s3_datasets_shares/services/s3_share_service.py
noah-paige: wanted to do this before but was having difficulty so went by index and meant to re-visit

have updated now by key name

backend/dataall/modules/s3_datasets_shares/services/share_managers/lf_share_manager.py
noah-paige: removed

tests/modules/s3_datasets_shares/tasks/test_lf_share_manager.py
noah-paige: resolved

tests/modules/s3_datasets_shares/tasks/test_lf_share_manager.py
noah-paige: resolved

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
noah-paige: I was originally thinking it woudl make more sense to move some methods from `ShareObjectRepository` to `ShareObjectItemRepository` -- such as:
-  `find_sharable_item` 
- `get_share_item_by_uri`
- `get_share_item_details`
- `remove_share_object_item`
- `paginated_list_shareable_items`
++ some more along with the filters with respect to share items

maybe it would make sense to have `ShareObjectRepository`, `ShareObjectItemRepository`, and `ShareObjectFilterRepository` as separate classes? Could we pick up in separate PR for this refactor if needed?

noah-paige:
> It is a great implementation of the feature, thanks @noah-paige! Some things that we should not forget to do:
> 
>     * [ ]  Add documentation!
> 
>     * [ ]  Add in the description of this PR some messages to warn users:
>       - Deprecated old shared db names
>       - force upgrade of lake formation to version 3 or more


Call outs in PR should already be there - I will update User Guide documentation and push as a separate PR

backend/dataall/modules/shares_base/db/share_object_item_repositories.py
dlpzx: I am running into the same conclusion when working with Redshift. Let's add it to next steps, but definitely out of this PR

frontend/src/modules/Shares/components/ShareItemFilterModal.js
dlpzx: Perfect!

backend/dataall/modules/shares_base/services/share_item_service.py
dlpzx: checked, all good!

backend/dataall/modules/s3_datasets_shares/services/s3_share_service.py
dlpzx: checked! all good!

dlpzx (approved):
Done a final quick test and reviewed comments, this PR is ready!

PR #1449, opened 2024-08-01
Redshift dataset module testing: Re-added client factories, mocking clients

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
- Use client factories in Redshift boto3 clients
- use single mock for the factory clients

### Relates
- #1424 

### Security

  - Is the input sanitized?

### DIFF
--- a/tests/modules/redshift_datasets/conftest.py
+++ b/tests/modules/redshift_datasets/conftest.py
@@ -19,7 +19,7 @@ def patch_sts_remote_session(module_mocker):
 
 
 @pytest.fixture(scope='function')
-def patch_redshift(mocker):
+def mock_redshift(mocker):
     redshiftClient = mocker.patch('dataall.modules.redshift_datasets.aws.redshift.RedshiftClient', autospec=True)
     redshiftClient.return_value.describe_cluster.return_value = {
         'ClusterIdentifier': 'cluster_id_1',
@@ -29,7 +29,7 @@ def patch_redshift(mocker):
 
 
 @pytest.fixture(scope='function')
-def patch_redshift_data(mocker):
+def mock_redshift_data(mocker):
     redshiftDataClient = mocker.patch(
         'dataall.modules.redshift_datasets.aws.redshift_data.RedshiftDataClient', autospec=True
     )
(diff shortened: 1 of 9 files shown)


dlpzx:
Tested in AWS:
image

tests/modules/redshift_datasets/conftest.py
```diff
@@ -67,25 +20,55 @@ def patch_sts_remote_session(module_mocker):
 
 @pytest.fixture(scope='function')
 def patch_redshift(mocker):
```
petrkalos, line 22: nit: instead of `patch_...` I'd use `mock_...`

tests/modules/redshift_datasets/test_unit_redshift_connection_service.py
```diff
@@ -207,10 +189,10 @@ def test_list_environment_redshift_connections_unauthorized(
...
 
-def test_list_connection_schemas(connection1_serverless, api_context_1, patch_redshift):
+def test_list_connection_schemas(connection1_serverless, api_context_1, patch_redshift_data):
     # When
     response = RedshiftConnectionService.list_connection_schemas(uri=connection1_serverless.connectionUri)
-    assert_that(response).is_equal_to(MockRedshiftDataClient().list_redshift_schemas())
+    assert_that(response).contains('public', 'dev')
 
 
 def test_list_connection_schemas_unauthorized(connection1_serverless, api_context_2):
```
petrkalos, line 198: `connection1_serverless` is setting `api_context_2` but you also use the `api_context_2` fixture. How does that work? Is there a a conflict?

dlpzx: What we need:
- create connection1 with api context 1
- list connections with api context 2
What happens:
- we use fixture connection1 - sets api context 1 and creates connection
- we use fixture api_context 2 - sets api context 2
- execute test code - list connections

If we defined the test as `
def test_list_connection_schemas_unauthorized(api_context_2, connection1_serverless, ):` then it would fail. 

I am not the biggest fun of it, but unless we are able to keep to api contexts in parallel I am not sure I know how to fix it

PR #1447, opened 2024-07-31
Redshift data sharing - Cluster encryption guardrails and information

dlpzx:
### Feature or Bugfix
- Feature
 TO BE MERGED AFTER https://github.com/data-dot-all/dataall/pull/1424 and after #1449 

### Detail

This PR is part of the design explained in detail in #955 
- Add check that verifies the Redshift cluster imported is encrypted before creating a redshift connection
- Store the Redshift encryption type in the redshift-connection table. This is needed because Redshift datashares only work between namespaces that use the same encryption type
- In the migration script and in the db model of the redshift-connection table we add the columns encryptionType and connectionType. ConnectionType is not needed in this particular PR, but it will be needed in the next PR. It is added to simplify the migration scripts

### Relates
- #955 - Sharing part

### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)?
  - Is the input sanitized?
  - What precautions are you taking before deserializing the data you consume?
  - Is injection prevented by parametrizing queries?
  - Have you ensured no `eval` or similar functions are used?
- Does this PR introduce any functionality or component that requires authorization?
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms?
  - Are you logging failed auth attempts?
- Are you using or adding any cryptographic features?
  - Do you use a standard proven implementations?
  - Are the used keys controlled by the customer? Where are they stored?
- Are you introducing any new policies/roles/users?
  - Have you used the least-privilege principle? How?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1451, opened 2024-08-02
Redshift data sharing - Redshift connection types and namespace Id

dlpzx:
### Feature or Bugfix
- Feature

⚠️ Needs to be merged after #1447 

### Detail
Complete design in https://github.com/data-dot-all/dataall/issues/955. This particular PR is part of the redshift dataset sharing implementation. 

To be able to do sharing, the dataset namespace and the consumer namespace need to be accessible by data.all. For this reason we introduce the concept of types of connections. When opening a share request, the datasets that can be shared are only those whose namespace has an admin connection + the requester redshift role will be selected from a list of namespaces with admin connections.

In addition, this PR adds the namespaceId in the RDS database for both serverless and provisioned clusters, as it is very handy information when working with datashares

### Relates
- #955 

### Security

  - Is the input sanitized?


dlpzx:
Testing:
- UI changes render a nice table in Environments > Connections 
- Add new Connection of Data_User type
- Add a new Connection of Admin type
- Add a new connection of provisioned cluster populates the namespaceId

Screenshot 2024-08-02 at 11 06 54

backend/dataall/modules/redshift_datasets/cdk/pivot_role_redshift_policy.py
```diff
@@ -7,6 +7,7 @@
 from dataall.core.environment.cdk.pivot_role_stack import PivotRoleStatementSet
 from dataall.modules.redshift_datasets.db.redshift_connection_repositories import RedshiftConnectionRepository
 from dataall.modules.redshift_datasets.aws.redshift_serverless import redshift_serverless_client
+from dataall.modules.redshift_datasets.services.redshift_enums import RedshiftConnectionTypes
```
noah-paige, line 10: nit: remove unused import

backend/dataall/modules/redshift_datasets/services/redshift_connection_service.py
```diff
@@ -57,6 +57,11 @@ def create_redshift_connection(uri, admin_group, data: dict) -> RedshiftConnecti
             RedshiftConnectionService._check_redshift_connection(
                 account_id=environment.AwsAccountId, region=environment.region, connection=connection
             )
+            if connection.redshiftType == RedshiftType.Cluster.value:
```
noah-paige, line 60: nit: for provisioned redshift cluster connections we call boto3 api `describe_cluster()` 3 times:
- _check_redshift_connection
- get_cluster_namespaceId
- _get_redshift_encryption

Maybe can optimize - but also the service methods support both serverless and cluster connections so may be a bit less straightforward

noah-paige (approved):
tested and able to select connection type and shows on FE properly

left optional comments - approving as is

noah-paige:
@dlpzx can you provide a list of tests you ahve performed on this PR

For me I tried testing the sharing of datasets but ran into some issues on share approval:

- For Redshift Serverless
    - `ERROR: You cannot share datashare with a namespace in a different account.`
    - Not sure why this is but I do see a step in [AWS Docs](https://docs.aws.amazon.com/redshift/latest/dg/producer-cluster-admin.html) first granting share access to AWS Account (which I am not sure we have that step in our process share approach)
    - I may also be very off base on the above but curious how you have been testing b/c possibly my set up is wrong

- For Redshift Cluster
     - `ERROR: CREATE DATASHARE is not enabled.`
     - After some debugging I believe this is because I need a ra3 type instance for my redshift cluster but could you confirm / let's make sure to make note of this in some form of documentation

PR #1452, opened 2024-08-05
pass ShareableType instead of it's value and log exception details

petrkalos:
### Feature or Bugfix
- Bugfix

### Detail
in update_share_item_status_batch we pass `processor.type.value` but  `processor.type` is expected

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/shares_base/services/sharing_service.py
+++ b/backend/dataall/modules/shares_base/services/sharing_service.py
@@ -112,26 +112,26 @@ class SharingService:
                             if not success:
                                 share_successful = False
                         except Exception as e:
-                            log.error(f'Error occurred during sharing of {type.value}: {e}')
+                            log.exception(f'Error occurred during sharing of {type.value}')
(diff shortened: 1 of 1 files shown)


backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -112,26 +112,26 @@ def approve_share(cls, engine: Engine, share_uri: str) -> bool:
...
                                 new_status=ShareItemStatus.Share_Failed.value,
-                                share_item_type=processor.type.value,
+                                share_item_type=processor.type,
                             )
                             share_successful = False
                 return share_successful
 
             except Exception as e:
-                log.error(f'Error occurred during share approval: {e}')
+                log.exception('Error occurred during share approval')
```
dlpzx, line 134: Why are we removing the logging of the exception message? is it included in log.exception?

petrkalos: Yes log.exception is printing both the stack trace and the message

PR #1458, opened 2024-08-06
Redshift data sharing - frontend changes in the Catalog - clean

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Complete design in https://github.com/data-dot-all/dataall/issues/955. This particular PR is part of the redshift dataset sharing implementation. (It is easier to see the snapshots)
- To be able to share redshift items this PR enables the request access button for redshift tables and datasets
- IMPORTANT: New request modal for Redshift Datasets that lists namespaces with connections that are of type ADMIN (we cannot open a share request for namespaces that are not "linked" to data.all). See a screenshot of this request below:
- It adds icons in the Catalog search view for each item
- It adds the type of item right below the Catalog item
- It standardizes the icons used in the dataset list and in the Catalog with the latests AWS public icons
- It fixes the principal type for principals that are not groups or consumption roles
- It moves the consumption data details outside of the share view as a child component and renders it conditionally only if the dataset type is S3.

Screenshot 2024-08-05 at 16 14 26
Screenshot 2024-08-05 at 17 45 15

### Key decisions
- Separate request modal for redshift - we could have re-used the request access modal, but it contains several elements that are very particular to S3 (e.g. consumption roles, checking that the policy is attached). So I decided to keep it separate. This way future changes to any of the sharing methods is independent from each other. 

### Relates
- #955 

### Security

  - Is the input sanitized?


frontend/src/modules/Catalog/components/GlossarySearchResultItem.js
```diff
@@ -320,7 +370,14 @@ export const GlossarySearchResultItem = ({ hit }) => {
               onClick={() =>
                 hit.resourceKind === 'dashboard'
                   ? handleRequestDashboardAccessModalOpen()
-                  : handleRequestAccessModalOpen()
```
SofiaSazonova, line 323: same approach

```
modal_open_func = {
'dashboard': handleRequestDashboardAccessModalOpen,
'dataset' : handleRequestAccessModalOpen,
...
}

```
`onClick = {modal_open_func[hit.resourceKind]}`

SofiaSazonova (changes requested):
Some minor changes requested

frontend/src/modules/Catalog/components/GlossarySearchResultItem.js
dlpzx: This one was causing a lot of issues on rendering

SofiaSazonova (approved):

PR #1467, opened 2024-08-07
Redshift data sharing - Add sharing tasks to process Redshift datashares

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Complete design in https://github.com/data-dot-all/dataall/issues/955. This particular PR is the CORE of the redshift dataset sharing implementation.

- Implement sharing logic in ECS task for approve, revoke and verify. ❗check the "Sharing with Redshift" section of the design with the key decisions on the sharing workflow
- Add the necessary redshift data API calls in the redshift_data client
- Move share alarm utils to shares_base so that they can be re-used in redshift sharing. It would be good to rename the file but it can wait.
- Includes tests for the processor functions: approve, revoke, verify

In contrast to the design in the Glue or S3 sharing mechanisms, in this case I decided to keep it simple and use the AWS client directly from the processor without a manager.

❗ I did not find a way to check permissions granted to redshift roles in Redshift. For this reason in the verification task we are not checking the last 2 steps of the share. In Redshift it is possible to check user permissions to tables (with [has_table_permissions](https://docs.aws.amazon.com/redshift/latest/dg/r_HAS_TABLE_PRIVILEGE.html)) and role permissions to datashares, databases and schemas woth some of the [info tables and views](https://docs.aws.amazon.com/redshift/latest/dg/cm_chap_system-tables.html); but when it comes to tables there is not a table to look up or a system function. For the moment I have not included this step, but I'll be meeting more Redshift experts for guidance. The "good" thing is that it is the last step for a share to succeed, so it will be a matter of users trying and getting an "Access denied for insufficient permissions", which can be troubleshooted

There are still a number of issues to be fixed on subsequent PRs:
- add guardrails to share creation
- polish FE (e.g. principal id, resource type)
- avoid IAM checks and dataset and IAM locks for Redshift

### Relates
- #955 

### Security

  - Is the input sanitized?


dlpzx:
Testing
## Approve share
- [ ] Approve share request with one single table - access the table from a user with that redshift role
- [ ] Approve a second table
- [ ] Approve 2 tables at once

## Revoke share
Create 3 share requests (between the same namespaces)
- R1: dataset1-table1-redshiftrole1
- R2: dataset1-table1-redshiftrole2
- R3: dataset2-table2-redshiftrole1

- [ ] Revoke R3 - table2 removed from datashare
- [ ] Revoke R2 - redshiftrole2 loses permissions to datashare/local_db/external_schema
- [ ] Revoke R1 - datashare is deleted


## Verify share
In an existing share request that is successful
- [ ] Delete datashare
- [ ] Remove schema from datashare
- [ ] Remove table from datashare
- [ ] Drop local_db
- [ ] Drop external_schema
- [ ] Revoke permissions from local_db
- [ ] Revoke permissions from external_schema

(after each test, reapply and ensure the share is healthy again)

backend/dataall/modules/redshift_datasets_shares/db/redshift_share_object_repositories.py
```diff
@@ -34,7 +34,7 @@ def count_redshift_principal_shares(session, environment_uri: str, group_uri: st
         share_item_shared_states = ShareStatusRepository.get_share_item_shared_states()
         return (
             session.query(ShareObject)
-            .join(
+            .outerjoin(
```
noah-paige, line 37: So are we using `outerjoin()` here not to include Shares with no Share Object Items? Would that work considering we are filtering by Share object Items type of `RedshiftTable` and for status in one of the shared states?

Also, looking at where we execute `count_redshift_principal_shares` --> i.e. where we call `EnvironmentResourceManager.count_group_resources(...)` it is for both `delete_env` and `remove_group` on env

While it is okay if we do not count share objects with no shared items for `deleve_env` because we execute `EnvironmentResourceManager.delete_env(...)` and clean up shares in other states afterwards - we do not do the same for `remove_group`

Meaning I believe there would be orphaned share object resources when removing an env group in that case

backend/dataall/modules/shares_base/db/share_object_state_machines.py
```diff
@@ -203,6 +203,7 @@ def __init__(self, state):
                     ShareItemStatus.Revoke_Failed.value: [
                         ShareItemStatus.Revoke_In_Progress.value,
                         ShareItemStatus.Revoke_Approved.value,
+                        ShareItemStatus.Revoke_Succeeded.value,
```
noah-paige, line 206: This is allowing an item that is in `Revoke_Succeeded` to transition to `Revoke_Failed` on `ShareItemActions.Failure` correct?

Should that be the case? I would think `Revoke_Failed` be a state that is unreachable from `Revoke_Succeeded`?

backend/dataall/modules/shares_base/services/share_object_service.py
```diff
@@ -220,11 +227,13 @@ def submit_share_object(cls, uri: str):
...
-                    action='Submit Share Object',
-                    message=f'The principal role {share.principalRoleName} is not found.',
-                )
+            if share.principalType in [PrincipalType.ConsumptionRole.value, PrincipalType.Group.value]:
+                # TODO make it generic to non IAM role principals
+                if not ShareObjectService.verify_principal_role(session, share):
+                    raise PrincipalRoleNotFound(
+                        action='Submit Share Object',
+                        message=f'The principal role {share.principalRoleName} is not found.',
+                    )
```
noah-paige, line 236: nit: move the line `if share.principalType in [PrincipalType.ConsumptionRole.value, PrincipalType.Group.value]:` and `return share.principalRoleName` if principalType is not one of the specified ones to verify

Instead of adding this condition to wrap around `if not ShareObjectService.verify_principal_role(session, share):` everywhere we call it

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -18,11 +33,477 @@ def __init__(self, session, share_data, shareable_items, reapply=False):
         self.tables: List[RedshiftTable] = shareable_items
         self.reapply: bool = reapply
 
+        dataset_connection = RedshiftConnectionRepository.get_redshift_connection(
+            self.session, self.dataset.connectionUri
+        )
+        self.source_connection = dataset_connection
```
noah-paige, line 39: nit:  no need for `dataset_connection` variable
```
self.source_connection = RedshiftConnectionRepository.get_redshift_connection(
    self.session, self.dataset.connectionUri
)
```

backend/dataall/modules/redshift_datasets_shares/aws/redshift_data.py
```diff
@@ -0,0 +1,408 @@
...
+    def _execute_statement(self, sql: str):
+        log.info(f'Executing {sql=} with connection {self.execute_connection_params}...')
+        execute_dict = self.execute_connection_params
+        execute_dict['Sql'] = sql
+        execute_statement_response = self.client.execute_statement(**execute_dict)
+
+        execution_finished = False
+        describe_statement_response = None
+        while not execution_finished:
+            describe_statement_response = self.client.describe_statement(Id=execute_statement_response['Id'])
```
noah-paige, line 38: nit: can use walrus operator here

```
execution_pending = True
while execution_pending := self.client.describe_statement(Id=execute_statement_response['Id']):
   ....
```

Also

backend/dataall/modules/redshift_datasets_shares/aws/redshift_data.py
```diff
@@ -0,0 +1,408 @@
...
+        try:
+            log.info(f'Checking {table_name=} in {datashare=}...')
+            sql_statement = f'DESC DATASHARE {RedshiftShareDataClient.parsed_name(datashare)};'
+            records = self._execute_statement_return_records(sql=sql_statement)
+            tables_in_datashare = [
+                [d for d in record][5]['stringValue'].split('.')[-1]  # "schemaname.tablename"
+                for record in records
+                if [d for d in record][4]['stringValue'] in ['table']
+            ]
+            log.info(f'Found {tables_in_datashare=}')
```
noah-paige, line 181: can you give an example of records to better understand this list comprehension we are doing? Do we need to use index based retrieval?

backend/dataall/modules/redshift_datasets_shares/aws/redshift_data.py
```diff
@@ -0,0 +1,408 @@
...
+            filtered_records = [[d for d in record if d.get('stringValue', False)] for record in records]
+            log.info(f'Returning {len(filtered_records)} records from executed statement')
+            return filtered_records
+        except Exception as e:
+            log.error(f'Failed to retrieve records for sql {id=}: {e}')
+            raise e
+
+    @staticmethod
+    def parsed_name(name: str) -> str:
+        return f'"{name}"'
```
noah-paige, line 69: nit: should this be named something other than parsed like `quoted_name` ? Also should we have an equivalent function for names that must be single quoted as well?

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -18,11 +33,477 @@ def __init__(self, session, share_data, shareable_items, reapply=False):
...
+                    revoked_item_SM.update_state(self.session, self.share.shareUri, new_state)
+                else:
+                    for table in self.tables:
+                        share_item = ShareObjectRepository.find_sharable_item(
+                            self.session, self.share.shareUri, table.rsTableUri
+                        )
+                        ShareStatusRepository.update_share_item_health_status(
+                            self.session, share_item, ShareItemHealthStatus.Unhealthy.value, str(e), datetime.now()
+                        )
+                return False
```
noah-paige, line 382: I see this block which is why I imagine we need to be able to transition from `Revoke_Succeeded` to `Revoke_Failed` if cannot clean up the DB / Datashare

I would think either (1) only call `update_state` once at the end of the entire revoke/ clean up rather than calling once after table clean up and again after datashare clean up OR (2) we do not update to `Revoke_Failed` of the non-table resources -- but this could leave some orphaned resources / not sure how it affects re-sharing same table to same target namespace later on

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -18,11 +33,477 @@ def __init__(self, session, share_data, shareable_items, reapply=False):
...
+            redshift_client_in_source = redshift_share_data_client(
+                account_id=self.share_data.source_environment.AwsAccountId,
+                region=self.share_data.source_environment.region,
+                connection=self.source_connection,
+            )
+            redshift_client_in_target = redshift_share_data_client(
+                account_id=self.share_data.target_environment.AwsAccountId,
+                region=self.share_data.target_environment.region,
+                connection=self.target_connection,
+            )
```
noah-paige, line 415: nit: can move init of source and target clients to class `init()` if is called on every method?

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -18,11 +33,477 @@ def __init__(self, session, share_data, shareable_items, reapply=False):
...
+                    datashare=self.datashare_name, schema=self.dataset.schema
+                ):
+                    ds_level_errors.append(
+                        ShareErrorFormatter.dne_error_msg(
+                            'Redshift schema added to datashare',
+                            f'datashare_name={self.datashare_name}, schema={self.dataset.schema}',
+                        )
+                    )
+                # 3) (in source namespace) Check the access is granted to the consumer cluster to the datashare
+                if not redshift_client_in_source.check_consumer_permissions_to_datashare(datashare=self.datashare_name):
```
noah-paige, line 433: should this be called from consumer (i.e. target) to see if datashare is accessible from consuemr side? Otherwise is this not similar to `check_datashare_exists` ?

noah-paige:
A couple other notes:
- (1) Will have to resolve conflicts that arose after merging granular shares PR
-  (2) Have we tested for cross account sharing? I was trying to set up with redshift serverless and I get an error `'You cannot share datashare with a namespace in a different account.'`
- (3) Call out [as per docs](https://docs.aws.amazon.com/redshift/latest/dg/considerations.html) for provisioned clusters that for CREATE DATASHARE - `'Data sharing is supported for all provisioned ra3 cluster types (ra3.16xlarge, ra3.4xlarge, and ra3.xlplus) and Amazon Redshift Serverless. It isn't supported for other cluster types.'`

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
dlpzx: Fixed!

backend/dataall/modules/redshift_datasets_shares/aws/redshift_data.py
dlpzx: A little different, but implemented :)

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
dlpzx: Yes, yes yes. Thank you!

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
dlpzx: local_db and external_schema absolutely yes.  As for the redshift_clients, I am not so sure. I do not want to introduce tightly coupled classes in our code, at least not in the constructor; the risk is to end up having the situation of the lf_share_managers. I can introduce a function to reduce the code duplication, let me know if that is ok

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
dlpzx: That was my dilemma. (1) is doable but trickier for the checks on the existence of other tables; I can look into the proper query that would be needed. (2) that was my thought, it might look as successful but in practice there might be leftovers. When it comes to re-sharing I made sure that it is possible even with orphan resources, but my concern is more on the orphan resources. a datashare shared with another cluster gives a lot of access rights.

In short, I will look at option (1) again and try to come up with the ultimate good solution


**by the way this issue is also present for Lake Formation glue sharing

backend/dataall/modules/shares_base/db/share_object_state_machines.py
dlpzx: See comment below!

backend/dataall/modules/redshift_datasets_shares/aws/redshift_data.py
dlpzx: Done! We don't but maybe we should, there are a number of places that require single quotes. I'll add that

backend/dataall/modules/redshift_datasets_shares/aws/redshift_data.py
dlpzx: It is a bit different from typical API calls, The API call is the [get_statement_result](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/redshift-data/client/get_statement_result.html) which returns:
```
{
    'ColumnMetadata': ....,
    'NextToken': 'string',
    'Records': [
        [
            {
                'blobValue': b'bytes',
                'booleanValue': True|False,
                'doubleValue': 123.0,
                'isNull': True|False,
                'longValue': 123,
                'stringValue': 'string'
            },
        ],
    ],
    'TotalNumRows': 123
}
```

dlpzx: In:
```
tables_in_datashare = [
                [d for d in record][5]['stringValue'].split('.')[-1]  # "schemaname.tablename"
                for record in records
                if [d for d in record][4]['stringValue'] in ['table']
            ]
```
We are processing a response of the type:
```
records = [
    [
        {
            'longValue': 1111111111111,
        }, #producer_account
        {
            'longValue': 12345689-1940u23085u2309,
        }, #producer_namespace
        {
            'stringValue': 'OUTBOUND'
        }, #share_type
        {
            'stringValue': 'string'
        }, #share_name
        {
            'stringValue': 'TABLE'
        }, #object_type
        {
            'stringValue': 'schema1.table_name_1'
        },#object_name
        {
            'booleanValue': False
        },#include_new
    ],  #Record1
]
```

dlpzx: In cases where we are doing a direct SELECT a, b, c from table, we have more control over the the response fields, but for the standard SHOW DATASHARES, DESC DATASHARE etc, the fields are already provided

backend/dataall/modules/shares_base/services/share_object_service.py
dlpzx: I am redoing all this piece of code in the last PR of this series, so I will take it into account in that PR, it already looks completely different with the interface to inject validation of create share object

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
dlpzx: I think the latest changes are what we are looking for. But I want your opinion on point 4) I think it is logical, if a revoke has failed, the clean-up does not happen until every table is revoked with 1, 2, 3

```
        For each table:
            Update table status with Start Action (Revoke_Approved ---> Revoke_In_Progress)
            try:
                1) 2) 3) 
            except:
                Update table status with Failure Action (Revoke_In_Progress ---> Revoke_Failed)
        If the previous is successful, we proceed to clean-up shared resources across datashares:
        try:
            4) (in target namespace) If no more tables are shared with the redshift role, revoke usage access to the external schema to the redshift role ----> IF A TABLE IS IN REVOKE_FAILED, IT WON'T CLEAN-UP THE SHARE
            5) 6) 7)
            Update NON-FAILED tables with Success Action (Revoke_In_Progress ---> Revoke_Succeeded)
        except:
            Update tables with Failure Action (Revoke_In_Progress ---> Revoke_Failed)
```

dlpzx:
> A couple other notes:
> 
>     * (1) Will have to resolve conflicts that arose after merging granular shares PR
> 
>     * (2) Have we tested for cross account sharing? I was trying to set up with redshift serverless and I get an error `'You cannot share datashare with a namespace in a different account.'`
> 
>     * (3) Call out [as per docs](https://docs.aws.amazon.com/redshift/latest/dg/considerations.html) for provisioned clusters that for CREATE DATASHARE - `'Data sharing is supported for all provisioned ra3 cluster types (ra3.16xlarge, ra3.4xlarge, and ra3.xlplus) and Amazon Redshift Serverless. It isn't supported for other cluster types.'`

- (1) All merged! - there are 2 issues in the merge/PR actions: one related to [an npm vulnerability found 3 hours ago](https://github.com/advisories/GHSA-8hc4-vh64-cxmj) in `axios`. There is no patched version yet, we need to figure out how to handle it for the moment. The other issue is on the migration scripts, there is a PR to fix it https://github.com/data-dot-all/dataall/pull/1478
- (2) Wrong assumption from my side. This testing is my #1 prio
- (3) this should be part of the add guardrails/validator PR. I added the docs in the design.To be honest I have not made up my mind of whether we should include absolutely all limitations in the guardrails. But we can discuss it in that PR

backend/dataall/modules/redshift_datasets_shares/aws/redshift_data.py
noah-paige: makes sense - seems like not much we can do here given API response

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -357,20 +398,32 @@ def process_revoked_shares(self) -> bool:
...
-                    f'with target {self.target_connection.name} in namespace {self.target_connection.nameSpaceId}'
+                    f'from source {self.source_connection.name} in namespace {self.source_connection.nameSpaceId} '
+                    f'with target {self.target_connection.name} in namespace {self.target_connection.nameSpaceId} '
                     f'due to: {e}'
                 )
                 if not self.reapply:
-                    new_state = revoked_item_SM.run_transition(ShareItemActions.Failure.value)
-                    revoked_item_SM.update_state(self.session, self.share.shareUri, new_state)
+                    all_failed_state = revoked_item_SM.run_transition(ShareItemActions.Failure.value)
+                    revoked_item_SM.update_state(self.session, self.share.shareUri, all_failed_state)
```
noah-paige, line 426: Not possible to have `self.reapply` be `True` for revoking shares items

Think we can safely always execute L425 + 426

backend/dataall/modules/redshift_datasets_shares/cdk/pivot_role_redshift_data_sharing_policy.py
```diff
@@ -10,12 +15,35 @@ class RedshiftDataSharingPivotRole(PivotRoleStatementSet):
...
-        return statements
+        additional_statements = []
+        engine = db.get_engine(envname=os.environ.get('envname', 'local'))
+        with engine.scoped_session() as session:
+            connections = RedshiftConnectionRepository.list_environment_redshift_connections(
+                session, environment_uri=self.environmentUri
+            )
+            if connections:
+                source_datashare_arns = [
+                    f'arn:aws:redshift:{self.region}:{self.account}:datashare:{conn.nameSpaceId}/*'
```
noah-paige, line 38: do we also need for `redshift-serverless`?

noah-paige: nevermind - seems datashares always use `redshift` and from testing look like this works as is

noah-paige:
Tested:

Serverless:
- [ ] Cross Account share approval (first time + query)
- [ ] Cross Account share verify 
- [ ] Cross Account share approval (add new item + query)
- [ ] Cross Account share revoke (one item + query other still) 
- [ ] Cross Account share reapply
- [ ] Cross Account share revoke all remaining items (delete resources)

Provisioned Cluster
 - [ ] Cross Account share approval (+ query)
- [ ] Cross Account share verify 
- [ ] Cross Account share revoke all

NOTE: Let's enforce all lowercase for connection names, can use `name = Column(String, nullable=False, default=slugifier('label'))` similar to how we do for data models that inherit `Resource` class... otherwise approval works one time but reapply, verify, revoke, and approval re-using same database all fail ( just calling out here but I imagine this is part of the guardrails PR)

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -338,38 +338,38 @@ def reapply_share(cls, engine: Engine, share_uri: str) -> bool:
...
-                            if not success:
-                                reapply_successful = False
-                        except Exception as e:
-                            log.error(f'Error occurred during share reapplying of {type.value}: {e}')
+                log.info(f'Verifying principal IAM Role {share_data.share.principalRoleName}')
+                # TODO make it generic to non IAM role principals
+                reapply_successful = ShareObjectService.verify_principal_role(session, share_data.share)
+                if not reapply_successful:
+                    log.error(f'Failed to get Principal IAM Role {share_data.share.principalRoleName}, exiting...')
+                    return False
```
noah-paige, line 346: some reason the skip `verify_principal` logic got reverted back to the old way for `reapply_share()`

To fix back to skip on RedshiftRole Type - otherwise cannot reapply

dlpzx: I did not want to interfere with this logic, reverted back to the quick principaltype check

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
dlpzx: Modified

dlpzx:
Hi @noah-paige thanks for the extensive testing.

I just changed the connection label/name treatment to use the default slugify for names:
image

As for the cross-account provisioned cluster datasharing errors, I'll keep posting here my findings:

The datashare is associated with the target account (you can see it in the generic datashares tab), but for some reason it associates the datashare with an empty namespace:
image

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -349,14 +356,12 @@ def process_revoked_shares(self) -> bool:
...
-                    else:
-                        ShareStatusRepository.update_share_item_health_status(
-                            self.session, share_item, ShareItemHealthStatus.Unhealthy.value, str(e), datetime.now()
-                        )
+                    failed_table_SM = ShareItemSM(started_state)
+                    failed_state = failed_table_SM.run_transition(ShareItemActions.Failure.value)
+                    failed_table_SM.update_state_single_item(self.session, share_item, failed_state)
+                    ShareStatusRepository.update_share_item_health_status(
+                        self.session, share_item, ShareItemHealthStatus.Unhealthy.value, str(e), datetime.now()
+                    )
```
noah-paige, line 364: should we set to UnHealthy if revoke fails? We do not do so for other revokes but I think makes sense

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -138,14 +145,14 @@ def process_approved_shares(self) -> bool:
...
                         region=self.share_data.target_environment.region,
                     )
-                    redshift_type = (
-                        'redshift-serverless'
+                    consumer_arn = (
+                        f'arn:aws:redshift-serverless:{self.share_data.target_environment.region}:{self.share_data.target_environment.AwsAccountId}:namespace/{self.target_connection.nameSpaceId}'
                         if self.target_connection.redshiftType == RedshiftType.Serverless.value
-                        else 'redshift'
+                        else f'arn:aws:redshift:{self.share_data.target_environment.region}:{self.share_data.target_environment.AwsAccountId}:namespace:{self.target_connection.nameSpaceId}'
                     )
```
noah-paige, line 152: do the same for verify on L512:516

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
dlpzx: I changed it while implementing :) If we were able to rollback all the steps before failure then we could leave it as healthy, but given that in the revoke we will break the share we can change it to unhealthy. I'll do

noah-paige (approved):
tested with latest changes - thanks for all the work here @dlpzx PR looks great

approving!

Issue #1468, opened 2024-08-07
Submit request on shares should redirect to share page


anushka-singh:
**Is your idea related to a problem? Please describe.**
Share request should redirect to share page and not catalog page

noah-paige:
This Issue was complete as of PR #1469 - thank you for your contribution @anushka-singh

PR #1469, opened 2024-08-07
Issue1468: Submit request redirect

anushka-singh:
### Feature or Bugfix
- Feature

### Detail
- Submit request redirects to share request page instead of catalog page

### Relates
- https://github.com/data-dot-all/dataall/issues/1468

### Testing
- Submit request button redirects to share page and not catalog anymore. 

### Security

  - Is the input sanitized?

### DIFF
--- a/frontend/src/modules/Shared/Shares/ShareEditForm.js
+++ b/frontend/src/modules/Shared/Shares/ShareEditForm.js
@@ -278,9 +278,9 @@ export const ShareEditForm = (props) => {
       onApply();
     }
     const targetPath = `/console/shares/${share.shareUri}`;
-      if (location.pathname !== targetPath) {
-        navigate(targetPath);
-      }
+    if (location.pathname !== targetPath) {
+      navigate(targetPath);
+    }
   };
 
   const draftRequest = async () => {
--- a/frontend/src/modules/Shared/Shares/ShareEditForm.js
+++ b/frontend/src/modules/Shared/Shares/ShareEditForm.js
@@ -276,11 +276,11 @@ export const ShareEditForm = (props) => {
 
     if (onApply) {
       onApply();
-      const targetPath = `/console/shares/${share.shareUri}`;
+    }
+    const targetPath = `/console/shares/${share.shareUri}`;
       if (location.pathname !== targetPath) {
         navigate(targetPath);
       }
-    }
   };
 
   const draftRequest = async () => {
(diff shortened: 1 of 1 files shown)


frontend/src/modules/Shared/Shares/ShareEditForm.js
```diff
@@ -268,6 +268,7 @@ export const ShareEditForm = (props) => {
         },
         variant: 'success'
       });
+      navigate(`/console/shares/${share.shareUri}`);
```
SofiaSazonova, line 271: On line 278 onApply function is called. 
If we redirect before it's called, than we break the logic.
Currently there is nothing very important there, but later we may want to add some logic in onApply.

SofiaSazonova (changes requested):
Left a comment about `onApply`

frontend/src/modules/Shared/Shares/ShareEditForm.js
zsaltys: @SofiaSazonova what should we do instead?

anushka-singh: @SofiaSazonova Are you suggesting I add `navigate(`/console/shares/${share.shareUri}`);` on line 279 after `onApply();` instead of on line 271?

noah-paige: @anushka-singh - yes I think just move the `navigate(...)` after calling `onApply();`

Also I imagine this works when submitting share from console - but what about when editing a share from `ShareView` form (we re-use the same `ShareEditForm`)

Maybe we can use `const location = useLocation();` and only redirect when not already on ShareView?

SofiaSazonova: Yes, conditional redirect is a good idea.

anushka-singh: Added conditional redirect. 
@noah-paige @SofiaSazonova Please take a look.

frontend/src/modules/Shared/Shares/ShareEditForm.js
```diff
@@ -275,6 +276,10 @@ export const ShareEditForm = (props) => {
 
     if (onApply) {
       onApply();
+      const targetPath = `/console/shares/${share.shareUri}`;
+      if (location.pathname !== targetPath) {
+        navigate(targetPath);
+      }
```
noah-paige, line 282: can we move L279-282 outside of the `if (onApply) { ...}` block

even if there is not a defined function `onApply` (coming from the parent component) I imagine we would want to redirect the user to the share view

anushka-singh: Okay that makes sense.

anushka-singh: @noah-paige Please review

noah-paige (approved):
looks good + tested redirect behavior - approving

Issue #1332, opened 2024-06-14
Support WRITE access for consumer roles


zsaltys:
**Is your idea related to a problem? Please describe.**
Currently data.all only grants READ only access to consumer IAM roles. However organizations like ours need to manage WRITE access as well to define which roles can write to which S3 buckets or which databases. Otherwise we end up managing read only access with data.all and all write access outside of data.all. We would like to unify and manage both read and write access via data.all.

**Describe the solution you'd like**
We need to think this through in context of:

- S3 bucket sharing
- Access points
- LakeFormation
- Incoming RedShift integration

**S3 bucket sharing**
Overall a very simple change. The IAM role just needs to be granted PUTObject permissions + KMS Encrypt permissions on the key etc. I would not want to grant DELETE or anything else besides PUT? Perhaps this can be configurable on config what permissions to grant so organizations can decide. Those extra permissions be granted by the user themselves on the IAM role using IaC when needed.

**Access points**
I don't know too much about these but I suspect they work similarly to S3 bucket sharing. Hope the team can clarify on this ticket.

**LakeFormation**
The question is what permissions we should grant here. We could limit only to just basic to add new partitions (don't know which permission controls that atm) which is what most of the writing roles will ever need. Or we can also grant things like DROP, INSERT, DELETE, CREATE. I think to be most useful to everyone WRITE should grant full write access with all the other permissions that are not currently granted. Though this could be inconsistent with S3 permissions if for example we only grant PUT on S3 but grant DELETE on the DB? Perhaps this could also be configurable but the default should be consistent for all.

**RedShift**
Don't know enough on this to comment.

My proposal would be that WRITE access is defined when creating a share. We don't want to do it when registering a consumer role because consumer role could also be used for both write access on same account and read access on other accounts (ex EMR role). We can either define WRITE access per share item or for the entire share. I think defining it for the entire SHARE makes sense. Default should be READ ONLY. Write access should only be selectable (and validated by backend) **if the consumer role and dataset belong to the same environment.** The only problem I see is currently share items let you select specific tables so we could grant write access to them. But how do we grant CREATE access on the DB? Do we just do it implicitly which can confuse the user.. What if he doesn't select any tables should we still grant WRITE access to the DB? The only way I can think of solving this is that there must be a new share item for the DATABASE itself.

We also must make sure the share validator / health checker is made aware of the new extra permissions.

anmolsgandhi:
[...]

PR #1407, opened 2024-07-11
Introduce check for IAM actions in share_verify bucket and access points + reapply with list of allowed actions

SofiaSazonova:
### Feature or Bugfix

- Bugfix

### Detail
- Share verifier now detects "s3:*" permission as an error 
- Share reapplied with enforced valid actions

### Relates
- <URL or Ticket>

### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)?
  - Is the input sanitized?
  - What precautions are you taking before deserializing the data you consume?
  - Is injection prevented by parametrizing queries?
  - Have you ensured no `eval` or similar functions are used?
- Does this PR introduce any functionality or component that requires authorization?
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms?
  - Are you logging failed auth attempts?
- Are you using or adding any cryptographic features?
  - Do you use a standard proven implementations?
  - Are the used keys controlled by the customer? Where are they stored?
- Are you introducing any new policies/roles/users?
  - Have you used the least-privilege principle? How?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1472, opened 2024-08-08
S3Bucket WRITE/MODIFY permissions

petrkalos:
### Feature or Bugfix
- Feature

### Detail
Adding support for requesting WRITE/MODIFY permissions for an S3Bucket

### Testing
Tested in local and dev delpoyment

### Relates
#1332 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/s3_datasets_shares/services/share_managers/lf_share_manager.py
+++ b/backend/dataall/modules/s3_datasets_shares/services/share_managers/lf_share_manager.py
@@ -1,29 +1,69 @@
 import logging
 import time
 from datetime import datetime
-from dataall.core.environment.services.environment_service import EnvironmentService
-from dataall.modules.s3_datasets_shares.aws.glue_client import GlueClient
-from dataall.modules.s3_datasets_shares.aws.lakeformation_client import LakeFormationClient
-from dataall.base.aws.quicksight import QuicksightClient
+from enum import Enum, auto
+from typing import List
+
 from dataall.base.aws.iam import IAM
+from dataall.base.aws.quicksight import QuicksightClient
 from dataall.base.aws.sts import SessionHelper
 from dataall.base.db import exceptions
+from dataall.core.environment.services.environment_service import EnvironmentService
+from dataall.modules.s3_datasets.db.dataset_models import DatasetTable
+from dataall.modules.s3_datasets_shares.aws.glue_client import GlueClient
+from dataall.modules.s3_datasets_shares.aws.lakeformation_client import LakeFormationClient
+from dataall.modules.s3_datasets_shares.services.s3_share_alarm_service import S3ShareAlarmService
+from dataall.modules.shares_base.db.share_object_models import ShareObjectItem, ShareObjectItemDataFilter
 from dataall.modules.shares_base.db.share_object_repositories import ShareObjectRepository
 from dataall.modules.shares_base.db.share_object_state_machines import ShareItemSM
 from dataall.modules.shares_base.db.share_state_machines_repositories import ShareStatusRepository
+from dataall.modules.shares_base.services.share_manager_utils import ShareErrorFormatter
 from dataall.modules.shares_base.services.shares_enums import (
     ShareItemStatus,
     ShareObjectActions,
     ShareItemActions,
     ShareItemHealthStatus,
+    ShareObjectDataPermission,
 )
-from dataall.modules.s3_datasets.db.dataset_models import DatasetTable
-from dataall.modules.s3_datasets_shares.services.s3_share_alarm_service import S3ShareAlarmService
-from dataall.modules.shares_base.db.share_object_models import ShareObjectItem, ShareObjectItemDataFilter
-from dataall.modules.shares_base.services.share_manager_utils import ShareErrorFormatter
 from dataall.modules.shares_base.services.sharing_service import ShareData
 
 logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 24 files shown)


petrkalos:
One change that I made from the demo video in #1332 is to switch the checkboxes to a multiselect text field for consistency
image

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    DATAALL_ACCESS_POINT_KMS_ENCRYPT_SID,
+)
+from dataall.modules.s3_datasets_shares.aws.s3_client import (
+    DATAALL_READ_ONLY_SID,
+    DATAALL_WRITE_ONLY_SID,
+    DATAALL_MODIFY_ONLY_SID,
+)
+from dataall.modules.shares_base.services.shares_enums import ShareObjectDataPermission
+
+SID_TO_ACTIONS = {
```
SofiaSazonova, line 18: It's really cool to have mapping
The only problem is that s3 actions are still duplicated here https://github.com/data-dot-all/dataall/pull/1407
So, we may have problems with share verification

SofiaSazonova (changes requested):

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_utils.py
petrkalos: That's a great catch Sofia, indeed it seems that we have 2 places (IAM policies and S3 policies) we set the permissions and I missed the first. Fortunately the first are being used only for cross-account sharing while the later for intra account. The initial requirement was to implement WRITE only for intra-account and as such we are good to go. The only piece that is missing is to add guardrails around this rule which I will implement and push soon.

SofiaSazonova:
Looks good. Are there going to be any integration tests?

SofiaSazonova:
I'll wait for all combinations, as you announced)

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -568,33 +575,31 @@ def update_dataset_bucket_key_policy(self):
...
-                )
-            else:
-                logger.info(
-                    f'KMS key does not contain share statement {DATAALL_ACCESS_POINT_KMS_DECRYPT_SID}, '
-                    f'generating a new one'
-                )
-                statements[DATAALL_ACCESS_POINT_KMS_DECRYPT_SID] = self.generate_default_kms_decrypt_policy_statement(
-                    target_requester_arn
-                )
+            for target_sid in perms_to_sids(self.share.permissions, SidType.KmsAccessPointPolicy):
```
noah-paige, line 578: do we not have to add logic for revoke (at `def delete_dataset_bucket_key_policy()`) similar to how we do for bucket share?

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -568,33 +575,31 @@ def update_dataset_bucket_key_policy(self):
...
+                if target_sid in statements.keys():
+                    logger.info(f'KMS key policy contains share statement {target_sid}, ' f'updating the current one')
+                    statements[target_sid] = add_target_arn_to_statement_principal(
+                        statements[target_sid], target_requester_arn
+                    )
+                else:
+                    logger.info(f'KMS key does not contain share statement {target_sid}, ' f'generating a new one')
+                    statements[DATAALL_ACCESS_POINT_KMS_DECRYPT_SID] = self.generate_default_kms_policy_statement(
```
noah-paige, line 586: should this be the same as with `s3_bucket_share_manager`? As in:
```
                    statements[target_sid] = self.generate_default_kms_policy_statement(
                        target_requester_arn, target_sid
                    )
```

noah-paige (changes requested):
left some comments and requested changes may be needed on access point manager/ folder sharing

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
petrkalos: good catch, will push fixes

noah-paige (approved):
changes look good! Tested the approve, verify, and revoke paths for shares in aws deployment - approving

backend/dataall/modules/s3_datasets_shares/services/share_managers/lf_share_manager.py
```diff
@@ -1,31 +1,71 @@
...
+        LfPermType.Filters: ['SELECT'],
+    },
+}
+
+
+def perms_to_lfperms(permissions: List[str], lf_perm_type: LfPermType) -> List[str]:
+    lfperms = list()
+    for p in permissions:
+        lfperms.extend(PERM_TO_LF_PERMS[p][lf_perm_type])
+    return list(dict.fromkeys(lfperms))
```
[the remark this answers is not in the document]
petrkalos: It is for dedup yes and this approach is used because a `set` doesn't preserve the order while `dict.fromkeys` does. Although order doesn't make any logical difference it does help with testing (i.e think `assert_called_once_with(['P1', 'P2])` etc)

Issue #1065, opened 2024-02-19
Implement metadata forms


zsaltys:
AWS DataZone has a feature called: **Metadata Forms**. Essentially it's a dynamic form creator that can be attached to datasets.
I think this is an excellent feature and it would help data.all solve some problems.

One problem is that at the moment we don't know what metadata users want to keep on their datasets. Currently data.all hardcodes Confidentiality and Topics. For our use case for example Topics make no sense and for Confidentiality we use our own levels.

I propose the following:

1) Create a new Play module Metadata Forms
2) Forms are owned by teams who created them
3) Forms should be visible to teams who created them but should also have an option to be shared with everyone
4) Forms should allow to require them to be used on all data.all datasets without exception. These would be very useful for organization wide metadata like confidentiality levels. This option should only be available to DA admins.
5) Datasets should allow to attach from 1 to N metadata forms.
6) Metadata forms should allow to create fields: strings, integers, booleans and most importantly using Glossary Terms as this allows to create specific dropdowns. Using glossary terms I can create a required form for my own confidentiality levels.
7) (Optional) Allow to specify required forms on Organization level or Environment level. This way an environment or organization can enforce certain metadata forms are always used on their datasets.
8) Form fields which were set as required by DAAdmin should be searchable as that way we can search by things like confidentiality level.
9) Remove confidentiality and topics dropdowns from datasets. Offer a migration path to users who already use it.
10) Refactor confidentiality with new checkboxes on datasets: 1) hide schema without having access etc.. Instead of hiding the meaning behind a confidentiality level make it explicit with checkboxes what is allowed on a dataset.

noah-paige:
Hi @zsaltys - I think the above idea is one that we would want to pick up for a future release of data.all as it would be very helpful for more dynamic metadata associations on data.all resources. I think this would be a good candidate to extend the `Discover` Section of the UI rather than the `Play` section which focuses more on data consumption use cases

Additionally, I would be curious to hear your thoughts on if you see a benefit of extending this Metadata Forms concept to more than just Datasets but also to other resources in data.all, such as Notebooks, Worksheets, etc. I think offering this type of flexibility everywhere could have value but curious to hear other opinions here

We will add this issue to our backlog for now and determine when we can prioritize

dlpzx:
For 2.6 we will work on the design, implementation will be carried out in 2.7

SofiaSazonova:
**Overview**
[...]

PR #1474, opened 2024-08-08
Metadata forms 4: Access Control

SofiaSazonova:
### Feature or Bugfix

- Feature

### Detail
- Restrict operations for non-Owners
- Display user role in MF list
- Hide edit buttons for non-Owners and non-Admins
- Test For unauth operations

### Relates
- #1065 

### Security

  - Is the input sanitized?


backend/dataall/modules/metadata_forms/services/metadata_form_service.py
```diff
@@ -37,6 +38,42 @@ def validate_create_field_params(data):
             raise exceptions.RequiredParameter('glossaryNodeUri')
 
 
+class MetadataFormAccessService:
```
noah-paige, line 41: Curious is there a reason to implement a new way of checking user permissions rather than use the existing `ResourcePolicyService` and `has_resource_permission` decorator?

backend/dataall/modules/metadata_forms/services/metadata_form_service.py
```diff
@@ -37,6 +38,42 @@ def validate_create_field_params(data):
...
+    def get_user_role(uri):
+        if MetadataFormAccessService.is_owner(uri):
+            return MetadataFormUserRoles.Owner.value
+        else:
+            return MetadataFormUserRoles.User.value
+
+
 class MetadataFormService:
     @staticmethod
     def create_metadata_form(data):
```
noah-paige, line 79: should creating metadata forms be able to be restricted via a TenantPermission enforced by DA Admins?

frontend/src/modules/Metadata_Forms/components/MetadataFormFields.js
```diff
@@ -261,16 +261,19 @@ const DisplayTable = (props) => {
...
-              startIcon={<PencilAltIcon size={15} />}
-              sx={{ mt: 1 }}
-              onClick={startEdit}
-              type="button"
-              variant="outlined"
-            >
-              Edit
-            </Button>
+            {(userRole === userRolesMF.Owner ||
+              userRole === userRolesMF.Admin) && (
```
noah-paige, line 265: what would be difference between Owner and Admin in this case? I see currently we are not resolving any user role as Admin but maybe you intend to in a future PR?

backend/dataall/modules/metadata_forms/services/metadata_form_service.py
SofiaSazonova: When MF is created we don't make an entry in resource_permission_policy table. So, we can't check it there.
On the other hand, MF access rules are easy and rely on group|env|org-membership

frontend/src/modules/Metadata_Forms/components/MetadataFormFields.js
SofiaSazonova: I didn't decide yet. So, just in case I decided to make it the same way as for e.g. datasets.
If it's not needed, I will reduce it later.

SofiaSazonova: I give it another thought and removed Admin role for now.

backend/dataall/modules/metadata_forms/services/metadata_form_service.py
SofiaSazonova: Oh, I see, what do you mean. I'll save it for another PR, I think.
Right now I found a bug in Admin Setting panel

SofiaSazonova: Bug fixed, tennant permissions added

noah-paige:
Tested 

- [x] User roles `USER` or `OWNER` resolve correctly and show on FE
- [x] Tenant Admin can list all MF Forms regardless of envs/orgs a part of
- [x] User can only view MF Forms that are global or a part of one of their envs, orgs, or groups/teams
- [x] USER roles cannot edit, delete, create fields, etc on MF Form
- [x] Cannot create MF Form without Tenant Permission MANAGE_METADATA_FORMS
- [x] Create, Update, Delete as OWNER works

frontend/src/modules/Metadata_Forms/components/MetadataFormFields.js
noah-paige, line 1: minor bug found on L125 in `MetadataFormFields` when trying to edit the `required` value. Currently the code is:

```
onChange={(event) => {
     updateField(index, 'required', event.target.value === 'on');
}}
```

But after testing it seems that `event.target.value` is always value `on`. Meaning after  changing the `required` field once to enable you can never re-edit and disable. I think this has to be changed to

```
                  onChange={(event, value) => {
                    updateField(index, 'required', value);
                  }}
```

after some testing - `value` will be boolean true/false depending on if checkbox is licked or not and will update `required` on and off appropriately

noah-paige:
Hi @SofiaSazonova - left one small comment on a bug found not introduced in this PR but would super if could quickly push a one line change to fix

After that I think this PR good to be merged

noah-paige (approved):
lgtm - approving

Issue #1285, opened 2024-05-17
Automated Share Correction with Re-Apply Share ECS Task


anushka-singh:
**Is your idea related to a problem? Please describe.**
We have Share Verifier ECS task scheduled to run weekly. Occasionally, this leads to shares being marked as unhealthy. Thankfully, the 'Re-Apply Share' button proves to be a valuable resource, resolving most of these issues upon activation.

**Describe the solution you'd like**
There is a need to introduce a new ECS task specifically for re-applying shares. This task can execute after the Share Verifier task, allowing for automatic correction of shares without the need for manual intervention.

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

dlpzx:
This is a great idea that does not require a lot of additional efforts on top of what we have already implemented :) Are you planning to take it?

noah-paige:
Hello - we have opened a PR that adds a step to the share verifier task that will also start the reapplier task to try and re-apply/fix for any unhealthy share items after verification of all share items is complete

Currently the implementation is default behavior when the verifier task is triggered - please let me know thoughts after taking a look at the PR

@anushka-singh @zsaltys

anushka-singh:
Added a comment on the PR

PR #1476, opened 2024-08-08
Run reapply automatically if Share Verifier Task detects Unhealthy Shared Items

noah-paige:
### Feature or Bugfix

- Feature/Enhancement 

### Detail
- After the scheduled task verifies all share objects --> automatically re-apply the unhealthy share items found in data.all

### Relates
- https://github.com/data-dot-all/dataall/issues/1285

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/shares_base/tasks/share_verifier_task.py
+++ b/backend/dataall/modules/shares_base/tasks/share_verifier_task.py
@@ -5,7 +5,7 @@ from dataall.modules.shares_base.db.share_object_repositories import ShareObject
 from dataall.modules.shares_base.db.share_object_models import ShareObject
 from dataall.modules.shares_base.services.shares_enums import ShareItemStatus
 from dataall.modules.shares_base.services.sharing_service import SharingService
-from dataall.modules.shares_base.tasks.share_reapplier_task import EcsBulkShareRepplyService
+from dataall.core.stacks.aws.ecs import Ecs
 
 from dataall.base.db import get_engine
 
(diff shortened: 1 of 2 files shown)


backend/dataall/modules/shares_base/tasks/share_verifier_task.py
```diff
@@ -37,8 +39,18 @@ def verify_shares(engine):
         return processed_share_objects
 
 
+def trigger_reapply_task():
```
anushka-singh, line 42: By only triggering the reapply task if unhealthy shares are detected, we can avoid unnecessary ECS task executions.
We could have something like: 
```
def trigger_reapply_task(should_trigger):
    if should_trigger:
        log.info("Unhealthy shares detected, triggering reapply task...")
        Ecs.run_ecs_task(
            task_definition_param='ecs/task_def_arn/share_reapplier',
            container_name_param='ecs/container/share_reapplier',
            context=[],
        )
    else:
        log.info("No unhealthy shares detected, skipping reapply task.")
```

And define should_trigger in verify_shares by fetching health status of items.

noah-paige: we currently don't return that information from `verify_shares` and don't have an straightforward mechanism to determine if there are unhealthy shares other than adding a new query to find all ShareObjectItems that have healthStatus of `Unhealthy`

While we can do that and populate the value of `should_trigger` to T/F based on the count of that query - I personally feel that the additional cost is very negligible and there is likely no need

ECS pricing model is variable per GB and CPU down to a per second granularity - and our ECS Task uses 1vCPU and 2GB of memory

Especially, since these tasks a scheduled on a weekly cadence - the additional pricing if any would be fractions of cents

Link to ECS Fargate Pricing model - https://aws.amazon.com/fargate/pricing/

If others feel strongly I can re-evaluate!

petrkalos (approved):

PR #1488, opened 2024-08-16
allow dbmigrations lambda to invoke any alembic command

petrkalos:
### Feature or Bugfix
- Feature

### Detail
Currently if things get mixed up (in the db) due to migrations it's very hard to revert the situation (because db isn't exposed to the internet). We have 3 options
* Do it manually via SQL Editor
* Create an EC2 instance to use as a jump proxy
* Create a VPN

With this change we can now run any alembic command directly from the lambda by providing args as event data.

#### Manual downgrade 

Let's assume that you are working on a feature and have a new alembic revision locally, let's say `rev1`. 
Also you have already deployed that to your dev AWS account.
The upstream migrations history looks like `rev0`
The local migations history looks like this `rev0`->`rev1`.
A PR gets merged with an alembic migration `rev2` and is now changing the head from `rev0` to `rev2`.
Now the upstream migrations history looks like `rev0` -> `rev2`.
In order to merge you must rebase `main` and as solve this conflict, so the migrations history should look like this `rev0`->`rev2`->`rev1`.
Upon deployment you dev deployment thinks that is already on `rev1` and hence pipeline will not run any migrations and will skip the changes in `rev2`.

With this tool you can pass (before deployment of backend) as dbmigrations event payload the following which will downgrade the DB to `rev0`.
```
{
  "command": "downgrade",
  "args": {
    "revision": -1
  }
}
```
Then when normal upgade runs it will succesfully apply first `rev2` and then `rev1`.

#### Check history and current revision
It allows you to quickly look at the alembic history  and indicate the current revision of the db by passing the following as event contest...
```
{
  "command": "history",
  "args": {
    "indicate_current": true
  }
}
```

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/migrations/versions/b833ad41db68_maintenance_window_schema.py
+++ b/backend/migrations/versions/b833ad41db68_maintenance_window_schema.py
@@ -11,14 +11,17 @@ import os
 from alembic import op
 import sqlalchemy as sa
 from sqlalchemy import Column, String, orm
+from sqlalchemy.ext.declarative import declarative_base
 
-from dataall.base.db import get_engine, has_table, Base
+from dataall.base.db import get_engine, has_table
 
 # revision identifiers, used by Alembic.
 revision = 'b833ad41db68'
 down_revision = '458572580709'
 branch_labels = None
 depends_on = None
+
+Base = declarative_base()
 
 
 class Maintenance(Base):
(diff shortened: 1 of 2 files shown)


backend/migrations/versions/b833ad41db68_maintenance_window_schema.py
```diff
@@ -11,15 +11,18 @@
...
-from dataall.base.db import get_engine, has_table, Base
+from dataall.base.db import get_engine, has_table
 
 # revision identifiers, used by Alembic.
 revision = 'b833ad41db68'
 down_revision = '458572580709'
 branch_labels = None
 depends_on = None
 
+Base = declarative_base()
```
petrkalos, line 24: irrelevant change but it was preventing alembic from loading because the `maintenance` table was trying to be defined twice. Following the pattern of the rest of the alembic script we redefine base instead of reusing.

SofiaSazonova (approved):
Cool! Very usefull!

SofiaSazonova:
Minor comment: @petrkalos could you please add this description to our 'dev hacks' and readme for migrations?

Issue #1083, opened 2024-03-01
Enable Data Producers to Set Expiration Periods for Shares:


anushka-singh:
**Is your feature request related to a problem? Please describe.**
- Shares lack expiration dates, leading to potential issues with stale or outdated data.
- There is no systematic process for removing invalid or outdated shares from the data pool.
- Lack of auditing mechanisms makes it challenging to track share validity and usage.
- Stale shares can clutter the system and may lead to misuse or confusion.
- There's a need for shares to have time-bound validity, with both producers and consumers reviewing them periodically.

**Describe the solution you'd like**

**1. Granular Expiration Settings:**
- Producers can set expiration periods at:

   1.1 - **Dataset level:** Define an expiration date range (e.g., 30-90 days) when creating a dataset.
   1.2 - **Share level:** Specify different expiration periods for individual shares based on requestor needs (e.g., 30 days for R1, 90 days for R2).

**2. Notification System:**

- **As expiration approaches** (e.g., two weeks prior), the system sends notifications to requestors and/or producers to reassess the share's validity.

**3. Handling Expiration:**

- Upon reaching expiration:
   3.1 - **Escalation**: Option to escalate to a higher authority in the chain for further review or action.
   3.2 - **Revocation**: Ability to revoke access to the share, potentially disrupting downstream processes dependent on the dataset.
   3.3 - **Visual Indicators**: Display a prominent indicator (e.g., red button) signaling the share's expired status without immediately revoking access, preventing disruptions to existing processes.

**4. Configurability:**
- All time ranges are configurable, allowing flexibility to adapt to different use cases and requirements.

****Describe alternatives you've considered****
- I will write a longer description in following days proposing different solutions and discuss with team
- All time ranges are up for discussion and will be configurable

*P.S. Please Don't attach files. Add code snippets directly in the message body instead.*

anushka-singh:
### Solution 2: Priority setting based system
![GH1083-2](https://github.com/data-dot-all/dataall/assets/26413731/139099ef-85b7-4316-b5c3-9397eb465712)

Please look at attached diagram for more context on what A1, A2, B1,B2 mean:



#### What is the priority based system?
- When user creates a share request, they can select one of 3 set priorities for their request. 
**A1. Highest priority: Priority 1** - access never expires. This is a critical dataset needed for a long term process that will never expire. User needs to add reasoning for choosing this priority that will be reviewed by producers. Producer might accept or downgrade request.
[...]

PR #1489, opened 2024-08-19
feat(GH-1083) share expiration

TejasRGitHub:
### Feature or Bugfix
- Feature
- 
### Detail

- Details mentioned in the GH issue - https://github.com/data-dot-all/dataall/issues/1083

### Relates
- https://github.com/data-dot-all/dataall/issues/1083

### Security

  - Is the input sanitized?
  - Have you used the least-privilege principle? How? Yes


TejasRGitHub:
Hi @noah-paige , @anushka-singh , @SofiaSazonova , @dlpzx , could you please take a look at the PR and let me know of any initial review comments. 

Upcoming :
1. Test case summary and tests I have performed
2. Video demonstrating end to end flow on the working of share expiration.
3. Small modification to the share request flow ( I will detail more about this in the GH issue - 1083 )

backend/dataall/modules/s3_datasets/api/dataset/resolvers.py
```diff
@@ -167,6 +173,29 @@ def validate_creation_request(data):
         ConfidentialityClassification.validate_confidentiality_level(data.get('confidentiality', ''))
         if len(data['label']) > 52:
             raise InvalidInput('Dataset name', data['label'], 'less than 52 characters')
+        if data.get('enableExpiration', False):
+            RequestValidator.validate_share_expiration_request(data)
+
+    @staticmethod
+    def validate_share_expiration_request(data):
+        if data.get('expiryMinDuration') < 0 or data.get('expiryMaxDuration') < 0:
```
SofiaSazonova, line 181: This expression will raise an error if `data.get('expiryMinDuration') is None`.

backend/dataall/modules/s3_datasets/api/dataset/resolvers.py
```diff
@@ -167,6 +173,29 @@ def validate_creation_request(data):
...
+            raise InvalidInput(
+                'expiration duration ',
+                '',
+                'must be greater than zero',
+            )
+        if data.get('expiryMinDuration') > data.get('expiryMaxDuration'):
```
SofiaSazonova, line 187: Also will raise an exception if one of those is None

SofiaSazonova:
@TejasRGitHub 
Could you please add some integration tests? At least for happy path

SofiaSazonova:
UI comments:
1. If share expiration is disabled it should be displayed as well. Right now there is no information, if expiration is disabled.
2. When we are on the step 2 of creating share (when we add items) we can see “Edit expiration” button, but no information of current requested expiration policy. It can be confusing especially if the user already edited it and they need to expand this menu to see if the changes are saved. I would suggest to show this controls without any additional button.
3. If on step 2 I change share expiration settings, I can’t “submit” request afterwards, only create a draft. But I should still have ‘submit’ button.
4. If I submitted non-expirable share, I can’t set it back to limited one. Is it on purpose? What If the owner of the share wants to limit access afterwards?

backend/dataall/modules/s3_datasets/db/dataset_repositories.py
```diff
@@ -183,6 +189,20 @@ def get_dataset_tables(session, dataset_uri):
...
+    @staticmethod
+    def update_dataset_shares_expiration(session, dataset):
+        """
+        When share expiration is enabled on the dataset while editing a dataset
+        update all the shares on that dataset and set minimum expiration on them
+        """
+        shares = session.query(ShareObject).filter(ShareObject.datasetUri == dataset.datasetUri).all()
+        for share in shares:
+            share.expiryDate = ShareObjectService.calculate_expiry_date(
+                expirySetting=dataset.expirySetting, expirationPeriod=dataset.expiryMinDuration
```
SofiaSazonova, line 201: What happens, if
1. Current expiration is less then new `dataset.expiryMinDuration`?
2. Current expiration is more then new `dataset.expiryMaxDuration`?

backend/dataall/modules/shares_base/services/share_object_service.py
```diff
@@ -454,3 +688,19 @@ def _validate_group_membership(session, share_object_group, environment_uri):
                 action=CREATE_SHARE_OBJECT,
                 message=f'Team: {share_object_group} is not a member of the environment {environment_uri}',
             )
+
+    @staticmethod
+    def calculate_expiry_date(expirationPeriod, expirySetting):
+        currentDate = date.today()
```
SofiaSazonova, line 694: Here `date.today()` is used, but it can be called for existing shares as well.
So, there can be a case
1. Share is about to expire in 2 days
2. We update dataset's share expiration settings
3. We call  calculate_expiry_date
4. With `currentDate = date.today()` share is automatically prolonged for at least a month.

backend/dataall/modules/s3_datasets/db/dataset_repositories.py
```diff
@@ -183,6 +189,20 @@ def get_dataset_tables(session, dataset_uri):
...
 
+    @staticmethod
+    def update_dataset_shares_expiration(session, dataset):
+        """
+        When share expiration is enabled on the dataset while editing a dataset
+        update all the shares on that dataset and set minimum expiration on them
+        """
+        shares = session.query(ShareObject).filter(ShareObject.datasetUri == dataset.datasetUri).all()
+        for share in shares:
+            share.expiryDate = ShareObjectService.calculate_expiry_date(
```
SofiaSazonova, line 200: service layer is called from repository layer -- it violates the structure of the code and can lead to nasty circular dependencies. Could you please untangle it?
I believe the logic part can be moved to dataset_service and we only leave actual object update in repository file.

backend/dataall/modules/shares_base/api/mutations.py
```diff
@@ -35,6 +40,13 @@
     type=gql.Boolean,
 )
 
+cancelShareExtensionObject = gql.MutationField(
```
SofiaSazonova, line 43: IMHO, `ShareExtensionObject` is a confusing naming.
We do not introduce another type of object, just change the properties of existing `ShareObject`.
Would you mind to call it here and in other mutations e.g. `_operation_ShareExtension`

SofiaSazonova (changes requested):
Left some comments 

General question:
What will happen if 
1. Request for extension is submitted
2. Share expired before request is approved
3. Request approved
?

backend/dataall/modules/s3_datasets/api/dataset/resolvers.py
TejasRGitHub: Thanks for pointing this out. I was thinking its good to raise an error in case an invalid input is given. But I can see that error which will get raised here would be unclear. I will add explicit type checking and raise a custom error in case a valid is not present

TejasRGitHub:
> UI comments:
> 
> 1. If share expiration is disabled it should be displayed as well. Right now there is no information, if expiration is disabled.

I can show that on the UI. Thanks
> 2. When we are on the step 2 of creating share (when we add items) we can see “Edit expiration” button, but no information of current requested expiration policy. It can be confusing especially if the user already edited it and they need to expand this menu to see if the changes are saved. I would suggest to show this controls without any additional button.

Okay. I can put a requested date on the UI. 
> 3. If on step 2 I change share expiration settings, I can’t “submit” request afterwards, only create a draft. But I should still have ‘submit’ button.

Will add this. Thanks. I had this, in one of the updates, this must have been removed

> 4. If I submitted non-expirable share, I can’t set it back to limited one. Is it on purpose? What If the owner of the share wants to limit access afterwards?

I wanted to make it not able to edit the share expiration once it is non-expiring share . Since a non-expiring wouldn't have an expiry. But I get your point in which if a share was given non-expiring access inadvertently then there should be a mechanism to make it a limited period access expiring share . 

Will make updates. Thanks a lot for the comments

backend/dataall/modules/s3_datasets/db/dataset_repositories.py
TejasRGitHub: Hmm. That's a good point. I will incorporate some checks 
I am thinking 
if the share already has an expiration date which is present, then I won't change the expiration period, but if the share expiration date is not present then I will give it minimum expiration.

backend/dataall/modules/shares_base/services/share_object_service.py
TejasRGitHub: Okay. Let me know if this is the correct scenario you are thinking,

1. Share is already approved with an expiration date
2. Someone modifies the Dataset and changes the expiration settings 
3. Then all shares are updated and thus this share will have a new expiration. 

If this is case you are thinking, I will be addressing this as a part of this comment - https://github.com/data-dot-all/dataall/pull/1489#discussion_r1727638952

backend/dataall/modules/shares_base/api/mutations.py
TejasRGitHub: sure. I can do that.

backend/dataall/modules/shares_base/services/share_object_service.py
```diff
@@ -297,6 +389,44 @@ def approve_share_object(cls, uri: str):
         Worker.queue(engine=context.db_engine, task_ids=[approve_share_task.taskUri])
         return share
 
+    @classmethod
+    @ResourcePolicyService.has_resource_permission(APPROVE_SHARE_OBJECT)
+    def approve_share_extension_object(cls, uri: str):
```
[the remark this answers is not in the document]
TejasRGitHub: Will do the necessary changes

TejasRGitHub:
> 2\. Share expired before request is approved

Since all the share requests expire at the end of the month ( for monthly & quarterly types) . There is a ECS task which will be running at the end of the month. This ECS task will revoke the share requests which have been expired.

TejasRGitHub:
> @TejasRGitHub Could you please add some integration tests? At least for happy path

Hi @SofiaSazonova , since I don't use Cognito let me check if I can run currently present tests. If I am able to I will add new tests as well . Thanks!

backend/dataall/modules/s3_datasets/db/dataset_repositories.py
TejasRGitHub: Oh Thanks @SofiaSazonova. Since it is a share expiration value I wanted it to be related in the share_base folder, and related services. 
How about if I place it in `share_object_repositories` ?

backend/dataall/modules/s3_datasets/db/dataset_repositories.py
TejasRGitHub: The reason I am not touching existing shares, as they were already approved before and this would create a problems in which the user would have to be informed and also that the requester user has to keep in mind changes in the share expiration.

backend/dataall/modules/s3_datasets/db/dataset_repositories.py
SofiaSazonova: If you touch only share (not the Dataset) -- it belongs to the ShareObjectService (logic part) and ShareObjectRepository(DB update part).

SofiaSazonova:
> > @TejasRGitHub Could you please add some integration tests? At least for happy path
> 
> Hi @SofiaSazonova , since I don't use Cognito let me check if I can run currently present tests. If I am able to I will add new tests as well . Thanks!

Oh, sure, we have only cognito-using tests. Then fine.

TejasRGitHub:
Hi @SofiaSazonova , Updated code with requested changes and improved UX at places.

One pending item - https://github.com/data-dot-all/dataall/pull/1489#discussion_r1726883795

Tested on local dev ( localhost:8080 ) on docker containers. 
1. Creating a dataset with share expiration ✅ 
2. Imporing a dataset with share expiration ✅ 
3. Creating a share for a dataset with Share Expiration ✅  ( Able to modify the share period till the time the share request is approved )
4. Edit the share expiration period request ( while the share is in 'Submitted', 'Draft' state ) -  ✅ 
5. Approving the share request ✅ 
6. Creating a share extension request ✅ 
7. Approveing/ Rejecting/ Cancelling a share request ✅ 
8. Adding another share Item to the shareand checking if I am able to create an extension request✅  ( Here a user should not be able to just request for extension. I am keeping it this way so that share extension workflow doesn't collide with the share approval workflow ) 
9. Converting a share request to non-expiring share request ( by clicking on extend share request and checking the box ) ✅  ( Able to submit this request and the UI looks good ) 
10. Approving the share extension request ✅ 
11. Converting a non-expiring share request to an expiring share request ✅ 
12. Creating/ importing a dataset without expiration ( TODO ) 
13. Approve / Reject and add item to the share request ( TODO )

.checkov.baseline
```diff
@@ -168,6 +168,12 @@
                     "check_ids": [
                         "CKV_AWS_158"
                     ]
+                },
```
TejasRGitHub, line 171: Adding this since the checkov were failing for the ECS log groups ( outputing that the ECS logs grooup need to be encrypted by KMS ). All ECS log groups are not encrypted rightnow and can be taken as another security task . Out of scope for this enhancement feature

backend/dataall/modules/s3_datasets/db/dataset_repositories.py
SofiaSazonova: In Dataset Service 
before the line 264
` DatasetRepository.update_dataset_shares_expiration(session=session, dataset=dataset)`

I would put calculate_expiry_date  logic -- it has nothing specific to share or any other object, just calculation of the dstes , so we can place it in this Service.
It shouldn't be part of other service to avoid any additional dependencies

and then `ShareObjectRepository.update_dataset_shares_expiration(session=session, dataset_uri=dataset, expiration = dataset.enableExpiration, endDate)`

With this solution we 1) detangle services and repos 2) keep logic in services, db updates in repos, 3) keep share object operations in ShareObjectRepository

SofiaSazonova (changes requested):

backend/dataall/modules/s3_datasets/db/dataset_repositories.py
TejasRGitHub: Hi @SofiaSazonova , Thanks! I think this makes sense.

anushka-singh:
Do we have validation on being able to set both expiration minimum and maximum to the same number?
For example, if expiration minimum months is 2, then expiration maximum months has to be greater than 2.

TejasRGitHub:
> Do we have validation on being able to set both expiration minimum and maximum to the same number? For example, if expiration minimum months is 2, then expiration maximum months has to be greater than 2.

There is no validation to prevent this. Minimum and maximum can be the same number. Do you think this should be blocked ? 

If the max and min is same then that would mean that the dataset has fixed expiration.

anushka-singh:
> > Do we have validation on being able to set both expiration minimum and maximum to the same number? For example, if expiration minimum months is 2, then expiration maximum months has to be greater than 2.
> 
> There is no validation to prevent this. Minimum and maximum can be the same number. Do you think this should be blocked ?
> 
> If the max and min is same then that would mean that the dataset has fixed expiration.

What would having min and max be the same number mean?
If we want to specify minimum expiration of one month, we will need to have min as x and max as x + 1 right?

I do think we need validation on this, and also validate that max is always > min.

TejasRGitHub:
> What would having min and max be the same number mean? If we want to specify minimum expiration of one month, we will need to have min as x and max as x + 1 right?
> 
> I do think we need validation on this, and also validate that max is always > min.

If a user wants a dataset to have fixed one month expiry , then in that case he can specify the min = 1 and max = 1. Now the requester has only one option to input in the share expiration period field i.e. 1. With min = max, if lets the user to have a fixed expiration period. This could be a valid use case for someone. 

>  and also validate that max is always > min.

Yes this validation is performed

backend/dataall/modules/s3_datasets/services/dataset_service.py
```diff
@@ -261,7 +262,7 @@ def update_dataset(uri: str, data: dict):
                     if k not in ['stewards', 'KmsAlias']:
                         setattr(dataset, k, data.get(k))
 
-                DatasetRepository.update_dataset_shares_expiration(session=session, dataset=dataset)
+                DatasetRepository.update_dataset_shares_expiration(session=session, enabledExpiration=dataset.enableExpiration, datasetUri=dataset.datasetUri, expirationDate=ShareObjectService.calculate_expiry_date(expirationPeriod=dataset.expiryMinDuration, expirySetting=dataset.expirySetting))
```
SofiaSazonova: Sorry, may be I wasn't 100% clear.
Let's not call one service form another.
Better make `calculate_expiry_date` a private function of DatasetService

backend/dataall/modules/s3_datasets/db/dataset_repositories.py
```diff
@@ -190,22 +190,20 @@ def get_dataset_tables(session, dataset_uri):
         return session.query(DatasetTable).filter(DatasetTable.datasetUri == dataset_uri).all()
 
     @staticmethod
-    def update_dataset_shares_expiration(session, dataset):
+    def update_dataset_shares_expiration(session, enabledExpiration , datasetUri, expirationDate):
```
SofiaSazonova: In this function we change only ShareObjects. So, I should be a part of ShareObjectRepository

SofiaSazonova (changes requested):

TejasRGitHub:
> > > What would having min and max be the same number mean? If we want to specify minimum expiration of one month, we will need to have min as x and max as x + 1 right?
> > > I do think we need validation on this, and also validate that max is always > min.
> > 
> > 
> > If a user wants a dataset to have fixed one month expiry , then in that case he can specify the min = 1 and max = 1. Now the requester has only one option to input in the share expiration period field i.e. 1. With min = max, if lets the user to have a fixed expiration period. This could be a valid use case for someone.
> > > and also validate that max is always > min.
> > 
> > 
> > Yes this validation is performed
> 
> And what happens if min is 2 and max is 3. Even then expiration period is 1 month or not?

If min is 2 and max is 3 then a requester can request for either 2 months access or 3 months access. Requester won't be able to request 1 month access

backend/dataall/modules/s3_datasets/db/dataset_repositories.py
TejasRGitHub: Okay. Let me move out of dataset repositories and move it to ShareObjectRepository.

backend/dataall/modules/s3_datasets/services/dataset_service.py
TejasRGitHub: This would cause repeating the same function in both the ShareObjectService and the DatasetService . Would that be fine ?

TejasRGitHub: Also, can you tell me why is it that internal services cannot call each other please ?

TejasRGitHub: Updated code. Moved `calculate_expiry_date ` to a `dataall\base\utils` package inside the common_module_utls.py file

TejasRGitHub:
Hi @SofiaSazonova , I have updated the PR with code corrections suggested in the comments . Also I have incorporated fetchEnums function in the frontend to get the enums.

SofiaSazonova (changes requested):
@TejasRGitHub thanks! It looks much better now. 
The only comments left is about naming: 
1. rename the `common_module_utils.py` to `expiration_util.py `
2. classname is also better be `ExpirationCalculator` or `ExpirationUtil`
3. place enum directly in there without additional file `backend/dataall/base/utils/enums.py`

TejasRGitHub:
> @TejasRGitHub thanks! It looks much better now. The only comments left is about naming:
> 
> 1. rename the `common_module_utils.py` to `expiration_util.py `
> 2. classname is also better be `ExpirationCalculator` or `ExpirationUtil`
> 3. place enum directly in there without additional file `backend/dataall/base/utils/enums.py`

Updated code!

SofiaSazonova (approved):
Now it's good to go)

SofiaSazonova:
@TejasRGitHub could you please resolve the conflicts with main for PR to be merged

TejasRGitHub:
Hi @SofiaSazonova , Resolved merge conflicts

PR #1495, opened 2024-08-21
Metadata forms 6: attach MF to Orgs, Envs and Datasets

SofiaSazonova:
### Feature or Bugfix

- Feature

### Detail
- MF now can be attached to Orgs, Envs and Datasets id user has edit permissions
- Everyone, who has read access, can view attached MFs
- Attached MF ca be deleted
- NO edit yet
- cascade delete MF => MF field => Attached MF => Attached MF Fields

### Relates
- #1065 

### Security

  - Is the input sanitized?


SofiaSazonova:
https://github.com/user-attachments/assets/c60e4b04-8dff-445b-9d6f-bf771ef4870a

frontend/src/modules/Metadata_Forms/components/metadataAttachement.js
```diff
@@ -0,0 +1,304 @@
...
+
+  const fetchAvailableForms = async () => {
+    const response = await client.query(listMetadataForms({}));
+    if (!response.errors) {
+      setAvailableForms(
+        response.data.listMetadataForms.nodes.map((form) => ({
+          label: form.name,
+          value: form.uri,
+          form: form
+        }))
```
noah-paige, line 57: should available forms only be forms that a user can see that are also not already attachd to the entity - compliment of the set?

frontend/src/modules/Metadata_Forms/components/renderedMetadataForm.js
```diff
@@ -0,0 +1,219 @@
...
+      setErrors(errors);
+      setSubmitting(false);
+      return;
+    }
+    const input = {
+      entityUri: entityUri,
+      entityType: entityType,
+      fields: localFields.map((field) => ({
+        fieldUri: field.uri,
+        value: JSON.stringify(field.value) || 'null'
```
noah-paige, line 118: should this be string `'null'`? Or actual `null` type

frontend/src/modules/Metadata_Forms/components/fields.js
```diff
@@ -0,0 +1,129 @@
...
+export const GlossaryTermField = (props) => {
+  const { field, client, dispatch, onChange, errors } = props;
+
+  const [glossaryOptions, setGlossaryOptions] = useState([]);
+
+  const fetchGlossaryTerms = async () => {
+    const response = await client.query(
+      getGlossaryTree({
+        nodeUri: field.glossaryNodeUri,
+        filter: { pageSize: 500 }
```
noah-paige, line 59: nit: maybe use `Defaults.selectListFilter` here

noah-paige: Also should be able to pass `nodeType` variable to filter on `Term` rather than filtering the response (looking at `GlossaryNodeSearchFilter`)

frontend/src/modules/Metadata_Forms/components/MetadataFormPreview.js
```diff
@@ -1,7 +1,70 @@
...
+        <CircularProgress size={100} />
+      </Box>
+    );
+  }
+
+  return (
+    <RenderedMetadataForm
+      fields={fields}
+      metadataForm={metadataForm}
+      preview="true"
```
noah-paige, line 65: can this be boolean value `true`? We pass `preview={false}` in metadataAttachement but string true here?

noah-paige (commented):
Left some comments on PR but overall the changes look good! 2 general Qs:

- (1) Should there be some permissioning around what user is allowed to attach MF Forms and delete them? These could be new permissions or re-use existing permissions somehow (such as if a user is able to Update Dataset - they should be able to attach / remove MF forms for that entity as well... and similar for other entity types) 
    - May need to think on how to implmement to avoid dependencies but I think we need to add some type of restriction as anyone can attach or delete forms from entities as of now


- (2) Should we add additional tests around attaching and deleting MF forms?

noah-paige:
Testing in AWS:

- [ ] Fail to Attach Form if Required Field Not Populated
- [ ] Populate Field Values Free Form + Drop Down possible Values
- [ ] Attach Form to Org
- [ ] Attach Form to Env
- [ ] Attach Form to Dataset
- [ ] Shared Dataset Cannot Delete (only READ)
- [ ] Read + Delete Attached Form to Org as OrgAdmin
- [ ] Read + Delete Attached Form to Env as EnvAdmin
- [ ] Read + Delete Attached Form to Dataset as DatasetAdmin

noah-paige:
Other comments:

- For boolean values it looks like it is defaulted to `False` value but when you first fill in a form it is `null` and need to click to `true` and click again to `false` to actually populate as `false` and not `null`

- When providing `availableForms` to attach to an entity, for an Organization I can select and attach a form with visibility of Environment. Similarly for Environment I can attach for Dataset visibility. While I think Global and Team may apply to all - I don't think all of the above forms should be given as options depending on entity type and visibility of form

frontend/src/modules/Metadata_Forms/components/renderedMetadataForm.js
noah-paige: nvm - I see what we are doing with JSON stirngify and loads on BE for this

frontend/src/modules/Organizations/views/OrganizationView.js
```diff
@@ -215,6 +217,13 @@ const OrganizationView = () => {
               {currentTab === 'environments' && (
                 <OrganizationEnvironments organization={org} />
               )}
+              {currentTab === 'metadata' && (
+                <MetadataAttachement
+                  entityType="Organization"
+                  entityUri={org.organizationUri}
+                  canEdit={true}
```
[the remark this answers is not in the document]
SofiaSazonova: Right now, whoever has access to Env|Org view page, can Edit it. So I just keep the same access level for MF

frontend/src/modules/Metadata_Forms/components/metadataAttachement.js
SofiaSazonova: listMetadataForms has a visibility filter by default there.

SofiaSazonova: Implemented 'see that are also not already attachd to the entity '

frontend/src/modules/Organizations/views/OrganizationView.js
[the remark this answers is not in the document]
noah-paige: non Env Admins will not be able to Update Env from backend and same with non Org Admins

also only Admins can do things like see `Stack` tab

I think the same should be the case for MF Forms Edit-ability (all can see but only admins can edit)

frontend/src/modules/Metadata_Forms/components/fields.js
```diff
@@ -56,15 +57,11 @@ export const GlossaryTermField = (props) => {
     const response = await client.query(
       getGlossaryTree({
         nodeUri: field.glossaryNodeUri,
-        filter: { pageSize: 500 }
+        filter: { ...Defaults.filter, nodeType: 'T' }
```
noah-paige, line 60: I think this should be `selectListFilter` (which has pageSize 10000) rather than `filter` (with pageSize (10) - or would need some pagination

noah-paige:
> Other comments:
> 
>     * For boolean values it looks like it is defaulted to `False` value but when you first fill in a form it is `null` and need to click to `true` and click again to `false` to actually populate as `false` and not `null`... UPDATE - I actually am not able to specify a value of false (i.e. checkbox deselected) for a required boolean value, can we confirm if that is possible?
> 
>     * When providing `availableForms` to attach to an entity, for an Organization I can select and attach a form with visibility of Environment. Similarly for Environment I can attach for Dataset visibility. While I think Global and Team may apply to all - I don't think all of the above forms should be given as options depending on entity type and visibility of form

From my pov I think these 2 pending comments and the comment on editability of attached forms are what remains. Then I think PR looks good

SofiaSazonova:
@noah-paige 
1. Boolean values are now 'false' explicitly by default
2. Added filter on visibility

noah-paige:
After testing on the latest changes:

For attaching to Organization (PROBLEM - Env Visibility Forms still show)
- Get all Global or Team related Forms
- Get all Forms related to the Current Organization
- NO Env Forms (Still Showing Env Forms)

For attaching to Environemnt (GOOD)
- Get all Global or Team related Forms
- Get all Forms related to the Parent Organization or Current Environment

For attaching to Dataset (GOOD)
- Get all Global or Team related Forms
- Get all Forms related to the Parent Organization or Parent Environment

As a User Listing Forms on MF View (GOOD)
- Get all Global or Team related Forms
- Get all Forms Related to All of User's Envs or Orgs

nit: For Admins do we need a separate way of handling this case?
- For instance when we do something like list Envs we do not handle da_admins separately and show all in `query_user_environments()`
- I get the idea of Admin should see all but it is not consistent with how we handle other queries in backend

noah-paige:
As an enhancement idea would it make sense to use some type of registry (similar to how we do with GlossaryRegistry or FeedRegistry) OR some type of Interface/ abstract class (like with `DatasetServiceInterface` or  `SharesProcessorInterface`) for MF Forms

Thinking as we continue to add attachable Entity Types that conditions and custom logic for each entity could get cumbersome and really is specific to each entity type rather than specific to MF Forms

May involve moving MF Forms to `core` to avoid circular deps but something to consider I think for future

SofiaSazonova:
I thinks, on next step I will revise the access resolvement slightly differently. 
I want to add 'target entity types' to MF, so e.g. a MF can be only attached to Dataset and other only to the Org.

noah-paige:
Changes look good! Last thing is around who can edit MF Forms attached to entities

https://github.com/data-dot-all/dataall/pull/1495#discussion_r1729267045

I think we should have it so only EnvAdmins can edit the Attached Forms on their Env and same with OrgAdmins with their Orgs - do you disagree?

SofiaSazonova:
@noah-paige I'm not sure about it. 
Right now, whoever view Env and Org is allowed to press 'Edit' and change it's name. e.g., which is also metadata, so to speak. So, why should we restrict access to MFs, if person can change name of the entity?

noah-paige:
> @noah-paige I'm not sure about it. Right now, whoever view Env and Org is allowed to press 'Edit' and change it's name. e.g., which is also metadata, so to speak. So, why should we restrict access to MFs, if person can change name of the entity?

@SofiaSazonova yes I see the same that anyone can access the `EnvironmentEditForm` but if a non EnvAdmin attempted to execute `updateEnvironment` they would be restricted because of the following decorator in EnvironmentService:

```
@ResourcePolicyService.has_resource_permission(environment_permissions.UPDATE_ENVIRONMENT)
def update_environment(uri, data=None):
```

Only Env Admins are given `UPDATE_ENVIRONMENT` permissions - so while maybe we should fix the FE conditional rendering to prevent even showing the `Edit` button, I do not believe any person can change name of entity or make any updates unless they are the EnvAdmin team.

The same should be the case for Organization Updates as well

I think we should also add some permission checks for any person who tries to `create_attached_metadata_form` or `delete_attached_metadata_form` to ensure the user is admin for the entity they are trying to attach form to

SofiaSazonova:
@noah-paige Sure.
Then I will change the frontend for Env and Org Views as well

backend/dataall/modules/metadata_forms/services/attached_metadata_form_service.py
```diff
@@ -60,9 +62,18 @@ def list_attached_forms(filter=None):
         if not filter:
             filter = {}
 
+        context = get_context()
+        groups = context.groups
+        is_da_admin = TenantPolicyValidationService.is_tenant_admin(groups)
+        filter = filter if filter is not None else {}
+        orgs, envs = MetadataFormAccessService.get_target_orgs_and_envs(
+            context.username, context.groups, is_da_admin, filter
+        )
```
noah-paige, line 71: would filtering the visibility of the attached forms make sense? 

Looking ahead - lets say a dataset has an attached MF Form with parent MF form having a visibility on Env level. Now a requestor comes along and tries to request access to dataset on a new environment...

If we use the MF Forms key-values as dynamic filters in future should they not be visible in attached form view too? Not sure what the right answer is here

backend/dataall/modules/metadata_forms/db/metadata_form_repository.py
```diff
@@ -1,5 +1,5 @@
 from sqlalchemy import or_, and_
-from sqlalchemy.orm import with_polymorphic
+from sqlalchemy.orm import with_polymorphic, load_only
```
noah-paige, line 2: nit: unused import `load_only`

frontend/src/modules/Environments/views/EnvironmentView.js
```diff
@@ -220,28 +220,31 @@ const EnvironmentView = () => {
...
-                  sx={{ m: 1 }}
-                  onClick={handleArchiveObjectModalOpen}
-                  type="button"
-                  variant="outlined"
-                >
-                  Delete
-                </Button>
-              </Box>
+              {(env.userRoleInEnvironment === 'Admin' ||
+                env.userRoleInEnvironment === 'Owner') && (
```
noah-paige, line 224: nit: Define once `isAdmin` once at top and re-use in both places

frontend/src/modules/Organizations/views/OrganizationView.js
```diff
@@ -171,27 +171,30 @@ const OrganizationView = () => {
...
-                    startIcon={<ArchiveOutlined />}
-                    sx={{ m: 1 }}
-                    variant="outlined"
-                    onClick={handleArchiveObjectModalOpen}
-                  >
-                    Archive
-                  </Button>
-                </Box>
+                {(org.userRoleInOrganization === 'Admin' ||
+                  org.userRoleInOrganization === 'Owner') && (
```
noah-paige, line 175: nit: Define once `isAdmin` once at top and re-use in both places

noah-paige (commented):
A couple of minor call outs and I am curious to hear your thoughts on the visibility restrictions for attached MF Forms?

Also we can pick up restricting MF Form Edit in backend as a separate PR but think it is important to get in for v2.7

noah-paige:
Hi @SofiaSazonova - coming back to restricting attached MF forms shown to a user, I have the following set up:

1 Dataset w/ 3 Attached Metadata Forms:
- 1 Form w/ Global Visibility
- 1 Form w/ Team Visibility of Datset Owner
- 1 Form w/ Org Visibility that Dataset is in

This dataset is shared with another Team (GroupB) in another Environment (a part of a separate Org in data.all). Meaning Group B should not be able to see 2 out of the 3 attached forms

When GroupB user navigates to shared Dataset View and clicks on Metadata Tab, we get the error:

```
An error occurred (UnauthorizedOperation) when calling GET METADATA FORM LIST operation: User XXXXX can not view organization ORG_URI
```

coming from `def get_target_orgs_and_envs(...)` when called as part of `def list_attached_forms(...)`

SofiaSazonova:
@noah-paige my last commit should resolve this error

noah-paige (commented):
Hi @SofiaSazonova - coming back to the same set up where I have:

1 Dataset w/ 4 Attached Metadata Forms:

    1 Form w/ Global Visibility
    1 Form w/ Team Visibility of Datset Owner
    1 Form w/ Org Visibility that Dataset is in
    1 Form w/ Env Visibility that Dataset is in

This dataset is shared with another Team (GroupB) in another Environment (a part of the SAME Org in data.all). Meaning Group B should be able to see 2 out of the 4 attached forms (the global and org ones)

When GroupB user navigates to shared Dataset View and clicks on Metadata Tab, we do not get any error but also do not see the org form (only Global attached form shows).

Also the org form does not show on Metadata Forms List Page. I believe this is because of some improper filtering on `def query_metadata_forms(...)`


By default sqlalchemy will combine each filter attached via:

```
query = query.filter(...)
query = query.filter(...)
```

with an AND operator which will then cause some issues showing results. I think the below code should fix 

```
        query = session.query(MetadataForm)
        filters = []

        if org_uris is not None:
            filters.append(
                and_(
                    MetadataForm.visibility == MetadataFormVisibility.Organization.value,
                    MetadataForm.homeEntity.in_(org_uris),
                )
            )

        if env_uris is not None:
            filters.append(
                and_(
                    MetadataForm.visibility == MetadataFormVisibility.Environment.value,
                    MetadataForm.homeEntity.in_(env_uris),
                )
            )

        if not is_da_admin:
            filters.append(MetadataForm.visibility == MetadataFormVisibility.Global.value)
            filters.append(MetadataForm.SamlGroupName.in_(groups))
            filters.append(
                and_(
                    MetadataForm.visibility == MetadataFormVisibility.Team.value,
                    MetadataForm.homeEntity.in_(groups),
                )
            )
        query = query.filter(or_(*filters))

        if filter and filter.get('hideAttached') and filter.get('entityType') and filter.get('entityUri'):
            query = query.filter(
                ~MetadataForm.uri.in_(
                    session.query(AttachedMetadataForm.metadataFormUri)
                    .filter(
                        AttachedMetadataForm.entityUri == filter.get('entityUri'),
                        AttachedMetadataForm.entityType == filter.get('entityType'),
                    )
                    .subquery()
                )
            )

        if filter and filter.get('search_input'):
            query = query.filter(
                or_(
                    MetadataForm.name.ilike('%' + filter.get('search_input') + '%'),
                    MetadataForm.description.ilike('%' + filter.get('search_input') + '%'),
                )
            )

        return query.order_by(MetadataForm.name)
```

SofiaSazonova:
I think, I solved it a bit easier:

Admins see all (if we are not restricted by entity, when list attached forms or possible forms)
Non Admins:
1.  If you are owner, you can see the form
2. If you are not owner:
    2.1. Filter by the team
    2.2. Filter by the orgs
    2.3. Filter by the envs

backend/dataall/modules/metadata_forms/db/metadata_form_repository.py
```diff
@@ -110,3 +174,51 @@ def update_metadata_form_field(session, fieldUri, data):
...
+    @staticmethod
+    def get_all_attached_metadata_form_fields(session, uri):
+        return session.query(all_fields).filter(AttachedMetadataFormField.attachedFormUri == uri).all()
+
+    @staticmethod
+    def query_attached_metadata_forms(session, is_da_admin, groups, envs, orgs, filter):
+        all_mfs = MetadataFormRepository.query_metadata_forms(
+            session, is_da_admin, groups, envs, orgs, filter
+        ).subquery()
+        print(f'{orgs=}')
```
noah-paige, line 214: nit: remove print

noah-paige:
I do believe it may make easier for both testing and implementation purposes if we split up the GQL Operations:

- (1) ListMetadataForms - lists Metadata Forms to populate Metadata Form View page (working as expected currently)
- (2) listAttachedMetadataForms - takes in resource URI, finds **user's** resource entities (Env + Org), query all attached forms for this resource filtered by what User can see
- (3) listAvailableMetadataForms - takes in resource URI, finds **target** resource entities (Env + Org), query all available MF Forms filtered by what is already attached

SofiaSazonova:
About deletion of entities and connected MFs -- I want to leave it for a separate PR. There are a lot of things to consider there.

SofiaSazonova:
> (3) If i have a MF Form created with visibility of EnvA I can attach that form to EnvA as well as all of its child datastets (same for Org visibility forms too and attachable to homeEntity Org)

Yes, I thinks, it's the case. Again, in future I want to add 'target entity type', so we can restrict it, if needed

(1), (2) --  This is a good call. I will consider it.
But I prefer to deal with all the deletions in separate PR

noah-paige (commented):
one minor bug and some `nit` comments left

Also I think we need to change the `list_metadata_form` query definition at `dataall/tests_new/integration_tests/modules/metadata_forms/queries.py` for integration tests and add additional query defs/update tests as needed

noah-paige:
> > (3) If i have a MF Form created with visibility of EnvA I can attach that form to EnvA as well as all of its child datastets (same for Org visibility forms too and attachable to homeEntity Org)
> 
> Yes, I thinks, it's the case. Again, in future I want to add 'target entity type', so we can restrict it, if needed
> 
> (1), (2) -- This is a good call. I will consider it. But I prefer to deal with all the deletions in separate PR

We can leave for a future PR - however do you see any risks with (1) and (2) as it is currently implemented along with create and delete attached MF forms not having a decorator for `has_resource_permissions` checks or similar permissions checks? I know restricted on FE but restrict directly on BE

noah-paige (approved):
think these final changes look good for this PR - tested in a deployment as well and all MF form queries working as expected

Approving

Issue #1304, opened 2024-05-30
Review Validations on input fields while importing of dataset ( can be extended to an environment, org, etc ) 


TejasRGitHub:
**Is your idea related to a problem? Please describe.**
While importing a dataset, the input fields containing bucket name, glue db, KMS key need to be validated so that the input fields conform to the expectation.

For eg. - a user could input ARN of the bucket name instead of the bucket name  and this would create a dataset but it would fail while creating a dataset stack.

Currently, there are few validations - i.e. checking if the KMS key exists or not for that bucket. 

**Describe the solution you'd like**
Add validations for inputs on forms where resource creation is present ( datasets , environment, etc )  

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

noah-paige:
I think we should build out a similar utility to our current `NamingConventionService`, such as `NamingValidationService` which is solely responsible for checking resource naming provided on forms, e.g. bucket names, role ARNs, glue DB names, etc.

Can then use this utility for all of the areas where we need this type of validation. Let me know if you had something similar in mind for implementing the above

To see if have room to pick up in v2.7 or v2.8... or always eager to accept external contributions!

PR #1501, opened 2024-08-27
Validate imported resource names  via NamingConventionService

SofiaSazonova:
### Feature or Bugfix

- Refactoring

### Detail
- new properties in NamingConventionPattern: valid_external_regex, max_imported_length. If not defined, internal regex and max_length will be used
- function validate_imported_name is used for resources that are created outside data.all: e.g. s3, kms-aliases and glue tables for imported datasets

### Relates
- #1304 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/base/utils/naming_convention.py
+++ b/backend/dataall/base/utils/naming_convention.py
@@ -8,7 +8,7 @@ class NamingConventionPattern(Enum):
         'regex': '[^a-zA-Z0-9-]',
         'separator': '-',
         'max_length': 63,
-        'valid_external_regex': '^[a-z0-9][a-z0-9.-]{1,61}[' 'a-z0-9]$',
+        'valid_external_regex': '(?!(^xn--|.+-s3alias$))^[a-z0-9][a-z0-9-]{1,61}[a-z0-9]$',
     }
     KMS = {'regex': '[a-zA-Z0-9_-]+$', 'separator': '-', 'max_length': 63}
     IAM = {'regex': '[^a-zA-Z0-9-_]', 'separator': '-', 'max_length': 63}  # Role names up to 64 chars
@@ -73,7 +73,7 @@ class NamingConventionService:
     def validate_imported_name(self):
         regex = NamingConventionPattern[self.service].value['regex']
         max_length = NamingConventionPattern[self.service].value['max_length']
-        valid_external_regex = NamingConventionPattern[self.service].value.get('valid_external_regex', regex)
+        valid_external_regex = NamingConventionPattern[self.service].value.get('valid_external_regex', '.*')
         if 'arn:aws:' in self.target_label:
             raise Exception('An error occurred (InvalidInput): name expected, arn-like string received')
         if not re.search(valid_external_regex, self.target_label):
--- a/backend/dataall/base/utils/naming_convention.py
+++ b/backend/dataall/base/utils/naming_convention.py
@@ -9,9 +9,8 @@ class NamingConventionPattern(Enum):
         'separator': '-',
         'max_length': 63,
         'valid_external_regex': '^[a-z0-9][a-z0-9.-]{1,61}[' 'a-z0-9]$',
-        'max_imported_length': 63,
     }
-    KMS = {'regex': '[a-zA-Z0-9/_-]+$', 'separator': '-', 'max_length': 63}
+    KMS = {'regex': '[a-zA-Z0-9_-]+$', 'separator': '-', 'max_length': 63}
     IAM = {'regex': '[^a-zA-Z0-9-_]', 'separator': '-', 'max_length': 63}  # Role names up to 64 chars
     IAM_POLICY = {'regex': '[^a-zA-Z0-9-_]', 'separator': '-', 'max_length': 128}  # Policy names up to 128 chars
     GLUE = {
@@ -75,14 +74,13 @@ class NamingConventionService:
         regex = NamingConventionPattern[self.service].value['regex']
         max_length = NamingConventionPattern[self.service].value['max_length']
         valid_external_regex = NamingConventionPattern[self.service].value.get('valid_external_regex', regex)
-        max_imported_length = NamingConventionPattern[self.service].value.get('max_length', max_length)
         if 'arn:aws:' in self.target_label:
             raise Exception('An error occurred (InvalidInput): name expected, arn-like string received')
         if not re.search(valid_external_regex, self.target_label):
             raise Exception(
                 f'An error occurred (InvalidInput): label value {self.target_label} must match the pattern {valid_external_regex}'
             )
-        elif len(self.target_label) > max_imported_length:
+        elif len(self.target_label) > max_length:
             raise Exception(
-                f'An error occurred (InvalidInput): label value {self.target_label} must be less than {max_imported_length} characters'
+                f'An error occurred (InvalidInput): label value {self.target_label} must be less than {max_length} characters'
             )
--- a/backend/dataall/base/utils/naming_convention.py
+++ b/backend/dataall/base/utils/naming_convention.py
@@ -10,7 +10,7 @@ class NamingConventionPattern(Enum):
         'max_length': 63,
         'valid_external_regex': '(?!(^xn--|.+-s3alias$))^[a-z0-9][a-z0-9-]{1,61}[a-z0-9]$',
     }
-    KMS = {'regex': '[a-zA-Z0-9_-]+$', 'separator': '-', 'max_length': 63, 'valid_external_regex': '[a-zA-Z0-9_-]+$'}
+    KMS = {'regex': '[^a-zA-Z0-9-]$', 'separator': '-', 'max_length': 63, 'valid_external_regex': '^[a-zA-Z0-9_-]+$'}
     IAM = {'regex': '[^a-zA-Z0-9-_]', 'separator': '-', 'max_length': 63}  # Role names up to 64 chars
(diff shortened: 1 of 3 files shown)


backend/dataall/base/utils/naming_convention.py
```diff
@@ -4,10 +4,22 @@
 
 
 class NamingConventionPattern(Enum):
-    S3 = {'regex': '[^a-zA-Z0-9-]', 'separator': '-', 'max_length': 63}
+    S3 = {
+        'regex': '[^a-zA-Z0-9-]',
+        'separator': '-',
+        'max_length': 63,
+        'valid_external_regex': '^[a-z0-9][a-z0-9.-]{1,61}[' 'a-z0-9]$',
```
noah-paige, line 11: should it match on trailing `' '` ?

backend/dataall/base/utils/naming_convention.py
noah-paige, line 1: should there ever be a difference between `regex` and `valid_external_regex` or `max_length` and `max_imported_length`?

backend/dataall/modules/s3_datasets/services/dataset_service.py
```diff
@@ -106,19 +106,30 @@ def check_dataset_account(session, environment):
 
     @staticmethod
     def check_imported_resources(dataset: S3Dataset):
+        # check that resource names are valid
+        if dataset.S3BucketName:
+            NamingConventionService(
+                target_uri=dataset.datasetUri,
+                target_label=dataset.S3BucketName,
+                pattern=NamingConventionPattern.S3,
+                resource_prefix='',
```
noah-paige, line 115: nit: omit specifying `resource_prefix` everywhere

backend/dataall/modules/s3_datasets/services/dataset_service.py
```diff
@@ -106,19 +106,30 @@ def check_dataset_account(session, environment):
...
+            ).validate_imported_name()
+
+        if dataset.importedGlueDatabase and dataset.GlueDatabaseName:
+            NamingConventionService(
+                target_uri=dataset.datasetUri,
+                target_label=dataset.GlueDatabaseName,
+                pattern=NamingConventionPattern.GLUE,
+                resource_prefix='',
+            ).validate_imported_name()
```
noah-paige, line 124: is `dataset.GlueDatabaseName` set at this time? I think it is set when `DatasetRepository.create_dataset()` is called which calls `cls._set_dataset_aws_resources()` and sets `GlueDatabaseName` AFTER this validation check happens 

I think the `target_label` for this code block should be `data.get('glueDatabaseName')` and not `dataset.GlueDatabaseName`

backend/dataall/base/utils/naming_convention.py
```diff
@@ -4,10 +4,22 @@
...
+        'max_imported_length': 63,
+    }
+    KMS = {'regex': '[a-zA-Z0-9/_-]+$', 'separator': '-', 'max_length': 63}
```
noah-paige, line 14: do we need a `/` as a supported character? If so do we need to escape it with `\/`?

noah-paige (commented):
left some some minor comments

backend/dataall/base/utils/naming_convention.py
SofiaSazonova: Hi! Yes, I suppose. E.g. for s3 buckets in data.all digits are required, so I can't import dataset from bucket named 'myAWSome-bucket'. As for max length -- idk.

backend/dataall/base/utils/naming_convention.py
SofiaSazonova: There is a thread with the discussion here
https://stackoverflow.com/questions/50480924/regex-for-s3-bucket-name/62673054#62673054

backend/dataall/base/utils/naming_convention.py
SofiaSazonova: We don't need it, true

backend/dataall/modules/s3_datasets/services/dataset_service.py
SofiaSazonova: Previously the check was 
`if len(dataset.GlueDatabaseName) > NamingConventionPattern.GLUE.value.get('max_length'):`
But you are right, GlueDatabaseName is set to 'undefined' by build_dataset function

noah-paige: yes it was likely a bug before too - good to fix now either way

backend/dataall/base/utils/naming_convention.py
noah-paige: From the above thread and other sites the regex I am seeing is `(?!(^xn--|.+-s3alias$))^[a-z0-9][a-z0-9-]{1,61}[a-z0-9]$` rather than the one above which also allows `.` and `' '`

backend/dataall/base/utils/naming_convention.py
noah-paige: I see we actually use `regex` to match all the disallowed characters and replace then in `slugify()` call

With that in mind maybe we should not default to `regex` if `valid_external_regex` is not defined

So `validate_imported_name` will require `valid_external_regex` or default to allowing all characters like `NamingConventionPattern[self.service].value.get('valid_external_regex', '.*')`


and I think max_length should always be same as max_imported_length so can keep only 1

backend/dataall/modules/s3_datasets/services/dataset_service.py
```diff
@@ -129,6 +138,12 @@ def check_imported_resources(dataset: S3Dataset):
                     param_value=dataset.KmsAlias,
                     constraint=f'empty, Bucket {dataset.S3BucketName} is encrypted with AWS managed key (SSE-S3). KmsAlias {kms_alias} should NOT be provided as input parameter.',
                 )
+            NamingConventionService(
+                target_uri=dataset.datasetUri,
+                target_label=kms_alias,
+                pattern=NamingConventionPattern.KMS,
+                resource_prefix='',
```
noah-paige, line 145: nit: remove `resource_prefix`

backend/dataall/base/utils/naming_convention.py
```diff
@@ -4,10 +4,21 @@
...
 
 class NamingConventionPattern(Enum):
-    S3 = {'regex': '[^a-zA-Z0-9-]', 'separator': '-', 'max_length': 63}
+    S3 = {
+        'regex': '[^a-zA-Z0-9-]',
+        'separator': '-',
+        'max_length': 63,
+        'valid_external_regex': '(?!(^xn--|.+-s3alias$))^[a-z0-9][a-z0-9-]{1,61}[a-z0-9]$',
+    }
+    KMS = {'regex': '[a-zA-Z0-9_-]+$', 'separator': '-', 'max_length': 63}
```
noah-paige, line 13: let's add oen more `valid_external_regex` for KMS service

backend/dataall/base/utils/naming_convention.py
```diff
@@ -56,3 +69,18 @@ def validate_name(self):
             raise Exception(
                 f'An error occurred (InvalidInput): label value {self.target_label} must be less than {max_length} characters'
             )
+
+    def validate_imported_name(self):
+        regex = NamingConventionPattern[self.service].value['regex']
```
noah-paige, line 74: nit: don;t need `regex`

backend/dataall/base/utils/naming_convention.py
```diff
@@ -4,10 +4,21 @@
...
 
 class NamingConventionPattern(Enum):
-    S3 = {'regex': '[^a-zA-Z0-9-]', 'separator': '-', 'max_length': 63}
+    S3 = {
+        'regex': '[^a-zA-Z0-9-]',
+        'separator': '-',
+        'max_length': 63,
+        'valid_external_regex': '(?!(^xn--|.+-s3alias$))^[a-z0-9][a-z0-9-]{1,61}[a-z0-9]$',
+    }
+    KMS = {'regex': '[a-zA-Z0-9_-]+$', 'separator': '-', 'max_length': 63, 'valid_external_regex': '[a-zA-Z0-9_-]+$'}
```
noah-paige, line 13: last thing - since `regex` is as `regex pattern for disallowed characters` in `def slugify(...)` I think best we remove the `regex` for `KMS` Pattern (or change to be the negation of what it is now)

backend/dataall/modules/s3_datasets/services/dataset_service.py
```diff
@@ -105,20 +105,29 @@ def check_dataset_account(session, environment):
...
+    def check_imported_resources(dataset: S3Dataset, data: dict = {}):
+        # check that resource names are valid
+        if dataset.S3BucketName:
+            NamingConventionService(
+                target_uri=dataset.datasetUri,
+                target_label=dataset.S3BucketName,
+                pattern=NamingConventionPattern.S3,
+            ).validate_imported_name()
+
+        if dataset.importedGlueDatabase and dataset.GlueDatabaseName:
```
noah-paige, line 117: nit: just check `if dataset.importedGlueDatabase:` no need to have `and dataset.GlueDatabaseName` based on previous discussion

backend/dataall/base/utils/naming_convention.py
noah-paige: ++ need to make `'valid_external_regex': '^[a-zA-Z0-9_-]+$'` to ensure it starts and ends with

noah-paige (approved):
looks good -approving

Issue #1502, opened 2024-08-27
Make look-up period for share logs configurable


SofiaSazonova:
Now share logs are looked up for latest 24 hours.
Make this parameter configurable

PR #1503, opened 2024-08-27
Make log query period configurable

SofiaSazonova:
### Feature or Bugfix

- Feature

### Detail
- log query period for shares and stacks are set in config.json
- if not set, log query period is 1 day

### Relates
- #1502 #1502
### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/core/stacks/api/resolvers.py
+++ b/backend/dataall/core/stacks/api/resolvers.py
@@ -68,7 +68,7 @@ def get_stack_logs(context: Context, source, targetUri: str = None, targetType: 
     query = StackService.get_stack_logs(target_uri=targetUri, target_type=targetType)
     envname = os.getenv('envname', 'local')
     log_group_name = f"/{Parameter().get_parameter(env=envname, path='resourcePrefix')}/{envname}/ecs/cdkproxy"
-    log_query_period_days = config.config.get_property('core.log_query_period_days', 1)
+    log_query_period_days = config.get_property('core.log_query_period_days', 1)
 
     results = CloudWatch.run_query(
         query=query,
--- a/backend/dataall/core/stacks/api/resolvers.py
+++ b/backend/dataall/core/stacks/api/resolvers.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 
+from dataall.base.config import config
 from dataall.base.api.context import Context
 from dataall.core.environment.services.environment_service import EnvironmentService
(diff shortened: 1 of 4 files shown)


SofiaSazonova:
@anushka-singh have a look please

config.json
```diff
@@ -70,6 +70,7 @@
             "env_aws_actions": true,
             "cdk_pivot_role_multiple_environments_same_account": false,
             "enable_quicksight_monitoring": false
-        }
+        },
+        "log_query_period_days": 1
```
anushka-singh, line 74: Maybe a better name for this would be logViewPeriodDays or logDisplayRangeDays since this variable is specifically for displaying logs going back "x" days.

backend/dataall/core/stacks/api/resolvers.py
```diff
@@ -67,10 +68,12 @@ def get_stack_logs(context: Context, source, targetUri: str = None, targetType:
     query = StackService.get_stack_logs(target_uri=targetUri, target_type=targetType)
     envname = os.getenv('envname', 'local')
     log_group_name = f"/{Parameter().get_parameter(env=envname, path='resourcePrefix')}/{envname}/ecs/cdkproxy"
+    log_query_period_days = config.config.get_property('core.log_query_period_days', 1)
```
anushka-singh, line 71: This should be config.get_property and not config.config.get_property.

anushka-singh (changes requested):
Added comments

config.json
noah-paige: ++ if this is only for shares I would move not under `core` - maybe `datasets_base` because we have share feature flags there already (and no `shares_base`)

SofiaSazonova: I would prefer to leave name as it is, since it's about query, not about view. Only the last one log is displayed, even if there were 10 of them.

@noah-paige as we discussed in slack, it's not only for shares, but for all logs, so we leave it in the core

noah-paige:
@SofiaSazonova - in `dataall/frontend/src/modules/Shares/components/ShareLogs.js` we hardcode a string if no logs `'No logs available for the last 24 hours. Logs may take few minutes after the share is processed...'`

Maybe we remove the piece `for the last 24 hours` given this change

SofiaSazonova:
I removed any specific time in this message.

noah-paige (approved):
looks good, approving

anushka-singh (approved):
LGTM! 🚀

Issue #1509, opened 2024-09-02
Redshift data sharing - Enhancements


dlpzx:
This issue keeps track of the [next steps ](https://github.com/data-dot-all/dataall/issues/955#issuecomment-2004237959)and suggestions arising from the implementation of https://github.com/data-dot-all/dataall/issues/955.

### Redshift next steps
- [ ] Add Connections of IAM Federation type - next steps!
- [ ] Use getEnums API call to return clusterTypes with utils implemented in #1435 
- [ ] Extract more common dataset_base code from redshift datasets and s3 datasets
     - [ ] Common FE elements in import/create S3 dataset and import Redshift dataset
     - [ ] Common FE elements in edit datasets
     - [ ] Common resolvers (resolve_dataset_environment, resolve_dataset_owners_group, resolve_dataset_stewards_group)
     - [ ] Common updateDataset API call
     - [ ] Common ModifyRedshiftDatasetInput
- [ ] Following the pattern set by @SofiaSazonova at #1435 I think we should start thinking how to detangle UI from the `config.json`. Here we could have a query that returns all the enabled modules._Originally posted by @petrkalos in https://github.com/data-dot-all/dataall/pull/1424#discussion_r1697871999_

PR #1511, opened 2024-09-02
Redshift data sharing - Added methods from sharing back to redshift datasets (check_on_delete, list_shared_datasets...)

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Complete design in https://github.com/data-dot-all/dataall/issues/955. This particular PR is focused on adding missing functionalities in redshift_datasets that need to be implemented inside redshift_datasets. 

For example, when we delete a redshift dataset we would want to first check if there are any share requests shared for that dataset. To avoid circular dependencies it is required to use an interface in the same way it was implemented for S3.

In this PR:
- Add `RedshiftShareDatasetService(DatasetServiceInterface)` class and implement required abstract methods (check_on_delete, resolve_user_shared_datasets.....
- Use this class in redshift_Datasets module in resolvers, on dataset deletion...
- Some of the code was very similar to the db queries implemented by S3 datasets; for this reason in this PR some of the queries are moved to the generic ShareObjectRepository to be reused by both types of dataset

### Relates
- #955 

### Security

  - Is the input sanitized?


dlpzx:
## Testing

### resolve_user_role and append_list_user_datasets
- [ ] User can see in the Dataset menu a list of Redshift owned and Redshift shared datasets. 
- [ ] For the Shared datasets, the user role is `Shared` 

### Check_on_delete, execute_on_delete
- [ ] When a Redshift dataset with no shares is deleted it deletes fine
- [ ] When a Redshift dataset with requested but not shared items is deleted it deletes the pending items and share requests and the dataset
- [ ] When a Redshift dataset with shared items is deleted it throws an error - shared items must be revoked first

### Share request permission transfer to stewards
- [ ] for a dataset with a share request open, change the steward team and verify that the new steward team got permissions to access and manage the share request
- [ ] for the same dataset, change the stewards team to the owners team and verify that the old stewards do not have permissions to the share anymore

backend/dataall/modules/redshift_datasets/services/redshift_dataset_service.py
```diff
@@ -45,6 +46,46 @@
 
 
 class RedshiftDatasetService:
```
petrkalos, line 48: nit: it might be nice to have an interface for DatasetServices as well

backend/dataall/modules/redshift_datasets/services/redshift_dataset_service.py
```diff
@@ -45,6 +46,46 @@
...
+    _interfaces: List[DatasetServiceInterface] = []
+
+    @classmethod
+    def register(cls, interface: DatasetServiceInterface):
+        cls._interfaces.append(interface)
```
petrkalos, line 53: more of a question rather than a comment: why do we need to keep a list? don't we always register just one?

petrkalos (approved):

backend/dataall/modules/redshift_datasets/services/redshift_dataset_service.py
dlpzx: Agree. In this PR I did not want to mix things too much (I am already creating validation code + moving code around). There is an item in the next steps to remove duplications, I'll reference this comment there https://github.com/data-dot-all/dataall/issues/1509

backend/dataall/modules/redshift_datasets/services/redshift_dataset_service.py
dlpzx: So far yes; we communicate from redshift-shares to redshift-dataset. I imagined other consumption modules that would need to interact with the dataset module. For example a data-pipeline that depends on datasets and the dataset should not be deleted if it has pipelines

PR #1521, opened 2024-09-05
Add External Id Conditions to Deployment Roles

noah-paige:
### Feature or Bugfix

- Enhancement

### Detail 
- Add external ID condition to roles created in deployment account and used by tooling account (frontend config role, stacks updater role, and S3 Deployment role)

### Relates
N/A

### Security

  - Is the input sanitized?


noah-paige:
Testing:

- [x] Deploy to AWS in us-east-1, cross account, internet facing deployment, stack updater enabled
- [x] Pipeline successful
- [x] FE accessible - APIs working as expected

deploy/stacks/iam_utils.py
```diff
@@ -0,0 +1,6 @@
+from aws_cdk import aws_iam as iam
```
dlpzx, line 1: nit: unused import

deploy/stacks/iam_utils.py
```diff
@@ -0,0 +1,6 @@
...
+import hashlib
+
+
+def get_tooling_account_external_id(account: str):
+    return hashlib.sha256(str(account).encode('UTF-8')).hexdigest()[:12]
```
dlpzx, line 6: If we are strict with external IDs and their purpose (like in this [blogpost](https://aws.amazon.com/blogs/apn/securely-using-external-id-for-accessing-aws-accounts-owned-by-others/)) :
- we would define the tooling account as customer and the deployment accounts as providers. --> I don't think it is necessary for us to store the external Ids necessarily in the deployment accounts because we are not working with "real" external customers performing manual actions
- external IDs being unique ---> I think it is too easy for anyone to obtain the external ID value with hashlib if they know the AWS deployment account. It is defeating the purpose of having a password(external Id)

noah-paige: - The current way external ID is generated is that the deployment account is needed as input to determine what the external ID is - so I believe this conforms to the idea that the deployment account is "providing" the external ID and the tooling account is using information provided from deployment account to determine external ID it must use to assume role 
- Also while it is not random how the external ID is created there is still an extra layer of security where a user must know the deployment account, how to generate the external ID, & what role to assume on top already having valid credentials to the tooling account that would be making the assume role request
-  The above plus the fact that the tooling and deployment accounts are unique and private to each deployed data.all application I think makes us safe enough 😄

deploy/stacks/iam_utils.py
noah-paige: removed

deploy/stacks/iam_utils.py
dlpzx: Thanks for the rationale @noah-paige. I think this is enough to prove that external Ids are not the unique security layer; but just another additional layer that we add to protect the accounts and IAM roles from malicious users. I will approve

dlpzx (approved):

PR #1529, opened 2024-09-10
Add bucket versioning pt 2

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- After adding explicitly created artifacts bucket for CodePipeline - began seeing errors when running `cdk synth` (ONLY FOR SPECIFIC DEPLOYMENTS)
    -  Error Message `Error: Artifact Bucket must have a KMS Key to add cross-account action 'Prepare' (pipeline account: 'XXXXX', action account: 'YYYYY'). Create Pipeline with 'crossAccountKeys-true'`

- This error only showed on some deployments and seemed to be throwing a validation error even though docs say providing a manually created bucket which is encrypted should suffice
   -  potentially could be a version mismatch for why it is working only on some deployments (??)

- After adding 'crossAccountKeys=true'` parameter `cdk synth` works appropriately for all

### Relates
- N/A

### Security

  - Is the input sanitized?


noah-paige:
tested in AWS deployment in `us-east-1` - still creates and uses the explicitly created artifacts bucket and KMS Key as expected

dlpzx:
I was able to replicate the error and I am now testing the fix while reviewing the PR

dlpzx (approved):
Looks good. Tested in my own AWS accounts 👍

PR #1531, opened 2024-09-10
Allow configurable Region to run CDK IaC checks

noah-paige:
### Feature or Bugfix

- Enhancement

### Detail
- Allow for `cdk synth` to run on a configurable region using env variables rather than fixing for `eu-west-1` for testing cases
- Change checkov scans to use `us-east-1`

### Relates
- N/A

### Security

  - Is the input sanitized?


deploy/app.py
```diff
@@ -67,7 +68,7 @@
 cdk_pipeline_region = app.node.try_get_context('tooling_region') or os.getenv('CDK_DEFAULT_REGION')
```
dlpzx, line 68: Looks good, I just have 2 questions: 
- why don't we set up the default eu-west-1 for the tooling?
- if this is used only for the IaC checks, do we even need to set the eu-west-1 region?

noah-paige: > why don't we set up the default eu-west-1 for the tooling?
This is in refernece to `cdk_pipeline_region` variable yes? I will add a default for `CDK_DEFAULT_REGION` to be `eu-west-1` for there as well

> if this is used only for the IaC checks, do we even need to set the eu-west-1 region?
It is only used for IaC checks so we could just not set a default in general. Currently we set `us-east-1` for checkov and `eu-west-1` for cdk-nag. We could standardize as all us-east-1 and then not set eu-west-1 but maybe we keep each for additional coverage?

dlpzx: Sounds good

dlpzx (approved):

PR #1391, opened 2024-07-09
Add Dataset integration tests - Tables, Folders

noah-paige:
### Feature or Bugfix
- Feature: Testing

### Detail

In this PR we add new fixtures for S3 datasets that are used for the tests in S3/tables/folders but also for the tests developed in #https://github.com/data-dot-all/dataall/pull/1389:
- Fix imported KMS dataset - there was an error in the KMS keys and in the registration of the Glue database
- Folders as separate fixture using create_folder data.all API
- Tables as separate fixture using boto3 calls to create the table, upload data and then use sync_tables data.all API - the data can be queried!

This PR moves dataset_base testing scenarios to datasets_base/test_dataset.py. Testing scenarios have been defined for the S3 datasets and the remaining test scenarios for the datasets_base APIs are defined with their signature and a TODO comment.

It also splits the S3 dataset tests into their corresponding API subcategories (in `backend/.../s3_datasets/api`)
- test_s3_datasets
- test_s3_tables
- test_s3_tables_profiling
- test_s3_tables_columns
- test_s3_folders

Implement testing scenarios for `test_s3_folders` covering all APIs and dataset types (parametrized tests). Note that to avoid duplication of tests, unauthorized test cases are tested with only one of the dataset types as the code executed is the same for all cases.

Implement testing scenarios for `test_s3_tables` covering all APIs and dataset types (parametrized tests). Same as folders, unauthorized tests are performed on a single dataset type. New tests include: sync_tables with real tables, preview tables with real tables, preview unauthorized depending on the confidentiality level, get_dataset_level, list_dataset_tables

For `test_s3_datasets` only test_create_dataset_unauthorized is added, but for other existing tests we add test for all dataset types (parametrized tests).

#### Next steps
In follow-up PRs we should implement the missing commented TODO tests for:
- datasets_base ---> list owned tests
- s3_datasets ---> list owned tests
- s3_tables ---> data filters tests
- s3_tables_profiling ---> some tests
- s3_tables_columns ---> all tests
- Review backwards compatibility tests and add table and folder test cases

### Relates
- https://github.com/data-dot-all/dataall/pull/1379

### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)?
  - Is the input sanitized?
  - What precautions are you taking before deserializing the data you consume?
[...]

Issue #1358, opened 2024-06-24
Integration tests executed on a real deployment as part of the CICD - Datasets


dlpzx:
Same as for https://github.com/data-dot-all/dataall/issues/1220.

This issue is to track the progress for the Datasets modules.
It has its own dedicated issue because of the challenge of pre-existing infrastructure needed to test datasets

dlpzx:
## Required tests for basic coverage

### For fresh deployments
For each of the following API calls we need to test authorized and unauthorized scenarios as well as all possible configurations (e.g. autoapproval...)
- [ ] Create Dataset - includes testing if dataset is indexed in Catalog
- [ ] Import Dataset - includes testing if dataset is indexed in Catalog
- [ ] List Datasets
- [ ] Edit Dataset
- [ ] Delete dataset
- [ ] Start crawler
- [ ] Sync tables
- [ ] Preview table
- [ ] Delete table
- [ ] Create folder
- [ ] Delete folder

### For backwards compatibility
- [ ] Update created Dataset stack
- [ ] Update imported Dataset stack
- [ ] Create Dataset in updated Environment
- [ ] Import Dataset in updated Environment

## Full coverage [WIP]

### For fresh deployments
- [ ] AWS access to Dataset - Credentials 
- [ ] AWS access to Dataset - S3 redirect
- [ ] Upload data
- [ ] Profile table

### For backwards compatibility
For the updated Dataset stacks:
- [ ] AWS access to Dataset - Credentials 
- [ ] AWS access to Dataset - S3 redirect
- [ ] Start crawler
- [ ] Sync tables

PR #1533, opened 2024-09-11
Add Dataset integration tests - Dataset missing tests, Table Profiling

dlpzx:
### Feature or Bugfix
- Feature: Tests
TO BE MERGED AFTER https://github.com/data-dot-all/dataall/pull/1391

### Detail
Follow-up of https://github.com/data-dot-all/dataall/pull/1391. This PR adds:
- Tests for profiling jobs - because it is an easy submodule I decided to "chain" the tests and make them one dependent on the next one. I could also create a fixture for a profiling job (check warning, profiling jobs cannot be deleted)
- Added missing tests in datasets_base - we still need to add redshift datasets and other types of datasets every time there is a new dataset added.
- Added missing tests in s3_datasets: test_list_s3_datasets_owned_by_env_group. 

⚠️ Issues discovered during testing.
They are not bugs, they are missing functionalities:
- Profiling jobs can never be deleted. It is just information on the RDS database, but nevertheless it cannot be deleted.
- It would be nice to have an API that checks the status of a Glue crawler

### Relates
- https://github.com/data-dot-all/dataall/issues/1358
- https://github.com/data-dot-all/dataall/pull/1391
### Security

  - Is the input sanitized?


tests_new/integration_tests/modules/s3_datasets/global_conftest.py
```diff
@@ -288,20 +400,72 @@ def persistent_s3_dataset1(client1, group1, persistent_env1, testdata):
 
 
 @pytest.fixture(scope='session')
-def persistent_imported_sse_s3_dataset1(client1, group1, persistent_env1, testdata):
+def persistent_imported_sse_s3_dataset1(client1, group1, persistent_env1, persistent_env1_aws_client, testdata):
+    bucket_name = 'dataalltestingpersistentimportedsses3'
```
noah-paige, line 404: won't this cause issues with globally unqiue requirement of S3 Bucket names?

dlpzx: Yes, I ran into that when testing the #1391. It should be fixed now

dlpzx:
Tested locally and currently being tested in real AWS CICD pipeline

dlpzx:
Tested in real CICD pipeline in AWS

tests_new/integration_tests/modules/datasets_base/test_dataset.py
```diff
@@ -6,39 +6,65 @@
...
+    session_s3_dataset1,
+    session_imported_sse_s3_dataset1,
+    session_imported_kms_s3_dataset1,
+    session_id,
+):
+    assert_that(
+        list_datasets_created_in_environment(
+            client1, environment_uri=session_env1.environmentUri, term=session_id
+        ).nodes
+    ).is_length(3)
```
SofiaSazonova, line 52: The same thing we already discussed in previous PR and I don't know how do deal with it better.
But if we add more datasets, we will have to check all the tests like this, where the number is fixed. Is it ok?

tests_new/integration_tests/modules/s3_datasets/test_s3_tables_profiling.py
```diff
@@ -8,12 +8,36 @@
...
 from integration_tests.errors import GqlError
-from integration_tests.modules.s3_datasets.conftest import DATASETS_TABLES_FIXTURES_PARAMS
+from integration_tests.utils import poller
 
 log = logging.getLogger(__name__)
 
 
-@pytest.mark.parametrize(*DATASETS_TABLES_FIXTURES_PARAMS)
+def has_job_finished(job):
+    return job.status not in ['STARTING', 'RUNNING', 'STOPPING', 'WAITING']
```
SofiaSazonova, line 17: Just curious, why do you check the statuses the job NOT IN rather then IN?

SofiaSazonova (approved):
It looks good, there are some questions but I don't think it's blockers

tests_new/integration_tests/modules/s3_datasets/test_s3_tables_profiling.py
dlpzx: I think it was because the "in-progress" where easy to write

tests_new/integration_tests/modules/datasets_base/test_dataset.py
dlpzx: Yes, it is a case that could happen but it should not be the norm. Other modules should use the global_conftest fixtures from other modules, instead of creating their own. If they create their own, then they should create them in conftest and with a limited scope or without the term=session_id

PR #1535, opened 2024-09-11
add mlstudio integ tests

### LINKED ISSUE #1534: Integration tests executed on a real deployment as part of the CICD - MLStudio
Same as for https://github.com/data-dot-all/dataall/issues/1220.

This issue is to track the progress for the MLStudio module.
It has its own dedicated issue because of the challenge of testing shares and the mutiple scenarios

petrkalos:
### Feature or Bugfix
Feature

### Detail
Adding integration tests for ML Studio

### Relates
related to #1220 and resolves #1534

### Security

  - Is the input sanitized?

### DIFF
--- a/tests_new/integration_tests/client.py
+++ b/tests_new/integration_tests/client.py
@@ -29,9 +29,9 @@ class Client:
         graphql_endpoint = os.path.join(os.environ['API_ENDPOINT'], 'graphql', 'api')
         headers = {'AccessKeyId': 'none', 'SecretKey': 'none', 'authorization': self.token}
         r = requests.post(graphql_endpoint, json=query, headers=headers)
-        r.raise_for_status()
         if errors := r.json().get('errors'):
             raise GqlError(errors)
+        r.raise_for_status()
 
         return DefaultMunch.fromDict(r.json())
 
--- a/tests_new/integration_tests/modules/mlstudio/conftest.py
+++ b/tests_new/integration_tests/modules/mlstudio/conftest.py
@@ -1,0 +1,20 @@
+import pytest
+
+from integration_tests.core.stack.utils import check_stack_ready
+from integration_tests.modules.mlstudio.mutations import create_smstudio_user, delete_smstudio_user
+from integration_tests.modules.mlstudio.queries import get_smstudio_user
+
+
+@pytest.fixture(scope='session')
+def smstudio_user1(session_id, client1, persistent_env1):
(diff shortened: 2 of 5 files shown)


tests_new/integration_tests/client.py
```diff
@@ -29,9 +29,9 @@ def query(self, query: str):
         graphql_endpoint = os.path.join(os.environ['API_ENDPOINT'], 'graphql', 'api')
         headers = {'AccessKeyId': 'none', 'SecretKey': 'none', 'authorization': self.token}
         r = requests.post(graphql_endpoint, json=query, headers=headers)
-        r.raise_for_status()
         if errors := r.json().get('errors'):
             raise GqlError(errors)
+        r.raise_for_status()
```
petrkalos, line 34: Irrelevant change to first check for any errors and then raise for status. Sometimes the returned code when there are errors is not 2xx.

tests_new/integration_tests/modules/mlstudio/conftest.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+from integration_tests.core.stack.utils import check_stack_ready
+from integration_tests.modules.mlstudio.mutations import create_smstudio_user, delete_smstudio_user
+from integration_tests.modules.mlstudio.queries import get_smstudio_user
+
+
+@pytest.fixture(scope='session')
+def smstudio_user1(session_id, client1, persistent_env1):
+    env_uri = persistent_env1.environmentUri
```
noah-paige, line 10: Don't we have to ensure the `create_environment` API call has `{'key': 'mlstudioEnabled', 'value': 'true'},` as part of parameters?

Otherwise we wouldn't be able to spin up mew sm studio users

noah-paige (commented):
Left a comment around creating env ensuring ML Studio is enabled.

 Ideally we only enable on the 1 persistent env and not on the session envs so we do not need to add additional wait time for create SM Studio Domains for each env which would be lengthy

Otherwise code looks good - will try out the tests in meantime

tests_new/integration_tests/modules/mlstudio/conftest.py
petrkalos: yes you are right, I enabled it manually on my persistent environment and forgot to add it in code

noah-paige (approved):
looks good and tests passing on my aws deployment too - approved

PR #1539, opened 2024-09-12
Feat/integration tests dataset filters

noah-paige:
### Feature or Bugfix

- Feature

### Detail
- Adding integration tests for Dataset Table Data Filters

- PENDING TESTS PASSING IN DEV AWS ENV
- Merge after https://github.com/data-dot-all/dataall/pull/1391/

### Relates
- related to https://github.com/data-dot-all/dataall/issues/1220 and #1358 

### Security

  - Is the input sanitized?


tests_new/integration_tests/modules/s3_datasets/test_s3_tables_filters.py
```diff
@@ -0,0 +1,149 @@
...
+    'includedCols': ['col_0'],
+}
+ROW_INPUT = {
+    'filterName': 'rowfilter',
+    'description': 'test row',
+    'filterType': 'ROW',
+    'rowExpression': 'col_1 LIKE "%value%" AND col_0 IS NOT NULL',
+}
+
+
```
dlpzx, line 28: nit: Although defining tests that depend on each other is not wrong, it is best practice to make tests not depend on each other if possible (for profiling jobs I ended up not doing it because there is no way of deleting a run). 

In this case it would mean having a data filter fixture that is checked in the `create-*` test, imported in the `list-*` test and `delete-*` test would create their own filter. 

The second advantage of creating the filter as a fixture is that we can add it to the globconftest and use it in the shares tests.

tests_new/integration_tests/modules/s3_datasets/test_s3_tables_filters.py
```diff
@@ -0,0 +1,149 @@
...
+from integration_tests.modules.s3_datasets.queries import (
+    create_table_data_filter,
+    delete_table_data_filter,
+    list_table_data_filters,
+)
+from integration_tests.errors import GqlError
+
+log = logging.getLogger(__name__)
+
```
dlpzx, line 14: It is not required yet, but if in the future we implement the shares_filter tests, the COL_INPUT and ROW_INPUT should include sample_data column names

tests_new/integration_tests/modules/s3_datasets/test_s3_tables_filters.py
dlpzx: If going this route and use the parametrized fixtures without dependencies, check the changes in #1391 that avoid duplicating lines of code with `@pytest.mark.parametrize(*TABLES_FIXTURES_PARAMS)`

PR #1540, opened 2024-09-12
Metadata form 7: Access control and deletion behaviour

SofiaSazonova:
### Feature or Bugfix

- Feature

### Detail
- When entity is deleted all attached MF are deleted
- When Org or Env is deleted MF with visibility in this org|env is deleted. And all connected attached MFs
- Resource policies: CREATE_METADATA_FORM (create MF with visibility inside this entity), ATTACH_METADATA_FORM 
- Migrations for entity owners to have all permissions 

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?


backend/dataall/modules/metadata_forms/services/attached_metadata_form_service.py
```diff
@@ -2,9 +2,12 @@
 from dataall.base.db import exceptions, paginate
 from dataall.core.environment.db.environment_repositories import EnvironmentRepository
 from dataall.core.organizations.db.organization_repositories import OrganizationRepository
+from dataall.core.permissions.services.resource_policy_service import ResourcePolicyService
 from dataall.core.permissions.services.tenant_policy_service import TenantPolicyValidationService
+from dataall.modules.metadata_forms.db.enums import MetadataFormVisibility
```
noah-paige, line 7: nit: remove all unused imports

backend/dataall/modules/metadata_forms/services/attached_metadata_form_service.py
```diff
@@ -35,13 +38,20 @@ class AttachedMetadataFormService:
...
             mf_fields = MetadataFormRepository.get_metadata_form_fields(session, uri)
             AttachedMetadataFormValidationService.validate_enrich_fields_params(mf_fields, data)
-
+            ResourcePolicyService.check_user_resource_permission(
+                session=session,
+                username=context.username,
+                groups=context.groups,
+                resource_uri=data.get('entityUri'),
+                permission_name=ATTACH_METADATA_FORM,
+            )
```
noah-paige, line 54: any reason we don't use `    @ResourcePolicyService.has_resource_permission(ATTACH_METADATA_FORM, parent_resource=_FUNC_RETURNING_ENTITY_URI)` and instead explicitly call `check_user_resource_permission`

backend/dataall/modules/metadata_forms/services/attached_metadata_form_service.py
```diff
@@ -76,5 +86,13 @@ def list_attached_forms(filter=None):
...
         mf = AttachedMetadataFormService.get_attached_metadata_form(uri)
-        with get_context().db_engine.scoped_session() as session:
+        context = get_context()
+        with context.db_engine.scoped_session() as session:
+            ResourcePolicyService.check_user_resource_permission(
+                session=session,
+                username=context.username,
+                groups=context.groups,
+                resource_uri=mf.entityUri,
+                permission_name=ATTACH_METADATA_FORM,  # attach and delete are the same for now
```
noah-paige, line 96: same as above - any reason we don't use `    @ResourcePolicyService.has_resource_permission(ATTACH_METADATA_FORM, parent_resource=_FUNC_RETURNING_ENTITY_URI)` and instead explicitly call `check_user_resource_permission`

backend/dataall/modules/metadata_forms/services/metadata_form_permissions.py
```diff
@@ -1,6 +1,50 @@
...
+DATASET_ALL.extend(ALL_METADATA_FORMS_ENTITY_PERMISSIONS)
+# ------------------------METADATA FORM----------------------------
+# permissions to change and delete metadata forms
+# these permissions are attached to MFs
+UPDATE_METADATA_FORM_FIELD = 'UPDATE_METADATA_FORM_FIELD'
+DELETE_METADATA_FORM_FIELD = 'DELETE_METADATA_FORM_FIELD'
+DELETE_METADATA_FORM = 'DELETE_METADATA_FORM'
+EDIT_METADATA_FORM = 'EDIT_METADATA_FORM'
+
+METADATA_FORM_PERMISSIONS_ALL = [UPDATE_METADATA_FORM_FIELD, DELETE_METADATA_FORM_FIELD, DELETE_METADATA_FORM]
```
noah-paige, line 42: should we add `EDIT_METADATA_FORM` here as well?

backend/migrations/versions/427db8f31999_backfill_MF_resource_permissions.py
```diff
@@ -0,0 +1,81 @@
...
+            session=session,
+            group=dataset.SamlGroupName,
+            resource_uri=dataset.datasetUri,
+            permissions=[ATTACH_METADATA_FORM],
+            resource_type=DatasetBase.__name__,
+        )
+
+
+def downgrade():
+    print('no downgrade supported')
```
noah-paige, line 81: nit: do we need to include downgrade similar to how `/Users/noahpaig/Desktop/dataall-v2m4-testing/dataall/backend/migrations/versions/852cdf6cf1e0_add_redshift_datasets.py` does it using `.find_resource_policies()` and then deleting?

backend/dataall/modules/metadata_forms/services/metadata_form_service.py
```diff
@@ -2,6 +2,7 @@
 from dataall.base.db import exceptions, paginate
 from dataall.core.organizations.db.organization_repositories import OrganizationRepository
 from dataall.core.environment.db.environment_repositories import EnvironmentRepository
+from dataall.core.permissions.services.resource_policy_service import ResourcePolicyService
 from dataall.core.permissions.services.tenant_policy_service import TenantPolicyValidationService, TenantPolicyService
```
noah-paige, line 6: nit remove unused import `TenantPolicyValidationService`

noah-paige (changes requested):
did initial review before testing functionality - requesting some changes

backend/dataall/modules/metadata_forms/services/metadata_form_service.py
```diff
@@ -238,8 +259,24 @@ def batch_metadata_form_field_update(uri, data):
...
+        result_permissions = []
+        with context.db_engine.scoped_session() as session:
+            for permissions in ALL_METADATA_FORMS_ENTITY_PERMISSIONS:
+                if ResourcePolicyService.check_user_resource_permission(
+                    session=session,
+                    username=context.username,
+                    groups=context.groups,
+                    resource_uri=entityUri,
+                    permission_name=permissions,
+                ):
```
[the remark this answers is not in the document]
SofiaSazonova: Good point

backend/dataall/modules/metadata_forms/services/attached_metadata_form_service.py
SofiaSazonova: because here we can't extract data.get('entityUri') as a parent resource.
We can, of course change the parameter of the function and pass 'entityUri' as a separate parameter, not as a part of data, but I assume it is more bulky, that just explicitly call `check_user_resource_permission`

backend/dataall/modules/metadata_forms/services/attached_metadata_form_service.py
SofiaSazonova: same as above

backend/dataall/core/environment/services/environment_service.py
```diff
@@ -46,6 +46,8 @@
 from dataall.core.permissions.services.tenant_permissions import MANAGE_ENVIRONMENTS
 from dataall.core.stacks.db.stack_repositories import StackRepository
 from dataall.core.vpc.db.vpc_repositories import VpcRepository
+from dataall.modules.metadata_forms.db.enums import MetadataFormEntityTypes, MetadataFormVisibility
+from dataall.modules.metadata_forms.db.metadata_form_repository import MetadataFormRepository
```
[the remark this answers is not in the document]
dlpzx: In this case the way to delete metadata forms from the environment would be by using the `EnvironmentResourceManager`

backend/dataall/modules/metadata_forms/db/metadata_form_repository.py
```diff
@@ -247,3 +247,27 @@ def query_attached_metadata_forms(session, is_da_admin, groups, user_envs_uris,
...
+
+    @staticmethod
+    def delete_all_home_metadata_forms(session, homeEntityUri, visibility):
+        mfs = (
+            session.query(MetadataForm)
+            .filter(and_(MetadataForm.homeEntity == homeEntityUri, MetadataForm.visibility == visibility))
+            .all()
+        )
+        for mf in mfs:
+            session.delete(mf)
```
dlpzx, line 273: Instead of querying and then looping and deleting items one by one you could use `delete` directly in the query as a DELETE sql statement:


Like here: 
```
session.query(DatasetTableDataFilter).filter(
            and_(
                DatasetTableDataFilter.tableUri == table.tableUri,
            )
        ).delete()
```

SofiaSazonova: It's not used now, but I refactored it anyway

backend/dataall/core/permissions/services/resource_policy_service.py
```diff
@@ -212,20 +212,21 @@ def associate_permission_to_resource_policy(session, policy, permission):
         session.commit()
 
     @staticmethod
-    def get_resource_policy_permissions(session, group_uri, resource_uri) -> List[ResourcePolicyPermission]:
+    def get_resource_policy_permissions(group_uri, resource_uri) -> List[ResourcePolicyPermission]:
```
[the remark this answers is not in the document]
SofiaSazonova: I need to revert this change. It crushes the stack deploy/update job

noah-paige:
With the new way of handling deletes of metadata forms with deletes of resources a couple things come to mind:
- (1) We still have a dependency between MF module and other resources (i.e. metadata form tables must exist for org, env, dataset deletes to work) but since we always create all tables in DB (and I do not think we plan to change this behavior in the foreseeable future) I think this is okay

- (2) Right now when we create a new form or attach a new form we do not create any permissions for the user/group on the form. I
    - Instead we protect backend resolvers for metadata forms by either the `MetadataFormAccessService.can_perform()` method (which checks if owner or not of the visibility entity) OR checking if has group has ATTACH_METADATA_FORM or CREATE_METADATA_FORM on the resource itself
    - I see the addition of some new permissions like UPDATE_METADATA_FORM_FIELD, EDIT_METADATA_FORM, etc. that are not currently used but if we do plan on assigning these permissions on the form moving forward this would have to also be cleaned up in DB trigger (just calling out the complexity of these triggers could expand)




What if we had a generic EventSystem class (in core or base?):

```
class DataallEventSystem:
    def __init__(self):
        self._subscribers = {}

    def subscribe(self, event: str, callback: Callable[..., None]) -> None:
        if event not in self._subscribers:
            self._subscribers[event] = []
        self._subscribers[event].append(callback)

    def publish(self, event: str, *args: Any, **kwargs: Any) -> None:
        if event in self._subscribers:
            for callback in self._subscribers[event]:
                callback(*args, **kwargs)

```

And then each MF Form resource could publish on delete:

```
class EnviornmentService:
    def __init__(self, event_system: DataallEventSystem):
        self.event_system = event_system
    
    ...

    def delete_environment(self, uri: str) -> None:
        self.event_system.publish("environment_deleted", uri)
```



And then MetadataFromService and AttachedMetadataFormService could import event system and subscribe:
```
class MetadataFormService:
    @classmethod
    def subscribe(cls, event_system: DataallEventSystem) -> None:
        event_system.subscribe("environment_deleted", cls.on_delete_env)
        event_system.subscribe("organization_deleted", cls.on_delete_org)
        event_system.subscribe("dataset_deleted", cls.on_delete_dataset)

    @classmethod
    def add_dependency(self, cls):
        self.dependencies.add(cls)

    def on_delete_env(self, env_uri):
        # Delete MF Forms and Attached Forms for env_uri
    ...

```

And lastly `__init__.py` for MF Module Interface would call:

```
MetadataFormService.subscribe(DataallEventSystem)
AttachedMetadataFormService(DataallEventSystem)
```


Not sure if this pub/sub approach is best but this way we can have multiple publishers and multiple subscribers (similar to EnvironmentResourceManager but not dedicated to just Environments and more flexible)


Just thinking out loud here - not sure if this is the right approach for the ask 
cc @SofiaSazonova @dlpzx @petrkalos

backend/dataall/modules/metadata_forms/services/attached_metadata_form_service.py
noah-paige: you can use `param_name` to resolve this, for example something similar to below:

```
    @ResourcePolicyService.has_resource_permission(
        ATTACH_METADATA_FORM, parent_resource=_get_entity_uri, param_name='data'
    )
    def create_attached_metadata_form(...):
```

where:
```
     def _get_entity_uri(data):
          reutnr data.get('entityUri')
```

`update_table_column_description` is an example where this `param_name` is used

backend/dataall/modules/metadata_forms/db/metadata_form_repository.py
noah-paige, line 1: in testing I saw that if I create 2 MF Forms for 2 dif Orgs and then try to attach one of the forms to a resource (i.e. a dataset) they both come up

this is because of L78 where we show the form if the user is the owner. I think we should just remove that as part of the filter which will resolve this bug on available form listing

noah-paige (commented):
Left some comments - after some minor fixes testing and functionality look good

Left a big comment on design of the delete triggers but on second thought it may not be worth the effort to implement it the way I propose unless we really see a need for future work on MF forms or elsewhere

Triggers seem easy enough to create / delete so easily reversible too if needed

backend/dataall/modules/metadata_forms/db/metadata_form_repository.py
dlpzx: If not used let's remove it

dlpzx:
> With the new way of handling deletes of metadata forms with deletes of resources a couple things come to mind:
> 
>     * (1) We still have a dependency between MF module and other resources (i.e. metadata form tables must exist for org, env, dataset deletes to work) but since we always create all tables in DB (and I do not think we plan to change this behavior in the foreseeable future) I think this is okay
> 
>     * (2) Right now when we create a new form or attach a new form we do not create any permissions for the user/group on the form. I
>       
>       * Instead we protect backend resolvers for metadata forms by either the `MetadataFormAccessService.can_perform()` method (which checks if owner or not of the visibility entity) OR checking if has group has ATTACH_METADATA_FORM or CREATE_METADATA_FORM on the resource itself
>       * I see the addition of some new permissions like UPDATE_METADATA_FORM_FIELD, EDIT_METADATA_FORM, etc. that are not currently used but if we do plan on assigning these permissions on the form moving forward this would have to also be cleaned up in DB trigger (just calling out the complexity of these triggers could expand)
> 
> 
> What if we had a generic EventSystem class (in core or base?):
> 
> ```
> class DataallEventSystem:
>     def __init__(self):
>         self._subscribers = {}
> 
>     def subscribe(self, event: str, callback: Callable[..., None]) -> None:
>         if event not in self._subscribers:
>             self._subscribers[event] = []
>         self._subscribers[event].append(callback)
> 
>     def publish(self, event: str, *args: Any, **kwargs: Any) -> None:
>         if event in self._subscribers:
>             for callback in self._subscribers[event]:
>                 callback(*args, **kwargs)
> ```
> 
> And then each MF Form resource could publish on delete:
> 
> ```
> class EnviornmentService:
>     def __init__(self, event_system: DataallEventSystem):
>         self.event_system = event_system
>     
>     ...
> 
>     def delete_environment(self, uri: str) -> None:
>         self.event_system.publish("environment_deleted", uri)
> ```
> 
> And then MetadataFromService and AttachedMetadataFormService could import event system and subscribe:
> 
> ```
> class MetadataFormService:
>     @classmethod
>     def subscribe(cls, event_system: DataallEventSystem) -> None:
>         event_system.subscribe("environment_deleted", cls.on_delete_env)
>         event_system.subscribe("organization_deleted", cls.on_delete_org)
>         event_system.subscribe("dataset_deleted", cls.on_delete_dataset)
> 
>     @classmethod
>     def add_dependency(self, cls):
>         self.dependencies.add(cls)
> 
>     def on_delete_env(self, env_uri):
>         # Delete MF Forms and Attached Forms for env_uri
>     ...
> ```
> 
> And lastly `__init__.py` for MF Module Interface would call:
> 
> ```
> MetadataFormService.subscribe(DataallEventSystem)
> AttachedMetadataFormService(DataallEventSystem)
> ```
> 
> Not sure if this pub/sub approach is best but this way we can have multiple publishers and multiple subscribers (similar to EnvironmentResourceManager but not dedicated to just Environments and more flexible)
> 
> Just thinking out loud here - not sure if this is the right approach for the ask cc @SofiaSazonova @dlpzx @petrkalos
Answering to the downsides of using database triggers:
(1) I do not think it will be a problem for our own development. 
(2) If triggers go further than deletes on a dependent table - if a delete implements business logic outside of database rules (foreign keys) I think we should avoid them. The main reason is that they can be difficult to debug.
(3) Additional concern - As a developer I might find it difficult to understand how metadata forms are deleted, it would not occur to me to look into each migration script for business logic code. Could we move the specific trigger definition to the metadata forms module? In addition, for the case we need to update a trigger, if we have the trigger code in the module we could track the history more easily (using git) than searching inside the migration scripts.


Finally, regarding @noah-paige pub/sub model vs current db triggers; db triggers are good enough for MF now. We should revisit the decision:
- if MF needs to be applied to new/unexpected resources (e.g. customer external modules)
- if the db triggers need to implement a complex logic or need to be updated often
In those cases I think the pub/sub model is a great candidate, it would be great if we could substitute other interfaces we are currently using for module communication (Environments - Resources, Datasets - Shares) and have a  generic way to handle intermodule event dependencies

SofiaSazonova:
@noah-paige @dlpzx 
I like the idea of pub/sub model. We can revisit EnvironmentResources as well to unify the idea of intermodule communications.
But I don't think, we should include it into MF Prs. It's quite a big work on it's own, that requires a lot of tests as well.
I tend to keep triggers for now and then revisit it with new mechanism when it's ready.

noah-paige:
One additional bug found during testing (thanks to @rbernotas) - when a user a part of the admin group tries to view a Metadata Tab for a resource, for instance a data.all Organization, the use receives an error `'NoneType' object is not iterable` back from `listEntityMetadataForm` 
- This error stems from  L144 at `dataall/backend/dataall/modules/metadata_forms/db/metadata_form_repository.py` (i.e. `orgs = list(set(user_org_uris).intersection(set(entity_orgs_uris)))`)
- It is because `user_org_uris` and `user_env_uris` are set to `None` by `get_user_admin_status_orgs_and_envs_()` for admin users
- I think we need to do something similar to what we did for `entity_org_uris` and `entity_envs_uris` and set the following:
```
        user_orgs_uris =  user_orgs_uris or []
         user_envs_uris = user_envs_uris or []
```

in `query_entity_metadata_forms`

backend/migrations/versions/427db8f31999_backfill_MF_resource_permissions.py
SofiaSazonova: done

backend/dataall/modules/metadata_forms/db/metadata_form_repository.py
SofiaSazonova: removed

backend/migrations/versions/075d344ae2cc_mf_triggers.py
```diff
@@ -87,6 +87,25 @@ def upgrade():
...
+            RETURN OLD;
+        END;
+        $$ LANGUAGE plpgsql;
+
+        -- Create the trigger for dataset table
+        CREATE TRIGGER metadata_form_delete_trigger
+        BEFORE DELETE ON metadata_form
+        FOR EACH ROW
+        EXECUTE FUNCTION metadata_form_delete_trigger_function();
+    """)
```
noah-paige, line 108: is this trigger used anywhere? Do we grant permissions on Metadata Form resources? 

From what I see we only grant permissions to CREATE or ATTACH Form on the parent resource (i.e. env, org, etc.) and then for UPDATE, DELETE, or similar on form itself we just check if owner or not

backend/dataall/modules/metadata_forms/db/metadata_form_repository.py
```diff
@@ -140,6 +143,8 @@ def query_entity_metadata_forms(
 
         entity_orgs_uris = entity_orgs_uris or []
         entity_envs_uris = entity_envs_uris or []
+        user_org_uris = user_org_uris or []
+        user_env_uris = user_env_uris or []
```
noah-paige, line 147: with this change if I am an admin - `user_org_uris` and `user_env_uris` will always be empty List and therefore `orgs` and `envs` will be empty and I will not be able to attach any org or env visibility forms to my resource

While it makes sense that admins are treated special for `query_user_metadata_forms` to see all - I think either:
 - Admins should be treated same as any other user for `query_entity_metadata_forms`
 - OR if admin - `orgs = entity_orgs_uris or []` and `envs = entity_envs_uris or []` ignoring the `user_org_uris` and `user_env_uris`

backend/migrations/versions/427db8f31999_backfill_MF_resource_permissions.py
```diff
@@ -70,12 +70,54 @@ def upgrade():
...
+            session=session,
+            group=env.SamlGroupName,
+            resource_uri=env.environmentUri,
+            resource_type=Environment.__name__,
+            permissions=[ATTACH_METADATA_FORM, CREATE_METADATA_FORM],
+        )
+        for policy in policies:
+            for permission in policy.permissions:
+                session.delete(permission)
+                session.commit()
```
noah-paige, line 95: this deletes all permissions on the `resource_policy`

When downgrading and re-upgrading my user no longer had other Env permissions like `GET_ENVIRONMENT` and similar - need to add a filter here (likely a bug in the other migration script where used as well)

noah-paige:
Thanks for the latest changes @SofiaSazonova. 3 pending items I am seeing that I think need to be resolved for this PR:

- (1) Downgrade Migration Script introduces errors ([link](https://github.com/data-dot-all/dataall/pull/1540#discussion_r1772025624))
- (2) DA Admins can not attach Org or Env visibility filters to an Entity ([link](https://github.com/data-dot-all/dataall/pull/1540#discussion_r1772011954))
- (3) Too many Forms showing as available when listing for an entity because of query filters ([link](https://github.com/data-dot-all/dataall/pull/1540/files/f63d539a44716527fbaf533865a970402475e2cc..7a04e7d3572af0d5e4a493b63e0f094cf65754de#r1759456519))


++ 1 `nit` comment on triggers migration ([link](https://github.com/data-dot-all/dataall/pull/1540/files/f63d539a44716527fbaf533865a970402475e2cc..7a04e7d3572af0d5e4a493b63e0f094cf65754de#r1771980371))

Afterwards I can re-test and think should be good for this PR

Additionally - there are some additional enhancements or fixes that we may want to pick up for MF Forms. I will add to the open issue https://github.com/data-dot-all/dataall/issues/1553

SofiaSazonova:
@noah-paige 
1. Downgrades: I added fixes for this migration and redshift one (discussed it with @dlpzx )
2. I rolled back the queries as they were a couple iterations ago. May be it's not short, but it works as it should. It's also not bulky. That solves both problems with visibility, you mentioned
3. About MF permissions: I just want to have all triggers in sam place, so I left it there.

backend/migrations/versions/427db8f31999_backfill_MF_resource_permissions.py
noah-paige: - Latest downgrade has a bug - ResourcePolicyPermission has no field `name`
    -  The following resolves for each instance

```
        for policy in policies:
            for resource_pol_permission in policy.permissions:
                if resource_pol_permission.permission.name in [ATTACH_METADATA_FORM, CREATE_METADATA_FORM]:
                    session.delete(resource_pol_permission)
                    session.commit()
```

noah-paige:
> Thanks for the latest changes @SofiaSazonova. 3 pending items I am seeing that I think need to be resolved for this PR:
> 
>     * (1) Downgrade Migration Script introduces errors ([link](https://github.com/data-dot-all/dataall/pull/1540#discussion_r1772025624))
> 
>     * (2) DA Admins can not attach Org or Env visibility filters to an Entity ([link](https://github.com/data-dot-all/dataall/pull/1540#discussion_r1772011954))
> 
>     * (3) Too many Forms showing as available when listing for an entity because of query filters ([link](https://github.com/data-dot-all/dataall/pull/1540/files/f63d539a44716527fbaf533865a970402475e2cc..7a04e7d3572af0d5e4a493b63e0f094cf65754de#r1759456519))
> 
> 
> ++ 1 `nit` comment on triggers migration ([link](https://github.com/data-dot-all/dataall/pull/1540/files/f63d539a44716527fbaf533865a970402475e2cc..7a04e7d3572af0d5e4a493b63e0f094cf65754de#r1771980371))
> 
> Afterwards I can re-test and think should be good for this PR
> 
> Additionally - there are some additional enhancements or fixes that we may want to pick up for MF Forms. I will add to the open issue #1553

@SofiaSazonova - latest testing 

- (1) Bug in Downgrade ([link](https://github.com/data-dot-all/dataall/pull/1540#discussion_r1773864992))
- (2) Fixed
- (3) Fixed
- ++ sounds good on keeping the trigger in migration

backend/migrations/versions/427db8f31999_backfill_MF_resource_permissions.py
SofiaSazonova: Oh, stupid me.

SofiaSazonova:
@noah-paige I hope, this is the last one and I haven't missed smth now.

Issue #1342, opened 2024-06-18
Worksheet Run Query - 504 Gateway Timeout


sandeephs1:
### Describe the bug

Running query in worksheet errors with the below messages -
"Network error occurred", Network error: Failed to fetch"

### How to Reproduce

Create a dataset, load dataset with parquet data
create a worksheet, run a query to fetch data from the dataset.
If the query is limited with 10 or 5000, it gets completed
if the limit exceeds certain number like 15000 etc it fails with error "Network error occurred", Network error: Failed to fetch"
NOTE: Limit number in the query cannot be determined

### Expected behavior

Worksheet should fetch the result and display it

### OS

Mac Sonoma 14.3

### Python version

3.9.13

### AWS data.all version

2.0.0

### Additional context

We noticed, if data is big, Run query ends up with “504 Gateway Timeout” error in Application console, though Athena Completes the query execution.

dlpzx:
Hi @sandeephs1, sorry for the late response. This error can be caused by API Gateway. 504 Gateway Timeout error is a very typical error that appears when the API takes longer than 29s to respond. What that means is that even if Lambda has a timeout of 15 minutes, it needs to return a response to the API in less than 29s

dlpzx:
HI @sandeephs1 any update on this issue?

sandeephs1:
Hi @dlpzx yew we were able to handle it, made async. 
For us it is 2 features - "Run Query" and "Download" the result output. Currently we are testing it. Will revert once the testing is complete

zsaltys:
I think it makes more sense to do a run + poll.. vs just pure RPC... The download approach also makes sense. To generate the result in some file on S3 .. The UI can offer a limited amount of lines preview of that file maybe 10K top lines.. And if you want more you can download the entire result. We should make sure this is super tight with security/permissions so users can't download something they're not supposed to etc.

dlpzx:
Hi @zsaltys, I like the approach of limiting the query result and using S3 to download the complete result. The actual query query results are stored in the Athena workgroup location:
- teamA in environmentA has a team IAM role roleA and a workgroup wgA created by data.all
- in Worksheets data.all uses that team IAM role and workgroup to run the query
- the results are stored in the workgroup S3 location = environment S3 bucket/ athenaQueries / teamAFolder/
- roleA can only use the workgroup wgA and can only access the S3 folder envBucket/ athenaQueries / teamAFolder/

We could make use of these pretty tight security set up to download results.

Issue #360, opened 2023-03-10
Add ability to download Worksheet results (CSV/XLS/Clipboard/Export)


rbernotas:
**Is your idea related to a problem? Please describe.**

The ability to query datasets through the worksheets is a great feature, making data.all more usable as a single pane of glass for users.  While once a user has been granted access to data, they may be able to work more directly with the data from S3 within an AWS account, non-technical or less technical users in an organization may need the ability to download query results directly from data.all, without having to work directly with the AWS account, in order to more easily retrieve result data as reports.

**Describe the solution you'd like**

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

Add a feature to the Worksheet result data display, where the user can choose to download the result data they have generated as a CSV formatted file.  XLS format, Copy results to clipboard, and other export options will also be useful.  The feature could be presented to the user as a menu or button which prompts the user for a local download location, and then processes the output file for the user to download.

dlpzx:
Hi @rbernotas thank you for adding it :) We will have a look

mvidhu:
Our organisation users are also interested in this feature.

anmolsgandhi:
#293

rbernotas:
Referencing https://github.com/data-dot-all/dataall/issues/293 as well, I do think we will have interest in this feature, as despite the security concerns, it is a use case that we already regularly support with other platforms. One thing we have done in the past is to limit the number of rows that can be downloaded at a given time, so it's not so easy for someone to just select star grab all the data from a table and walk away with it. Ideally the CSV download would be a configurable feature, but also configurable in the maximum number of rows that can be downloaded for a query result. The idea being, that the CSV download would be for more ad-hoc usage / reporting, as opposed to bulk data export and data transfer (which would not be desirable).

We also had another set of use cases for CSV download that could be applied with such export support. It would be nice to be able to CSV export info about all the datasets in the catalog, or even your search result from the catalog (again, for reporting purposes, so teams can pass around the list of onboarded datasets, accounts that have been onboarded, etc.). Same for environments and organizations, and perhaps also available shares. This same functionality could be used in those cases as well. And future auditing feature where we can pull activity data from RDS can also benefit from this.

SofiaSazonova:
I have several thoughts on this feature:

1. We can do it as another group-permission. So, not everyone will have an opportunity to download the data


[...]

Issue #293, opened 2023-02-07
Download button from data.all UI for shared folders/tables


dlpzx:
This feature has been requested by a customer. In their organization non-technical personas are also data.all users. These users will request access to data and will need to download the particular files shared with them. 

**Describe the solution you'd like**
At the moment in data.all users can get the files by getting AWS credentials and running AWS CLI commands against the S3 access point of the folder shared with them. But ideally, they would be provided a button "download" to get this files directly from the UI.

**Describe alternatives you've considered**
This can be achieved in a similar way that the "Upload" functionality in the Dataset UI for Dataset owners. 
 - Add "download" button next to shared tables and folders (for example in the sharedItems table)
 - Create API call and write boto3 call in S3 handlers.

Open question: should we download one file, the complete folder content, a fix number of files? 

**Additional context**

*P.S. Please Don't attach files. Add code snippets directly in the message body instead.*

mvidhu:
We are also interested in this feature

dlpzx:
@anmolsgandhi to keep in mind if there is spare-time in v2.1.0

anmolsgandhi:
Prioritizing for v2.2.0, if there is spare time in v2.1.0 we can triage and bump it up.

zsaltys:
@dlpzx We would definitely not be interested in this feature for security reasons as it gives attackers a direct way to access data. I would like to ask that the permissions for this feature and the feature itself are behind a feature flag that can be enabled/disabled. This can be the default if we want to but we should release it in a state where it can be disabled.

rbernotas:
I do think we will have interest in this feature, as despite the security concerns, it is a use case that we already regularly support with other platforms.  One thing we have done in the past is to limit the number of rows that can be downloaded at a given time, so it's not so easy for someone to just select star grab all the data from a table and walk away with it.  Ideally the CSV download would be a configurable feature, but also configurable in the maximum number of rows that can be downloaded for a query result.  The idea being, that the CSV download would be for more ad-hoc usage / reporting, as opposed to bulk data export and data transfer (which would not be desirable).

[...]

PR #1547, opened 2024-09-13
Query result download

anilsoni7:
### Feature or Bugfix

- Feature

### Detail
- teamA in environmentA has a team IAM role roleA and a workgroup wgA created by data.all
- in Worksheets data.all uses that team IAM role and workgroup to run the query
- the results are stored in the workgroup S3 location = environment S3 bucket/ athenaQueries / teamAFolder/
- Create the presigned url for the csv present under the s3 location with 2 hours of expiration
- Storing presigned urls generated in database so that can be reused in future

### Relates
- https://github.com/data-dot-all/dataall/issues/293
- https://github.com/data-dot-all/dataall/issues/360
- https://github.com/data-dot-all/dataall/issues/1342

### Testcase

- [x]  tests/modules/worksheets/test_worksheet.py/test_create_query_download_url

### Security

fetching data from storage outside the application (e.g. a database, an S3 bucket)? - Yes
  - Is the input sanitized?
  - Is injection prevented by parametrizing queries? - Yes
  - Have you ensured no `eval` or similar functions are used? - Yes
- Does this PR introduce any functionality or component that requires authorization? - Yes
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms? - Yes
- Are you using or adding any cryptographic features? - N/A
  - Do you use a standard proven implementations? - Yes
  - Are the used keys controlled by the customer? Where are they stored? - N/A

### DIFF
--- a/backend/dataall/base/aws/s3_client.py
+++ b/backend/dataall/base/aws/s3_client.py
@@ -3,6 +3,7 @@ from botocore.config import Config
 
 from botocore.exceptions import ClientError
 import logging
+from dataall.base.db.exceptions import AWSResourceNotFound
 
 log = logging.getLogger(__name__)
 
@@ -32,4 +33,32 @@ class S3_client:
             return presigned_url
         except ClientError as e:
             log.error(f'Failed to get presigned URL due to: {e}')
+            raise e
+
+    @staticmethod
+    def object_exists(region, bucket, key) -> bool:
+        try:
+            S3_client.client(region, None).head_object(Bucket=bucket, Key=key)
+            return True
+        except ClientError as e:
+            log.error(f'Failed to check object existence due to: {e}')
+            if e.response['Error']['Code'] == '404':
+                return False
+            raise AWSResourceNotFound('s3_object_exists', f'Object {key} not found in bucket {bucket}')
+
+    @staticmethod
+    def put_object(region, bucket, key, body):
+        try:
+            S3_client.client(region, None).put_object(Bucket=bucket, Key=key, Body=body)
+        except ClientError as e:
+            log.error(f'Failed to put object due to: {e}')
+            raise e
+
+    @staticmethod
+    def get_object(region, bucket, key):
+        try:
+            response = S3_client.client(region, None).get_object(Bucket=bucket, Key=key)
+            return response['Body'].read().decode('utf-8')
+        except ClientError as e:
+            log.error(f'Failed to get object due to: {e}')
             raise e
--- a/backend/dataall/modules/worksheets/api/resolvers.py
+++ b/backend/dataall/modules/worksheets/api/resolvers.py
@@ -4,6 +4,7 @@ from dataall.modules.worksheets.db.worksheet_models import Worksheet
 from dataall.modules.worksheets.db.worksheet_repositories import WorksheetRepository
 from dataall.modules.worksheets.services.worksheet_service import WorksheetService
 from dataall.base.api.context import Context
+from dataall.modules.worksheets.services.worksheet_query_result_service import WorksheetQueryResultService
 
 
 def create_worksheet(context: Context, source, input: dict = None):
@@ -69,3 +70,10 @@ def run_sql_query(context: Context, source, environmentUri: str = None, workshee
 def delete_worksheet(context, source, worksheetUri: str = None):
     with context.engine.scoped_session() as session:
         return WorksheetService.delete_worksheet(session=session, uri=worksheetUri)
+
+
+def create_athena_query_result_download_url(context: Context, source, input: dict = None):
+    WorksheetQueryResultService.validate_input(input)
+
+    with context.engine.scoped_session() as session:
+        return WorksheetQueryResultService.download_sql_query_result(session=session, data=input)
--- a/backend/dataall/modules/worksheets/api/types.py
+++ b/backend/dataall/modules/worksheets/api/types.py
@@ -87,14 +87,16 @@ WorksheetQueryResult = gql.ObjectType(
     fields=[
         gql.Field(name='worksheetQueryResultUri', type=gql.ID),
         gql.Field(name='queryType', type=gql.NonNullableType(gql.String)),
-        gql.Field(name='sqlBody', type=gql.NonNullableType(gql.String)),
+        gql.Field(name='sqlBody', type=gql.String),
         gql.Field(name='AthenaQueryId', type=gql.NonNullableType(gql.String)),
         gql.Field(name='region', type=gql.NonNullableType(gql.String)),
         gql.Field(name='AwsAccountId', type=gql.NonNullableType(gql.String)),
-        gql.Field(name='AthenaOutputBucketName', type=gql.NonNullableType(gql.String)),
-        gql.Field(name='AthenaOutputKey', type=gql.NonNullableType(gql.String)),
-        gql.Field(name='timeElapsedInSecond', type=gql.NonNullableType(gql.Integer)),
+        gql.Field(name='ElapsedTimeInMs', type=gql.Integer),
         gql.Field(name='created', type=gql.NonNullableType(gql.String)),
+        gql.Field(name='downloadLink', type=gql.String),
+        gql.Field(name='OutputLocation', type=gql.String),
+        gql.Field(name='expiresIn', type=gql.AWSDateTime),
+        gql.Field(name='fileFormat', type=gql.String),
     ],
 )
 
--- a/backend/dataall/modules/worksheets/db/worksheet_repositories.py
+++ b/backend/dataall/modules/worksheets/db/worksheet_repositories.py
@@ -53,3 +53,17 @@ class WorksheetRepository(EnvironmentResource):
             page=data.get('page', WorksheetRepository._DEFAULT_PAGE),
             page_size=data.get('pageSize', WorksheetRepository._DEFAULT_PAGE_SIZE),
         ).to_dict()
+
+    @staticmethod
+    def find_query_result_by_format(
+        session, worksheet_uri: str, athena_query_id: str, file_format: str
+    ) -> WorksheetQueryResult:
+        return (
+            session.query(WorksheetQueryResult)
+            .filter(
+                WorksheetQueryResult.worksheetUri == worksheet_uri,
+                WorksheetQueryResult.AthenaQueryId == athena_query_id,
+                WorksheetQueryResult.fileFormat == file_format,
+            )
+            .first()
+        )
--- a/backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
+++ b/backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
@@ -1,0 +1,139 @@
+import csv
+import io
+import os
+from datetime import datetime, timedelta
+from typing import TYPE_CHECKING
+
+from openpyxl import Workbook
+
+from dataall.base.aws.s3_client import S3_client
+from dataall.base.db import exceptions
+from dataall.core.environment.services.environment_service import EnvironmentService
+from dataall.modules.worksheets.db.worksheet_models import WorksheetQueryResult
+from dataall.modules.worksheets.db.worksheet_repositories import WorksheetRepository
+from dataall.modules.worksheets.services.worksheet_service import WorksheetService
+
+if TYPE_CHECKING:
+    try:
+        from sqlalchemy.orm import Session
+        from mypy_boto3_s3.client import S3Client
+    except ImportError:
+        print('skipping type checks as stubs are not installed')
+        S3Client = None
+        Session = None
+
+
+class WorksheetQueryResultService:
+    SupportedFormats = {'csv', 'xlsx'}
+
+    @staticmethod
+    def validate_input(data):
+        if not data:
+            raise exceptions.InvalidInput('data', data, 'input is required')
+        if not data.get('athenaQueryId'):
+            raise exceptions.RequiredParameter('athenaQueryId')
+        if not data.get('fileFormat'):
+            raise exceptions.RequiredParameter('fileFormat')
+        if data.get('fileFormat', '').lower() not in WorksheetQueryResultService.SupportedFormats:
+            raise exceptions.InvalidInput(
+                'fileFormat', data.get('fileFormat'), ', '.join(WorksheetQueryResultService.SupportedFormats)
+            )
+
+    @staticmethod
+    def get_output_bucket(session: 'Session', environment_uri: str) -> str:
+        environment = EnvironmentService.get_environment_by_uri(session, environment_uri)
+        bucket = environment.EnvironmentDefaultBucketName
+        return bucket
+
+    @staticmethod
+    def create_query_result(
+        environment_bucket: str, athena_workgroup: str, worksheet_uri: str, region: str, aws_account_id: str, data: dict
+    ) -> WorksheetQueryResult:
+        sql_query_result = WorksheetQueryResult(
+            worksheetUri=worksheet_uri,
+            AthenaQueryId=data.get('athenaQueryId'),
+            fileFormat=data.get('fileFormat'),
+            OutputLocation=f's3://{environment_bucket}/athenaqueries/{athena_workgroup}/',
+            region=region,
+            AwsAccountId=aws_account_id,
+            queryType='data',
+        )
+        return sql_query_result
+
+    @staticmethod
+    def get_file_key(
+        workgroup: str, query_id: str, file_format: str = 'csv', athena_queries_dir: str = 'athenaqueries'
+    ) -> str:
+        return f'{athena_queries_dir}/{workgroup}/{query_id}.{file_format}'
+
+    @staticmethod
+    def convert_csv_to_xlsx(csv_data) -> io.BytesIO:
+        wb = Workbook()
+        ws = wb.active
+        csv_reader = csv.reader(csv_data.splitlines())
+        for row in csv_reader:
+            ws.append(row)
+
+        excel_buffer = io.BytesIO()
+        wb.save(excel_buffer)
+        excel_buffer.seek(0)
+        return excel_buffer
+
+    @staticmethod
+    def handle_xlsx_format(output_bucket: str, file_key: str) -> bool:
+        aws_region_name = os.getenv('AWS_REGION_NAME', 'eu-west-1')
+        file_name, _ = file_key.split('.')
+        csv_data = S3_client.get_object(region=aws_region_name, bucket=output_bucket, key=f'{file_name}.csv')
+        excel_buffer = WorksheetQueryResultService.convert_csv_to_xlsx(csv_data)
+        S3_client.put_object(region=aws_region_name, bucket=output_bucket, key=file_key, body=excel_buffer)
+        return True
+
+    @staticmethod
+    def download_sql_query_result(session: 'Session', data: dict = None):
+        # # default timeout for the download link is 2 hours(in minutes)
+        default_timeout = os.getenv('QUERY_RESULT_TIMEOUT_MINUTES', 120)
+
+        environment = EnvironmentService.get_environment_by_uri(session, data.get('environmentUri'))
+        worksheet = WorksheetService.get_worksheet_by_uri(session, data.get('worksheetUri'))
+        env_group = EnvironmentService.get_environment_group(
+            session, worksheet.SamlAdminGroupName, environment.environmentUri
+        )
+        output_file_key = WorksheetQueryResultService.get_file_key(
+            env_group.environmentAthenaWorkGroup, data.get('athenaQueryId'), data.get('fileFormat')
+        )
+        sql_query_result = WorksheetRepository.find_query_result_by_format(
+            session, data.get('worksheetUri'), data.get('athenaQueryId'), data.get('fileFormat')
+        )
+        if data.get('fileFormat') == 'xlsx':
+            is_job_failed = WorksheetQueryResultService.handle_xlsx_format(
+                environment.EnvironmentDefaultBucketName, output_file_key
+            )
+            if is_job_failed:
+                raise ValueError('Error while preparing the xlsx file')
+
+        if not sql_query_result:
+            sql_query_result = WorksheetQueryResultService.create_query_result(
+                environment.EnvironmentDefaultBucketName,
+                env_group.environmentAthenaWorkGroup,
+                worksheet.worksheetUri,
+                environment.region,
+                environment.AwsAccountId,
+                data,
+            )
+        S3_client.object_exists(
+            region=environment.region, bucket=environment.EnvironmentDefaultBucketName, key=output_file_key
+        )
+        if sql_query_result.is_download_link_expired():
+            url = S3_client.get_presigned_url(
+                region=environment.region,
+                bucket=environment.EnvironmentDefaultBucketName,
+                key=output_file_key,
+                expire_minutes=default_timeout,
+            )
+            sql_query_result.downloadLink = url
+            sql_query_result.expiresIn = datetime.utcnow() + timedelta(seconds=default_timeout)
+
+        session.add(sql_query_result)
+        session.commit()
+
+        return sql_query_result
--- a/backend/migrations/versions/d1d6da1b2d67_add_columns_worksheet_query_result_model.py
+++ b/backend/migrations/versions/d1d6da1b2d67_add_columns_worksheet_query_result_model.py
@@ -1,0 +1,37 @@
+"""add_columns_worksheet_query_result_model
+
+Revision ID: d1d6da1b2d67
+Revises: d274e756f0ae
+Create Date: 2024-09-10 14:34:31.492186
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'd1d6da1b2d67'
+down_revision = 'd274e756f0ae'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.alter_column('worksheet_query_result', 'status', nullable=True)
+    op.alter_column('worksheet_query_result', 'sqlBody', nullable=True)
+    op.alter_column('worksheet_query_result', 'DataScannedInBytes', type_=sa.BigInteger(), nullable=True)
+    op.add_column('worksheet_query_result', sa.Column('downloadLink', sa.String(), nullable=True))
+    op.add_column('worksheet_query_result', sa.Column('expiresIn', sa.DateTime(), nullable=True))
+    op.add_column('worksheet_query_result', sa.Column('updated', sa.DateTime(), nullable=False))
+    op.add_column('worksheet_query_result', sa.Column('fileFormat', sa.String(), nullable=True))
+
+
+def downgrade():
+    op.drop_column('worksheet_query_result', 'fileFormat')
+    op.drop_column('worksheet_query_result', 'updated')
+    op.drop_column('worksheet_query_result', 'expiresIn')
+    op.drop_column('worksheet_query_result', 'downloadLink')
+    op.alter_column('worksheet_query_result', 'DataScannedInBytes', type_=sa.Integer(), nullable=True)
+    op.alter_column('worksheet_query_result', 'sqlBody', nullable=False)
+    op.alter_column('worksheet_query_result', 'status', nullable=False)
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -16,3 +16,4 @@ requests_aws4auth==1.1.1
 sqlalchemy==1.3.24
 starlette==0.36.3
 alembic==1.13.1
+openpyxl==3.1.5
--- a/frontend/src/modules/Worksheets/components/WorksheetResult.js
+++ b/frontend/src/modules/Worksheets/components/WorksheetResult.js
@@ -12,12 +12,66 @@ import {
   TableRow
 } from '@mui/material';
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useState, useCallback } from 'react';
 import { FaBars } from 'react-icons/fa';
 import * as ReactIf from 'react-if';
 import { Scrollbar } from 'design';
 
-export const WorksheetResult = ({ results, loading }) => {
+import Stack from '@mui/material/Stack';
+import Radio from '@mui/material/Radio';
+import RadioGroup from '@mui/material/RadioGroup';
+import FormControlLabel from '@mui/material/FormControlLabel';
+import { Download } from '@mui/icons-material';
+import { LoadingButton } from '@mui/lab';
+import {
+  useClient
+} from 'services';
+import {
+  createWorksheetQueryResultDownloadUrl
+} from '../services';
+import { SET_ERROR, useDispatch } from 'globalErrors';
+
+export const WorksheetResult = ({ results, loading, currentEnv, athenaQueryId, worksheetUri }) => {
+  const [runningDownloadQuery, setRunningDownloadQuery] = useState(false);
+  const [fileType, setFileType] = useState('csv');
+  const client = useClient();
+  const dispatch = useDispatch();
+
+  const handleChange = (event) => {
+    setFileType(event.target.value);
+  };
+
+  const downloadFile = useCallback(async () => {
+    try {
+      setRunningDownloadQuery(true);
+      const response = await client.query(
+        createWorksheetQueryResultDownloadUrl({
+          fileFormat: fileType,
+          environmentUri: currentEnv.environmentUri,
+          athenaQueryId: athenaQueryId,
+          worksheetUri: worksheetUri
+        })
+      );
+
+      if (!response.errors) {
+        const link = document.createElement('a');
+        link.href = response.data.createWorksheetQueryResultDownloadUrl.downloadLink;
+        // Append to html link element page
+        document.body.appendChild(link);
+        // Start download
+        link.click();
+        // Clean up and remove the link
+        link.parentNode.removeChild(link);
+      } else {
+        dispatch({ type: SET_ERROR, error: response.errors[0].message });
+      }
+    } catch (e) {
+      dispatch({ type: SET_ERROR, error: e.message });
+    } finally {
+      setRunningDownloadQuery(false);
+    }
+  }, [client, dispatch, currentEnv, athenaQueryId, fileType])
+
   if (loading) {
     return <CircularProgress />;
   }
@@ -33,6 +87,34 @@ export const WorksheetResult = ({ results, loading }) => {
               <Box>
                 <FaBars /> Query Results
               </Box>
+            }
+            action={
+              <>
+                <Stack direction="row" spacing={2}>
+                  <RadioGroup
+                    row
+                    aria-labelledby="demo-row-radio-buttons-group-label"
+                    name="row-radio-buttons-group"
+                    value={fileType}
+                    onChange={handleChange}
+                  >
+                    <FormControlLabel value="csv" control={<Radio />} label="CSV" />
+                    <FormControlLabel value="xlsx" control={<Radio />} label="XLSX" />
+                  </RadioGroup>
+
+                  <LoadingButton
+                    loading={runningDownloadQuery}
+                    color="primary"
+                    onClick={downloadFile}
+                    startIcon={<Download fontSize="small" />}
+                    sx={{ m: 1 }}
+                    variant="contained"
+                  >
+                    Download
+                  </LoadingButton>
+                </Stack>
+
+              </>
             }
           />
           <Divider />
--- a/frontend/src/modules/Worksheets/services/createWorksheetQueryResultDownloadUrl.js
+++ b/frontend/src/modules/Worksheets/services/createWorksheetQueryResultDownloadUrl.js
@@ -1,0 +1,40 @@
+import { gql } from 'apollo-boost';
+
+export const createWorksheetQueryResultDownloadUrl = ({
+  fileFormat,
+  environmentUri,
+  athenaQueryId,
+  worksheetUri
+}) => ({
+  variables: {
+    fileFormat,
+    environmentUri,
+    athenaQueryId,
+    worksheetUri
+  },
+  query: gql`
+    mutation CreateWorksheetQueryResultDownloadUrl(
+      $fileFormat: String!
+      $environmentUri: String!
+      $athenaQueryId: String!
+      $worksheetUri: String!
+    ) {
+      createWorksheetQueryResultDownloadUrl(
+        input: {
+          fileFormat: $fileFormat
+          environmentUri: $environmentUri
+          athenaQueryId: $athenaQueryId
+          worksheetUri: $worksheetUri
+        }
+      ) {
+        downloadLink
+        AthenaQueryId
+        expiresIn
+        fileFormat
+        OutputLocation
+      }
+    }
+  `
+});
+
+
--- a/tests/modules/worksheets/test_worksheet.py
+++ b/tests/modules/worksheets/test_worksheet.py
@@ -1,5 +1,9 @@
 import pytest
 
+from unittest.mock import MagicMock
+
+from future.backports.datetime import datetime
+
 from dataall.modules.worksheets.api.resolvers import WorksheetRole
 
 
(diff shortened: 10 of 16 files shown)


backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+        from mypy_boto3_s3.client import S3Client
+    except ImportError:
+        print('skipping type checks as stubs are not installed')
+        S3Client = None
+        Session = None
+
+
+class WorksheetQueryResultService:
+    SupportedFormats = {'csv', 'xlsx'}
+
```
dlpzx, line 28: As a convention, in data.all backed the `api` layer is responsible of the input validation of the API calls; in the `service` layer we verify permissions and execute business logic. Which means that in this case we should place lines 29-40 in `dataall/modules/worksheets/api/resolvers.py`

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+            raise exceptions.InvalidInput('data', data, 'input is required')
+        if not data.get('athenaQueryId'):
+            raise exceptions.RequiredParameter('athenaQueryId')
+        if not data.get('fileFormat'):
+            raise exceptions.RequiredParameter('fileFormat')
+        if data.get('fileFormat', '').lower() not in WorksheetQueryResultService.SupportedFormats:
+            raise exceptions.InvalidInput(
+                'fileFormat', data.get('fileFormat'), ', '.join(WorksheetQueryResultService.SupportedFormats)
+            )
+
```
dlpzx, line 41: This method `get_output_bucket` seems unused, are we safe to we remove it?

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+        from sqlalchemy.orm import Session
+        from mypy_boto3_s3.client import S3Client
+    except ImportError:
+        print('skipping type checks as stubs are not installed')
+        S3Client = None
+        Session = None
+
+
+class WorksheetQueryResultService:
+    SupportedFormats = {'csv', 'xlsx'}
```
dlpzx, line 27: Let's use an Enum for the supported formats to avoid string issues. To avoid circular dependencies with `api`, we should move `dataall/modules/worksheets/api/enums.py` to `backend/dataall/modules/worksheets/services/worksheet_enums.py` and add
```
class WorksheetResultsFormat(GraphQLEnumMapper:
    CSV = 'csv'
    XLSX= 'xlsx'
```

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+
+from openpyxl import Workbook
+
+from dataall.base.aws.s3_client import S3_client
+from dataall.base.db import exceptions
+from dataall.core.environment.services.environment_service import EnvironmentService
+from dataall.modules.worksheets.db.worksheet_models import WorksheetQueryResult
+from dataall.modules.worksheets.db.worksheet_repositories import WorksheetRepository
+from dataall.modules.worksheets.services.worksheet_service import WorksheetService
+
```
dlpzx, line 15: Why cannot we use the imported S3_Client and the Session directly in the type checks?

backend/migrations/versions/d1d6da1b2d67_add_columns_worksheet_query_result_model.py
```diff
@@ -0,0 +1,37 @@
...
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'd1d6da1b2d67'
+down_revision = 'd274e756f0ae'
+branch_labels = None
+depends_on = None
+
+
```
dlpzx, line 19: We should also drop the now unused columns: sqlBody, queryType?

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+        environment_bucket: str, athena_workgroup: str, worksheet_uri: str, region: str, aws_account_id: str, data: dict
+    ) -> WorksheetQueryResult:
+        sql_query_result = WorksheetQueryResult(
+            worksheetUri=worksheet_uri,
+            AthenaQueryId=data.get('athenaQueryId'),
+            fileFormat=data.get('fileFormat'),
+            OutputLocation=f's3://{environment_bucket}/athenaqueries/{athena_workgroup}/',
+            region=region,
+            AwsAccountId=aws_account_id,
+            queryType='data',
```
dlpzx, line 59: I think we do not longer need queryType. We can safely remove it from the db model and avoid writing it here

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+        return sql_query_result
+
```
dlpzx, line 62: nit: in line 65 use Enum value instead of directly writing `csv` and for the athena_queries_dir we could set a constant at the top level of the file, and re-use it
```
ATHENA_QUERIES_PATH = 'athenaqueries'

class WorksheetQueryResultService:
    @staticmethod
    def get_environment_bucket(environment) -> str:
```

backend/dataall/base/aws/s3_client.py
dlpzx, line 1: Data.all "typical" AWS clients are boto3 clients that assume the dataallPivotRole (and optionally other roles) in the Environment accounts. There are examples in almost all modules, for example for S3 there is `backend/dataall/modules/s3_datasets/aws/s3_profiler_client.py` that assumes the pivot role; the `dataall/modules/s3_datasets/aws/s3_location_client.py` that assumes the dataset role and the `dataall/modules/s3_datasets/aws/s3_dataset_client.py` that assumes the dataset role specifying the VPC endpoint and config.

The S3_Client in this file is a "special" type of client in data.all as it assumes the role of the backend resource that is executing it. The use-case for this client is to read files in S3 Buckets in the infra account where the backend resources are. 

In the use-case of WorksheetsQueryResults we need to read from the environment bucket which is in the environment account, so we need to use a "typical" boto3 client, otherwise we would get AccessDenied errors

dlpzx: You can create an S3Client in `worksheers/aws`

backend/dataall/base/aws/s3_client.py
```diff
@@ -33,3 +34,31 @@ def get_presigned_url(region, bucket, key, expire_minutes: int = 15):
         except ClientError as e:
             log.error(f'Failed to get presigned URL due to: {e}')
             raise e
+
+    @staticmethod
+    def object_exists(region, bucket, key) -> bool:
+        try:
+            S3_client.client(region, None).head_object(Bucket=bucket, Key=key)
+            return True
+        except ClientError as e:
```
dlpzx, line 43: nit: the error logs here might be a bit misleading:
- 404 Not Found is "ok", we should just return False
- for everything else we could raise an `UnauthorizedOperation` for the action `CHECK_S3_OBJECT_EXISTS` and log the error

backend/requirements.txt
```diff
@@ -15,4 +15,5 @@ requests==2.32.2
 requests_aws4auth==1.1.1
 sqlalchemy==1.3.24
 starlette==0.36.3
-alembic==1.13.1
\ No newline at end of file
+alembic==1.13.1
+openpyxl==3.1.5
```
dlpzx, line 19: I agree on using `openpyxl` :) Just curious, did you choose it because it does not require MS Excel to be installed?

backend/dataall/modules/worksheets/api/types.py
```diff
@@ -87,14 +87,16 @@
...
         gql.Field(name='queryType', type=gql.NonNullableType(gql.String)),
-        gql.Field(name='sqlBody', type=gql.NonNullableType(gql.String)),
+        gql.Field(name='sqlBody', type=gql.String),
         gql.Field(name='AthenaQueryId', type=gql.NonNullableType(gql.String)),
         gql.Field(name='region', type=gql.NonNullableType(gql.String)),
         gql.Field(name='AwsAccountId', type=gql.NonNullableType(gql.String)),
-        gql.Field(name='AthenaOutputBucketName', type=gql.NonNullableType(gql.String)),
-        gql.Field(name='AthenaOutputKey', type=gql.NonNullableType(gql.String)),
-        gql.Field(name='timeElapsedInSecond', type=gql.NonNullableType(gql.Integer)),
+        gql.Field(name='ElapsedTimeInMs', type=gql.Integer),
```
dlpzx, line 94: super nit: except names (AWS, Athena) name fields with initial letter in lowercase

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+            )
+
+    @staticmethod
+    def get_output_bucket(session: 'Session', environment_uri: str) -> str:
+        environment = EnvironmentService.get_environment_by_uri(session, environment_uri)
+        bucket = environment.EnvironmentDefaultBucketName
+        return bucket
+
+    @staticmethod
+    def create_query_result(
```
dlpzx, line 49: nit: internal functions should be named with prefix `_` (e.g. `_create_query_result`)

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @staticmethod
+    def handle_xlsx_format(output_bucket: str, file_key: str) -> bool:
+        aws_region_name = os.getenv('AWS_REGION_NAME', 'eu-west-1')
+        file_name, _ = file_key.split('.')
+        csv_data = S3_client.get_object(region=aws_region_name, bucket=output_bucket, key=f'{file_name}.csv')
+        excel_buffer = WorksheetQueryResultService.convert_csv_to_xlsx(csv_data)
+        S3_client.put_object(region=aws_region_name, bucket=output_bucket, key=file_key, body=excel_buffer)
+        return True
+
```
dlpzx, line 90: external functions used by resolvers should be decorated with permission checks. In other words, we need to ensure from the server-side that the API call to download_sql_query_results can only be executed by users with permissions.

```
    @staticmethod
    @ResourcePolicyService.has_resource_permission(RUN_ATHENA_QUERY)
    def download_sql_query_result(session, uri, data: dict = None):
```

dlpzx: I think the `RUN_ATHENA_QUERY` permission can be reused. If a user has permissions to run the query, then he should have permissions to download the results. It can be implemented later if needed, but do you think there will be a case in which we need to distinguish between users that can run queries and users that can download the results?

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+            fileFormat=data.get('fileFormat'),
+            OutputLocation=f's3://{environment_bucket}/athenaqueries/{athena_workgroup}/',
+            region=region,
+            AwsAccountId=aws_account_id,
+            queryType='data',
+        )
+        return sql_query_result
+
+    @staticmethod
+    def get_file_key(
```
dlpzx, line 64: super nit: rename to a more meaningful name `build_file_key_string`

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+                key=output_file_key,
+                expire_minutes=default_timeout,
+            )
+            sql_query_result.downloadLink = url
+            sql_query_result.expiresIn = datetime.utcnow() + timedelta(seconds=default_timeout)
+
+        session.add(sql_query_result)
+        session.commit()
+
+        return sql_query_result
```
dlpzx, line 139: Suggestion to improve download_sql_query_results (some are very nit)
1. first step, read data from db
```
        # # default timeout for the download link is 2 hours(in minutes)
        default_timeout = os.getenv('QUERY_RESULT_TIMEOUT_MINUTES', 120)

        environment = EnvironmentService.get_environment_by_uri(session, data.get('environmentUri'))
        worksheet = WorksheetService.get_worksheet_by_uri(session, data.get('worksheetUri'))
        env_group = EnvironmentService.get_environment_group(
            session, worksheet.SamlAdminGroupName, environment.environmentUri
        )
        sql_query_result = WorksheetRepository.find_query_result_by_format(
            session, data.get('worksheetUri'), data.get('athenaQueryId'), data.get('fileFormat')
        )
```
2. create sql_query_results, right after checking it and use it from now on instead of reading from data
```
        if not sql_query_result:
            sql_query_result = WorksheetQueryResultService.create_query_result(
                environment.EnvironmentDefaultBucketName,
                env_group.environmentAthenaWorkGroup,
                worksheet.worksheetUri,
                environment.region,
                environment.AwsAccountId,
                data,
            )
```
3.  build output_file_key without format to avoid creating string and then splitting it + Use enum in comparison and sql_query_results.fileFormat + avoid unnecessary middle step `handle_xlsx_format` and read directly. The if clause for `is_job_failed` won't be reached if there is an exception. 
```
        output_file_key = WorksheetQueryResultService.get_file_key(
            env_group.environmentAthenaWorkGroup, data.get('athenaQueryId')
        )

        if sql_query_results.fileFormat == WorksheetResultsFormat.XLSX.value:
            csv_data = S3_client.get_object(region=environment.region, bucket=environment.EnvironmentDefaultBucketName, key=f'{output_file_key}..{WorksheetResultsFormat.CSV.value}')
            excel_buffer = WorksheetQueryResultService.convert_csv_to_xlsx(csv_data)
            S3_client.put_object(region=environment.region, bucket=environment.EnvironmentDefaultBucketName, key=f'{output_file_key}.{WorksheetResultsFormat.XLSX.value}', body=excel_buffer)


```
4. Check if s3 file exists and if the download link is expired create a new one. Add and commit changes
```

        S3_client.object_exists(
            region=environment.region, bucket=environment.EnvironmentDefaultBucketName, key=f'{output_file_key}.{sql_query_result.fileFormat}'
        )
        if sql_query_result.is_download_link_expired():
            url = S3_client.get_presigned_url(
                region=environment.region,
                bucket=environment.EnvironmentDefaultBucketName,
                key=f'{output_file_key}.{sql_query_result.fileFormat}',
                expire_minutes=default_timeout,
            )
            sql_query_result.downloadLink = url
            sql_query_result.expiresIn = datetime.utcnow() + timedelta(seconds=default_timeout)

        session.add(sql_query_result)
        session.commit()

        return sql_query_result
```

frontend/src/modules/Worksheets/services/createWorksheetQueryResultDownloadUrl.js
dlpzx, line 1: This API call can be simplified by using the `WorksheetQueryResultDownloadUrlInput` defined in the backend API definition. The same we do for createDatasets, where we do not declare each variable, but just the input as `NewDatasetInput`

```
export const createDataset = (input) => {
  return {
    variables: {
      input
    },
    mutation: gql`
      mutation CreateDataset($input: NewDatasetInput!) {
        createDataset(input: $input) {
          datasetUri
          label
          userRoleForDataset
        }
      }
    `
  };
};
```

frontend/src/modules/Worksheets/components/WorksheetResult.js
dlpzx, line 1: the frontend files are not formatted according to prettier. Once I approve the PR they will run into formatting issues. To check and fix formatting issues you can run the following commands from frontend/package.json respectively:
```
    "lint": "eslint --ext js src",
    "lint:fix": "eslint --fix --ext js src",
```

dlpzx: or directly in the terminal inside the `frontend/` folder: `npm run lint:fix`

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+        bucket = environment.EnvironmentDefaultBucketName
+        return bucket
+
+    @staticmethod
+    def create_query_result(
+        environment_bucket: str, athena_workgroup: str, worksheet_uri: str, region: str, aws_account_id: str, data: dict
+    ) -> WorksheetQueryResult:
+        sql_query_result = WorksheetQueryResult(
+            worksheetUri=worksheet_uri,
+            AthenaQueryId=data.get('athenaQueryId'),
```
dlpzx, line 54: There is a small bug: the AthenaQueryId is the primary key of the table, which means that there cannot be 2 records with the same AthenaQueryId value. If a user tries to first download the results for CSV and then for XLSX the database will throw an error:
Screenshot 2024-09-16 at 18 29 30

tests/modules/worksheets/test_worksheet.py
```diff
@@ -30,6 +34,38 @@ def worksheet(client, tenant, group):
...
+
+    # s3_client.client.return_value = s3_client
+
+    s3_client().object_exists.return_value = True
+    s3_client().put_object.return_value = None
+    s3_client().get_object.return_value = '123,123,123'
+    s3_client.get_presigned_url.return_value = 'https://s3.amazonaws.com/file/123.csv'
+    yield s3_client
+
+
```
dlpzx, line 50: Remove commented code if it is not used

tests/modules/worksheets/test_worksheet.py
```diff
@@ -30,6 +34,38 @@ def worksheet(client, tenant, group):
     return response.data.createWorksheet
 
 
+@pytest.fixture(scope='module', autouse=True)
+def mock_s3_client(module_mocker):
+    s3_client = MagicMock()
+    module_mocker.patch('dataall.modules.worksheets.services.worksheet_query_result_service.S3_client', s3_client)
+
```
dlpzx, line 41: nit: use autospec=True in mocker.patch method - autospec ensures that only functions defined in the class are patched. For example if in the tests we call the funciton s3_client.do_something it will throw and error with autospec, while with just MagicMock it won't

tests/modules/worksheets/test_worksheet.py
```diff
@@ -30,6 +34,38 @@ def worksheet(client, tenant, group):
...
+    # s3_client.client.return_value = s3_client
+
```
dlpzx, line 43: I think the correct syntax is `s3_client.get_presigned_url.return_value` without the brackets

tests/modules/worksheets/test_worksheet.py
```diff
@@ -145,3 +181,36 @@ def test_update_worksheet(client, worksheet, group):
...
+            'athenaQueryId': '123',
+            'fileFormat': 'csv',
+            'environmentUri': env_fixture.environmentUri,
+        },
+    )
+
+    expires_in = datetime.strptime(response.data.createWorksheetQueryResultDownloadUrl.created, '%Y-%m-%d %H:%M:%S.%f')
+    assert response.data.createWorksheetQueryResultDownloadUrl.downloadLink is not None
+    assert response.data.createWorksheetQueryResultDownloadUrl.fileFormat == 'csv'
+    assert expires_in > datetime.utcnow()
```
dlpzx, line 216: After implementing the permission check, see comment in the service file, we need to implement a test scenario for the unauthorized not-happy-path

dlpzx (changes requested):
Hi @anilsoni7, first of all, thank you for opening the PR! I have left some comments and requested changes, but it is less than it looks. The technical design of how results are downloaded and the user experience are great :) 🚀 Feel free to request more info in the comments if anything is unclear

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: The imported S3_client has limited functionality implemented. The auto complete provided with respected to imported S3_Client would be limited.
on other hand s3_client from mypy provides the types for every operation supported under boto3 s3 client.

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: done

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: check the comment in [backend/dataall/modules/worksheets/services/worksheet_query_result_service.py](https://github.com/data-dot-all/dataall/pull/1547/files/01d681db39722f3ad04493271cf2b02533789e92#diff-cdc94a4ea25d93c1b5b7ae5bf723c13b0b1fcb2e9dd392764ec0645b655c039c)

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: changes are done

tests/modules/worksheets/test_worksheet.py
anilsoni7: I tried without brackets but that didn't worked and was throwing error.
so fall backed to the approach which is working

backend/dataall/base/aws/s3_client.py
anilsoni7: "typical" AWS S3 client has been created under `worksheets/aws` which supports the operations required for download features

backend/dataall/base/aws/s3_client.py
anilsoni7: changes are done, implemented under `worksheets/aws/s3_client.py`

backend/requirements.txt
anilsoni7: Yes! We want to convert the CSV data to XLSX programatically. Therefore does not require MS Excel interface.
Therefore `openpyxl` was chosen.
There are different library which support the same operations and has good performance as compared to `openpyxl` but `openpyxl` provides the stable library while maintain the good performance

backend/dataall/modules/worksheets/api/types.py
anilsoni7: changes are done

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: Right now! I do not see any such use-case for distinguishing the user for running the query and downloading the result.
We can safely proceed ahead with `RUN_ATHENA_QUERY` permission

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: changes are done

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: changes are done

frontend/src/modules/Worksheets/services/createWorksheetQueryResultDownloadUrl.js
anilsoni7: changes done

backend/migrations/versions/d1d6da1b2d67_add_columns_worksheet_query_result_model.py
anilsoni7: we can drop queryType column from the table.
sqlBody and status can be kept as an reserved columns which will be used in future

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: removed queryType from model

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,139 @@
...
+            ws.append(row)
+
+        excel_buffer = io.BytesIO()
+        wb.save(excel_buffer)
+        excel_buffer.seek(0)
+        return excel_buffer
+
+    @staticmethod
+    def handle_xlsx_format(output_bucket: str, file_key: str) -> bool:
+        aws_region_name = os.getenv('AWS_REGION_NAME', 'eu-west-1')
```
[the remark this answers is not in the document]
anilsoni7: refer to the comments on `backend/dataall/base/aws/s3_client.py`

backend/migrations/versions/d1d6da1b2d67_add_columns_worksheet_query_result_model.py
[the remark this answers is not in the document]
anilsoni7: the revision has been updated and migrations are made into sequence

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: created a new columns named `worksheetQueryResultUri` which is assigned primary key

tests/modules/worksheets/test_worksheet.py
anilsoni7: commented code is removed

tests/modules/worksheets/test_worksheet.py
anilsoni7: changes are done

frontend/src/modules/Worksheets/components/WorksheetResult.js
anilsoni7: frontend code is formatted using `npm run lint:fix`

backend/dataall/modules/worksheets/api/resolvers.py
```diff
@@ -73,7 +73,18 @@ def delete_worksheet(context, source, worksheetUri: str = None):
 
 
 def create_athena_query_result_download_url(context: Context, source, input: dict = None):
-    WorksheetQueryResultService.validate_input(input)
+
+    if not input:
+        # raise exceptions.InvalidInput('data', input, 'input is required')
```
dlpzx, line 77: Remove commented code

backend/dataall/modules/worksheets/aws/s3_client.py
```diff
@@ -0,0 +1,73 @@
...
+    try:
+        from mypy_boto3_s3 import S3Client as S3ClientType
+    except ImportError:
+        S3ClientType = None
+
+log = logging.getLogger(__name__)
+
+
+class S3Client:
+
```
dlpzx: just curious: why are you defining type hints with quotes? (e.g. lin 23: env: 'Environment'):)

backend/dataall/base/aws/s3_client.py
dlpzx: Now the Client methods `object_exists`, `put_object` and  `get_object` in the generic base/aws/s3_client.py are not used anywhere in the repo. We should delete them from here to avoid dead code

backend/dataall/modules/worksheets/aws/s3_client.py
```diff
@@ -0,0 +1,71 @@
+import boto3
+from botocore.config import Config
```
dlpzx, line 2: remove unused imports (line 1 and 2)

backend/dataall/modules/worksheets/db/worksheet_repositories.py
```diff
@@ -53,3 +53,17 @@ def paginated_user_worksheets(session, username, groups, uri, data=None, check_p
...
+    ) -> WorksheetQueryResult:
+        return (
+            session.query(WorksheetQueryResult)
+            .filter(
+                WorksheetQueryResult.worksheetUri == worksheet_uri,
+                WorksheetQueryResult.AthenaQueryId == athena_query_id,
+                WorksheetQueryResult.fileFormat == file_format,
+            )
+            .first()
+        )
```
dlpzx, line 69: This query can be optimized by using the primary key of the table. We could directly use a `find_query_result_by_uri` function as 

```
    @staticmethod
    def find_worksheet_result_by_uri(session, uri: str) -> WorksheetQueryResult
        return session.query(WorksheetQueryResult).get(uri)
```

backend/migrations/versions/d1d6da1b2d67_add_columns_worksheet_query_result_model.py
```diff
@@ -0,0 +1,44 @@
...
+# revision identifiers, used by Alembic.
+revision = 'd1d6da1b2d67'
+down_revision = 'f87aecc36d39'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.alter_column('worksheet_query_result', 'status', nullable=True)
```
dlpzx, line 21: Don't we need to add the `worksheetQueryResultUri` column? or was it already created?

backend/dataall/modules/worksheets/services/worksheet_permissions.py
```diff
@@ -38,12 +38,17 @@
...
 ENVIRONMENT_INVITATION_REQUEST.append(RUN_ATHENA_QUERY)
 
 ENVIRONMENT_ALL.append(RUN_ATHENA_QUERY)
+ENVIRONMENT_ALL.append(RUN_ATHENA_QUERY_TENANT)
 
 RESOURCES_ALL.append(RUN_ATHENA_QUERY)
 RESOURCES_ALL_WITH_DESC[RUN_ATHENA_QUERY] = 'Run Worksheet Athena queries on this environment'
+
+TENANT_ALL.append(RUN_ATHENA_QUERY_TENANT)
+TENANT_ALL_WITH_DESC[RUN_ATHENA_QUERY_TENANT] = 'Run Worksheet Athena queries on any environment'
```
dlpzx, line 54: The permission `RUN_ATHENA_QUERY_TENANT` is not used for what the definition says. It is used to restrict who can download the results of a worksheet. It would be better to have a DOWNLOAD_ATHENA_QUERY_RESULTS permission

dlpzx: or re-use the RUN_ATHENA_QUERY permission

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @staticmethod
+    def _create_query_result(
+        environment_bucket: str, athena_workgroup: str, worksheet_uri: str, region: str, aws_account_id: str, data: dict
+    ) -> WorksheetQueryResult:
+        sql_query_result = WorksheetQueryResult(
+            worksheetUri=worksheet_uri,
+            AthenaQueryId=data.get('athenaQueryId'),
+            fileFormat=data.get('fileFormat'),
+            OutputLocation=f's3://{environment_bucket}/athenaqueries/{athena_workgroup}/',
```
dlpzx, line 41: nit: use `_DEFAULT_ATHENA_QUERIES_PATH` instead of hardcoded `athenaqueries` in line 41

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
```diff
@@ -0,0 +1,121 @@
...
+        for row in csv_reader:
+            ws.append(row)
+
+        excel_buffer = io.BytesIO()
+        wb.save(excel_buffer)
+        excel_buffer.seek(0)
+        return excel_buffer
+
+    @staticmethod
+    @TenantPolicyService.has_tenant_permission(RUN_ATHENA_QUERY_TENANT)
```
dlpzx, line 66: Tenant permissions are generic permissions to restrict a complete module permissions. In this case we are defining permissions for a very particular functionality that affects a single worksheet.


We can use `MANAGE_WORKSHEETS` to restrict/grant access at the tenant level in a very generic way. And then add a resource policy `DOWNLOAD_ATHENA_QUERY_RESULT` that protects the worksheet. We need to modify the signature of the function and define the `uri` input, which should correspond to the worksheet_uri, which no longer will be passed in `data`.

```
@TenantPolicyService.has_tenant_permission(MANAGE_WORKSHEETS)
@ResourcePolicyService.has_resource_permission(DOWNLOAD_ATHENA_QUERY_RESULT)
def download_sql_query_result(session: 'Session', uri: str, env_uri: str, data: dict = None):
```

backend/dataall/modules/worksheets/services/worksheet_permissions.py
dlpzx: It should not be added to TENANT_ALL, instead is should be treated as any other Worksheet resource permission.  It should be defined in line 24 as `DOWNLOAD_ATHENA_QUERY_RESULTS = "DOWNLOAD_ATHENA_QUERY_RESULTS"` and then added to `WORKSHEET_ALL` and that's it for the definition

dlpzx: Read the comments on the usage of this permission

dlpzx (changes requested):
Hi @anilsoni7, the PR looks great, but there are still some changes regarding permissions that we need to solve

backend/dataall/modules/worksheets/api/resolvers.py
anilsoni7: done

backend/dataall/modules/worksheets/aws/s3_client.py
anilsoni7: There are two type of type hints available in python.

1. Backward Reference `env: Environment`
2. Forward Reference `env: 'Environment'`

The major difference between two is
**Backward Reference** requires Type hints to be present in the current local or global scope but
**Forward Reference** resolves type hints run time and does not requires type hints to be present in scope.

backend/dataall/modules/worksheets/aws/s3_client.py
```diff
@@ -0,0 +1,71 @@
...
+
+class S3Client:
+    def __init__(self, env: 'Environment'):
+        self._client = SessionHelper.remote_session(env.AwsAccountId, env.region).client('s3', region_name=env.region)
+        self._env = env
+
+    @property
+    def client(self) -> 'S3ClientType':
+        return self._client
+
```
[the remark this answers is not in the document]
anilsoni7: `S3Client.get_presigned_url` is used under `worksheets/services/worksheet_query_result_service.py` line no 108 to generate download url

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: done

tests/modules/worksheets/test_worksheet.py
```diff
@@ -30,6 +34,19 @@ def worksheet(client, tenant, group):
...
+    s3_client = module_mocker.patch(
+        'dataall.modules.worksheets.services.worksheet_query_result_service.S3Client', autospec=True
+    )
+
+    s3_client.object_exists.return_value = True
+    s3_client.put_object.return_value = None
+    s3_client.get_object.return_value = '123,123,123'
+    s3_client.get_presigned_url.return_value = 'https://s3.amazonaws.com/file/123.csv'
+    yield s3_client
+
```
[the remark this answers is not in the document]
anilsoni7: Please refer the comments on `backend/dataall/modules/worksheets/aws/s3_client.py`

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: done

backend/migrations/versions/d1d6da1b2d67_add_columns_worksheet_query_result_model.py
anilsoni7: Column is added, so silly of me.

backend/dataall/modules/worksheets/db/worksheet_repositories.py
anilsoni7: At the point where we are requesting to download the data we do not have `uri` for `WorksheetQueryResult`,
so we have to check if the url is already generated for `worksheetUri`, `athenaQueryId` and `fileFormat`.

so the query cannot be optimized only using `uri`. 
To optimize the query we can create index on `worksheetUri` and `fileFormat`

backend/dataall/modules/worksheets/services/worksheet_permissions.py
anilsoni7: Agreed

backend/dataall/modules/worksheets/services/worksheet_permissions.py
```diff
@@ -47,3 +45,13 @@
 
 RESOURCES_ALL.append(RUN_ATHENA_QUERY)
 RESOURCES_ALL_WITH_DESC[RUN_ATHENA_QUERY] = 'Run Worksheet Athena queries on this environment'
+
```
dlpzx, line 48: The code block from lines 48 to 57 is unnecessary. `DOWNLOAD_ATHENA_QUERY_RESULTS` is a permission that is applied on a Worksheet Resource.  `ENVIRONMENT_X` permissions always refer to permissions applied to Environment Resource. RESOURCES_ALL_WITH_DESC is already defined in line 32, but for Environment permissions such as `RUN_ATHENA_QUERY` in this case, we would want to display another description in the invite-team window.

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
dlpzx: Same applies to other `WorksheetQueryResultService` methods: `build_s3_file_path`, `convert_csv_to_xlsx`

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
dlpzx: At the end there is a dedicated permission applied on the worksheet, but I think it is good. We can resolve

backend/dataall/base/aws/s3_client.py
dlpzx: we should remove these methods in  backend/dataall/base/aws/s3_client.py as they are no longer used

frontend/src/modules/Worksheets/components/WorksheetResult.js
dlpzx: I built the frontend and it succeeds! Resolving

backend/dataall/modules/worksheets/db/worksheet_repositories.py
dlpzx: My bad, I understood it wrong! it is the WorksheetUri not the query results uri! Resolving

dlpzx (changes requested):
hi @anilsoni7, I reviewed and tested the functionality locally and resolved all the comments that are solved or were misunderstandings or questions. Can you take a look at the pending open comments? Thank you again for the good work :)

backend/dataall/base/aws/s3_client.py
anilsoni7: removed the unused code

backend/dataall/modules/worksheets/services/worksheet_query_result_service.py
anilsoni7: changes are done

backend/dataall/modules/worksheets/services/worksheet_permissions.py
anilsoni7: remove the unnecessary code

dlpzx: Hi Anil, I think you forgot to push the changes :) there are also some conflicts to resolve. Could you merge main to your branch?

anilsoni7:
Hi @dlpzx, I have pushed the latest changes resolving the comments which you mentioned. Can you review them

dlpzx (approved):
Looks good! Ready to be merged. Thank you @anilsoni7  🚀

anilsoni7:
Fixed 
- Integration test
- alembic test
- ruff checks & format


need to have a discussion with you regarding `npm audit` failing.
there is an package which is falling into critical severity

dlpzx:
Hi @anilsoni7 sorry for bothering you again. We have fixed the issue with the npm-audit, could you merge the `main` branch into this branch? There is also an issue on the migration scripts that needs to be fixed before merging

PR #1550, opened 2024-09-13
Add Permissions integration tests

dlpzx:
### Feature or Bugfix
- Feature: testing

### Detail
Implement tests for Permissions api calls (inside core/permissions) as part of https://github.com/data-dot-all/dataall/issues/1220

!Excludes updateSSMParameter mutation - I think it is unused

### Relates
- https://github.com/data-dot-all/dataall/issues/1220

### Security

  - Is the input sanitized?


tests_new/integration_tests/core/permissions/test_permissions.py
```diff
@@ -0,0 +1,55 @@
...
+        'UnauthorizedOperation', 'LIST_TENANT_TEAM_PERMISSIONS'
+    )
+
+
+def test_list_tenant_groups(clientTenant):
+    response = list_tenant_groups(clientTenant)
+    assert_that(response.count).is_greater_than_or_equal_to(4)
+    assert_that(response.nodes).is_not_empty()
+    admin_group = next(group for group in response.nodes if group.groupUri == 'DHAdmins')
+    assert_that(admin_group).contains_key('tenantPermissions')
```
noah-paige, line 30: nit: any reason we filter for `DHAdmins` and not just take the first entry like `response.nodes[0]` to ensure it contains `tenantPermissions` key?

noah-paige (approved):
Left one minor nit but this PR looks good and tested for checks passing

tests_new/integration_tests/core/permissions/test_permissions.py
dlpzx: The DHAdmins is a group created during deployment; it is the only one with the tenant permissions. It is true this is testing something else. We could break it down into 2 tests - test_tenant_group_exists and test_list_tenant_groups

dlpzx: I ended up adding a comment to explain the purpose of that last check

noah-paige: okay I think makes sense - may be no need to include `response.nodes[0]` since it is duplicate chck of `admin_group` which is testing the same and more, but PR looks good

PR #1556, opened 2024-09-17
Add Glossaries integration tests

dlpzx:
### Feature or Bugfix
- Feature:Testing

### Detail
Implement tests for Glossaries/Catalog as part of https://github.com/data-dot-all/dataall/issues/1220

⚠️ To test glossary associations, we need resources that use the glossary terms (associate a term to a resource). Possible resources include: s3_datasets, s3_tables, s3_folders, dashboards, redshift_datasets and redshift_tables. All of which are part of other modules that might be enabled or disabled. In this PR we assume that the s3_datasets module is enabled! If it was not enabled, then the glossary tests would fail!

During the implementation several enhancement ideas came up and are collected in #1557 

Tested in local with connection to real AWS deployment:
image

### Relates
- #1220 

### Security

  - Is the input sanitized?


tests_new/integration_tests/modules/catalog/conftest.py
```diff
@@ -0,0 +1,133 @@
...
+Session glossary elements needed if using associations
+
+WARNING!
+Associations are applied to the S3_Datasets module
+Glossaries can only be tested if the S3_datasets module is enabled in the deployment used for testing!
+"""
+
+
+@pytest.fixture(scope='session')
+def session_glossary1(client1, group1):
```
noah-paige, line 81: is there a reason to have 1 glossary to test CRUD operations (i.e. `glossary1`) and another glossary (i.e. `session_glossary1`) to test associations? Could this not all be handled with 1?

tests_new/integration_tests/modules/catalog/test_glossaries.py
```diff
@@ -0,0 +1,216 @@
...
+):
+    response = approve_term_association(client1, link_uri=dataset_association_with_glossary_term1.linkUri)
+    assert_that(response).is_true()
+    response = list_glossary_associations(client1, node_uri=session_glossary1.nodeUri)
+    association = next(
+        (n for n in response.associations.nodes if n.linkUri == dataset_association_with_glossary_term1.linkUri), None
+    )
+    assert_that(association.approvedBySteward).is_equal_to(True)
+
+
```
noah-paige, line 192: approve and dismiss term association we may want to test the unauthorized path - I believe only glossary owner or admin can approve or dismiss via `_verify_term_assocaition_approver_role()`

tests_new/integration_tests/modules/catalog/test_glossaries.py
```diff
@@ -0,0 +1,216 @@
...
+def test_start_reindex_catalog_unauthorized(client1):
+    assert_that(start_reindex_catalog).raises(GqlError).when_called_with(client1, handle_deletes=True).contains(
+        'Only data.all admin', 're-index catalog'
+    )
+
+
+def test_start_reindex_catalog_handle_deletes(clientTenant):
+    response = start_reindex_catalog(clientTenant, handle_deletes=True)
+    assert_that(response).is_true()
```
noah-paige, line 211: as a future enhancement (not this PR) may want a way to poll the ecs task to ensure it completes and then a way to ensure catalog populated appropriately

but out of scope for this PR imo

noah-paige (commented):
left some comments - mostly suggested additions but overall it looks good, will give a test and wait to hear your thoughts on the pending comments

noah-paige:
Can we also add to https://github.com/data-dot-all/dataall/issues/1557 to separate out glossary from catalog - do not think they should be the same module

tests_new/integration_tests/modules/catalog/conftest.py
```diff
@@ -0,0 +1,133 @@
...
+            'autoApprovalEnabled': False,
+            'enableExpiration': False,
+            'KmsAlias': session_s3_dataset1.KmsAlias,
+        },
+    )
+    response = list_glossary_associations(client1, node_uri=session_glossary1.nodeUri)
+    ds_association = next(
+        (assoc for assoc in response.associations.nodes if assoc.targetUri == session_s3_dataset1.datasetUri), None
+    )
+    yield ds_association
```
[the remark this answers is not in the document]
dlpzx: I did not do it because the dataset is deleted anyway; but thinking of it, it would be good to clean up the orphan associations

tests_new/integration_tests/modules/catalog/conftest.py
dlpzx: The idea was to be able to reuse the glossary tag dataset outside of the module. But I think it is better to keep it simple for the moment

tests_new/integration_tests/modules/catalog/test_glossaries.py
dlpzx: Great catch, you are right, the permissions are checked in the repositories layer!

dlpzx:
Implemented changes for the PR and checked again the tests:
image

noah-paige (approved):
looks good and tested -approving!

Issue #1528, opened 2024-09-10
Hide stack logs from all users except Data.all admin users 


TejasRGitHub:
**Is your idea related to a problem? Please describe.**
Stack logs ( Environment, Dataset ) and also share logs are accessible to some users based on if they are owner, business owner, stewards group etc.

These stack logs help in debugging any issues for a data.all developer or an admin but for other users do not provide much info. Apart from that the stack logs also output roles and stack data which might contain sensitive / secret information.

**Describe the solution you'd like**
To better protect information about AWS accounts and roles, it is better to provide a config which lets data.all admins decide if they want data.all users to see these logs. 

The configs allows data.all admins to only allow data.all admins to view the logs and no one else.

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

TejasRGitHub:
This would be configurable with config like : 
```json
{
    "modules": {
        ........,
        "s3_datasets": {
            "active": true,
            "features": {
                "file_uploads": true,
                "file_actions": true,
                "aws_actions": true,
                "preview_data": true,
                "glue_crawler": true,
                "show_logs" : false
            }
        },
        "s3_datasets_shares": {
            "active": true,
             "features": {
                 "show_logs" : false
             }
        },
       ,..
    },
    "core": {
        "features": {
            "env_aws_actions": false,
            .....,
            "show_logs" : false
        },
    }
}

```

noah-paige:
Hi @TejasRGitHub - currently in data.all DA Admins cannot see all data.all Envs, Dataasets, and other resources with IaC Stacks which the Admin team is not a part of and thus can not see the logs for these data.all resource stacks as well

For the above example when `show_logs` is set to `false` would this feature flag mean that non-Admin teams are no longer able to view logs only or would it also be additional changes to allow admins to not view data.all resources / logs from UI?

TejasRGitHub:
Hi @noah-paige , in our usecase , our admin team is a part of all the teams present in data.all. What we want to achieve is to block non-admin user(s) / group(s) from viewing the logs on data.all UI 

Thus, when show_logs : false would not show logs on UI to non-Admin users but it will still show logs to data.all admins

noah-paige:
@TejasRGitHub okay makes sense - to make it more generic I would have 3 options  for enabled/disabled/admin-only

but the config change sounds good

TejasRGitHub:
That's even better. Thanks @noah-paige. Will make the changes accordingly

PR #1559, opened 2024-09-17
[Gh-1528] Configurable stack logs display

TejasRGitHub:
### Feature or Bugfix
- Feature

### Detail

As described in this issue - https://github.com/data-dot-all/dataall/issues/1528 , stack logs provide some key info. 

Creating a feature config called `show_stack_logs` for s3_datasets, s3_dataset_shares, and core ( for environment stack logs ) and adding logic in frontend to hide/show the logs button. 
Using decorator in backend to restrict access based on config.

### Relates
- https://github.com/data-dot-all/dataall/issues/1528

### Security

  - Is the input sanitized?


config.json
```diff
@@ -47,11 +47,15 @@
...
+                "glue_crawler": true,
+                "show_stack_logs": "admin-only"
             }
         },
         "s3_datasets_shares": {
-            "active": true
+            "active": true,
+            "features": {
+                "show_stack_logs": "admin-only"
+            }
```
dlpzx, line 58: From the user experience  show_stack_logs is very similar to show_stack_logs, but implementation-wise they are pretty different. share-logs is a module specific feature that should be defined completely inside the shares module. 
- the config.json parameter should reflect the types of logs we are limiting the visibility for `show_stack_logs` vs `show_share_logs`
- `show_stack_logs` will be implemented in the Stacks core module and `show_share_logs` in the Shares module

backend/dataall/base/utils/stack_logs_utils.py
```diff
@@ -0,0 +1,33 @@
+from functools import wraps
+from dataall.base.config import config
+from dataall.base.context import get_context
+
+
```
dlpzx, line 5: See comment in config.json first!
We can make the utility more generic. At the moment it is very very focused in the stacks and it contains a lot of business info (where anything in `base` should be usable outside of data.all concepts, dataset, shares...)

We could add a utility similar to `dataall.base.feature_toggle_checker.is_feature_enabled`, maybe even in the same file

```
def is_feature_allowed_value(config_property: str, allowed_value: Any):
    def decorator(f):
        fn, fn_decorator = process_func(f)

        def decorated(*args, **kwargs):
            value = config.get_property(config_property)
            if value != allowed_value
                raise Exception(f'Disabled by config {config_property}')
            return fn(*args, **kwargs)

        return fn_decorator(decorated)

    return decorator
```

backend/dataall/modules/shares_base/services/share_logs_service.py
```diff
@@ -15,6 +16,7 @@
 
 class ShareLogsService:
     @staticmethod
+    @is_stack_logs_visible(targetType='shares')
```
dlpzx, line 19: With the utility proposed in the comment in stack_logs_util we could rewrite this as:
```
@is_feature_enabled('modules.shares_base.features.show_share_logs', allowed_values=['enabled', 'admin_only'])
```

config.json
dlpzx: - Since the logs seen are share generic logs the config.json parameter should not be in `s3_datasets_shares` but in `shares_base`

backend/dataall/modules/shares_base/services/share_logs_service.py
dlpzx: Checking if a user is part of the admin group is a pretty repeated task that could be further extracted in another utility. But that is step 2 :)

backend/dataall/modules/shares_base/services/share_logs_service.py
dlpzx, line 1: Shouldn't we also check the permissions and the visibility of the logs in `get_share_logs`?

backend/dataall/core/stacks/api/resolvers.py
```diff
@@ -64,6 +65,7 @@ def resolve_task_id(context, source: Stack, **kwargs):
         return source.EcsTaskArn.split('/')[-1]
```
dlpzx, line 67: Move this check to the service layer (`StackService.get_stack_logs`). Despite being a bit polluted in this case, resolvers should just check if the input is valid and call the corresponding service.

dlpzx: Check comments in `stack_logs_util` --->
we could use the decorator as:
```
def map_target_stack_to_config(target_type):
    if target_type == 'environment':
        return 'core.features.show_stack_logs'
    if target_type == 'dataset':
        return 'modules.s3_datasets.features.show_stack_logs'


@is_feature_allowed_values(resolve_config=map_target_stack_to_config, allowed_values=['enabled', 'admin_only'])
```

frontend/src/modules/Shared/Stack/Stack.js
```diff
@@ -96,11 +100,36 @@ export const Stack = (props) => {
       .catch((e) => dispatch({ type: SET_ERROR, error: e.message }));
   };
```
dlpzx, line 102: Instead of re-creating the logic of the backend in the frontend, we could do the same as in the `frontend/src/modules/Shares/views/ShareView.js` (for share.canViewLogs) and use a resolver in the stack object

dlpzx (changes requested):
Hi @TejasRGitHub thank you so much for the cool feature! I left some comments and suggestions!

config.json
TejasRGitHub: Hi @dlpzx , totally agreed on the naming part. I will make the changes. 
I can put in in another config json - `shares_base`

frontend/src/modules/Shared/Stack/Stack.js
TejasRGitHub: Will do

backend/dataall/modules/shares_base/services/share_logs_service.py
TejasRGitHub: That's a great catch. I will add the new decorator to it.

TejasRGitHub: Oh wait thought. `get_share_logs` has a check `ShareLogsService.check_view_log_permissions` and I am adding a decorator to `check_view_log_permissions` so this should take care of blocking the share logs access

backend/dataall/core/stacks/api/resolvers.py
TejasRGitHub: @dlpzx , thanks for all the code snippets to give me a very clear understanding. I now understand how you would want to make it generic. I really like it. 

Where do you think we should keep this `map_target_stack_to_config ` function since it is mapping specifics of dataset and environment. Should I place it here ?

backend/dataall/modules/shares_base/services/share_logs_service.py
TejasRGitHub: I see what you meant by checking the `get_share_logs` method. I have added a check in  `check_view_log_permissions` which will take care of the allow/disallowing user

backend/dataall/modules/shares_base/services/share_logs_service.py
TejasRGitHub: I have updated the code as per above comment. Thanks once again for a details explanation and providing a better way to do this. Also this aligns with the structure used in data.all ( base, core, modules )

backend/dataall/base/utils/stack_logs_utils.py
TejasRGitHub: @dlpzx , for now, I have added the decorator function sugggested by you in the same file. Please let me know if you have a preference to keep it in another file

TejasRGitHub:
Hi @dlpzx , Thanks a lot for the detailed review and suggestions . I have incorporated those changes and updated code. Please let me know if this looks good now

config.json
```diff
@@ -47,7 +47,14 @@
                 "file_actions": true,
                 "aws_actions": true,
                 "preview_data": true,
-                "glue_crawler": true
+                "glue_crawler": true,
+                "show_stack_logs": "admin-only"
```
dlpzx, line 51: The parameter `show_stack_logs` could be applied to any module that deploys a CloudFormation template upon module-object-creation. So in theory we could also add the `show_stack_logs` to the modules: ml_studio, notebooks and datapipelines. 
- important: what is the default value? does it change in comparison to the previous state for these modules?
- let's not forget to clearly explain this in the github pages :)

backend/dataall/modules/shares_base/services/share_logs_service.py
dlpzx: Sounds good, I will review the current implementation, but it makes sense

backend/dataall/base/utils/logs_utils.py
dlpzx, line 1: nit: rename this file to make it unrelated to logs (potentially we can use this in absolutely any feature!)
nit2: for the same reason rename `check_if_user_allowed_view_logs` to `check_if_user_is_admin`

backend/dataall/base/utils/logs_utils.py
```diff
@@ -0,0 +1,33 @@
...
+                raise Exception(
+                    f'Disabled since incorrect values in config {config_property_value}. Correct config values {allowed_values}'
+                )
+            return fn(*args, **kwargs)
+
+        return fn_decorator(decorated)
+
+    return decorator
+
+
```
dlpzx, line 29: I think I was not very clear with the purpose of `is_feature_has_allowed_values`. It was not only to check the validity of the config values (although I really like that you added it). It was to enable or disable a feature depending on the value, same as is_feature_enabled but with string values instead of booleans. So it would be more a `is_feature_enabled_values`. And in this case the enabled_values=['enabled', 'admin-only']. We will implement the difference between `enabled` and `admin-only` separately

backend/dataall/base/utils/logs_utils.py
```diff
@@ -0,0 +1,33 @@
...
+def check_if_user_allowed_view_logs(groups, config):
+    if (config == 'admin-only' and 'DAAdministrators' not in groups) or config == 'disabled':
+        return False
+    return True
```
dlpzx, line 33: I would make this generic and only check if the user is admin and move the 'admin-only' logic to the stack-logs code:

def check_if_user_is_admin(groups):
    if 'DAAdministrators' not in groups:
        return False
    return True

dlpzx: It is out of the scope for this PR (I'll open the issue once this one is completed) but the next step would be to assess all the places where we raise an Unauthorized error after checking if the user is admin and modify the utility accordingly (maybe it goes in the direction of another decorator that can be parametrized on config values)

backend/dataall/core/stacks/api/resolvers.py
```diff
@@ -57,6 +58,13 @@ def resolve_events(context, source: Stack, **kwargs):
     return json.dumps(source.events or {})
 
 
+def resolve_stack_visibility(context, source: Stack, **kwargs):
+    if not source:
+        return False
+    log_config = config.get_property(StackServiceUtils.map_target_to_config(target_type=source.stack))
+    return check_if_user_allowed_view_logs(context.groups, log_config)
```
dlpzx, line 65: Implement here the logic of if (config == 'admin-only' and not is_user_admin) or config == 'disabled':
        return False
    return True

backend/dataall/base/utils/logs_utils.py
dlpzx, line 33: Unless we make this logic very generic and turn it into a decorator that disables service calls depending of if user is admin (out of scope for this PR) this code should be implemented in the stack module.

dlpzx: In the same way there is a `ShareLogsService.check_view_log_permissions`

backend/dataall/core/stacks/api/resolvers.py
```diff
@@ -57,6 +58,13 @@ def resolve_events(context, source: Stack, **kwargs):
     return json.dumps(source.events or {})
```
dlpzx, line 60: We can take inspiration from the resolver in shares_base/api and add guardrails for exceptions + move all logic to service.

shares_base resolver:
```
def resolve_can_view_logs(context: Context, source: ShareObject):
    try:
        return ShareLogsService.check_view_log_permissions(context.username, context.groups, source.shareUri)
    except Exception as e:
        log.error(f'Failed to resolve if user can view share logs due to: {e}')
        return False
```
equivalent for resolve_stack_visibility:
```
```
def resolve_can_view_logs(context: Context, source: Stack, **kwargs):
   if not source:
        return False
    try:
        return StackService.check_if_user_allowed_view_logs(context.username, context.groups, source.stackUri)
    except Exception as e:
        log.error(f'Failed to resolve if user can view logs due to: {e}')
        return False
```

File: backend/dataall/modules/shares_base/services/share_logs_service.py [hunk 11]
```diff
@@ -15,8 +16,16 @@
...
+        allowed_values=['admin-only', 'enabled', 'disabled'],
+        config_property='modules.shares_base.features.show_share_logs',
+    )
     def check_view_log_permissions(username, groups, shareUri):
-        with get_context().db_engine.scoped_session() as session:
+        context = get_context()
+        log_config = config.get_property('modules.shares_base.features.show_share_logs')
+        if not check_if_user_allowed_view_logs(context.groups, log_config):
+            return False
+        with context.db_engine.scoped_session() as session:
```
dlpzx, line 28: replace line 26-27 by:
```
       if config == 'admin-only' and 'DAAdministrators' not in groups:
            return False
```

config.json
TejasRGitHub: @dlpzx, we don't use the MLstudio, notebooks, pipeline module and thus I didn't know there are stack logs for them. Let me check quickly how much is the effort. 

> important: we are not adding those modules in the map_target_to_config, what would happen to their stack logs?
That's a great point. For now, I can make it return None/ or some form of incorrect string so that `is_feature_allowe...` will throw and error. Also, in the error I can specify a way to implement this

> let's not forget to clearly explain this in the github pages :) 
Yes I will add those in the readme for sure.

backend/dataall/base/utils/logs_utils.py
TejasRGitHub: Yes it would be more like `is_feature_enable_on_allowed_values` ( I can rename the function according to what makes sense ). As you said, this decorator is just to check if the config has correct config values since we are departing from true/false ( boolean values ) . The logic to check based on the config value can be done in the service layer. 

Initially I was trying not to touch code in the service layer and just make use of decorator to allow/disallow some feature, but this was conflicting with how the code is structured.

backend/dataall/base/utils/logs_utils.py
TejasRGitHub: Will do.  Since the code was getting repeated , here`StackService.check_if_user_allowed_view_logs` and here `ShareLogsService.check_view_log_permissions` I thought of putting it here.

backend/dataall/base/utils/logs_utils.py
```diff
@@ -0,0 +1,33 @@
+from typing import Optional, Callable, List, Any
+from dataall.base.config import config
+from dataall.base.utils.decorator_utls import process_func
+
+
+def is_feature_has_allowed_values(
```
TejasRGitHub, line 6: Adding `default_value` to this function

backend/dataall/base/utils/logs.py
```diff
@@ -0,0 +1,32 @@
+from typing import Optional, Callable, List, Any
+from dataall.base.config import config
+from dataall.base.utils.decorator_utls import process_func
+
+
+def is_feature_has_allowed_values(
+    allowed_values: List[Any],
+    default_value: Any,
```
TejasRGitHub, line 8: Added this default value param

config.json
dlpzx: Redshift does not have an associated CDK stack; this is good and in case the parameter is not set it should default to enabled

backend/dataall/base/utils/logs.py
dlpzx, line 1: nit: rename file or move the code to `backend/dataall/base/feature_toggle_checker.py`

backend/dataall/modules/shares_base/services/share_logs_service.py
```diff
@@ -43,7 +57,7 @@ def get_share_logs_query(log_stream_name):
     @staticmethod
     def get_share_logs(shareUri):
         context = get_context()
-        if not ShareLogsService.check_view_log_permissions(context.username, context.groups, shareUri):
+        if not ShareLogsService.check_view_logs_permissions(context.username, context.groups, shareUri):
             raise exceptions.ResourceUnauthorized(
```
dlpzx, line 61: Small enhancement: the exception in lines 61-65 is not needed. The exception is raised in check_view_logs_permissions

backend/dataall/base/utils/logs.py
```diff
@@ -0,0 +1,32 @@
...
+                config_property_value = resolve_property(*args, **kwargs)
+            if config_property:
+                config_property_value = config_property
+            value = config.get_property(config_property_value, default_value)
+            if value not in allowed_values:
+                raise Exception(
+                    f'Disabled since incorrect values in config {config_property_value}. Correct config values {allowed_values}'
+                )
+            return fn(*args, **kwargs)
+
```
dlpzx, line 29: I thought we had agreed to use this to enable/disable feature, not for only validation of config params. 

```
def is_feature_enabled_for_values(
    allowed_values: List[Any],
    enabled_values: List[Any],
    default_value: Any,
    resolve_property: Optional[Callable] = None,
    config_property: Optional[str] = None,
):
    def decorator(f):
        fn, fn_decorator = process_func(f)

        def decorated(*args, **kwargs):
            config_property_value = None
            if config_property is None and resolve_property is None:
                raise Exception('Config property not provided')
            if resolve_property:
                config_property_value = resolve_property(*args, **kwargs)
            if config_property:
                config_property_value = config_property
            value = config.get_property(config_property_value, default_value)
            if value not in allowed_values:
                raise Exception(
                    f'Disabled since incorrect values in config {config_property_value}. Correct config values {allowed_values}'
                )
            if value not in enabled_values:
                raise Exception(f'Disabled by config for {value=} . Enabled config values {enabled_values}')
            return fn(*args, **kwargs)

        return fn_decorator(decorated)

    return decorator
```
This removes the need to implement enabled/disabled logic in the particular shares/log/X-service

backend/dataall/core/stacks/api/resolvers.py
dlpzx: Coming back to the example of shares_base/resolver ---> the resolver does not contain any business logic, it just calls the service. The service `check_if_user_allowed_view_logs` is then the one decorated and contains the permission logic. The service is used by the resolver and also by other services such as `get_stack_logs` which does not need to be decorated

dlpzx: class StackService:
    @is_feature_has_allowed_values(
        allowed_values=['admin-only', 'enabled', 'disabled'],
        enabled_values=['admin-only', 'enabled'],
        default_value='enabled',
        resolve_property=StackServiceUtils.map_target_to_config,
    )
    @staticmethod
    def check_if_user_allowed_view_logs(groups, target_type):
        config = config.get_property(StackServiceUtils.map_target_to_config(target_type), 'enabled')
        if config == 'admin-only' and 'DAAdministrators' not in groups:
            return False
        return True

backend/dataall/modules/shares_base/services/share_logs_service.py
TejasRGitHub: There is an exception raised inside the `check_view_logs_permissions`, in case the admin-only setting is on. If it is set to enabled, it should follow the previous logic of checking of the  user belongs to dataset owner and/or stewards groups. In this case, a false is returned .

backend/dataall/base/utils/logs.py
TejasRGitHub: @dlpzx, I was thinking that the decorator just be used as not a feature enable / disable checker but just to check if the values are correctly put. The logic to enable and disable would then be handled in the service layers. 
I like this idea in which it checks if the values are relating to enabled/disabled config values, will make the change. I forgot / didn't think like this

backend/dataall/core/stacks/api/resolvers.py
TejasRGitHub: I think I misunderstood you in your previous comment. I thought you mean to replicate logic I had `resolve_can_view_logs` to the other resolver. Now I can get what you mean. 
No exceptions ( guard rails ) , no business logic but only validations in the resolver layer.

TejasRGitHub: I saw this comment - https://github.com/data-dot-all/dataall/pull/1559#discussion_r1766409253 and thought you like the exceptions in the resolver and thus implemented this way. 
I will make changes

backend/dataall/modules/shares_base/services/share_logs_service.py
TejasRGitHub: I see where the issue is . Corrected code in the next update

backend/dataall/base/utils/logs.py
TejasRGitHub: Moved to feature_toggle_checker.py

backend/dataall/modules/shares_base/services/share_logs_service.py
```diff
@@ -15,8 +16,18 @@
...
     @staticmethod
-    def check_view_log_permissions(username, groups, shareUri):
-        with get_context().db_engine.scoped_session() as session:
+    @is_feature_enabled_for_allowed_values(
+        allowed_values=['admin-only', 'enabled', 'disabled'],
+        enabled_values=['admin-only', 'enabled'],
+        default_value='enabled',
+        config_property='modules.shares_base.features.show_share_logs',
+    )
+    def check_view_logs_permissions(username, groups, shareUri):
```
dlpzx, line 25: nit: we do ont need to pass the username and groups. We can get them directly from the context object `context.groups` and `context.username`

backend/dataall/modules/shares_base/api/resolvers.py
```diff
@@ -214,7 +213,7 @@ def resolve_user_role(context: Context, source: ShareObject, **kwargs):
 
 
 def resolve_can_view_logs(context: Context, source: ShareObject):
-    return ShareLogsService.check_view_log_permissions(context.username, context.groups, source.shareUri)
+    return ShareLogsService.check_view_logs_permissions(context.username, context.groups, source.shareUri)
```
dlpzx, line 216: Why have you changed the previous? Right now getting a shareObject is going to result in exception if it is disabled due to this resolver - which means that we cannot read the rest of the info of the ShareObject
```
def resolve_can_view_logs(context: Context, source: ShareObject):
    try:
        return ShareLogsService.check_view_logs_permissions(context.username, context.groups, source.shareUri)
    except Exception as e:
        log.error(f'Failed to resolve if user can view share logs due to: {e}')
        return False
```

backend/dataall/core/stacks/api/resolvers.py
dlpzx: Almost almost :) Let's do the analysis again:

In the resolver that resolves the stack/share_logs visibility (canViewLogs value):
- call the StackLogsService or the ShareLogsService `check_view_logs_permissions`
- add try/except clause to ensure it defaults to returning False in case there in an exception - otherwise the whole ShareObject/Environment/Dataset won't be able to be returned

The `check_view_logs_permissions`:
- should be decorated with `is_feature_enabled_for_allowed_values` - raises Exceptions for disabled config values
- inside it should contain additional logic for `admin-only` and other ownership checks - returns False when not alowed - ⚠️ We should change this behavior and raise an Exception as we end up doing in `get_share_logs` for example:
```
exceptions.ResourceUnauthorized(
                username=context.username,
                action='View Share Logs',
                resource_uri=shareUri,
            )
```

Any other resolver that uses `check_view_logs` to return logs (e.g. `get_share_logs`):
- directly use check_view_logs - which should return the exception

dlpzx (changes requested):
Hi Tejas, thanks again for all the effort put in this PR. Have a look at the summary in this comment: https://github.com/data-dot-all/dataall/pull/1559/files#r1770948390 for the changes needed :)

backend/dataall/core/stacks/api/resolvers.py
TejasRGitHub: Thanks once again @dlpzx , for the review and the explanation. I have made the updates. Please let me know if they look good

backend/dataall/modules/shares_base/api/resolvers.py
TejasRGitHub: Updated code with the suggested changes

TejasRGitHub:
Hi @dlpzx , I have made the changes and tested the code locally and on aws data.all deployment. Please let me know if it looks good and can be merged

backend/dataall/core/stacks/services/stack_service.py
```diff
@@ -213,3 +233,21 @@ def get_stack_logs(target_uri, target_type):
...
+    # Adding target_type as an argument as this is needed by the function `map_target_type_to_log_config_path` when resolving path to config
+    @staticmethod
+    @is_feature_enabled_for_allowed_values(
+        allowed_values=['admin-only', 'enabled', 'disabled'],
+        enabled_values=['admin-only', 'enabled'],
+        default_value='enabled',
+        resolve_property=map_target_type_to_log_config_path,
+    )
+    def check_if_user_allowed_view_logs(target_type: str, target_uri: str, config: str):
+        context = get_context()
```
dlpzx, line 246: nit: instead of passing the config as argument we can directly look it up inside the check_if_user_allowed_view_logs function.

dlpzx:
Hi @TejasRGitHub, this is a great PR! I have left one last nit comment and resolved all other comments. I have approved it and will merge with your response.

dlpzx (approved):

Issue #1546, opened 2024-09-12
Integration tests executed on a real deployment as part of the CICD - Dashboards


noah-paige:
Same as for https://github.com/data-dot-all/dataall/issues/1220.

This issue is to track the progress for the Dashboards.
It has its own dedicated issue for tracking

PR #1560, opened 2024-09-17
Feat/integration tests dashboards

noah-paige:
### Feature or Bugfix

- Feature

### Detail
- Add integration tests for dashboards
- Remove `shareDashboard` unused API
- Small update to notebook tests to now use `set_env_params` util

### Relates
- https://github.com/data-dot-all/dataall/issues/1546

### Note

For dashboard tests to run successfully the follow pre-requisite work must be complete:
- QS Enterprise Account created in `session_env1` with Capacity Pricing
- QS Dashboard created with `dataall` QS Group having `Owner` permissions
    - And `dashboardId` added to the `testdata.json` as shown in `README`

### Security

  - Is the input sanitized?


noah-paige:
Tests passing

Screenshot 2024-09-17 at 4 56 46 PM

backend/dataall/modules/dashboards/api/mutations.py
dlpzx, line 1: We should also share the resolver function and the service function. The sub-functions used in the service are used in the real handling of the share

tests_new/integration_tests/README.md
```diff
@@ -62,6 +62,11 @@ Currently **we support only Cognito based deployments** but support for any IdP
           "accountId": "...",
           "region": "eu-west-1"
         }
+      },
+      "dsabhoards": {
```
dlpzx, line 66: typo: dsabhoards -> dashboards

tests_new/integration_tests/modules/dashboards/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+    finally:
+        if ds:
+            delete_dashboard(client1, ds.dashboardUri)
+
+
+@pytest.fixture(scope='function')
+def dashboard1_share(client1, client2, dashboard1, group2):
+    try:
+        share = request_dashboard_share(client2, dashboard1.dashboardUri, group2)
+        yield share
```
dlpzx, line 43: nit: share referenced before assignment. it is missing the share=None

tests_new/integration_tests/modules/dashboards/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+        'terms': [],
+    }
+    ds = import_dashboard(client, dashboard_input)
+    return get_dashboard(client, ds.dashboardUri)
+
+
+@pytest.fixture(scope='session')
+def dashboard1(session_id, client1, persistent_env1, testdata):
+    set_env_params(client1, persistent_env1, dashboardsEnabled='true')
+    dashboardId = testdata.dashboards['persistent_env1'].dashboardId
```
dlpzx, line 29: In the README the environment for the dashboards was the session_env1, we need to change the one here or the one there

tests_new/integration_tests/modules/dashboards/test_dashboard.py
```diff
@@ -0,0 +1,113 @@
...
+    delete_dashboard,
+    approve_dashboard_share,
+    reject_dashboard_share,
+)
+from integration_tests.modules.dashboards.conftest import create_dataall_dashboard
+from integration_tests.errors import GqlError
+
+UPDATED_DESC = 'new description'
+
+
```
dlpzx, line 21: I see that we are using the persistent environment to test the dashboards, same way for ml_studio (FYI @petrkalos); but for s3_datasets and notebooks we use session environments. I think we should agree on which one to use in the modules.

The main difference lies in the cdk extensions deployed as part of the environment stack: inside some modules (ml_studio, s3_datasets) we deployed additional infra as part of the environment stack (mlstudio domain; s3/glue custom resources). If we use the session environment we use the latest version of these "cdk-extensions" and we test how these are deployed from scratch. If we use the persistent environment we need to update the environment first and check it suceeded (which tests the backwards compatibility of the "cdk-extension") and then we can perform the test

In the most perfect scenario we would test the fresh-functionality of the module on one side and the backwards compatibility on the other. But it might be a bit too much. wdyt?

tests_new/integration_tests/modules/dashboards/test_dashboard.py
```diff
@@ -0,0 +1,113 @@
...
+def test_get_author_session(client1, persistent_env1):
+    assert_that(get_author_session(client1, persistent_env1.environmentUri)).starts_with('https://')
+
+
+def test_get_author_session_unauthorized(client2, persistent_env1):
+    assert_that(get_author_session).raises(GqlError).when_called_with(client2, persistent_env1.environmentUri).contains(
+        'UnauthorizedOperation', 'CREATE_DASHBOARD'
+    )
+
```
dlpzx, line 30: I have not tested yet, but the `DashboardQuicksightService.get_quicksight_author_session` is not decorated. Instead it uses `DashboardQuicksightService._check_user_must_be_admin` which uses `TenantPolicyRepository.is_tenant_admin` where is_tenant_admin does not exist. I am not sure how this is working at the moment

tests_new/integration_tests/modules/dashboards/test_dashboard.py
```diff
@@ -0,0 +1,113 @@
...
+    filter = {'term': session_id}
+    assert_that(search_dashboards(client1, filter).nodes).is_length(1)
+    assert_that(search_dashboards(client2, filter).nodes).is_length(0)
+
+
+def test_get_dashboard_unauthorized(client2, dashboard1):
+    assert_that(get_dashboard).raises(GqlError).when_called_with(client2, dashboard1.dashboardUri).contains(
+        'UnauthorizedOperation', 'GET_DASHBOARD'
+    )
+
```
dlpzx, line 46: nit: add the dashboard uri to the assertions to make sure we are checking the correct resource (Applicable to other tests)

dlpzx (commented):
Left some comments!

backend/dataall/modules/dashboards/api/mutations.py
noah-paige: you mean remove the resolver function and service function right?

tests_new/integration_tests/modules/dashboards/conftest.py
noah-paige: let's keep as `persistent_env` and I can change in README but we also do not have `persistent_env` as part of the `envs` part of json as well so I will add both

tests_new/integration_tests/modules/dashboards/test_dashboard.py
noah-paige: `get_quicksight_author_session` is resolver for `getPlatformAuthorSession` (only used in Admin Monitoring view if enabled) while `get_quicksight_designer_url` is resolver for `getAuthorSession` (used for start QS Session for any user)

tests_new/integration_tests/modules/dashboards/test_dashboard.py
noah-paige: In the most perfect I would say `persistent_env` should have an existing data.all resource of each type to update/test on and then `session_env` stands up a new one, tests, and tears down as well (and in a perfect world it all runs super fast 😄) 

For us I think the high priority items may need that type of testing (both persistent and session), including:
- S3 and Redshift Datasets
- Shares

For the play tool modules I think it is good to newly create, test, and tear down each time. Really this could happen in either session or persistent and has no real difference unless the module impacts Env CDK (which is ML Studio and Datasets only curently). Our currently set up is:
- ML Studio (currently does persistent **)
- Notebooks (currently does both persistent and session - might as well keep)
- Dashboards (currently does persistent)
- Data Pipelines (not implemented yet)
- Worksheets (N/A - not env or IaC related to creation)
- Omics  (not implemented yet)


** To note - for ML Studio we do stand up new user profiles, test and then tear down but do so in `persistent_env1`. I would think largely because the long run time needed to stand up new ML Studio Domain as IaC in env CDK and the fact that deleting a studio Domain right now does not fully clean up after itself (EFS left behind preventing VPC deletion)

noah-paige: My suggestion is that we use session at a minimum everywhere (for play tools) and do a full create, test, and clean up -- except for MLStudio until we do some looking into some of the above

And for the high priority should have both session and persistent

I can switch dashboards to session_env and re-test (it should not change anything and would validate some of my pre-reqs again) 

Thoughts? @dlpzx

tests_new/integration_tests/modules/dashboards/conftest.py
noah-paige: after repsonding to comment below - may switch all to `session_env1` depending on outcome of that thread

tests_new/integration_tests/core/environment/queries.py
```diff
@@ -61,10 +61,14 @@ def create_environment(client, name, group, organizationUri, awsAccountId, regio
                 'region': region,
                 'description': 'Created for integration testing',
                 'tags': tags,
+                'type': 'IntegrationTesting',
                 'parameters': [
                     {'key': 'notebooksEnabled', 'value': 'true'},
+                    {'key': 'dashboardsEnabled', 'value': 'true'},
+                    {'key': 'mlStudiosEnabled', 'value': 'false'},
```
noah-paige, line 68: Note: need to provide env params to create env otherwise they do not get created and then causes issues with `set_env_params(...)` util for testing

tests_new/integration_tests/modules/dashboards/conftest.py
noah-paige: yup - added now good catch

tests_new/integration_tests/modules/dashboards/test_dashboard.py
noah-paige: done!

backend/dataall/modules/dashboards/api/mutations.py
noah-paige: done!

tests_new/integration_tests/README.md
noah-paige: updated, good catch

tests_new/integration_tests/modules/dashboards/test_dashboard.py
dlpzx: I like the conclusions :) 
- splitting for Datasets/Shares(high-prio, infra dependent) vs Play tools
- Use session at a minimum
- Look at special cases with care - ML Studio dirty teardown of resources

tests_new/integration_tests/core/environment/queries.py
dlpzx: All env params?

petrkalos: I wasn't aware of this limitation, would it make sense to adjust the `set_env_params` accordingly?

noah-paige: `set_env_params` takes all the `env.parameters` returned from backend and if any of the values differ - sets the new value and updates the environment via API request

As long as the create_environment input had all the parameters as part of the input the functionality of `set_env_params` should be good as is - unless the expectation is for `set_env_params` to also identify and add env parameters to the environment if they do not exist already?

tests_new/integration_tests/README.md
```diff
@@ -62,6 +66,11 @@ Currently **we support only Cognito based deployments** but support for any IdP
           "accountId": "...",
           "region": "eu-west-1"
         }
+      },
+      "dasbhoards": {
```
dlpzx, line 70: nit: typo dasbhoards --> dashboards

dlpzx: nit: in the README we are saying we use the session_env but then in the code we are using the persistent environment

dlpzx (approved):
approving with 2 nit comments + @petrkalos comments no the set_env_params

tests_new/integration_tests/README.md
noah-paige: sorry forgot to push typo fix - pushed now

we use `session_env1` everywhere now and not `persistent_env1`

tests_new/integration_tests/core/environment/queries.py
noah-paige: @petrkalos - I think we are good as is but please confirm

Will merge this PR Monday unless you or others suggest additional changes

Issue #1562, opened 2024-09-18
Separate responsibilities of Redshift connection types


dlpzx:
**Is your idea related to a problem? Please describe.**

**Current connection model**
Right now we have 2 types of connections `DATA_USER` and `ADMIN` and connections are owned by a single Team (IdP group) 

* `DATA_USER` connections are fine. They are owned by data producers which are the ones publishing the data
* `ADMIN` connections at the moment are owned by data consumer teams and needed as a requisite to create a share → wrong! they should be owned by Redshift admin teams. They can also be used to publish data.

**New proposed connection model**

* DATA_USER connections - created and owned by data producers - used by the data producers to publish data
* ADMIN connections - created and owned by Redshift admins. ONLY USED TO PROCESS SHARE REQUESTS. 
    * option 1: all teams in the Environment can use the connection-details (namespace) to create share requests to their own redshift roles
    * option2: redshift admins decide which teams within the Environment can create share requests to their own redshift roles

Changes needed:

* Enforce that only DATA_USER connections are the only ones used to import datasets - effort: low
* Change ownership model of ADMIN connections - option 1: effort low; option 2: effort: medium

**Describe the solution you'd like**

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

PR #1565, opened 2024-09-20
Add Redshift connection tooltips and info + restrict to DATA_USER connections for import Redshift Dataset

dlpzx:
### Feature or Bugfix
- Feature: enhancement

### Detail
This feature is an enhancement suggested by Redshift experts on #955, which is well explained in #1562. 
This PR:
- adds more info and tooltips that explain details about Redshift Connections on the UI
- restricts the type of connection that can be used to import a dataset: ONLY DATA_USER CONNECTIONS CAN BE USED TO IMPORT DATASETS. It implements this logic both in the frontend and backend

FIRST VERSION:
image

UPDATED:
image

### Relates
#1562 

### Security

  - Is the input sanitized?

### DIFF
--- a/frontend/src/modules/Environments/components/EnvironmentRedshiftConnectionAddForm.js
+++ b/frontend/src/modules/Environments/components/EnvironmentRedshiftConnectionAddForm.js
@@ -256,11 +256,11 @@ export const EnvironmentRedshiftConnectionAddForm = (props) => {
                         variant="outlined"
                       >
                         {connectionOptions.map((r) => (
-                          <Tooltip title={r.info} placement="right-end">
-                            <MenuItem key={r.value} value={r.value}>
-                              {r.label}
-                            </MenuItem>
-                          </Tooltip>
+                          <MenuItem key={r.value} value={r.value}>
+                            <Tooltip title={r.info} placement="right-end">
+                              <div>{r.label}</div>
+                            </Tooltip>
+                          </MenuItem>
                         ))}
                       </TextField>
                     </CardContent>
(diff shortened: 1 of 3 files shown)


frontend/src/modules/Environments/components/EnvironmentRedshiftConnectionAddForm.js
```diff
@@ -104,9 +115,25 @@ export const EnvironmentRedshiftConnectionAddForm = (props) => {
...
         </Typography>
+      </Box>
+      <Box sx={{ p: 3 }}>
+        <Box sx={{ pl: 5, pr: 5 }}>
+          <Alert
+            severity="info"
+            variant="outlined"
+            sx={{ whiteSpace: 'pre-line' }}
+          >
+            <AlertTitle>Types of Connections</AlertTitle>- DATA USER: This
```
petrkalos, line 129: I am not a big fun of this chunky text appearing at the top, I'd enhance and keep only the tooltips per option and maybe even maybe make the longer text appear if you click a button.

Something likes [this](https://stackoverflow.com/questions/77184297/how-to-embed-a-tooltip-icon-in-a-select-material-ui) or [this](https://stackoverflow.com/questions/59154960/how-to-add-icons-and-tooltip-on-react-select-component)

petrkalos (approved):

frontend/src/modules/Environments/components/EnvironmentRedshiftConnectionAddForm.js
dlpzx: Good suggestions! I made some changes. I tried to implement the tooltip for the complete select row, but it was struggling to update the value once selected. That's why I changed it to the word only. But i think with the info icon we are good

PR #1577, opened 2024-09-25
migrate local server to FastAPI

petrkalos:
### Feature or Bugfix
- Refactoring

### Detail
* Migrate local server Flask to FastAPI
* Upgrade and align versions of important libraries (aws-cdk-lib/boto3)
* Simplify requirements files (removed explicitly defined deps that were not required or were implicitly pulled by other packages).

### Testing
Penting succesful run on dev pipeline

### Security

  - Is the input sanitized?


tests/client.py
```diff
@@ -72,6 +77,6 @@ def graphql_server():
```
dlpzx, line 77: In line 75 `debug=app.debug` we are defining the parameter passed to graphql_sync based on an attribute of app, which does exist for a Flask app but not for a FastAPI app

tests/client.py
```diff
@@ -22,40 +26,41 @@ def query(
...
 
-    @app.route('/graphql', methods=['POST'])
-    def graphql_server():
+    @app.post('/graphql')
+    async def graphql_server(request: Request):
         # GraphQL queries are always sent as POST
         # Note: Passing the request to the context is optional.
-        # In Flask, the current request is always accessible as flask.request
-        data = request.get_json()
+        data = await request.json()
```
dlpzx, line 63: Just curious, why is this request async? I might be getting it wrong, but is it just waiting for the request data? Because the graphql_sync call is synchronous no?

tests/client.py
petrkalos: It exists as a field https://github.com/fastapi/fastapi/blob/master/fastapi%2Fapplications.py#L823

tests/client.py
petrkalos: `.json()` is declared as async thus to get the results we must await it, otherwise we get back a promise.  Thus if you have to call await you must declare the calling method as async.

tests/client.py
dlpzx: In the docs is not there, but the code is the code :)

deploy/stacks/cdk_nag_exclusions.py
```diff
@@ -28,6 +28,10 @@
         'id': 'AwsSolutions-CB3',
         'reason': 'Access to docker daemon is required to build docker images',
     },
+    {
+        'id': 'AwsSolutions-SMG4',
+        'reason': 'Database is used for test purposes',
+    },
 ]
```
dlpzx, line 36: Why do we get this exception in this PR?

petrkalos: I think because I upgraded the aws-cdk-lib

dlpzx (approved):
Looks good! Tested locally!

PR #1579, opened 2024-09-25
Add integration tests feed

noah-paige:
### Feature or Bugfix

- Feature

### Detail
- Add integration tests for `feed` module

### Relates
- #1220 

### Security

  - Is the input sanitized?

### DIFF
--- a/tests_new/integration_tests/modules/feed/queries.py
+++ b/tests_new/integration_tests/modules/feed/queries.py
@@ -1,0 +1,64 @@
+# TODO: This file will be replaced by using the SDK directly
+
+
+def post_feed_message(client, target_uri, target_type, content):
+    query = {
+        'operationName': 'PostFeedMessage',
+        'variables': {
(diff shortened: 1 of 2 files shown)


SofiaSazonova (approved):
I know, that it's not the most interesting part of data.all, but... Don't we need to test non-happy path like 'no permissions to post feed' or smth?

noah-paige:
> I know, that it's not the most interesting part of data.all, but... Don't we need to test non-happy path like 'no permissions to post feed' or smth?

Currently we do not protect against `postFeedMessage` or `getFeed` - meaning any data.all users can execute those APIs 

I can add an additional check to ensure error thrown when `targetUri` and `targetType` not specified at a bare minimum at this point to add some coverage

But overall may want to make `feed` more robust and add some additional checks at a later date

noah-paige:
@SofiaSazonova - pushed some unhappy path checks, please re-review when you have some time!

PR #1580, opened 2024-09-26
CICD Integration tests: s3 dataset shares, persistent shares

SofiaSazonova:
### Feature or Bugfix
- Feature

### Detail
- test for persistent shares
- common test functions are places in separate file
- updated envs/dataset fixtures

### Relates
- #1376

### Security

  - Is the input sanitized?


tests_new/integration_tests/aws_clients/utils.py
```diff
@@ -0,0 +1,43 @@
+import json
+import boto3
+
+from tests_new.integration_tests.aws_clients.sts import StsClient
+
+
+# it's here and not in Env test module, because it's used only here and we don't want circular dependencies
+def get_environment_access_token(client, env_uri, group_uri):
```
petrkalos, line 8: I wouldn't mix that. To avoid the circular dependency I'd add access_token as a parameter in the get_group_session

tests_new/integration_tests/core/environment/test_environment.py
```diff
@@ -51,15 +54,18 @@ def test_list_envs_invited(client2, session_env1, session_env2, session_id):
...
+        persistent_env1.environmentUri,
+        persistent_env1.stack.stackUri,
+        persistent_env1.environmentUri,
+        target_type='environment',
+    )
+    updated_before = datetime.strptime(stack.updated, '%Y-%m-%d %H:%M:%S.%f').timestamp()
+    stack = update_env_stack(client1, persistent_env1)
+    assert_that(stack).contains_entry(status='UPDATE_COMPLETE')
+    updated = datetime.strptime(stack.updated, '%Y-%m-%d %H:%M:%S.%f').timestamp()
+    assert_that(updated).is_greater_than_or_equal_to(updated_before)
```
petrkalos, line 68: 1. we write ISO dates so you can use `datetime.fromisoformat(...)`
2. you don't have to convert to timestamps, datetime objects are comparable and errors will much easier to read 
3. depending on whether `updated` is updated based on CFN or not (see my comment below) we might not want the "equal" part

i.e...
```python
nn1 = datetime.datetime.fromisoformat('2024-10-01 08:12:27.732312') # this is the format we write in the db
nn2 = datetime.datetime.now()
assert_that(nn2).is_greater_than(nn1)
```

petrkalos (changes requested):

tests_new/integration_tests/core/environment/test_environment.py
SofiaSazonova: 1. ok
2. ok
3. We need equal)

tests_new/integration_tests/conftest.py
```diff
@@ -167,6 +167,11 @@ def session_id() -> str:
     return datetime.datetime.utcnow().isoformat()
 
 
+@pytest.fixture(scope='session', autouse=True)
+def session_start_timestamp(session_id) -> float:
```
petrkalos, line 171: this is unused

tests_new/integration_tests/core/environment/test_environment.py
petrkalos: wrt `3` and for the record (not to be solved in this PR).

Ideally we would want to be able to tell wether an update was issued but at the moment I don't think it's possible as this is a limitation of the backend.

tests_new/integration_tests/core/environment/test_environment.py
```diff
@@ -51,15 +54,18 @@ def test_list_envs_invited(client2, session_env1, session_env2, session_id):
...
-    stack = check_stack_ready(client1, env_uri, stack_uri)
-    assert_that(stack.status).is_equal_to('UPDATE_COMPLETE')
+    stack = get_stack(
+        client1,
+        persistent_env1.environmentUri,
+        persistent_env1.stack.stackUri,
+        persistent_env1.environmentUri,
+        target_type='environment',
+    )
+    updated_before = datetime.strptime(stack.updated, '%Y-%m-%d %H:%M:%S.%f')
```
petrkalos, line 64: nit: you could use `datetime.fromisoformat`

Issue #1572, opened 2024-09-24
ConcurrentModificationException during the share processing


SofiaSazonova:
### Describe the bug

While share of the glue table is processed, the Lake Formation returns an error
`Could not grant principal arn:aws:iam::ACCOUNT_ID:role/dataall-test-ShareTestConsumptionRole permissions ['DESCRIBE', 'SELECT'] and permissions with grant options None to {'TableWithColumns': {'DatabaseName': 'dataall_session_s3_dataset1_tneblegb', 'Name': 'integrationtest1', 'ColumnWildcard': {}, 'CatalogId': 'ACCOUNT_ID'}}  due to: An error occurred (ConcurrentModificationException) when calling the GrantPermissions operation: Please retry in a short while`

### How to Reproduce

Repeated via integration tests, since they create and submit shares quickly (faster than the human user)

1. Create shares of the same dataset for a group and consumption role (`session_s3_dataset_1`)
2. Submit and approve both shares (`test_submit_object, test_approve_share`)
3. One of the share fails 

### Expected behavior

the root cause is a bit unclear because in the testing cases the share requests should happen one after the other, not concurrently, since we have resource lock

To avoid this errors we have to add retry block as [recommended here](https://docs.aws.amazon.com/lake-formation/latest/dg/troubleshooting.html#trouble-cross-account)

### OS

macOs Sonoma 14.2

### Python version

3.11

### AWS data.all version

2.3.0

### Additional context

the root cause is a bit unclear because in the testing cases the share requests should happen one after the other, not concurrently, since we have resource lock

PR #1585, opened 2024-09-26
retry for LF grant_permissions

SofiaSazonova:
### Feature or Bugfix
- Bugfix

### Detail
- Add retries to grant_permissions boto3 call in lakeformation client
- timeout is increased by 10% every retry

### Relates
- #1572

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/s3_datasets_shares/aws/lakeformation_client.py
+++ b/backend/dataall/modules/s3_datasets_shares/aws/lakeformation_client.py
@@ -10,6 +10,10 @@ from dataall.base.aws.sts import SessionHelper
 log = logging.getLogger('aws:lakeformation')
 
 
+def _retry_if_concurrency_error(exception):
+    return 'ConcurrentModificationException' in str(exception)
+
+
 class LakeFormationClient:
     def __init__(self, account_id, region):
         self._session = SessionHelper.remote_session(accountid=account_id, region=region)
@@ -132,9 +136,6 @@ class LakeFormationClient:
                 check_resource=data_filter_resource,
             )
         return True
-
-    def _retry_if_concurrency_error(self, exception):
-        return isinstance(exception, self._client.exceptions.ConcurrentModificationException)
 
     @retry(
         retry_on_exception=_retry_if_concurrency_error,
--- a/backend/dataall/modules/s3_datasets_shares/aws/lakeformation_client.py
+++ b/backend/dataall/modules/s3_datasets_shares/aws/lakeformation_client.py
@@ -3,14 +3,11 @@ from typing import List
 import time
 
 from botocore.exceptions import ClientError
+from retrying import retry
 
 from dataall.base.aws.sts import SessionHelper
 
 log = logging.getLogger('aws:lakeformation')
-
-NUMBER_OF_RETRIES = 5
-SLEEP_TIME = 5
-BACKOFF_COEFF = 1.1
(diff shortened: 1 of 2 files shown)


backend/dataall/modules/s3_datasets_shares/aws/lakeformation_client.py
```diff
@@ -160,16 +164,29 @@ def _grant_permissions_to_resource(
...
+                        try:
+                            response = self._client.grant_permissions(**grant_dict)
+                            last_error = None
+                            log.info(
+                                f'Successfully granted principal {principal} '
+                                f'permissions {permissions} '
+                                f'and permissions with grant options {permissions_with_grant_options} '
+                                f'to {str(resource)}  '
+                                f'response: {response}'
+                            )
```
dlpzx, line 179: Why do we need this `time.sleep(2)` in line 180? If it succeeds then it could continue without waiting no?

backend/dataall/modules/s3_datasets_shares/aws/lakeformation_client.py
```diff
@@ -160,16 +164,29 @@ def _grant_permissions_to_resource(
...
+                            time.sleep(2)
+                            break
+
+                        except ClientError as e:
+                            last_error = e
```
dlpzx, line 184: I think we should only retry for `ConcurrentModificationException`, not for all ClientErrors

SofiaSazonova: You are right

petrkalos (changes requested):
why not use @retrying decorator? check how we use it [here](https://github.com/data-dot-all/dataall/blob/3d4d648131a08f3c9ea3c3088a083b2c1fbd9e14/tests_new/integration_tests/client.py#L22). I'd also suggest to use exponential backoff, check the documentaiton [here](https://pypi.org/project/retrying/)

SofiaSazonova:
@petrkalos  have look)

petrkalos (approved):

dlpzx (approved):
Love the changes

petrkalos:
After reflecting and thinking about it a bit I think it would be better if we use the client's `ConcurrentModificationException`...

Two ways I thought of doing it...

1. Decorator+inner method (for example...)
    ```python
        class RetryTest:
            def __init__(self):
                self._client = boto3.client('lakeformation')
                self.counter = 0
            def _retry_if_concurrency_error(self, exception):
                print(f'retry callback {self.counter=}')
                return isinstance(exception, self._client.exceptions.ConcurrentModificationException)
            def test(self):
                @retry(
                    retry_on_exception=self._retry_if_concurrency_error,
                    stop_max_attempt_number=5,
                    wait_random_min=1000,
                    wait_random_max=3000,
                )
                def __call():
                    print(f'call... {self.counter=}')
                    self.counter += 1
                    raise self._client.exceptions.ConcurrentModificationException({}, '')
                __call()
    ```
3. Use of `Retrying` directly without decorator
    ```python
        Retrying(
            retry_on_exception=self._retry_if_concurrency_error,
            stop_max_attempt_number=5,
            wait_random_min=1000,
            wait_random_max=3000,
        ).call(self._client.grant_permissions, **grant_dict)
    ```

petrkalos (approved):
:ship:

Issue #1590, opened 2024-09-27
Enable hyperlinks in dataset description text


rbernotas:
**Is your idea related to a problem? Please describe.**

Our users need to be able to provide hyperlinks in the dataset description text.  This may be for prerequisite or associated documents with the data asset, forms to fill out, or other documentation.

**Describe the solution you'd like**

Dataset description text is parsed for hyperlinks.  Links in text are automatically converted to hyperlinks when the description is viewed in the dataset overview tab.

rbernotas:
We've had this solution internally for a bit, and would like to contribute (most of) it back.

The solution edits:

frontend/src/design/components/ObjectBrief.js
frontend/src/design/components/SanitizedHTML.js
frontend/src/modules/Shares/views/ShareView.js
frontend/src/utils/helpers/index.js

and additionally adds a file:

frontend/src/utils/helpers/linkMarkup.js

...where the logic is applied to parse for hyperlinks.  linkMarkup.js is where others may wish to apply additional regex and logic to sanitize the URLs (something we have customized for our purposes), but what we are contributing will be something that can be more broadly applicable to everyone.

rbernotas:
PR:  https://github.com/data-dot-all/dataall/pull/1591

PR #1591, opened 2024-09-27
Enable hyperlinks in dataset description

rbernotas:
### Feature or Bugfix

- Feature

### Detail
- Enables conversion of hyperlinks in dataset description text.

### Relates
- https://github.com/data-dot-all/dataall/issues/1590

### Security

fetching data from storage outside the application (e.g. a database, an S3 bucket)?  Dataset description field is parsed for hyperlinks
  - Is the input sanitized? Yes, and additional sanitization can take place in linkMarkup.js
  - What precautions are you taking before deserializing the data you consume?  N/A
  - Is injection prevented by parametrizing queries?  N/A
  - Have you ensured no `eval` or similar functions are used? Yes

### DIFF
--- a/frontend/src/utils/helpers/linkMarkup.js
+++ b/frontend/src/utils/helpers/linkMarkup.js
@@ -1,0 +1,33 @@
+export const createLinkMarkup = (text, color) => {
+  // Define the components of the regex pattern
+
+  // Matches optional protocol (http:// or https://)
+  const protocol = '(https?://)?';
+
+  // Matches any domain or subdomain
+  const domain = '([a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,}';
+
+  // Matches paths and query strings, excluding certain special characters
+  const pathAndQuery = '(\\/[^\\s()$@]*)?(\\?[\\w=&-]*)?(#[\\w\\-]*)?';
+
+  // Combine all parts into the final regex pattern
+  const urlRegex = new RegExp(
+    `(${protocol})?(${domain})(?=\\s|\\/|$)${pathAndQuery}(?=\\s|$)`,
+    'gi'
+  );
+
+  return text.replace(urlRegex, (fullMatch) => {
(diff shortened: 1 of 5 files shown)


frontend/src/utils/helpers/linkMarkup.js
```diff
@@ -0,0 +1,33 @@
+export const createLinkMarkup = (text, color) => {
```
dlpzx, line 1: nit: Is the color ever going to change? If not, could we get it from the theme directly in the linkMarkup util?

dlpzx (approved):
Hi @rbernotas, left some minor comments, but the functionality looks good. Thanks for the PR

frontend/src/utils/helpers/linkMarkup.js
rbernotas: I did look into this.  The color does need to change when you swap from light to dark theme in the settings.  But eslint doesn't like it when the useTheme hook is called within linkMarkup as createLinkMarkup isn't a React component or custom hook.  I looked into wrapping the function in a component, but then calling that started to get a little hacky just for the purpose of avoiding the param.

rbernotas:
integration tests had a failure in test_submit_share_extension_request_with_auto_approval.  The error doesn't look related to this code change... is this a known issue?

frontend/src/utils/helpers/linkMarkup.js
dlpzx: Understood, seems like the solution is more complex than the problem itself. I think we can leave it as is.

dlpzx:
> integration tests had a failure in test_submit_share_extension_request_with_auto_approval. The error doesn't look related to this code change... is this a known issue?

Yes, there was a small issue on the expiration date calculation for the last day of a month. The fix is in this [PR](https://github.com/data-dot-all/dataall/pull/1594). I re-run the failed job for this PR because today, 1st of October, we don't run into the issue

PR #1603, opened 2024-10-02
Clean up S3 Buckets in integration test

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Environment and Dataset stacks leave S3 Buckets to be deleted even when the CloudFormation stack is deleted.
This PR deletes the S3 Buckets when the session and temp fixtures for datasets and environments are deleted.

### Testing
Tested that environment and dataset buckets are all deleted - in real AWS CICD pipeline with NOT-empty buckets.
Tested that dataset buckets with some manually created access points succeed.

### Relates
- #1220 

### Security

  - Is the input sanitized?


tests_new/integration_tests/aws_clients/s3.py
```diff
@@ -0,0 +1,29 @@
...
+log = logging.getLogger(__name__)
+
+
+class S3Client:
+    def __init__(self, session, region):
+        self._client = session.client('s3', region_name=region)
+        self._resource = session.resource('s3', region_name=region)
+        self._region = region
+
+    def delete_bucket(self, bucket_name):
```
petrkalos, line 13: you need to account for access points as well (if bucket have them it will fail)

dlpzx: In a happy path world, we should not run into that situation because a dataset can only be deleted if all shares are revoked and if shares are revoked the access points are deleted. However, if the tests for revoking access points fail, then we would face this situation when tearing down resources. I am going to check if it is  easy to implement :)

tests_new/integration_tests/aws_clients/s3.py
```diff
@@ -0,0 +1,35 @@
...
+            # Delete any remaining objects (if versioning was not enabled)
+            bucket.objects.all().delete()
+            # Delete any remaining access point
+            paginator = self._control_client.get_paginator('list_access_points')
+            for page in paginator.paginate(AccountId=self._account, Bucket=bucket_name):
+                for access_point in page['AccessPointList']:
+                    self._control_client.delete_access_point(AccountId=self._account, Name=access_point['Name'])
+            bucket.delete()
+        except ClientError as e:
+            log.exception(f'Error deleting S3 bucket: {e}')
```
petrkalos, line 35: this will log the exception 2 times. `.exception` prints the message and stacktrace by default

petrkalos (approved):

dlpzx:
Not merging yet because while testing the code for access points a small error appeared. Wait for updates on this PR

tests_new/integration_tests/aws_clients/s3.py
dlpzx: Tested! Now it is ready to be merged. Do you want to have a final look @petrkalos ?

PR #1607, opened 2024-10-03
Integration tests - refresh tokens of AWS Clients

dlpzx:
### Feature or Bugfix
- Feature

### Detail
The new integration tests take quite some time (>1h) to get executed. For this reason the AWS clients passed as fixtures to perform different actions (e.g. create S3 Buckets) have expired STS tokens when we reach the teardown operations (e.g. delete S3 Buckets).

This PR implements some logic similar to the one proposed in https://github.com/benkehoe/aws-assume-role-lib which was referenced by the boto3 maintainers in the [GitHub issue](https://github.com/boto/boto3/issues/443) for this topic. 

### Relates
- #1220 

### Security

  - Is the input sanitized?


dlpzx:
🧪 Tested in AWS - executed the dataset tests with sleep in-between so that they would take longer than 1h to execute. The client for creation and deletion of S3 Buckets uses the refresh token client and is able to delete the buckets 1h after the client is defined.

Issue #1608, opened 2024-10-03
NoSuchAccessPoint exception during share processing


SofiaSazonova:
### Describe the bug

Sometimes during the share processing though the freshly created access point exists, API call `put_access_point_policy` fails with error `NoSuchAccessPoint`.

The investigation with S3 team reveals, that
the initial GETAccessPoint was added to NegativeCache
and then one of the PutAccessPointPolicy also made to the same backend host that was added to the negative Cache
This lead to 404

### How to Reproduce

Run share integration tests in eu-west-2 after 2 pm

### Expected behavior

drop the GetAccessPoint call before create one and just look for an exception in case it exists

### OS

macOs Sonoma 14.2

### Python version

3.11

### AWS data.all version

2.6.0

PR #1609, opened 2024-10-03
try to create AP every time, catch if already exists

SofiaSazonova:
### Feature or Bugfix

- Bugfix

### Detail
- Try to create AP every time we process the share
- Catch error if it already exists
- Retry on put_access_point_policy

### Relates
- #1608 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/s3_datasets_shares/aws/s3_client.py
+++ b/backend/dataall/modules/s3_datasets_shares/aws/s3_client.py
@@ -1,5 +1,5 @@
 import logging
-import time
+from retrying import retry
 from typing import List
 
 from dataall.base.aws.sts import SessionHelper
@@ -56,15 +56,14 @@ class S3ControlClient:
 
         return self.try_get_bucket_access_point_arn(bucket_name, access_point_name)
 
+    @retry(retry_on_result=lambda arn: arn is None, stop_max_attempt_number=10, wait_fixed=30000)
     def try_get_bucket_access_point_arn(self, bucket_name: str, access_point_name: str):
-        for attempt in range(5):
-            all_acess_points = self._client.list_access_points(
-                AccountId=self._account_id, Bucket=bucket_name, MaxResults=1000
-            )
-            for ap in all_acess_points['AccessPointList']:
-                if ap['Name'] == access_point_name:
-                    return ap['AccessPointArn']
-            time.sleep(30)
+        all_access_points = self._client.list_access_points(
+            AccountId=self._account_id, Bucket=bucket_name, MaxResults=1000
+        )
+        for ap in all_access_points['AccessPointList']:
+            if ap['Name'] == access_point_name:
+                return ap['AccessPointArn']
         return None
 
     def delete_bucket_access_point(self, access_point_name: str):
--- a/backend/dataall/modules/s3_datasets_shares/aws/s3_client.py
+++ b/backend/dataall/modules/s3_datasets_shares/aws/s3_client.py
@@ -1,4 +1,5 @@
 import logging
+import time
 from typing import List
 
 from dataall.base.aws.sts import SessionHelper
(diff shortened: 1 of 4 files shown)


backend/dataall/modules/s3_datasets_shares/aws/s3_client.py
```diff
@@ -49,6 +49,9 @@ def create_bucket_access_point(self, bucket_name: str, access_point_name: str):
                 Bucket=bucket_name,
             )
         except Exception as e:
+            if 'AccessPointAlreadyOwnedByYou' in str(e):
+                log.info(f'S3 bucket access point {access_point_name} already exists on {self._account_id}')
+                return self.get_bucket_access_point_arn(access_point_name)
             log.error(f'S3 bucket access point creation failed for location {bucket_name} : {e}')
```
petrkalos, line 55: nit: would convert to `log.exception(S3 bucket access point creation failed for location {bucket_name})`

backend/dataall/modules/s3_datasets_shares/aws/s3_client.py
```diff
@@ -49,6 +49,9 @@ def create_bucket_access_point(self, bucket_name: str, access_point_name: str):
                 Bucket=bucket_name,
```
petrkalos, line 49: nit: method name is a bit misleading now, is it a big change to rename to `try_create...` or `create_or_get...`?

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -505,9 +485,17 @@ def manage_access_point_and_policy(self):
                 self.s3_prefix,
                 perms_to_actions(self.share.permissions, SidType.BucketPolicy),
             )
-        s3_client.attach_access_point_policy(
-            access_point_name=self.access_point_name, policy=json.dumps(access_point_policy)
-        )
+
```
petrkalos, line 488: why do we need this now and didn't need it previously?

petrkalos (commented):
some nits and a question

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
SofiaSazonova: This is in case AP is not yet created. Before the retry part was in creation of of AP

petrkalos: I would explicitly encapsulate the Retry mechanism in the `create_bucket_access_point` to ensure that all calls to it return when the bucket is created.

For example something like the below....
```python
    def create_bucket_access_point(self, bucket_name: str, access_point_name: str):
        try:
            self._client.create_access_point(AccountId=self._account_id, Name=access_point_name, Bucket=bucket_name)
        except self._client.exceptions.AccessPointAlreadyOwnedByYou:
            ...
        except Exception as e:
            log.exception(f'S3 bucket access point creation failed for location {bucket_name=}')
            raise e
        return Retrying(
            retry_on_exception=self._client.exceptions.NotFoundException,
            stop_max_attempt_number=5,
            wait_random_min=1000,
            wait_random_max=3000,
        ).call(self.get_bucket_access_point_arn, access_point_name)['AccessPointArn']
```

Also you must make sure that you wait enough time. Previous code was waiting for 300secs++, yours wait from 5 to 15 seconds.
Also I think that the wait random is not a good idea, and should use exponential back off.

SofiaSazonova: Won't it lead us to the same error? If get_access_point_arn will fail, we again receive the problem of Negative Cash?

SofiaSazonova: also, there is no `self._client.exceptions.AccessPointAlreadyOwnedByYou:` so, I have to check the string

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -1,8 +1,9 @@
 import logging
 import json
-import time
 from itertools import count
 
+from retrying import Retrying
```
petrkalos, line 5: unused

backend/dataall/modules/s3_datasets_shares/aws/s3_client.py
```diff
@@ -43,16 +44,28 @@ def get_bucket_access_point_arn(self, access_point_name: str):
...
-            raise e
-        else:
-            return access_point['AccessPointArn']
+            if 'AccessPointAlreadyOwnedByYou' not in str(e):
+                raise e
+
+        return self.try_get_bucket_access_point_arn(bucket_name, access_point_name)
+
+    def try_get_bucket_access_point_arn(self, bucket_name: str, access_point_name: str):
+        for attempt in range(5):
```
petrkalos, line 60: you could use `@retry(retry_on_result=lambda arn: arn is None, stop_max_attempt_number=10, wait_fixed=30000)`

petrkalos (changes requested):

petrkalos (approved):

Issue #1626, opened 2024-10-09
Modal dialog to list users from Team names


rbernotas:
**Is your idea related to a problem? Please describe.**

We have an internal change for which we would like to propose contribution in part.  Users would like to understand who is in the Teams, without having to access the separate roles based access system (Cognito or otherwise), for purposes of locating contacts for governors/stewards/owners/users of a dataset/share/env/org, so that they can contact individuals to ask questions or get support.

**Describe the solution you'd like**

The feature would turn various appearances of the Team names text throughout data.all, into clickable links that pop up a modal dialog showing the users in the Teams.

For default data.all this would use Cognito for looking up the users; for custom roles based access systems it would use that custom system (and need to be implemented by the admins of that data.all instance).

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

rbernotas:
PR is available here:  https://github.com/data-dot-all/dataall/pull/1627

Since we don't have Cognito available to us, we could use some help with feedback and testing on the Cognito aspects of this change, if there is interest in accepting this contribution.

noah-paige:
Closing as completed in #1627 - thanks for the contribution @rbernotas

PR #1627, opened 2024-10-09
User modal dialog - team link

rbernotas:
### Feature or Bugfix
- Feature

### Detail
- Turns various appearances of Team name text into clickable links, which pop up a modal dialog that lists the users in the team.
- For more detail on the proposal, see: https://github.com/data-dot-all/dataall/issues/1626
- As we don't have Cognito available to us and use a separate custom system for roles management, we could use some help with feedback and testing on the Cognito piece of this, if there is interest in accepting this contribution.

### Relates
- https://github.com/data-dot-all/dataall/issues/1626

### Security

fetching data from storage outside the application (e.g. a database, an S3 bucket)?  NO
  - Is the input sanitized?  N/A
  - What precautions are you taking before deserializing the data you consume?  N/A
  - Is injection prevented by parametrizing queries?  N/A
  - Have you ensured no `eval` or similar functions are used?  YES
- Does this PR introduce any functionality or component that requires authorization?  NO
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms?  N/A
  - Are you logging failed auth attempts?  N/A
- Are you using or adding any cryptographic features?  NO
  - Do you use a standard proven implementations?  N/A
  - Are the used keys controlled by the customer? Where are they stored?  N/A
- Are you introducing any new policies/roles/users?  NO
  - Have you used the least-privilege principle? How?  N/A

### DIFF
--- a/backend/dataall/base/aws/cognito.py
+++ b/backend/dataall/base/aws/cognito.py
@@ -40,7 +40,7 @@ class Cognito(ServiceProvider):
             return group_email_ids
 
     def get_user_list_from_group(self, groupName):
-        return get_user_emailids_from_group(groupName)
+        return self.get_user_emailids_from_group(groupName)
 
     def list_groups(self, envname: str, region: str):
         user_pool_id = None
--- a/backend/dataall/base/aws/cognito.py
+++ b/backend/dataall/base/aws/cognito.py
@@ -40,7 +40,26 @@ class Cognito(ServiceProvider):
             return group_email_ids
 
     def get_user_list_from_group(self, groupName):
-        return self.get_user_emailids_from_group(groupName)
+        try:
+            envname = os.getenv('envname', 'local')
+            parameter_path = f'/dataall/{envname}/cognito/userpool'
+            ssm = boto3.client('ssm', region_name=os.getenv('AWS_REGION', 'eu-west-1'))
+            user_pool_id = ssm.get_parameter(Name=parameter_path)['Parameter']['Value']
+            paginator = self.client.get_paginator('list_users_in_group')
+            pages = paginator.paginate(UserPoolId=user_pool_id, GroupName=groupName)
+            cognito_user_list = []
+            for page in pages:
+                cognito_user_list += page['Users']
+            group_usernames = [user['Username'] for user in cognito_user_list]
+        except Exception as e:
+            envname = os.getenv('envname', 'local')
+            if envname in ['local', 'dkrcompose']:
+                log.error('Local development environment does not support Cognito')
+                return ['anonymous@amazon.com']
+            log.error(f'Failed to get email ids for Cognito group {groupName} due to {e}')
+            raise e
+        else:
+            return group_usernames
 
     def list_groups(self, envname: str, region: str):
         user_pool_id = None
--- a/backend/dataall/base/aws/cognito.py
+++ b/backend/dataall/base/aws/cognito.py
@@ -11,17 +11,22 @@ class Cognito(ServiceProvider):
     def __init__(self):
         self.client = boto3.client('cognito-idp', region_name=os.getenv('AWS_REGION', 'eu-west-1'))
 
+    def get_cognito_users(self, groupName):
+        envname = os.getenv('envname', 'local')
+        parameter_path = f'/dataall/{envname}/cognito/userpool'
+        ssm = boto3.client('ssm', region_name=os.getenv('AWS_REGION', 'eu-west-1'))
+        user_pool_id = ssm.get_parameter(Name=parameter_path)['Parameter']['Value']
+        paginator = self.client.get_paginator('list_users_in_group')
+        pages = paginator.paginate(UserPoolId=user_pool_id, GroupName=groupName)
+        cognito_user_list = []
+        for page in pages:
+            cognito_user_list += page['Users']
+
+        return cognito_user_list
+
     def get_user_emailids_from_group(self, groupName):
         try:
-            envname = os.getenv('envname', 'local')
-            parameter_path = f'/dataall/{envname}/cognito/userpool'
-            ssm = boto3.client('ssm', region_name=os.getenv('AWS_REGION', 'eu-west-1'))
-            user_pool_id = ssm.get_parameter(Name=parameter_path)['Parameter']['Value']
-            paginator = self.client.get_paginator('list_users_in_group')
-            pages = paginator.paginate(UserPoolId=user_pool_id, GroupName=groupName)
-            cognito_user_list = []
-            for page in pages:
-                cognito_user_list += page['Users']
+            cognito_user_list = self.get_cognito_users(groupName)
             group_email_ids = []
             attributes = []
             # Make a flat list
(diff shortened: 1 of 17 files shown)


backend/dataall/base/aws/cognito.py
```diff
@@ -34,11 +34,33 @@ def get_user_emailids_from_group(self, groupName):
...
+        try:
+            envname = os.getenv('envname', 'local')
+            parameter_path = f'/dataall/{envname}/cognito/userpool'
+            ssm = boto3.client('ssm', region_name=os.getenv('AWS_REGION', 'eu-west-1'))
+            user_pool_id = ssm.get_parameter(Name=parameter_path)['Parameter']['Value']
+            paginator = self.client.get_paginator('list_users_in_group')
+            pages = paginator.paginate(UserPoolId=user_pool_id, GroupName=groupName)
+            cognito_user_list = []
+            for page in pages:
+                cognito_user_list += page['Users']
```
noah-paige, line 52: nit: L16-24 and L44-52 are the same - can move to a separate function and define once

backend/dataall/base/aws/cognito.py
```diff
@@ -34,11 +34,33 @@ def get_user_emailids_from_group(self, groupName):
...
+            group_usernames = [user['Username'] for user in cognito_user_list]
+        except Exception as e:
+            envname = os.getenv('envname', 'local')
+            if envname in ['local', 'dkrcompose']:
+                log.error('Local development environment does not support Cognito')
+                return ['anonymous@amazon.com']
+            log.error(f'Failed to get email ids for Cognito group {groupName} due to {e}')
```
noah-paige, line 59: nit: should this be `usernames` not `email ids`

backend/dataall/base/aws/cognito.py
```diff
@@ -34,11 +34,33 @@ def get_user_emailids_from_group(self, groupName):
             if envname in ['local', 'dkrcompose']:
                 log.error('Local development environment does not support Cognito')
                 return ['anonymous@amazon.com']
-            log.error(f'Failed to get email ids for Cognito group {groupName} due to {e}')
+            log.error(f'Failed to get usernames for Cognito group {groupName} due to {e}')
```
noah-paige, line 37: nit: why this change? Should this be `email ids` not `usernames`

frontend/src/design/components/UserModal.js
```diff
@@ -0,0 +1,102 @@
...
+  TableCell,
+  TableBody,
+  Dialog,
+  CircularProgress
+} from '@mui/material';
+import { useClient } from '../../services';
+import { useDispatch } from '../../globalErrors';
+import { listUserForGroup } from '../../services/graphql/Groups/listUserForGroup';
+
+const UserModal = ({ teams, open, onClose }) => {
```
noah-paige, line 18: nit: plural `teams` but we only accept a singular team/group for the API to return successfully. Think that is fine design wise but can we rename to `team` everywhere?

frontend/src/design/components/UserModal.js
```diff
@@ -0,0 +1,102 @@
...
+                <TableRow key={index}>
+                  <TableCell>
+                    <Typography color="textSecondary" variant="body2">
+                      {userId}
+                    </Typography>
+                  </TableCell>
+                </TableRow>
+              ))}
+            </TableBody>
+          </Table>
```
noah-paige, line 83: Is there a need to add a `<Pager>` element here and paginate the response returned from the backend (++ would likely have to add some sort condition on backend to ensure consistent results)

Otherwise what if a group have 100+ users?

frontend/src/modules/DatasetsBase/components/DatasetGovernance.js
```diff
@@ -20,6 +22,14 @@ export const DatasetGovernance = (props) => {
       : [{ label: '-', nodeUri: '-' }];
   const tags = dataset.tags.length > 0 ? dataset.tags : ['-'];
 
+  const [ModalOpen, setIsModalOpen] = useState(false);
```
noah-paige, line 25: nit: Here and all other uses of `ModalOpen` - for consistency the syntax we have for state variables would be `  const [modalOpen, setIsModalOpen] = useState(false);` (lower case `m`)

frontend/src/modules/Shares/components/ShareBoxListItem.js
```diff
@@ -23,6 +24,18 @@ export const ShareBoxListItem = ({ share }) => {
       ? '/static/icons/Arch_Amazon-Redshift_64.svg'
       : '-';
 
+  const [ModalOpen, setIsModalOpen] = useState(false);
+  const handleOpenModal = () => setIsModalOpen(true);
+  const handleCloseModal = () => setIsModalOpen(false);
+
+  const [DataOwnerModalOpen, setIsDataOwnerModalOpen] = useState(false);
```
noah-paige, line 31: nit: `dataOwnerModalOpen` (lowercase first char)

frontend/src/modules/Shares/views/ShareView.js
```diff
@@ -796,6 +797,18 @@ const ShareView = () => {
   const [isReApplyShareItemModalOpen, setIsReApplyShareItemModalOpen] =
     useState(false);
 
+  const [ModalOpen, setIsModalOpen] = useState(false);
+  const handleOpenModal = () => setIsModalOpen(true);
+  const handleCloseModal = () => setIsModalOpen(false);
+
+  const [RequestTeamModalOpen, setIsRequestTeamModalOpen] = useState(false);
```
noah-paige, line 804: nit: `requestTeamModalOpen` (lowercase first char)

noah-paige (commented):
Left some comments - mostly nits on syntax 

One high level question I want to ask is whether every user should be allowed to view which members belong in each/any group? Is there a chance that that is too much information for any given user to be able to get? Should there be some auth controls around this?

Also, this is a good start to add this feature for Orgs, Envs, Datasets, and Shares. Ultimately we can also add to other modules like Notebooks, Dashboard, etc. if we think should be used everywhere

rbernotas:
> Left some comments - mostly nits on syntax
> 
> One high level question I want to ask is whether every user should be allowed to view which members belong in each/any group? Is there a chance that that is too much information for any given user to be able to get? Should there be some auth controls around this?
> 
> Also, this is a good start to add this feature for Orgs, Envs, Datasets, and Shares. Ultimately we can also add to other modules like Notebooks, Dashboard, etc. if we think should be used everywhere

Thanks @noah-paige for the comments.  For us, having transparency to who is in the teams is actually desirable so that users can reach out to members of those teams to ask questions--they can also see members of those teams in our RBAC system anyways, so this just makes it convenient.  From our perspective I can't think of a reason we would want to limit this information, but others elsewhere may have a different view.

backend/dataall/base/aws/cognito.py
rbernotas: Updated to split it out to a separate function.

backend/dataall/base/aws/cognito.py
rbernotas: Thanks, I changed the wrong error message.  Fixed.

frontend/src/design/components/UserModal.js
rbernotas: Thanks, changed this everywhere.

frontend/src/modules/DatasetsBase/components/DatasetGovernance.js
rbernotas: Changed to lowercase in all cases.

frontend/src/modules/Shares/components/ShareBoxListItem.js
rbernotas: Changed, thanks.

frontend/src/design/components/UserModal.js
rbernotas: We don't currently have pagination support in our custom authorizer to match this.  In practice we don't have much more than about 20 users in any given team at the moment--the modal dialog will also just scroll through the list of users and the api calls are pretty quick.  Maybe an enhancement for the future?

rbernotas:
Addressed PR comments.

noah-paige (commented):
Overall this PR looks good, just wanted to document a few thoughts/enhancement ideas moving forward:

- (1) Places that do not have UserModal component (but a part of Datasets, Envs, Orgs, Shares)
    - Environment —> Dataset Tab —> “Datasets You Own” Table (Owners Column)
    - Dataset - Table Overview
    - Dataset - Folder Overview
    - NOTE: There are likely more missing places … to have a more consistent approach maybe we have this feature on the following areas
        - Any Resource and/or Share Overview/View Page 
        - Any Teams Tabs (including Admin Teams Tab)
        - (… and remove from the List Item Pages for Env, Org or Share and similar views that are not included in the above)


- (2) It is not clear that the Team value is “clickable” and opens a model unless a user hovers over the text and even then there is no change is UI for the text - may want to add a box/button or similar to make it clear that this feature exists

- (3) Some type of pager on the frontend (I tested adding 25 users to a group and scroll on UI worked as expected so this is more minor)


Before approving and merging, I want to align on a consistent approach on where this feature should be implemented (i.e. number 1 above). This PR can still be in the scope of org, env, dataset, and share only as long as there is a consistent approach to extend further in future PRs.

Otherwise I think the remainder of the above comments can be picked up as enhancements and this PR could be good to merge

rbernotas:
Hi @noah-paige, I agree with you on 1, 2, and 3 regarding approach for further extension in future PRs.  Were you able to check if this works with Cognito?  That's one area we aren't able to test it against, as we are limited to our custom authorizer.

noah-paige:
@rbernotas - yes I tested with Cognito, below is a screenshot of the User Modal (please confirm this is as expected)

Screenshot 2024-10-14 at 12 13 00 PM


For 1 above would you want to make changes such that UserModal is only on the following pages:
- EnvironmentOverview
- EnvironmentTeams
- OrganizationOverview
- OrganizationTeams
- ShareView
- DatasetGovernance (which is used in DatasetOverview)

future PRs can extend further to otehr pieces like TableOverview, WorksheetView, AdministrationTeams, etc etc

rbernotas:
@noah-paige confirming both your Cognito screenshot is works as expected, and I will also send a commit in a bit that will limit the UserModal to those specific pages.

rbernotas:
@noah-paige , removed the modal dialog from the various views discussed.

noah-paige (approved):
working as expected - changes look good

PR #1594, opened 2024-09-30
Fix share expiration date calculation for end-of-month days

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
Solves #1592 

With this PR for the 15 first days of the month, extending the share request for another month means extending it until the end of the month. For the second half of the month, extending the share request for another month means extending it until the end of the NEXT month.

### Relates
- #1592 

### Security
Please answer the questions below briefly where applicable, or write `N/A`. Based on
[OWASP 10](https://owasp.org/Top10/en/).

- Does this PR introduce or modify any input fields or queries - this includes
fetching data from storage outside the application (e.g. a database, an S3 bucket)?
  - Is the input sanitized?
  - What precautions are you taking before deserializing the data you consume?
  - Is injection prevented by parametrizing queries?
  - Have you ensured no `eval` or similar functions are used?
- Does this PR introduce any functionality or component that requires authorization?
  - How have you ensured it respects the existing AuthN/AuthZ mechanisms?
  - Are you logging failed auth attempts?
- Are you using or adding any cryptographic features?
  - Do you use a standard proven implementations?
  - Are the used keys controlled by the customer? Where are they stored?
- Are you introducing any new policies/roles/users?
  - Have you used the least-privilege principle? How?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1635, opened 2024-10-10
Changes to the logic of calculating expiration date

TejasRGitHub:
### Feature or Bugfix
- Feature 

### Detail

Changes to the logic of calculating expiration. 

Initially current month was considered into the calculation for expiration date. But this resulted in a bug while performing unit tests here https://github.com/data-dot-all/dataall/pull/1594 and also this was not correct in the sense that if a user requests a month of expiration at the end of month, that user will get expiration at the end of the same month 

But with the new logic - https://github.com/data-dot-all/dataall/pull/1594, more than half a month is granted extra . 

**To minimize this extra access period on share, this PR proposes new logic,** 

1. If the user is in the last week of the month and if the user requests 1 month of data, allocate the end of the next month. 
2. If the user is in NOT in the last week of the months and if the user requests 1 month of data, allocate the end of same month as expiration. 

### Relates
- https://github.com/data-dot-all/dataall/issues/1083

### Security

  - Is the input sanitized?


TejasRGitHub:
@dlpzx , @noah-paige , @petrkalos  can you please take a look at this PR and its description. Making changes to the share expiration calculation logic.

backend/dataall/base/utils/expiration_util.py
```diff
@@ -8,29 +8,41 @@ class ExpirationUtils:
...
 
         return shareExpiryDate
 
+    @staticmethod
+    def get_last_week_date_range(year, month):
+        last_day_of_month = calendar.monthrange(year, month)[1]
+        last_date = date(year, month, last_day_of_month)
+        last_week_day = last_date.weekday()
+        start_date = last_date - timedelta(days=last_week_day)
+
```
dlpzx, line 43: Let me confirm what this function returns: start_date should be the last week_start date, it is calculated by substracting weekdays from the end_date of the month. the thing is that weekdays returns 0-6 (0 =Monday, 1=Tuesday.... 6=Sunday) if the end_date is a Monday, is it going to substract anything at all?

dlpzx (commented):
Need clarification on one calculation, other than that, the idea of using the last week to decide the expiration period seems more logical.

backend/dataall/base/utils/expiration_util.py
dlpzx: And it won't fail as we are using <= in the comparison. I think it is good, I was just a bit confused. End of week is end of week (Sunday for most of the people) so I think it is good.

dlpzx (approved):

TejasRGitHub:
@dlpzx  Updated branch and merged code from `main` branch

Issue #1621, opened 2024-10-08
Metadata Form Versioning


SofiaSazonova:
The target features are: Make MFs editable.
But what to do with already attached forms?
Solution: enable versioning of Metadata Forms.

![MF DB](https://github.com/user-attachments/assets/bf1c1af4-b1fd-43ce-ab48-409cbea78b9c)

1. Attached MF should be connected with version, not the MF directly
2. If there are attached entities and user edits the parent MF, new version is created
3. Entity owners are notified about new version of MF
4. When no attached MF left for the older version, it's automatically deleted.

SofiaSazonova:
How to edit MF:
![edit MF](https://github.com/user-attachments/assets/82cb7652-c010-40da-8839-ab80a1382ad3)

SofiaSazonova:
Attach MF: Only latest version is available

Edit attached entity: 
1. New version if opened for editing, the fields are pre-filled basing on field name, so user can adjust
2. Old attached MF is deleted
3. New created

Delete attached MF: When no attached MF left for the older version, it's automatically deleted.

noah-paige:
Some Comments/Qs:
- (1) What action publishes a new metadata form/ what properties of the form are editable? 
    - If a metadata form is edited for a new description will this publish a new version? Is description editable? Is it only for edits to the form fields?
- (2) Confirming my understanding - When a new form version is created this will
    - New record in `metadata_from_version`
    - New record(s) in `metadata_from_field` (there will be no overlap / re-use of form field records between different form versions right?)
- (3) I see both `metadata_form_field` and `attached_metadata_form` tables have both `metadataFormUri` and `metadataFormVersionUri` in their data model... is this just for easy look up? The one of significance would the the `metadataFormVersionUri` one right?
- (4) Will enforcement rules be with respect to each version of the metadata form? I think the above makes sense for a change in form fields after the form is in use - but have we thought about a change in enforcement rules after a form is in use?
- (5) Could version be a version Number that is a composite key along with metadataFormUri for readability - and then we can expose version # on the attached form for transparency / better UX
- (6) On the point of `'When no attached MF left for the older version, it's automatically deleted.'` - I think it may be easiest to have users manually delete version of forms and have an upper limit of 10 or so versions (this is similar to how IAM Policy Versions work) rather than some scheduled backend running task to garbage collect 
- (7) On editing of attached form what do you mean on `'New version if opened for editing, the fields are pre-filled basing on field name, so user can adjust'`?

[...]

PR #1637, opened 2024-10-11
Metadata form versioning - 1

SofiaSazonova:
### Feature or Bugfix

- Feature

### Detail
- new db model
- migration:
  -- create default version one
  -- backfill existing MF fields and attached MFs
- when MF is created, service automatically create version 1
- fields and attached metadata forms automatically use the latest version (for now version === 1)

### Relates
- #1621

### Security

  - Is the input sanitized?


backend/migrations/versions/5a798acc6282_create_version_mf.py
```diff
@@ -0,0 +1,101 @@
...
+    for mf in session.query(MetadataForm).all():
+        version = MetadataFormVersion(metadataFormUri=mf.uri, version=1)
+        session.add(version)
+        session.commit()
+
+    op.add_column('attached_metadata_form', sa.Column('version', sa.Integer()))
+    for amf in session.query(AttachedMetadataForm).all():
+        amf.version = 1
+        session.commit()
+    op.alter_column('attached_metadata_form', 'version', nullable=False)
```
noah-paige, line 50: nit: just a callout I think may be able to do same with `server_default` parameter such as 
```
op.add_column('attached_metadata_form', sa.Column('version', sa.Integer(), server_default=1))
```

But changing nullable value after would think works as well

backend/migrations/versions/5a798acc6282_create_version_mf.py
```diff
@@ -0,0 +1,101 @@
...
+    op.drop_constraint('f_key_enforcement_version_metadata', 'metadata_form_enforcement_rule', type_='foreignkey')
+    op.drop_column('metadata_form_enforcement_rule', 'version')
+    op.drop_constraint('fk_attached_mf_version_uri', 'attached_metadata_form', type_='foreignkey')
+    op.drop_column('attached_metadata_form', 'version')
+    op.create_table(
+        'maintenance',
+        sa.Column('status', sa.VARCHAR(), autoincrement=False, nullable=False),
+        sa.Column('mode', sa.VARCHAR(), autoincrement=False, nullable=True),
+        sa.PrimaryKeyConstraint('status', name='maintenance_pkey'),
+    )
```
noah-paige, line 99: I think this should be removed

noah-paige (commented):
left a few comments on code review - will test and come back with any other findings on migration script

backend/dataall/modules/metadata_forms/db/metadata_form_models.py
```diff
@@ -62,6 +90,12 @@ class AttachedMetadataForm(Base):
         ForeignKeyConstraint(
             ('metadataFormUri',), ('metadata_form.uri',), name='fk_attached_mf_uri', ondelete='CASCADE'
         ),
```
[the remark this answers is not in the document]
SofiaSazonova: On the one hand it's overkill to have additional constraint, but It won't hurt. 
I preferred not to touch it, but yes, we can drop it.

backend/migrations/versions/5a798acc6282_create_version_mf.py
SofiaSazonova: I don't want the column to have the default value, I want it to be explicitly filled. Otherwise we can have troubles with debugging.

backend/migrations/versions/5a798acc6282_create_version_mf.py
SofiaSazonova: Oh, yes. Removed it from update, forgot about downgrade

noah-paige (approved):
looks good - tested upgrade and downgrade paths

PR #1644, opened 2024-10-16
Fix: check if Redshift table exists before publishing it to data.all

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
Currently when we hit the addRedshiftTable api, we do not check if the provided table exists in Redshift. In the UI the users must select from a list that retrieves the Redshift tables, but programmatically they could introduce any table, which can result in non-existent tables being added to the data.all Catalog.

This PR 
1) checks if a table exists in Redshift and is accessible by the Redshift connection before adding it to data.all
2) in the UI, it shows a success banner with the list of tables added and an error banner with the list of error tables if any.
 
### Testing
- Tested in AWS 
      - UI views show banner with added tables for importDataset and addDataset
      - UI import Dataset does not show anything if no tables are added
- Added unit tests - checked as part of PR checks 🟢 
- Added integration tests -  attached screenshot of redshift dataset tests

image

### Relates
- #1424 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/redshift_datasets/api/datasets/types.py
+++ b/backend/dataall/modules/redshift_datasets/api/datasets/types.py
@@ -70,6 +70,8 @@ RedshiftDataset = gql.ObjectType(
             type=gql.Ref('RedshiftConnection'),
             resolver=resolve_dataset_connection,
         ),
+        gql.Field('successTables', gql.ArrayType(gql.String)),
+        gql.Field('errorTables', gql.ArrayType(gql.String)),
     ],
 )
 
@@ -150,4 +152,9 @@ RedshiftDatasetTableColumnSearchResult = gql.ObjectType(
         gql.Field('hasNext', gql.Boolean),
         gql.Field('hasPrevious', gql.Boolean),
     ],
+)
+
+RedshiftAddTableResult = gql.ObjectType(
+    name='RedshiftAddTableResult',
+    fields=[gql.Field('successTables', gql.ArrayType(gql.String)), gql.Field('errorTables', gql.ArrayType(gql.String))],
 )
--- a/backend/dataall/modules/redshift_datasets/api/datasets/types.py
+++ b/backend/dataall/modules/redshift_datasets/api/datasets/types.py
@@ -1,5 +1,3 @@
-from email.policy import default
-
 from dataall.base.api import gql
 from dataall.modules.datasets_base.services.datasets_enums import DatasetRole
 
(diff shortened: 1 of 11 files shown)


backend/dataall/modules/redshift_datasets/api/datasets/types.py
```diff
@@ -70,6 +70,8 @@
             type=gql.Ref('RedshiftConnection'),
             resolver=resolve_dataset_connection,
         ),
+        gql.Field('successTables', gql.ArrayType(gql.String)),
```
noah-paige, line 73: would this cause any issues for other queries/mutations that return `RedshiftDataset` but do not have `successTables` or `errorTables` defined (such as `updateRedshiftDataset` or `getRedshiftDataset`)?

backend/dataall/modules/redshift_datasets/services/redshift_dataset_service.py
```diff
@@ -112,17 +112,30 @@ def import_redshift_dataset(uri, admin_group, data: dict):
...
+                        session=session,
+                        username=context.username,
+                        dataset_uri=dataset.datasetUri,
+                        data={'name': table},
+                    )
+                    RedshiftDatasetService._attach_table_permissions(session, dataset, environment, rs_table)
+                    DatasetTableIndexer.upsert(session=session, table_uri=rs_table.rsTableUri)
+                    success_tables.append(table)
+            dataset.successTables = success_tables
+            dataset.errorTables = error_tables
```
noah-paige, line 138: can this chunk of code and L224-252 be moved to function and defined once?

noah-paige (commented):
left a couple comments/qs on the fix

backend/dataall/modules/redshift_datasets/api/datasets/types.py
dlpzx: No, because we are not trying to fetch that data in the definition of those API calls

backend/dataall/modules/redshift_datasets/services/redshift_dataset_service.py
dlpzx: Good call out, I will push another version. I also want to add some unit and integration tests for this PR

backend/dataall/modules/redshift_datasets/api/datasets/types.py
noah-paige: but programmatically we could/would be with SDK/CLI

backend/dataall/modules/redshift_datasets/api/datasets/types.py
```diff
@@ -1,3 +1,5 @@
+from email.policy import default
```
noah-paige, line 1: remove unused import

noah-paige (commented):
think there is one unused import accidentally added in this PR - otherwise look good

Issue #1645, opened 2024-10-16
Force deletion of data.all Resources


dlpzx:
**Is your idea related to a problem? Please describe.**
I am frustrated when a share object cannot be deleted because some items are shared. The same applies to Datasets or Environments. In some cases I want to delete the objects from data.all and perform the manual clean-up in AWS myself.

This scenario can happen:
- a share had some problem and is very very stuck in Revoke_Failed
- during integration testing some test fails and some resources are orphan (e.g Dataset without bucket)

**Describe the solution you'd like**
An option to:
- force share deletion (even if items are shared), 
- force dataset deletion (even if shares are shared),
-  force environment deletion (even if resources exist)
This should be optional and enough warnings should be provided so that users understand that the underlying AWS infra is still there.
*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

dlpzx:
Closing for the time-being because #1646 is enough. If needed this issue will be re-opened in the future

PR #1646, opened 2024-10-17
Add forceDelete to shareObjects to clean-up all shareItems

dlpzx:
### Feature or Bugfix
- Feature

### Detail

Implements #1645 for share requests. 

This PR adds a new input field in the deleteShareObject API called `forceDelete` that allows data.all users to force the deletion of a share object and all of its share items even if there are share items that are in shared states (e.g. Share_Succeeded, Revoke_Failed, Share_In_Progress...).
- [X] Delete of the metadata in RDS
- [X] Clean-up share AWS resources/actions as much as possible

The use-case is explained in #1645, but more specifically for shares, the main scenario are shares that have orphan resources, like an S3 Bucket or an IAM role that has been deleted. In this cases the users would have to re-create the resources and then revoke them and delete them, which is quite cumbersome. An easier way forward is to force deletion and clean-up. 

This feature is intended for developers and maintainers that call the APIs programmatically and only for this specific error handling situations.

### Testing
- [X] Deployed in AWS
- [X] A share with shared items throws an error when force_delete = False
- [X] A share without shared items is deleted without any clean-up tasks for any value of force_delete
- [X] A share with shared items is cleaned up when force_delete = True - the ECS triggers clean-up task that successfully wipes:
    - [X] Broken S3 Bucket share - remove bucket policy statements
    - [X] Broken S3 Access Points share - delete folder
    - [X] Broken Glue table share - remove Lake Formation permissions to target role in source account
    - [X] Broken Redshift table share - delete one of 2 shared tables
For each of the above cases I verified in the CloudWatch logs that the manual errors introduced were handled gracefully and that they clean-up task was able to clean-up the shares.
### Relates
- #1645 

### Security

  - Is the input sanitized?


tests/modules/s3_datasets_shares/test_share.py
```diff
@@ -2682,7 +2682,7 @@ def test_delete_share_object_remaining_items_error(
         client=client, user=user2, group=group2, shareUri=share3_processed.shareUri
     )
     # Then we get an error of the type
-    assert 'UnauthorizedOperation' in delete_share_object_response.errors[0].message
+    assert 'ShareItemsFound' in delete_share_object_response.errors[0].message
```
petrkalos, line 2685: I guess this is fixing an existing bug (we return the wrong message), right?

backend/dataall/modules/shares_base/tasks/share_manager_task.py
```diff
@@ -39,6 +39,10 @@
             log.info(f'Starting re-apply task for share : {share_uri}...')
             SharingService.reapply_share(engine=ENGINE, share_uri=share_uri)
 
+        elif handler == 'cleanup_share':
+            log.info(f'Starting clean-up task for share : {share_uri}...')
+            SharingService.cleanup_share(engine=ENGINE, share_uri=share_uri)
```
petrkalos, line 44: nit: this entire block can be simplified with the following...

```python
getattr(SharingService, handler)(engine=ENGINE, share_uri=share_uri)
```

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -603,3 +604,155 @@ def verify_shares(self) -> bool:
...
+                local_db_exists = self.redshift_data_client_in_target.check_database_exists(database=self.local_db)
+                # 1) (in target namespace) Revoke access to the revoked tables to the redshift role in external schema (if schema exists)
+                if local_db_exists and self.redshift_data_client_in_target.check_schema_exists(
+                    schema=self.external_schema, database=self.target_connection.database
+                ):
+                    try:
+                        self.redshift_data_client_in_target.revoke_select_table_access_to_redshift_role(
+                            schema=self.external_schema, table=table.name, rs_role=self.redshift_role
+                        )
+                    except Exception:
```
petrkalos, line 649: You repeat this exception suppression a couple of times in this PR. Perhaps you can create a utility function that takes a callble as an input and reuse it, something like...

```python
def suppress(func: Callable, exc: Type[Exception] = Exception, *args, **kwargs):
    try:
        func(*args, **kwargs)
    except exc:
        logging.exception()
```

petrkalos (approved):

tests/modules/s3_datasets_shares/test_share.py
dlpzx: Yes, this was a small bug

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -603,3 +604,147 @@ def verify_shares(self) -> bool:
...
+            self._initialize_clients()
+            for table in self.tables:
+                log.info(f'Revoking access to table {table}...')
+                local_db_exists = self.redshift_data_client_in_target.check_database_exists(database=self.local_db)
+                # 1) (in target namespace) Revoke access to the revoked tables to the redshift role in external schema (if schema exists)
+                if local_db_exists and self.redshift_data_client_in_target.check_schema_exists(
+                    schema=self.external_schema, database=self.target_connection.database
+                ):
+                    execute_and_suppress_exception(
+                        func=self.redshift_data_client_in_target.revoke_select_table_access_to_redshift_role(
```
petrkalos, line 646: you need to invoke it like `suppress(self.redshift_data_client_in_target, schema=self.extrernal_schema, table=table.name, ...)`, essentially defering the execution to the `suppress` util. Your current code will execute the method and then pass the returned value into the util thus if it throws it won't work.

petrkalos (changes requested):
usage of the util needs fixing

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
dlpzx: Uff, my bad, I was too fast and did not pay enough attention. It is fixed now but just in case I am doing some testing

Issue #1459, opened 2024-08-06
Consumer roles list page is very slow


fourtyplustwo:
We have about 15 total IAM roles on a single environment, 9 of them are data.all managed. The list loads very slowly about 13 seconds or so. I would like the roles to load much faster maybe 2-3 seconds tops and ideally less.

noah-paige:
Thanks for raising this concern @zsaltys - I believe @SofiaSazonova had worked on a similar issue where fetching the managed policy status for many consumption roles when requesting share access ([PR](https://github.com/data-dot-all/dataall/pull/1303)) 

I would think we can re-use a similar approach where we load all of the role information besides this additional fields which is resolved via boto3 call and likely causing the long runtime

That information can be run afterwards via separate API call or on displayed on demand via some mechanism

Please let us know if there is any particular expected behavior(s) you had in mind other than faster runtimes?

zsaltys:
@noah-paige I would say we should check if role has policy attached async in background ECS jobs and allow user to refresh a status if they need to.. Another option is to load it asynchronously from the list. Meaning we render the consumer roles list FIRST ... and then we show spinning circle that we're pulling information to figure out if the policy is attached.. I think maybe the second option would be my preferred one as it's going to show the truth just a bit slower...

Issue #884, opened 2023-11-17
IAM Policy enhancements - Split policy statements in chunks


anushka-singh:
**Is your idea related to a problem? Please describe.**
The limit for managed IAM policies is 6144 characters. Based on a dirty approximation, each team can request access to around 35 buckets based on the number of characters in policy statements. 

**Describe the solution you'd like**
In v2.1.0 we added some utilities in dataall/base/utils/iam_policy_utils.py. We can reuse them to limit the resources per policy and splitting policies.

dlpzx:
Thanks for opening the issue for visibility @anushka-singh. We will work together on introducing this feature because as you noticed it includes some scenarios out of the ones considedred in the iam_policy_utils

zsaltys:
@dlpzx @SofiaSazonova @anmolsgandhi  we ran into this issue on IAM policies when a role accumulated 38+ shares. Would be nice to prioritize this one.

petrkalos:
I started an investigation into it and here are my findings so far...

When using data.all generated IAM roles we already use all the policies (max 10 per role) for services hence we cannot add anymore. We can assume that not all deployment will have all the features enabled and those 10 might be fewer and as such have some space but I don't consider this a good solution.

When using consumption roles we don't have those service policies hence in theory we have space for adding more managed policies but keeping in mind that those roles are not data.all managed and hence users might have already added their own policies.

Overall to tackle this issue we need to come up with a smarter (which probably means dynamic) policy manager which will maximise the capacity based on the following limitations
* 10 managed policies per role
* 6144 chars per managed policy
* 10240 chars in total across all the inline policies per role

TejasRGitHub:
Hi @petrkalos , 

We are seeing IAM policy limit getting exceeded on a managed roles. I think an initial version where if the policy gets exceeded, another policy is created and attached would be helpful. I think data.all would somehow ( maybe ) have to remember the policy name it created and try to modify the policy which still has space left. 

FYI , if needed the limit can be pushed to 20 managed policies per role- https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-entity-length:~:text=Managed%20policies%20per%20role . Please correct me if I am wrong

TejasRGitHub:
[**WIP**]
####  Current IAM Roles and Policy management in data.all 

Data.all creates managed roles for data.all constructs like Environment, Dataset. These managed roles contain managed and inline policies. 

[...]

PR #1650, opened 2024-10-17
[Gh 884] IAM policy splitting for requestor IAM policies

TejasRGitHub:
### Feature or Bugfix
- Feature

### Detail

- Updates the process of modifying the IAM policies after approve / revoke to add / delete resources and also split the policies into chunks 
- Updates the managed IAM policies to have indexes at the end 
- Contains backward compatibility and other additional checks to make sure correct policies are created and older policies are deleted
- Contains a UI update to address this issue - https://github.com/data-dot-all/dataall/issues/1459

### Relates

- https://github.com/data-dot-all/dataall/issues/884
- https://github.com/data-dot-all/dataall/issues/1459

### Tests

- Inviting / removing team to an environment ✅ 
- Adding/ Removing consumption role to an environment ✅ 
- Creating / revoking a share with consumption role for S3 bucket with bucketpolicy sharing ✅ 
- Creating / revoking a share with environment team for S3 bucket with bucketpolicy sharing  ✅ 
- Creating / revoking a share with consumption role for S3 access point  ✅ 
- Creating / revoking a share with environment team for S3 access point  ✅ 
- Creating a share with requestors IAM policy being on the brink of policy length limits and checking if new indexed policy is getting created ✅ 
- Revoking a share with requestors IAM policy being on the brink of policy length limits and checking if extra policies are deleted ✅ 
- Checking if all consumption roles are loaded quickly on the environment teams UI asynchronously  ✅ 
- Check if requestor’s receive emails when an IAM role is about to reach the managed policy attachment limit ( Email Notification Changes ) 
- Migrating Shares from old to new policy management. via share validator and share verifier ( Share verifier successfully mentions error message and Share re-applier successfully converts policies to indexed managed policies ) ✅ 
- Creating new environment and deleting an environment ✅ 

### Security

  - Is the input sanitized?
  - Are the used keys controlled by the customer? Where are they stored?No 
- Are you introducing any new policies/roles/users? **yes**
  - Have you used the least-privilege principle? How? **yes**


anushka-singh:
Can we fix the integration tests? Why are they failing?

deploy/stacks/container.py
```diff
@@ -260,7 +260,7 @@ def add_share_management_task(self):
             f'ShareManagementTaskContainer{self._envname}',
             container_name='container',
             image=ecs.ContainerImage.from_ecr_repository(repository=self._ecr_repository, tag=self._cdkproxy_image_tag),
-            environment=self._create_env('DEBUG'),
+            environment=self.env_vars,
```
TejasRGitHub, line 263: Adding self.env_vars as they contain necessary configs for email notification code to work. 
Email notifications are sent if attachable managed IAM policy quota is exceeded

frontend/src/modules/Environments/services/listAllEnvironmentConsumptionRoles.js
```diff
@@ -29,11 +29,6 @@ export const listAllEnvironmentConsumptionRoles = ({
           groupUri
           IAMRoleArn
           dataallManaged
-          managedPolicies {
```
TejasRGitHub, line 32: removing this from here as this is what was causes the gql to load slowly. Replacing this with an asynchronous way of fetching this same information

backend/dataall/base/aws/service_quota.py
```diff
@@ -0,0 +1,56 @@
...
+    def get_service_quota_value(self, service_code, service_quota_code):
+        try:
+            log.info(
+                f'Getting service quota for service code: {service_code} and service quota code: {service_quota_code}'
+            )
+            response = self.client.get_service_quota(ServiceCode=service_code, QuotaCode=service_quota_code)
+            return response['Quota']['Value']
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'AccessDenied':
+                raise Exception(f'Data.all Environment Pivot Role does not have permissions to list quota codes: {e}')
```
dlpzx, line 54: nit: exception error message does not match this API call

backend/dataall/core/environment/cdk/pivot_role_core_policies/service_quota.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+
+    def get_statements(self):
+        statements = [
+            # Service Quota - Needed to determine the number of service quotas for managed policies which can be attached
+            iam.PolicyStatement(
+                sid='ServiceQuotaListGet',
+                effect=iam.Effect.ALLOW,
+                actions=['servicequotas:List*', 'servicequotas:Get*'],
+                resources=['*'],
```
dlpzx, line 19: I confirm that we cannot restrict this policy to specific resources. It requires a wildcard

backend/dataall/core/environment/cdk/pivot_role_stack.py
```diff
@@ -36,7 +41,14 @@ def generate_policies(self) -> List[iam.ManagedPolicy]:
             statements.extend(service.get_statements(self))
             logger.info(f'statements: {str(service.get_statements(self))}')
 
-        statements_chunks = split_policy_statements_in_chunks(statements)
+        statements_json = [statement.to_json() for statement in statements]
+        statement_chunks_json = split_policy_statements_in_chunks(statements_json)
+        statements_chunks = []
```
dlpzx, line 46: Can we move the logic between 46 and 51 into the iam_cdk_utils? As a user I want a utility to convert a already-chunked json policy to cdk, I don't need to know about conditions or resources...

backend/dataall/base/utils/iam_policy_utils.py
```diff
@@ -9,7 +8,7 @@
 MAXIMUM_NUMBER_MANAGED_POLICIES = 20  # Soft limit 10, hard limit 20
 
 
-def split_policy_statements_in_chunks(statements: List):
+def split_policy_statements_in_chunks(statements: List[Any]):
     """
```
dlpzx, line 12: In the typing hints, isn't it a List[Dict]?

backend/dataall/base/utils/naming_convention.py
```diff
@@ -56,6 +56,18 @@ def build_compliant_name(self) -> str:
...
         return f"{slugify(self.resource_prefix + '-' + self.target_label[:(max_length - len(self.resource_prefix + self.target_uri))] + suffix, regex_pattern=fr'{regex}', separator=separator, lowercase=True)}"
 
+    def build_compliant_name_with_index(self, index: int = None) -> str:
+        """
+        Builds a compliant AWS resource name
+        """
+        regex = NamingConventionPattern[self.service].value['regex']
+        separator = NamingConventionPattern[self.service].value['separator']
+        max_length = NamingConventionPattern[self.service].value['max_length']
+        index_string = f'-{index}' if index is not None else '- '
```
dlpzx, line 66: nit: the mix with the conditionals and what is a suffix is a bit confusing. We could improve it by directly:
```
        suffix = f'-{self.target_uri}' if len(self.target_uri) else ''
        index_suff = f'-{index}' if index is not None else ''
        return f"{slugify(self.resource_prefix + '-' + self.target_label[:(max_length - len(self.resource_prefix + self.target_uri + index_string))] + suffix + index_suff, regex_pattern=fr'{regex}', separator=separator, lowercase=True)}"

```

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -99,12 +156,11 @@ def create_all_policies(self, managed) -> bool:
         Manager that registers and calls all policies created by data.all modules and that
         need to be created for consumption roles and team roles
         """
-        try:
-            for Policy in self.initializedPolicies:
-                empty_policy = Policy.generate_empty_policy()
-                policy_name = Policy.generate_policy_name()
-                logger.info(f'Creating policy {policy_name}')
-
+        for policy_manager in self.initializedPolicies:
```
dlpzx, line 159: nit: why the rename? the inializedPolicies are policies not the manager

dlpzx, line 159: Why this name change? in initializedPolicies we still have ManagedPolicies instances

dlpzx: I see, the ManagedPolicy ABC class now handles single policies as well as list of policies. We need to change this a bit as the result is a very complex interface that is doing too many things and is very focused on the S3SharePolicyService(ManagedPolicy) service too much.

dlpzx: We could update ManagedPolicy to ManagedPolicyList:
- `generate_policy_name` can be renamed as `generate_base_policy_name`
- if possible (sometimes it is not the case), the logic of migrating should not be in the interface and we should avoid methods such as `generate_old_policy_name` or limit it to one single method (e.g. create_indexed_policies_from_single_policy_delete_single_policy)
- `generate_indexed_policy_name` should not be an abstract method, cannot it be constructed with the base name directly?
- for the non-abstract methods - we are duplicating the methods: check_if_policy_exists and check_if_managed_policies_exists. We should have a single ManagedPolicyList with check_if_policies_exist, check_if_policies_attached
- there should not be any references to shares (e.g. `share_policy_name`)

dlpzx (changes requested):
Hi @TejasRGitHub thanks for the contribution! I reviewed the IAM chunking with the new utilities and left some easy-fix comments. I stopped after reviewing the  the `ManagedPolicy` interface as it requires some re-work that could affect the rest of the implementation.

backend/dataall/base/aws/service_quota.py
TejasRGitHub: Do you mean instead of `list quota` I should add `list_service_quota`

backend/dataall/core/environment/cdk/pivot_role_core_policies/service_quota.py
TejasRGitHub: Hmm.. Can you tell me how to go about finding this ? 
Also, I do not get any exception thrown for this at all 😅  . Maybe because I am getting service quotas for iam and there is a `iam:Get*` permission somewhere

backend/dataall/core/environment/cdk/pivot_role_stack.py
TejasRGitHub: Oh yes. Will do

TejasRGitHub:
@dlpzx , pushed updated PR. Integration tests are failing and will fix them once the PR is reviewed and in a good shape.

backend/dataall/base/utils/naming_convention.py
```diff
@@ -56,6 +56,20 @@ def build_compliant_name(self) -> str:
         suffix = f'-{self.target_uri}' if len(self.target_uri) else ''
         return f"{slugify(self.resource_prefix + '-' + self.target_label[:(max_length - len(self.resource_prefix + self.target_uri))] + suffix, regex_pattern=fr'{regex}', separator=separator, lowercase=True)}"
 
+    def build_compliant_name_with_index(self, index: int = None) -> str:
+        """
+        Builds a compliant AWS resource name with an index at the end of the policy name
```
TejasRGitHub, line 61: Added comments to knw the developer / users to let them know how this functions is different from the other ( build_compliant_name ) and also added a note to let the user know why index string length is considered into generating the naming convention

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -29,42 +29,69 @@ def policy_type(self):
...
-    def generate_empty_policy(self) -> dict:
+    def generate_indexed_policy_name(self, index) -> str:
         """
-        Returns dict and needs to be implemented in the ManagedPolicies inherited classes
+        Returns string of policy name with index at the end .Needs to be implemented in the ManagedPolicies inherited classes
         """
         ...
 
     @abstractmethod
-    def create_managed_policy_from_inline_and_delete_inline(self) -> str:
```
TejasRGitHub, line 46: Simplifying interface and removing method not related to the interface

backend/dataall/core/environment/services/managed_iam_policies.py
TejasRGitHub: Updated ManagedPolicy class to remove the duplicating methods and also removed methods from interface which are not needed. Kept the name of the class same for now

backend/dataall/modules/s3_datasets_shares/services/s3_share_validator.py
```diff
@@ -122,30 +122,3 @@ def _validate_iam_role_and_policy(
                 action=principal_type,
                 message=f'The principal role {principal_role_name} is not found.',
             )
-
```
TejasRGitHub, line 125: Removed code is used while creating share object in the graphql lambda calls . I think this is not needed since even if the share policies don't exists and are not created / attached, a share object should be created. 
Once the share object is approved, revoked , then the code for processing the share does always take care of this thing - migrating share policy from inline/ managed ( w/o index ) to managed ( w/ index ) and attaching them at the end.

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -29,42 +29,69 @@ def policy_type(self):
...
-
-    def attach_policy(self):
-        policy_arn = f'arn:aws:iam::{self.account}:policy/{self.generate_policy_name()}'
-        try:
-            IAM.attach_role_policy(self.account, self.region, self.role_name, policy_arn)
-        except Exception as e:
-            raise Exception(f"Required customer managed policy {policy_arn} can't be attached: {e}")
+    def get_managed_policies(self) -> List[str]:
+        policy_pattern = self.generate_base_policy_name()
+        share_policies = self._get_policy_names(policy_pattern)
```
dlpzx, line 65: Remove specific references to share_managed_policies. Instead of `share_policies` --> `managed_policies`

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -29,42 +29,69 @@ def policy_type(self):
...
+        return share_policies
+
+    def check_if_policy_attached(self, policy_name):
+        is_policy_attached = IAM.is_policy_attached(self.account, self.region, policy_name, self.role_name)
+        return is_policy_attached
+
+    def get_policies_unattached_to_role(self):
+        policy_pattern = self.generate_base_policy_name()
+        share_policies = self._get_policy_names(policy_pattern)
```
dlpzx, line 74: Remove specific references to share_managed_policies.  Instead of share_policies --> managed_policies

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -29,42 +29,69 @@ def policy_type(self):
...
+    def attach_policies(self, managed_policies_list: List[str]):
+        for policy_name in managed_policies_list:
+            policy_arn = f'arn:aws:iam::{self.account}:policy/{policy_name}'
+            try:
+                IAM.attach_role_policy(self.account, self.region, self.role_name, policy_arn)
+            except Exception as e:
+                raise Exception(f"Required customer managed policy {policy_arn} can't be attached: {e}")
+
+    def _get_policy_names(self, policy_pattern):
+        share_policies = IAM.list_policy_names_by_policy_pattern(self.account, self.region, policy_pattern)
```
dlpzx, line 90: Remove specific references to share_managed_policies. Instead of `share_policies` --> `managed_policies`

backend/dataall/core/environment/services/managed_iam_policies.py
dlpzx: It happens in a couple of places in the abstract class

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -29,42 +29,69 @@ def policy_type(self):
...
-            raise Exception(f"Required customer managed policy {policy_arn} can't be attached: {e}")
+    def get_managed_policies(self) -> List[str]:
+        policy_pattern = self.generate_base_policy_name()
+        share_policies = self._get_policy_names(policy_pattern)
+        return share_policies
+
+    def check_if_policy_attached(self, policy_name):
+        is_policy_attached = IAM.is_policy_attached(self.account, self.region, policy_name, self.role_name)
+        return is_policy_attached
+
```
dlpzx, line 71: In places such as `backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py:407` we are iterating over the policies and calling `check_if_policy_attached`, which we could do by calling `get_policies_unattached_to_role` directly. If we only use that function in other services, `check_if_policy_attached` becomes a private method of the abstract class

backend/dataall/modules/s3_datasets_shares/services/s3_share_managed_policy_service.py
```diff
@@ -1,9 +1,19 @@
...
 from dataall.base.aws.iam import IAM
+from dataall.base.aws.service_quota import ServiceQuota
+from dataall.base.db.exceptions import AWSResourceQuotaExceeded
+from dataall.base.utils.iam_policy_utils import (
+    split_policy_statements_in_chunks,
+    split_policy_with_resources_in_statements,
+)
 from dataall.base.utils.naming_convention import NamingConventionService, NamingConventionPattern
 from dataall.core.environment.services.managed_iam_policies import ManagedPolicy
 import logging
```
dlpzx, line 14: Unused import ShareNotificationService

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -119,32 +145,45 @@ def create_all_policies(self, managed) -> bool:
...
         Manager that registers and calls all policies created by data.all modules and that
         need to be deleted for consumption roles and team roles
         """
-        try:
-            for Policy in self.initializedPolicies:
-                policy_name = Policy.generate_policy_name()
+        for policy_manager in self.initializedPolicies:
+            policy_name_list = policy_manager.get_managed_policies()
+
+            # Check if policy with old naming format exists
```
dlpzx, line 161: Although in theory it is impossible that there is an old_policy co-existing with new_indexed_policies, I would prefer to not make this assumption and remove 162 if-conditional as it is not needed and we ensure that no matter what we delete the old_policy

backend/dataall/core/environment/services/managed_iam_policies.py
```diff
@@ -153,13 +192,22 @@ def get_all_policies(self) -> List[dict]:
...
-                'policy_type': Policy.policy_type,
-                'exists': Policy.check_if_policy_exists(),
-                'attached': Policy.check_if_policy_attached(),
-            }
-            all_policies.append(policy_dict)
-        logger.info(f'All policies currently added to role {str(all_policies)}')
+        for policy_manager in self.initializedPolicies:
+            policy_name_list = policy_manager.get_managed_policies()
+
+            # Check if policy with old naming format exists
```
dlpzx, line 198: Same as in line 162

backend/dataall/modules/s3_datasets_shares/services/s3_share_validator.py
dlpzx: That is not entirely True, if in the share object creation request the customer does not select "attachMissingPolicies" then the managed policies won't be attached during the share task

backend/dataall/base/db/exceptions.py
```diff
@@ -147,6 +147,18 @@ def __str__(self):
         return f'{self.message}'
 
 
+class AWSResourceQuotaExceeded(Exception):
```
dlpzx, line 150: nit: `ServiceQuota` instead of `ResourceQuota` to match the actual name on AWS

backend/dataall/base/aws/service_quota.py
dlpzx: yes, it was a copy/paste from line 41

backend/dataall/base/utils/naming_convention.py
dlpzx: Understood!

backend/dataall/core/environment/cdk/pivot_role_core_policies/service_quota.py
dlpzx: Exactly, we must be granting iam:Get somewhere (which is fine, it is read only) This list comes from granting the permissions manually to any role in the AWS Console. It provides hints on the types of resources that can be used in a policy and other associated policies that are needed

backend/dataall/modules/s3_datasets_shares/services/s3_share_managed_policy_service.py
```diff
@@ -18,6 +28,9 @@
 EMPTY_STATEMENT_SID = 'EmptyStatement'
 
 S3_ALLOWED_ACTIONS = ['s3:List*', 's3:Describe*', 's3:GetObject']
+IAM_SERVICE_NAME = 'AWS Identity and Access Management (IAM)'
+IAM_SERVICE_QUOTA_NAME = 'Managed policies per role'
+DEFAULT_MAX_ATTACHABLE_MANAGED_POLICIES_ACCOUNT = 10
```
dlpzx, line 33: If in the future we need to implement other Service Quota exceptions, then we will create something similar to the NamingConvention utility but for Service quotas where we defined the service names and quotas. But we don't need it at this moment, so having the base exceptions + listing the constants up here as constants is enough. Just keep it in mind for future implementations (and reviews ;) )

backend/docker/dev/Dockerfile
```diff
@@ -55,12 +54,10 @@ RUN /bin/bash -c ". ~/.nvm/nvm.sh && \
...
-/root/.nvm/versions/node/${NODE_VERSION}/bin:\
-$(${PYTHON_VERSION} -m site --user-base)/bin:\
-$(python3 -m site --user-base)/bin:\
-$(python -m site --user-base)/bin:\
-/home/${CONTAINER_USER}/.local/bin:\
-$PATH" >> ~/.bashrc && \
+    /root/.nvm/versions/node/${NODE_VERSION}/bin:\
+    $(${PYTHON_VERSION} -m site --user-base)/bin:\
+    $(python3 -m site --user-base)/bin:\
+    $PATH" >> ~/.bashrc && \
```
dlpzx, line 60: Are these changes part of this PR?

docker-compose.yaml
dlpzx, line 1: Are these changes part of the PR?

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -32,13 +35,16 @@
...
 )
+from dataall.modules.shares_base.services.share_notification_service import ShareNotificationService
 from dataall.modules.shares_base.services.shares_enums import PrincipalType
 from dataall.modules.s3_datasets.db.dataset_models import DatasetStorageLocation, S3Dataset
 from dataall.modules.shares_base.services.sharing_service import ShareData
 
 logger = logging.getLogger(__name__)
+ACCESS_POINT_CREATION_TIME = 30
+ACCESS_POINT_CREATION_RETRIES = 10
+ACCESS_POINT_BACKOFF_COEFFICIENT = 1.1  # every time increase retry delay by 10%
```
dlpzx, line 47: This logic related to the S3 access points creation retries seems to not belong to this PR

frontend/src/modules/Environments/services/listAllEnvironmentConsumptionRoles.js
dlpzx: If we are not going to use it in the frontend, we should also remove it from the backend gql type where we are fetching it. (`backend/dataall/core/environment/api/types.py:183`) Please ensure that there is no other place where we get or list ConsumptionRoles and use the managedPolicies field

deploy/stacks/container.py
dlpzx: Aren't `self.env_vars = self._create_env()` ? Does this change change anything? It just replaces the calling the function by using the variable directly no?

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -165,26 +171,45 @@ def check_target_role_access_policy(self) -> None:
...
+        if not is_managed_policies_exists:
+            warn(
+                "Convert all your share's requestor policies to managed policies with indexes.",
+                DeprecationWarning,
+                stacklevel=2,
             )
-            self.folder_errors.append(
-                ShareErrorFormatter.dne_error_msg('IAM Policy attached', share_resource_policy_name)
+            old_managed_policy_name = share_policy_service.generate_old_policy_name()
+            if not share_policy_service.check_if_policy_exists(policy_name=old_managed_policy_name):
```
dlpzx, line 193: We can call `share_policy_service.check_if_policy_exists(policy_name=old_managed_policy_name)` just once and use an if-else clause to avoid re-calling it in line 200

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -165,26 +171,45 @@ def check_target_role_access_policy(self) -> None:
...
+                logger.info(
+                    f'No managed policy exists for the role: {self.target_requester_IAMRoleName}, Reapply share create managed policies.'
+                )
```
dlpzx, line 196: Nit: improve error message "Reapply share create managed policies." ---> "Reapply the share to re-create the necessary managed policies."

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -165,26 +171,45 @@ def check_target_role_access_policy(self) -> None:
...
+                self.folder_errors.append(ShareErrorFormatter.dne_error_msg('IAM Policy', share_resource_policy_name))
+                return
+
```
dlpzx, line 199: Should `if share_policy_service.check_if_policy_exists(policy_name=old_managed_policy_name):` throw an verification error? verify-shares checks that the share is healthy, if the share has an not-indexed policy it is healthy, the data can be accessed no?

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -194,101 +219,97 @@ def check_target_role_access_policy(self) -> None:
...
-        s3_statement_index = S3SharePolicyService._get_statement_by_sid(
-            policy_document, f'{IAM_S3_ACCESS_POINTS_STATEMENT_SID}S3'
-        )
-
-        if s3_statement_index is None:
-            logger.info(f'IAM Policy Statement {IAM_S3_ACCESS_POINTS_STATEMENT_SID}S3 does not exist')
+        if not S3SharePolicyService.check_if_sid_exists(
+            f'{IAM_S3_ACCESS_POINTS_STATEMENT_SID}S3', share_policy_service.total_s3_access_point_stmts
+        ):
+            logger.info(f'IAM Policy Statement with base Sid: {IAM_S3_ACCESS_POINTS_STATEMENT_SID}S3 does not exist')
```
dlpzx, line 225: nit: can we change the logs a bit? At first I assumed we were checking only if `{IAM_S3_ACCESS_POINTS_STATEMENT_SID}S3` exists, but it turns out that we are checking if there is any sid that starts with `{IAM_S3_ACCESS_POINTS_STATEMENT_SID}S3`

dlpzx: Similar to the log error in line 290

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -77,122 +77,149 @@ def check_s3_iam_access(self) -> None:
...
+        if not is_managed_policies_exists:
+            warn(
+                "Convert all your share's requestor policies to managed policies with indexes. Deprecation >= ?? ",
+                DeprecationWarning,
+                stacklevel=2,
             )
-            self.bucket_errors.append(
-                ShareErrorFormatter.dne_error_msg('IAM Policy attached', share_resource_policy_name)
+            old_managed_policy_name = share_policy_service.generate_old_policy_name()
+            if not share_policy_service.check_if_policy_exists(policy_name=old_managed_policy_name):
```
dlpzx, line 100: Same comments as in the access_points_manager: https://github.com/data-dot-all/dataall/pull/1650#discussion_r1822289919 and https://github.com/data-dot-all/dataall/pull/1650#discussion_r1822300141

backend/dataall/modules/shares_base/services/share_notification_service.py
```diff
@@ -24,6 +24,7 @@ class DataSharingNotificationType(enum.Enum):
     SHARE_OBJECT_EXTENSION_REJECTED = 'SHARE_OBJECT_EXTENSION_REJECTED'
     SHARE_OBJECT_REJECTED = 'SHARE_OBJECT_REJECTED'
     SHARE_OBJECT_PENDING_APPROVAL = 'SHARE_OBJECT_PENDING_APPROVAL'
+    SHARE_OBJECT_FAILED = 'SHARE_OBJECT_FAILED'
```
dlpzx, line 27: unused variable

backend/dataall/modules/shares_base/services/share_notification_service.py
```diff
@@ -109,6 +110,38 @@ def notify_persistent_email_reminder(self, email_id: str):
...
+        share_link_text = ''
+        if os.environ.get('frontend_domain_url'):
+            share_link_text = (
+                f'<br><br>Please visit data.all <a href="{os.environ.get("frontend_domain_url")}'
+                f'/console/shares/{self.share.shareUri}">share link</a> '
+                f'to view more details.'
+            )
+
+        msg_intro = f"""Dear User,
+
```
dlpzx, line 123: weird sentence: `We are contacting you because **for** a share requested by {email_id} failed` ---> We are contacting you because a share requested by {email_id} failed

backend/dataall/modules/shares_base/services/share_notification_service.py
```diff
@@ -109,6 +110,38 @@ def notify_persistent_email_reminder(self, email_id: str):
...
+            share_link_text = (
+                f'<br><br>Please visit data.all <a href="{os.environ.get("frontend_domain_url")}'
+                f'/console/shares/{self.share.shareUri}">share link</a> '
+                f'to view more details.'
+            )
+
+        msg_intro = f"""Dear User,
+        
+        We are contacting you because for a share requested by {email_id} failed because no new managed policy can be attached to your IAM role {self.share.principalRoleName}.
+        Please check the service quota for the managed policies that can be attached to a role in your aws account and increase the limit.
```
dlpzx, line 125: nit: for the IAM Managed Policies that can be attached to an IAM role in your AWS account

backend/dataall/modules/shares_base/services/share_notification_service.py
```diff
@@ -109,6 +110,38 @@ def notify_persistent_email_reminder(self, email_id: str):
...
+        For reference please take a look at this link - https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-entities
+    
+        Note - Previously made shares are not affected but all new shares will be failed till the time you increase the IAM quota limit.
```
dlpzx, line 128: nit: all new shares will fail if the IAM Managed Policies quota is not increased or if you do not dettach other IAM policies from the requester IAM role

dlpzx (commented):
Hi @TejasRGitHub thank you for the changes, it is way more in shape :) I added some additional comments with suggestions and improvements. I am also going to request you to 1) fix the failing integration tests and 2) list in the PR description the testing that you have done to ensure this PR is not introducing any error. include all testing scenarios: approve-verify-revoke new share, approve-verify-revoke old share, upgrade an old share....

backend/dataall/modules/s3_datasets_shares/services/s3_share_validator.py
TejasRGitHub: Yeah. You make a good point here. Let me undo those changes

backend/dataall/core/environment/services/managed_iam_policies.py
TejasRGitHub: Oh . Gotcha. Will do!

backend/dataall/core/environment/services/managed_iam_policies.py
TejasRGitHub: @dlpzx , Thanks for pointing this out. I missed this during refactoring and will make the changes. 

At places I do want to only check single policies ( like if old policy is attached, etc ) that's why I need the `check_if_policy_attached`

backend/dataall/modules/s3_datasets_shares/services/s3_share_managed_policy_service.py
TejasRGitHub: Thanks. You have hawk eyes 😄

backend/dataall/core/environment/services/managed_iam_policies.py
TejasRGitHub: There are situations in which we have seen some shares have old managed policies but no new managed policies. For example, an environment was onboarded in v2.6 then in v2.7 this IAM splitting change was deployd but they didn't migrate to new policy. Now in this case, they have a residual old managed policy which is not deleted. 
I don't see any problems arising due to this. But I thought this should also be deleted. 
Do you think this check needs to be removed ?

backend/dataall/base/db/exceptions.py
TejasRGitHub: I was thinking of this file containing sort of generic exception ( but user defined ), and so I named this as ResourceQuota and it could applied to anywhere quota is exceeded. Do you still prefer to name it with ServiceQuota

backend/docker/dev/Dockerfile
TejasRGitHub: Let me remove that. Its not part of this PR. Something I was trying to do during development

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
TejasRGitHub: Yes. Maybe I had a merge conflict. Need to check why this shows up

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
TejasRGitHub: Will make this change. Thanks!!

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
TejasRGitHub: Function `share_policy_service.check_if_policy_exists(policy_name=old_managed_policy_name)` doesn't raise Client error exception. For any other exceptions, that exception error will be caught here - `dataall/modules/s3_datasets_shares/services/share_processors/s3_access_point_share_processor.py` and the share will be unhealthy

TejasRGitHub: Let me know if you are alluding to something else which I am not getting

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
TejasRGitHub: Made changes to make it more clear

backend/dataall/modules/shares_base/services/share_notification_service.py
TejasRGitHub: Removed unused variable

backend/dataall/modules/shares_base/services/share_notification_service.py
TejasRGitHub: Yes. I added this sentence. Thanks @dlpzx

backend/dataall/base/aws/service_quota.py
TejasRGitHub: Alright made changes

deploy/stacks/container.py
TejasRGitHub: self.env_vars are modified during the initialization code to include the email notification env variables. Function _create_env doesn't have those so I used self.env_vars

backend/dataall/base/aws/iam.py
```diff
@@ -66,6 +67,25 @@ def get_role_policy(
...
+    def list_policy_names_by_policy_pattern(account_id: str, region: str, policy_filter_pattern: str):
+        try:
+            client = IAM.client(account_id, region)
+            # Setting Scope to 'Local' to fetch all the policies created in this account
+            paginator = client.get_paginator('list_policies')
+            policies = []
+            for page in paginator.paginate(Scope='Local'):
+                policies.extend(page['Policies'])
+            policy_names = [policy.get('PolicyName') for policy in policies]
+            return [policy_nm for policy_nm in policy_names if re.search(policy_filter_pattern, policy_nm)]
```
TejasRGitHub, line 80: Modified this to filter through the list of policies which match the regex patter `{base_policy_name}-\d` ( e.g. ; datall-env-<uri>-role_name-0, datall-env-<uri>-role_name-1 , ... )

frontend/src/modules/Environments/services/listAllEnvironmentConsumptionRoles.js
TejasRGitHub: Let me check and remove it

TejasRGitHub:
@dlpzx , I have updated the PR and replied to your comments where I thought more discussion is needed. 

I have updated the tests that I had done while submitting the PR. I will perform those tests once the PR is ready for merging. 
Also, I will fix all the integration test which are failing once the PR is in good shape.

Thanks Again for taking the time to review and providing your insightful review comments

docker-compose.yaml
```diff
@@ -7,7 +7,7 @@ services:
       dockerfile: docker/dev/Dockerfile
       args:
         CONTAINER_UID: ${UID}
-    entrypoint: /bin/bash -c 'aws configure set region "eu-west-1" &&. ~/.nvm/nvm.sh && uvicorn cdkproxymain:app --host 0.0.0.0 --port 2805 --reload'
+    entrypoint: /bin/bash -c 'aws configure set region "eu-west-1" && . ~/.nvm/nvm.sh && uvicorn cdkproxymain:app --host 0.0.0.0 --port 2805 --reload'
```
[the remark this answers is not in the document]
dlpzx: it is fine :)

backend/dataall/base/db/exceptions.py
dlpzx: I would go for ServiceQuota, because of the AWS prefix. We can make it generic when we need it

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
dlpzx: I was doubting whether having old-policy-not-indexed should end up in an unhealthy stage. 
- CONS: the share is healthy, it still works. We are saying to the user that the share is broken when in reality is not
- PROS: we force the users to upgrade their shares

After thinking about it, I think the PROS surpass the CONS. Better to force the users to upgrade :) Resolving

backend/dataall/core/environment/services/managed_iam_policies.py
dlpzx: Ahhhhh, I understand now! This is actually preventing the deletion of the old policy if the upgrade to indexed policies was not successful or did not happen for any reason. Seeing this point of view I think it is more important to ensure the sharing policy is not deleted than to clean it up. Let's leave the if-clause. This issue can be resolved

dlpzx (commented):
Left one ultra minor change comment about the exceptions name; but other than that the rest looks good! The only missing thing before approve are the integration tests.

Thank you for the great work @TejasRGitHub

TejasRGitHub:
Hi @dlpzx , I have updated the PR and fixed unit tests. ✅ . Also , I have tested the code with the test cases ( as described in the PR details section ). 
I have updated the PR with a few code changes - for some fixes I found during testing - and also refactored code to simplify. 
Can you please take a look at the latest changes and let me know if this PR still looks good

backend/dataall/base/utils/iam_cdk_utils.py
```diff
@@ -1,34 +1,67 @@
...
+            else:
+                statements.append(convert_from_json_to_iam_policy_statement(statement))
+        statements_chunks.append(statements)
     return statements_chunks
+
+
+def process_and_split_policy_with_resources_in_statements(base_sid: str, effect: str, actions: List[str],
+                                                          resources: List[str], condition_dict: Dict = None):
+    if condition_dict is not None:
+        print(f"Condition dictionary is: {condition_dict}")
```
dlpzx: Let's remove the print or replace for log

backend/dataall/modules/s3_datasets/cdk/pivot_role_datasets_policy.py
```diff
@@ -1,7 +1,10 @@
 import os
 from dataall.base import db
+from dataall.base.utils.iam_cdk_utils import (
+    convert_from_json_to_iam_policy_statement_with_conditions,
+    process_and_split_policy_with_resources_in_statements,
+)
 from dataall.base.utils.iam_policy_utils import (
-    split_policy_with_resources_in_statements,
     split_policy_with_mutiple_value_condition_in_statements,
 )
```
dlpzx, line 9: Some unused imports: convert_from_json_to_iam_policy_statement_with_conditions, split_policy_with_mutiple_value_condition_in_statements

backend/dataall/base/utils/iam_cdk_utils.py
```diff
@@ -0,0 +1,71 @@
...
+        statements: List[iam.PolicyStatement] = []
+        for statement in statement_js_chunk:
+            if statement.get('Condition', None):
+                statements.append(convert_from_json_to_iam_policy_statement_with_conditions(statement))
+            else:
+                statements.append(convert_from_json_to_iam_policy_statement(statement))
+        statements_chunks.append(statements)
+    return statements_chunks
+
+
```
dlpzx, line 43: `process_and_split_policy_with_resources_in_statements` is a function doing 2 separate things at the same time: 1) policies with resources and 2) policies with conditions. Let's make 2 functions each doing one single thing:
`process_and_split_policy_with_resources_in_statements` and `process_and_split_policy_with_conditions_in_statements`

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -351,21 +351,6 @@ def grant_target_role_access_policy(self):
         if kms_key_id:
             kms_target_resources = [f'arn:aws:kms:{self.dataset_region}:{self.dataset_account_id}:key/{kms_key_id}']
 
-        managed_policy_exists = True if share_policy_service.get_managed_policies() else False
```
dlpzx: Just to confirm my understanding. This code block is no longer needed because in `merge_statements_and_update_policies` we already create the IAM policy: `dataall.modules.s3_datasets_shares.services.s3_share_managed_policy_service.S3SharePolicyService._create_empty_policies_with_indexes`

dlpzx (commented):
Left some minor changes!

backend/dataall/modules/shares_base/tasks/share_manager_task.py
```diff
@@ -7,10 +7,11 @@
 from dataall.base.loader import load_modules, ImportMode
 
 root = logging.getLogger()
+root.setLevel(logging.INFO)
 if not root.hasHandlers():
     root.addHandler(logging.StreamHandler(sys.stdout))
 log = logging.getLogger(__name__)
-log.setLevel(os.environ.get('LOG_LEVEL', 'INFO'))
+# log.setLevel(os.environ.get('LOG_LEVEL', 'INFO'))
```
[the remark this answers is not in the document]
TejasRGitHub: I commented this and added this line - root.setLevel(logging.INFO) ( which was previously present ) since with the new changes the logger was not working. I think this should be fixed. I will revert back the changes . This is the GH issue were I have raise the issue with logging - https://github.com/data-dot-all/dataall/issues/1680

backend/dataall/base/utils/iam_cdk_utils.py
TejasRGitHub: updated the PR and separated this. Thanks for pointing this out.

petrkalos (approved):

TejasRGitHub:
@petrkalos , @dlpzx , Thanks for approving this PR. I have resolved the merge conflicts which were present.

PR #1657, opened 2024-10-24
Added error view and unified utility to check tenant user

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Users that do not belong to the DAAdministrators team can still navigate to the admin console. They cannot do anything as all the API calls will return errors. This PR takes it a step further and shows an error message and does not render any of the components if the admin view. Here is an snapshot of what users view:

I considered restricting the route in routes, but because we need to use the useGroups hook, it was giving a lot of errors.

image

In addition, we have created a utility that unifies the checks to this Cognito group in other places of the frontend.

### Security

  - Is the input sanitized?


noah-paige (approved):
working as expected - like the view for non admins

PR #1665, opened 2024-10-25
Limit Response info dataset queries

noah-paige:
### Feature or Bugfix

- Refactor

### Detail
- Change Response Types for Dataset, Dashboard, and Shares Modules:
  - S3Datasets: `getDataset`, `getDatasetTables`, and `getDatasetStorageLocation` to only return Simplified Env / Org 
return types
  - ++ Redshift Datasets
  - ++ Dataset Shares

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/modules/dashboards/api/types.py
+++ b/backend/dataall/modules/dashboards/api/types.py
@@ -19,17 +19,13 @@ Dashboard = gql.ObjectType(
         gql.Field('DashboardId', type=gql.String),
         gql.Field('tags', type=gql.ArrayType(gql.String)),
         gql.Field('created', type=gql.String),
+        gql.Field('AwsAccountId', type=gql.String),
         gql.Field('updated', type=gql.String),
         gql.Field('owner', type=gql.String),
         gql.Field('SamlGroupName', type=gql.String),
         gql.Field(
-            'organization',
-            type=gql.Ref('Organization'),
-            resolver=get_dashboard_organization,
-        ),
-        gql.Field(
             'environment',
-            type=gql.Ref('Environment'),
+            type=gql.Ref('EnvironmentSimplified'),
             resolver=resolve_environment,
         ),
         gql.Field(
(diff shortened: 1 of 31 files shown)


noah-paige:
Testing Completed:

- [x] All Dataset Tabs Load += `getDataset` returns with correct information on FE
- [x] All Dataset Tabs Load += `getDatasetTables` returns with correct information on FE
- [x] All Dataset Tabs Load += `getDatasetStorageLocation` returns with correct information on FE
- [ ] All Dataset Tabs Load += `getRedshiftDataset` returns with correct information on FE
- [ ] All Dataset Tabs Load += `getRedshiftDatasetTable` returns with correct information on FE

SofiaSazonova:
Error in query `ListDatasets` (/console/datasets) "Cannot query field 'organization' on type 'DatasetBase'."

SofiaSazonova:
Is it necessary to change `Environment` => `SimplifiedEnvironment` also in queries?
```
ListDataPipelines
GetDataPipeline
ListOmicsRun
getSagemakerNotebook
ListSagemakerNotebooks
ListSagemakerStudioUsers
getSagemakerStudioUser
```

SofiaSazonova (changes requested):
An error and a question

SofiaSazonova:
Also, a comment: in dataset e.g. if we change name, label doesn't change (or it was a bug). Will it work the same for environments? If so, we risk to have confusing data displayed if use label instead of name

noah-paige:
> Error in query `ListDatasets` (/console/datasets) `"Cannot query field 'organization' on type 'DatasetBase'."`

Good catch resolved now

noah-paige:
> Also, a comment: in dataset e.g. if we change name, label doesn't change (or it was a bug). Will it work the same for environments? If so, we risk to have confusing data displayed if use label instead of name

This is outside of the scope of this PR yes? I actually think we should prevent updates to label for datasets or envs or any resources in dataall unless we are certain they have no impact on provisioned resources via CDK

SofiaSazonova (approved):

PR #1667, opened 2024-10-28
 Fix: Remove optional AllowWrites 2 - not supported in all regions 

dlpzx:
### Feature or Bugfix
- Bugfix

### Detail
Follow up of #1664 but for the associate_data_share API call. In addition this PR removes a test that is already covered by the previous test.

### Relates
- #1664 

### Security

  - Is the input sanitized?


tests_new/integration_tests/modules/shares/redshift_datasets_shares/test_redshift_shares.py
```diff
@@ -252,27 +253,6 @@ def test_create_redshift_share_invalid_target_connection(
     )
 
 
-def test_create_redshift_share_unauthorized_target_connection(
```
SofiaSazonova, line 255: Is this test about AllowWrites? 
I can't see the connection

SofiaSazonova (approved):

PR #1669, opened 2024-10-28
Tests/extend token validity

noah-paige:
### Feature or Bugfix

- Tests

### Detail
- Extend Token Validity if integration tests enabled to account for tests runtime
- Add Env Variable change to cognito URLs Custom Resource to force re-run and setting callback URLs
    - @petrkalos - let me know if you had come up with a better solution to this part of custom resource execution

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?


deploy/stacks/frontend_cognito_config.py
```diff
@@ -105,6 +105,7 @@ def __init__(
                 'envname': envname,
                 'deployment_region': backend_region,
                 'custom_domain': str(bool(custom_domain)),
+                'redeploy': 'true',
```
petrkalos, line 108: Unfortunately this will only work the first time you deploy it and then it will stop finding diffs and won't redeploy.
The env variables solutions is to pass something random to an env variable, i.e timestamp or uuid but this will always make CDK return a diff which isn't healthy.

petrkalos (approved):
Env variable was merged by me already so rebase and merge

PR #1674, opened 2024-10-30
add salt to FrontendCognitoConfig to make it always run

petrkalos:
### Feature or Bugfix
- Bugfix

### Detail
Problem:
We modify the Cognito UserPoolClient manually with CustomResources (not through CDK deployment) and as such everytime we make a change to the UserPoolClient CDK code (i.e enabling/disabling integtests, changing the timeout etc) the new deployment is overrriding the manual changes.
We could specify the callback URLs in CDK but we don't know the URLs until later in the pipeline when the CDN is being deployed. Additionally AuthAtEdge requires Cognito to be already setup and as such we have a chicken and egg situation.

Solution:
Adding a random salt (timestamp) as an environment variable to force triggering the CustomResource on every deployment.
The downside of this solution is that the particular stack will always create a diff but due to the structure of the stacks it is very complicated to work around the issue.

### Security

  - Is the input sanitized?

### DIFF
--- a/deploy/stacks/frontend_cognito_config.py
+++ b/deploy/stacks/frontend_cognito_config.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from aws_cdk import (
     aws_lambda as _lambda,
     aws_iam as iam,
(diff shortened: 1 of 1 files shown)


noah-paige (approved):
Testing Cases Covered:

- [x] Pipeline Deployed Twice --> Triggered 2 Executions of Lambda Custom Resource
- [x] Lambda Env Var Updated with Timestamp (for each pipeline execution)
- [x] Cognito URLs configured Correctly for App Client (for each pipeline execution)


Fix works as expected

PR #1678, opened 2024-10-31
Lambda Event Logs Handling

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Ensure any potential sensitive information redacted from logged events

### Relates
- 

### Security

  - Is the input sanitized?

### DIFF
--- a/backend/dataall/base/utils/api_handler_utils.py
+++ b/backend/dataall/base/utils/api_handler_utils.py
@@ -26,6 +26,14 @@ ENGINE = get_engine(envname=ENVNAME)
 ALLOWED_ORIGINS = os.getenv('ALLOWED_ORIGINS', '*')
 
 
+def redact_creds(event):
+    if 'headers' in event and 'Authorization' in event['headers']:
+        event['headers']['Authorization'] = 'XXXXXXXXXXXX'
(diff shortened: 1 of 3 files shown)


noah-paige:
Testing Covered:

- [x] Deployed to AWS
- [x] CloudWatch Logs redacting any credentials or sensitive information

backend/dataall/base/utils/api_handler_utils.py
```diff
@@ -26,6 +26,14 @@
 ALLOWED_ORIGINS = os.getenv('ALLOWED_ORIGINS', '*')
 
 
+def redact_creds(event):
+    if 'headers' in event and 'Authorization' in event['headers']:
```
petrkalos, line 30: nit: probably something like `if event.get('headers', {}).get('Authorization') ...` is cleaner

petrkalos (approved):
The usual way to deal with this kind of situation is to use lower log levels (i.e debug) for something that might contain sensitive data and higher for those that they are not (i.e info). Then you configure the non-prod stage level to `debug` and prod to `info`.
That gives you the benefit of still seeing all the information in non-prod envs which might help with debugging but also hide them from prod which might reveal critical information.

I am approving now to move faster but I think we should revisit this in the future.

PR #1682, opened 2024-11-04
Added Token Validations

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- This PR does the following w.r.t Cognito IdP and Auth Flow
    - Changes Authorizer from built-in Cognito Authorizer to Custom Authorizer to validate token signature, issuer, and expiry time, etc.
    - Adds aditional step to execute GET API on Cognito's `/oauth/userInfo/` endpoint to ensure access Token validity 

Allows data.all API request to execute if the above criteria are met

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?


noah-paige:
Testing Completed:

- [ ] CodePipeline Deploys Successfully (for Cognito Auth with already existing data.all instance - i.e. UPDATE)
- [ ] Login + Store Access Token
- [ ] SignOut (clear tokens)
- [ ] ReAuth Workflow Successfuly 


- [ ] Post SignOut Run data.all API with same Tokens -- returns _______
- [ ] User removed from Cognito IdP --> User no longer can perform actions on data.all UI immediately

deploy/stacks/lambda_api.py
```diff
@@ -223,23 +228,25 @@ def __init__(
...
             )
+        )
+        ## GET CONGITO USERL POOL ID and APP CLIENT ID
 
-            if not os.path.isdir(custom_authorizer_assets):
-                raise Exception(f'Custom Authorizer Folder not found at {custom_authorizer_assets}')
+        if not os.path.isdir(custom_authorizer_assets):
+            raise Exception(f'Custom Authorizer Folder not found at {custom_authorizer_assets}')
 
+        if custom_auth:
```
petrkalos, line 245: Just an idea for the future:
What makes cognito a special authorizer? Perhaps we should treat it as a generic OAuth2.0 provider and avoid duplicating logic. For convenience customers can chose to enable cognito and then we can automatically populate the env variables, otherwise they have to provide them,

deploy/stacks/lambda_api.py
```diff
@@ -248,48 +255,60 @@ def __init__(
 
             for claims_map in custom_auth.get('claims_mapping', {}):
                 custom_lambda_env[claims_map] = custom_auth.get('claims_mapping', '').get(claims_map, '')
+        else:
+            custom_lambda_env = {
+                'envname': envname,
+                'LOG_LEVEL': 'DEBUG',
+                'user_info_url': f'https://{user_pool_domain.domain_name}.auth.{self.region}.amazoncognito.com/oauth2/userInfo',
+                'custom_auth_provider': 'Cognito',
+                'custom_auth_url': f'https://cognito-idp.{self.region}.amazonaws.com/{user_pool.user_pool_id}',
```
petrkalos, line 264: instead of passing the user_info_url explicitly we can use `https://{custom_auth_url}/.well-known/openid-configuration` and read the userinfo_endpoint field.

noah-paige: using `https://{custom_auth_url}/.well-known/openid-configuration` for userinfo URL now

deploy/stacks/lambda_api.py
noah-paige: agreed we should just document it as a generic OAuth2.0 irrespective of whatever IdP

backend/dataall/base/utils/api_handler_utils.py
```diff
@@ -24,13 +24,22 @@
 ]
 ENGINE = get_engine(envname=ENVNAME)
 ALLOWED_ORIGINS = os.getenv('ALLOWED_ORIGINS', '*')
+AWS_REGION = os.getenv('AWS_REGION')
 
 
 def redact_creds(event):
```
petrkalos, line 30: nit: That's why I don't like this approach, you have to remember to obfuscate stuff

deploy/stacks/lambda_api.py
petrkalos: should we remove the `custom_auth_jwks_url` since you are pulling this from `openid-configuration` endpoint?

deploy/custom_resources/custom_authorizer/custom_authorizer_lambda.py
```diff
@@ -19,18 +20,25 @@
 
 def lambda_handler(incoming_event, context):
     # Get the Token which is sent in the Authorization Header
+    logger.debug(incoming_event)
```
petrkalos, line 23: this will print the JWT token, are we using proper log-level to protect prod deployments?

deploy/custom_resources/custom_authorizer/jwt_services.py
```diff
@@ -53,49 +20,55 @@ def fetch_public_keys():
...
-    'leeway': 0,
+    'verify_at_hash': True,
+    'require': ['aud', 'iat', 'exp', 'iss', 'sub', 'jti'],  # "nbf", "at_hash"
 }
 
 
 class JWTServices:
+    @staticmethod
+    def _fetch_openid_url(key):
+        response = requests.get(f'{os.environ.get("custom_auth_url")}/.well-known/openid-configuration')
```
petrkalos, line 31: I'd use `os.path,join` one extra `/` in the `custom_auth_url` env and this will break.

frontend/src/authentication/hooks/useToken.js
```diff
@@ -21,14 +23,18 @@ export const useToken = () => {
               await auth.signinSilent();
             }
             const t = auth.user.id_token;
+            const at = auth.user.access_token;
```
TejasRGitHub, line 26: Are we now using both Access and Identity token ?

TejasRGitHub: Do we need any particular token for any specific use ? Like a specific gql endpoint needs an identity token and cannot work with access token ?

deploy/stacks/lambda_api.py
```diff
@@ -50,6 +50,8 @@ def __init__(
         apig_vpce=None,
         prod_sizing=False,
         user_pool=None,
+        user_pool_client=None,
+        user_pool_domain=None,
```
dlpzx, line 54: unused parameter `user_pool_domain`

deploy/custom_resources/custom_authorizer/jwt_services.py
```diff
@@ -53,49 +20,55 @@ def fetch_public_keys():
...
-    'require_sub': True,
-    'require_jti': True,
-    'require_at_hash': False,
-    'leeway': 0,
+    'verify_at_hash': True,
+    'require': ['aud', 'iat', 'exp', 'iss', 'sub', 'jti'],  # "nbf", "at_hash"
 }
 
 
 class JWTServices:
```
petrkalos, line 28: I would convert the static functions to member methods and make this class instantiable.
Then in `__init__`... 
* I would `_fetch_openid_url` and store it in an attribute
* I would instantiate the PyJWKClient (which I assume is fetching the public key)

Then in the lambda handler I would instantiate the JWTServices() as a global variable to allow me to reuse it (which will avoid 2 network requests per invocation when the lambda is hot)

deploy/custom_resources/custom_authorizer/jwt_services.py
```diff
@@ -53,49 +20,55 @@ def fetch_public_keys():
...
-        except JWTError as e:
+        except jwt.exceptions.PyJWTError as e:
             logger.error(f'JWT token validation failed: {str(e)}')
             return None
         except Exception as e:
             logger.error(f'Failed to validate token - {str(e)}')
             return None
+
+    @staticmethod
+    def validate_access_token(access_token):
```
petrkalos, line 69: Can we put a [TTLCache](https://cachetools.readthedocs.io/en/latest/#cachetools.TTLCache) decorator around it with a configurable `ttl`?

deploy/custom_resources/custom_authorizer/custom_authorizer_lambda.py
noah-paige: good call out - adding log_level setting on custom Authorizer Function in  `/deploy/stacks/lambda_api.py`

backend/dataall/base/utils/api_handler_utils.py
```diff
@@ -24,13 +24,22 @@
...
+
+    if event.get('multiValueHeaders', {}).get('Authorization'):
         event['multiValueHeaders']['Authorization'] = 'XXXXXXXXXXXX'
+
+    if event.get('multiValueHeaders', {}).get('accesskeyid'):
+        event['multiValueHeaders']['accesskeyid'] = 'XXXXXXXXXXXX'
+
+    if event.get('headers', {}).get('accesskeyid'):
+        event['headers']['accesskeyid'] = 'XXXXXXXXXXXX'
+
```
dlpzx, line 42: in `deploy/custom_resources/custom_authorizer/custom_authorizer_lambda.py` the only values that we actually use are coming from the headers. Is it possible that the API succeeds with multiVValueHeaders? and if not, maybe we don't need these checks here. (It might be out of scope for this PR)
```
auth_token = incoming_event['headers']['Authorization']
access_token = incoming_event['headers']['accesskeyid']
```

deploy/stacks/lambda_api.py
noah-paige: removed in latest

deploy/stacks/backend_stack.py
```diff
@@ -35,6 +35,7 @@ def __init__(
         id,
         envname: str = 'dev',
         resource_prefix='dataall',
+        tooling_region=None,
```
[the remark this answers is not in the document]
noah-paige: removed!

deploy/stacks/lambda_api.py
noah-paige: removed in latest

deploy/custom_resources/custom_authorizer/jwt_services.py
noah-paige: removed

deploy/custom_resources/custom_authorizer/custom_authorizer_lambda.py
```diff
@@ -19,18 +20,25 @@
...
-    logger.debug(verified_claims)
     if not verified_claims:
         raise Exception('Unauthorized. Token is not valid')
+    logger.debug(verified_claims)
 
+    # Generate Allow Policy w/ Context
     effect = 'Allow'
     policy = AuthServices.generate_policy(verified_claims, effect, incoming_event['methodArn'])
-    logger.debug('Generated policy is ', policy)
+    logger.debug('Generated policy is ', json.dumps(policy))
```
TejasRGitHub, line 41: logger.debug(f'Generated policy is json.dumps(policy)').

using f-string,. I think this `logger.debug('Generated policy is ', json.dumps(policy))` doesn't work

deploy/custom_resources/custom_authorizer/jwt_services.py
```diff
@@ -53,49 +20,55 @@ def fetch_public_keys():
...
-                audience=public_key.get('audience'),
+                issuer=os.environ.get('custom_auth_url'),
+                audience=os.environ.get('custom_auth_client'),
+                leeway=0,
                 options=jwt_options,
             )
-
             return payload
-        except ExpiredSignatureError:
+        except jwt.exceptions.ExpiredSignatureError:
```
dlpzx, line 60: nit: I would directly raise these Exceptions here instead of passing the None and then raising the general exception

deploy/custom_resources/custom_authorizer/jwt_services.py
noah-paige: for TTL Cache - please note custom authorizer already ahs a results_cache_ttl (set to 1 minute for now) at `deploy/stacks/lambda_api.py`:

```        
custom_authorizer = apigw.RequestAuthorizer(
            self,
            'CustomAuthorizer',
            handler=self.authorizer_fn,
            identity_sources=[apigw.IdentitySource.header('Authorization')],
            authorizer_name=f'{resource_prefix}-{envname}-custom-authorizer',
            assume_role=custom_authorizer_role,
            results_cache_ttl=Duration.minutes(1),
        )
```

Would adding an additional caching mechanism at the above be useful?

deploy/stacks/lambda_api.py
```diff
@@ -248,48 +251,66 @@ def __init__(
...
+                'custom_auth_jwks_url': f'https://cognito-idp.{self.region}.amazonaws.com/{user_pool.user_pool_id}/.well-known/jwks.json',
+                'email': 'email',
+                'user_id': 'email',
+            }
+
+        layer = _lambda.LayerVersion.from_layer_version_arn(
+            self,
+            'CryptographyLayerPy39',
+            f'arn:aws:lambda:{self.region}:770693421928:layer:Klayers-p39-cryptography:19',
+        )
```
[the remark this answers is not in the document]
noah-paige: Adding to `requirements.txt` led to custom authorizer Lambda raising a `Runtime Error ... GLIBC_2.26' not found`. Seemed to be due to the compiled compiled binary of the cryptography library is incompatible with the runtime environment of AWS Lambda
- cryptography package is compiled on a system with a newer version of glibc than what's available in the Lambda environment
- AWS Lambda with Python 3.9 uses Amazon Linux 2, which as of April 2024, uses an older version of glibc.

I can look into instead using `cdk-klayers` I think that should suffice if good with you?

deploy/custom_resources/custom_authorizer/jwt_services.py
```diff
@@ -1,46 +1,13 @@
...
-from jose import jwk
-from jose.jwt import get_unverified_header, decode, ExpiredSignatureError, JWTError
+import jwt
+
 import logging
 
 logger = logging.getLogger()
 logger.setLevel(os.environ.get('LOG_LEVEL', 'INFO'))
 
-# Configs required to fetch public keys from JWKS
```
TejasRGitHub, line 11: We don't need this because pyJWT does it with some inbuilt method ?

deploy/stacks/lambda_api.py
```diff
@@ -223,73 +224,94 @@ def __init__(
...
+        if not os.path.isdir(custom_authorizer_assets):
+            raise Exception(f'Custom Authorizer Folder not found at {custom_authorizer_assets}')
 
-            custom_lambda_env = {
-                'envname': envname,
-                'LOG_LEVEL': 'DEBUG',
-                'custom_auth_provider': custom_auth.get('provider'),
-                'custom_auth_url': custom_auth.get('url'),
-                'custom_auth_client': custom_auth.get('client_id'),
-                'custom_auth_jwks_url': custom_auth.get('jwks_url'),
```
TejasRGitHub, line 246: Are we removing the `custom_auth_jwks_url` URL completely ? and using the `.well-known/openid-configuration` instead ?

frontend/src/authentication/hooks/useToken.js
```diff
@@ -45,5 +51,5 @@ export const useToken = () => {
       );
```
TejasRGitHub, line 51: In the if (!token ) should it be instead if ( !token and !accessToken ) ?

backend/dataall/base/utils/api_handler_utils.py
```diff
@@ -24,13 +24,22 @@
 ]
 ENGINE = get_engine(envname=ENVNAME)
 ALLOWED_ORIGINS = os.getenv('ALLOWED_ORIGINS', '*')
+AWS_REGION = os.getenv('AWS_REGION')
 
 
 def redact_creds(event):
-    if 'headers' in event and 'Authorization' in event['headers']:
+    if event.get('headers', {}).get('Authorization'):
```
TejasRGitHub, line 31: Can anyone please tell why are we redacting these ?

frontend/src/authentication/hooks/useToken.js
noah-paige: debating whether we should just send access token and forget about ID

currently we send both which may be overkill

ID token contains headers like `email` that we do need. However when we run a request with access token against `userinfo` endpoint we get that information (such as `email`) in response that we can add to context


Additionally, access tokens (for cognito at least) contain user and group information - not sure if that is standard

noah-paige: At a high level the token uses are:
-  the access token is needed to invoke the `/userinfo` endpoint
- the verified claims of the ID Token are needed to populate context information (email, user, groups) for downstream lambda (this theoretically could be sourced elsewhere)

deploy/custom_resources/custom_authorizer/custom_authorizer_lambda.py
noah-paige: yup I made this change earlier and just ran into this - updating now

noah-paige: `logger.debug('Generated policy is ', policy)` was also not showing policy so opting for f string approach

deploy/custom_resources/custom_authorizer/jwt_services.py
noah-paige: yes - https://pyjwt.readthedocs.io/en/latest/usage.html (re: `PyJWKClient()`)

deploy/stacks/lambda_api.py
noah-paige: correct

backend/dataall/base/utils/api_handler_utils.py
noah-paige: user creds information does not need to be logged and is no longer relevant for the remaining request lifecycle - opting to redact that info

TejasRGitHub: Instead of redacting , can we just extract useful information and then pass it onto our lambda ?

deploy/custom_resources/custom_authorizer/custom_authorizer_lambda.py
```diff
@@ -16,21 +17,32 @@
...
+        raise Exception('Unauthorized. Missing identity or access JWT')
 
-    verified_claims = JWTServices.validate_jwt_token(auth_token)
-    logger.debug(verified_claims)
+    # Validate User is Active with Proper Access Token
+    user_info = jwt_service.validate_access_token(auth_token)
+
+    # Validate JWT
+    verified_claims = jwt_service.validate_jwt_token(auth_token[7:])
     if not verified_claims:
```
TejasRGitHub, line 36: Since the method `validate_jwt_token` is now raising exception this if block can be removed

frontend/src/authentication/hooks/useToken.js
noah-paige: updated only using access token now

petrkalos (changes requested):

tests_new/integration_tests/client.py
petrkalos, line 1: @TejasRGitHub, take a look at this change as I think it will enable support for any IdP

deploy/custom_resources/custom_authorizer/custom_authorizer_lambda.py
```diff
@@ -16,21 +17,33 @@
 Custom Lambda Authorizer is attached to the API Gateway. Check the deploy/stacks/lambda_api.py for more details on deployment
 """
 
+OPENID_CONFIG_PATH = os.path.join(os.environ.get('custom_auth_url', ''), '.well-known', 'openid-configuration')
+jwt_service = JWTServices(OPENID_CONFIG_PATH)
```
petrkalos, line 21: nit: I would capitalize global variables

tests_new/integration_tests/client.py
```diff
@@ -27,24 +30,58 @@ def __init__(self, username, password):
...
+    def _get_jwt_tokens(self):
+        token = uuid.uuid4()
+        scope = 'aws.cognito.signin.user.admin openid'
+
+        idp_domain_url = os.environ['IDP_DOMAIN_URL']
+
+        token_url = os.path.join(idp_domain_url, 'oauth2', 'token')
+        login_url = os.path.join(idp_domain_url, 'login')
+
+        client_id = os.environ['COGNITO_CLIENT']
```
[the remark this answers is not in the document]
noah-paige: there are still some particularities that makes integration tests only work with cognito. I am inclined to make it fully generic in a separate PR out of scope from here and leave it as it was inherited in this PR for now if that is okay

tests_new/integration_tests/client.py
```diff
@@ -27,24 +30,58 @@ def __init__(self, username, password):
...
-            'AuthFlow': 'USER_PASSWORD_AUTH',
-            'AuthParameters': {
-                'USERNAME': self.username,
-                'PASSWORD': self.password,
-            },
+    def _get_jwt_tokens(self):
+        token = uuid.uuid4()
+        scope = 'aws.cognito.signin.user.admin openid'
+
+        idp_domain_url = os.environ['IDP_DOMAIN_URL']
```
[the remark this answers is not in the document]
noah-paige: believe I already do that in the `deploy/stacks/pipeline.py` changes in this PR - will tag you on those lines for ref

deploy/custom_resources/custom_authorizer/custom_authorizer_lambda.py
```diff
@@ -16,21 +17,33 @@
 Custom Lambda Authorizer is attached to the API Gateway. Check the deploy/stacks/lambda_api.py for more details on deployment
 """
 
+OPENID_CONFIG_PATH = os.path.join(os.environ.get('custom_auth_url', ''), '.well-known', 'openid-configuration')
```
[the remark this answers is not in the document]
noah-paige: sure that is a good call out - just made that change to `os.environ['custom_auth_url']`

petrkalos (approved):

deploy/stacks/lambda_api.py
```diff
@@ -228,73 +230,95 @@ def __init__(
...
+                    'custom_auth_url': f'https://cognito-idp.{self.region}.amazonaws.com/{user_pool.user_pool_id}',
+                    'custom_auth_client': user_pool_client.user_pool_client_id,
+                    'email': 'email',
+                    'user_id': 'email',
+                }
+            )
+
+        # Initialize Klayers
+        runtime = _lambda.Runtime.PYTHON_3_9
+        klayers = Klayers(self, python_version=runtime, region=self.region)
```
dlpzx, line 275: nice!

tests_new/integration_tests/client.py
TejasRGitHub: Thanks @petrkalos for pointing at this. Will take a look

PR #1693, opened 2024-11-12
Fix/input validation

noah-paige:
### Feature or Bugfix

- Bugfix

### Detail
- Allow `:` and `.` chars

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?


dlpzx (approved):
Fixes the integration tests

dlpzx:
Although this PR fixes the current bug of tags with `:` and `.`, this issue might arise with other tags that contain especial characters as we are not validating the tag names or cleaning them when they are created. We only sanitize the search input but not the actual value upon creation. I think that ideally we would apply the naming standard to both processes

noah-paige:
> Although this PR fixes the current bug of tags with `:` and `.`, this issue might arise with other tags that contain especial characters as we are not validating the tag names or cleaning them when they are created. We only sanitize the search input but not the actual value upon creation. I think that ideally we would apply the naming standard to both processes

Will merge this PR as is to get integration tests working - let's have a separate PR to enforce naming convention of tags

PR #1694, opened 2024-11-13
tenant-permission tests 

dlpzx:
### Feature or Bugfix
- Feature

### Detail
Add unit tests that verify that MANAGE_X permissions are applied to all Mutations except for an OPT_OUT list of Mutations and to a subset of OPT_IN queries. 

The OPT_OUT mutations are either:
- admin actions that can only be performed by the tenants. Applying permissions in this case does not make sense.
- platform "support" features such as feed, notification, votes. No object needs to be protected in this case. 
- v2.7.0 features which will be addressed in a separate PR

The OPT_IN queries are operations that retrieve credentials or redirect URLs that allow the user to effectively create/update data.all objects.

### Security

  - Is the input sanitized?


tests/test_tenant_unauthorized.py
```diff
@@ -0,0 +1,37 @@
...
+from dataall.base.db.exceptions import TenantUnauthorized
+
+from pytest import fixture
+
+load_modules(modes={ImportMode.API})
+
+CHECK_PERMS = ['archiveOrganization']
+
+
+@fixture(scope='function')
```
petrkalos, line 17: nit: an alternative and maybe more flexible version of this doesn't require a fixture...

```python
@pytest.mark.parametrize('field_name,resolver', ...)
def test_generic(field_name, resolver, db, client, org1, groupNoTenantPermissions, userNoTenantPermissions):
    local = MagicMock()
    local.context = RequestContext(db, userNoTenantPermissions.username, [groupNoTenantPermissions.groupUri], userNoTenantPermissions)
    with patch('dataall.base.context._request_storage', local):
        # execute test command
```

tests/test_tenant_unauthorized.py
```diff
@@ -0,0 +1,37 @@
...
+    mocker.patch('dataall.base.context._request_storage', mock)
+
+
+@pytest.mark.parametrize(
+    'name,field_resolver',
+    [
+        (f'{_type.name}.{field.name}', field.resolver)
+        for _type in bootstrap().types
+        for field in _type.fields
+        if field.resolver and _type.name in ['Mutation', 'Query'] and field.name in CHECK_PERMS
```
petrkalos, line 33: I think we should inverse the for-loop and make pytest fail in case any of the listed CHECK_PERMS queries are missing. i.e...

```python
ALL_RESOLVERS = {
    f'{_type.name}.{field.name}': field.resolver
    for _type in bootstrap().types
    for field in _type.fields
    if field.resolver
}

CHECK_PERMS = [
    'Mutation.archiveOrganization',
    'Mutation.createOrganization',
]


@pytest.mark.parametrize('field_name,resolver', [(perm, ALL_RESOLVERS[perm]) for perm in CHECK_PERMS])
```

dlpzx: I implemented something similar in https://github.com/data-dot-all/dataall/pull/1694/commits/75d30c148250894bf791db2bef905c347ed8a3b0 adding an assertion to check that the resolver does not exist but continue for other tests

tests/test_tenant_unauthorized.py
dlpzx: Done in https://github.com/data-dot-all/dataall/pull/1694/commits/75d30c148250894bf791db2bef905c347ed8a3b0

dlpzx:
@petrkalos I have intentionally left the commented mutations and opened the PR to review. I think that the commented mutations will require some changes and I don't want to make the PR too big. If you think we should tackle all at once I am happy to keep working on this PR

tests/test_tenant_unauthorized.py
```diff
@@ -0,0 +1,155 @@
...
+import inspect
+
+
+load_modules(modes={ImportMode.API})
+
+NO_CHECK_PERMS = [
+    'Mutation.updateGroupTenantPermissions',
+    'Mutation.updateSSMParameter',
+    'Mutation.createNetwork',
+    'Mutation.deleteNetwork',
```
noah-paige, line 17: is this to say Operations like `createNetwork` and `deleteNetwork` do not have `MANAGE_XXXX` checks on them? They do have `MANAGE_ENVIRONMENTS` ... and I think a handful of tohers on this list are simialr 

Or are these mutations that would need some additional work on the backend app logic side to conform to this pattern?

backend/dataall/modules/notebooks/services/notebook_service.py
```diff
@@ -42,10 +42,10 @@
...
     """A request dataclass for notebook creation. Adds default values for missed parameters"""
 
-    label: str
-    VpcId: str
-    SubnetId: str
-    SamlAdminGroupName: str
+    label: str = None
+    VpcId: str = None
+    SubnetId: str = None
+    SamlAdminGroupName: str = None
```
noah-paige, line 48: why these changes to default `None`?

dlpzx: Yes, let me revert that and add these exception to the edge cases of PR part2

tests/test_tenant_unauthorized.py
```diff
@@ -0,0 +1,115 @@
...
+
+ALL_MUTATIONS = {
+    f'{_type.name}.{field.name}': field.resolver
+    for _type in bootstrap().types
+    for field in _type.fields
+    if field.resolver and _type.name == 'Mutation'
+}
+
+
+@pytest.mark.parametrize(
```
petrkalos, line 80: I think we should have one global variable let's say ALL_RESOLVERS and then filter accordingly on each test and if they don't pass the filter we raise a `pytest.skip(...)`

The reason I suggest that is because
1. We will be able to reuse this variable in other tests that require resolvers
2. It becomes obvious that we run those tests against ALL resolvers but we explicitly skip some

Wdyt?

petrkalos (approved):

noah-paige (approved):

PR #1711, opened 2024-11-21
add resource permission checks

petrkalos:
### Feature or Bugfix
Feature

### Detail
* introducing a test that is going through all the nested SQL queries and assert that `ResourcePolicyService.check_user_resource_permission` have been called with the expected permission name OR explicitly ignore the test.
  *  New subqueries will be tested automatically and fail if the expected permission is missing
  *  Removed queries will make the test suite fail to avoid keeping stale permissions
* UI: Make handling responses (i.e ListDatasets, GetDataset) more tolerant to missing information (i.e missing Stack) by doing conditional rendering. 
  Example usecase: A dataset is being shared by a user but only owners have permissions to see stack and environment info.
* Override config.json and enable all modules when running the tests. As a result checkov now synths the pipeline module that throws some errors (added in the baseline). @noah-paige 
* Make TestClient more tolerant to GQLErrors previously it would always throw if errors, now it will throw if there are only erros (and no data) allowing for partial information to be returned to the caller

### Security

  - Is the input sanitized?


tests_new/integration_tests/client.py
```diff
@@ -28,15 +32,18 @@ def __init__(self, username, password):
...
-    def query(self, query: str):
+    def query(self, query: dict):
         graphql_endpoint = os.path.join(os.environ['API_ENDPOINT'], 'graphql', 'api')
         headers = {'accesskeyid': 'none', 'SecretKey': 'none', 'Authorization': f'Bearer {self.access_token}'}
         r = requests.post(graphql_endpoint, json=query, headers=headers)
-        if errors := r.json().get('errors'):
-            raise GqlError(errors)
+        response = r.json()
+        if errors := response.get('errors'):
+            if any((response.get('data', {}) or {}).values()):  # check if there are data
```
dlpzx, line 41: Why do we need to check for `.values()`? wouldn't it be enough with `if any((response.get('data', {}):`

tests/test_tenant_unauthorized.py
```diff
@@ -1,49 +1,436 @@
...
+    SUPPORT = 'permissions do not apply to support notifications'
+    FEED = 'permissions do not apply to support feed messages'
+    VOTES = 'permissions do not apply to support votes'
+    BACKPORT = 'outside of this PR to be able to backport to v2.6.2'
+    INTRAMODULE = 'returns intra-module data'
+    PERMCHECK = 'checks user permissions for a particular feature'
+    CATALOG = 'catalog resources are public by design'
+    SIMPLIFIED = 'simplified response'
+    NOTREQUIRED = 'permission check is not required'
+
```
dlpzx, line 48: I really like that we specify IgnoreReasons! But I think we need to add some and maybe we can merge some together:

- `ADMIN` - all good! I would maybe rename it as `TENANT`
- `APPSUPPORT - permissions do not apply to application support features` - merging SUPPORT + FEED+ VOTES
- `BACKPORT`, `INTRAMODULE`
- `PERMCHECK` ---> it is very generic, we just want to account for the USERROLEINRESOURCE checks right?
- `PUBLIC` - Public by design  - It will be used in other places not only for Catalog
- ....

tests/test_tenant_unauthorized.py
```diff
@@ -62,27 +449,77 @@ def mock_input_validation(mocker):
...
-    if _type.name == 'Query' and f'{_type.name}.{field.name}' not in OPT_IN_QUERIES:
-        pytest.skip(f'Skipping test for {field.name}: This Query does not require a tenant permission check.')
+    fid = request.node.callspec.id
+    tdata = EXPECTED_RESOLVERS[fid]
+    msg = f'{fid} -> {field.resolver.__code__.co_filename}:{field.resolver.__code__.co_firstlineno}'
+    expected_perm = tdata.tenant_perm
+    if not expected_perm:
+        pytest.skip(msg + f' Reason: {tdata.tenant_ignore.value}')
+    logging.info(msg)
+
```
dlpzx, line 476: ultra nit: move msg creation inside if-clause where it is used

tests/test_tenant_unauthorized.py
dlpzx: TestData should define either permissions or the ignore reason, right now we are falling back to default NOTREQUIRED. Could we instead have some logic that throws an error for not-correctly defined TestData classes?

```
@dataclass
class TestData:
    tenant_perm: Optional[str] = None
    tenant_ignore: Optional['IgnoreReason'] = None
    resource_perm: Optional[str] = None 
    resource_ignore: Optional['IgnoreReason'] = None

    def __post_init__(self):
        # Validate that at least one of resource_perm or resource_ignore is set
        if self.resource_perm is None and self.resource_ignore is None:
            raise ValueError("Either resource_perm or resource_ignore must be set")
        if self.tenant_perm is None and self.tenant_ignore is None:
            raise ValueError("Either tenant_perm or tenant_ignore must be set")


```

tests/test_tenant_unauthorized.py
```diff
@@ -62,27 +449,77 @@ def mock_input_validation(mocker):
...
+    expected_perm = tdata.resource_perm
+    if not expected_perm:
+        pytest.skip(msg + f' Reason: {tdata.tenant_ignore.value}')
+    logging.info(msg)
+
+    assert_that(field.resolver).is_not_none()
+    username = 'ausername'
+    groups = ['agroup']
+    mock_local.context = RequestContext(MagicMock(), username, groups, 'auserid')
+    mock_local.context.db_engine.scoped_session().__enter__().query().filter().all.return_value = [MagicMock()]
```
dlpzx, line 514: why do we need to mock the db session queries, alls and filters?

tests/test_tenant_unauthorized.py
```diff
@@ -62,27 +449,77 @@ def mock_input_validation(mocker):
...
+    mock_check.side_effect = ResourceUnauthorized(groups, 'test_action', 'test_uri')
+    iargs = {arg: MagicMock() for arg in inspect.signature(field.resolver).parameters.keys()}
+    with suppress(ResourceUnauthorized):
+        field.resolver(**iargs)
```
dlpzx, line 518: Just to confirm my understanding. Here we are ignoring the exception because it is an artificial exception created by the mocker side effect in line 515. With this test we are assuming that the `ResourcePolicyService` returns `ResourceUnauthorized`, while for the tenantUnauthorized tests we explicitly checked that the resolver raised the expected exception

tests_new/integration_tests/client.py
petrkalos: It won't be enough, a GQL response looks like this so we have to check if subkeys (query1,query2,...) are non empty
```json
{
"data" : {
  "query1": {"key1": "data..."}
  "query2": None
}
```

tests/test_tenant_unauthorized.py
petrkalos: Correct, the important part of both tests is that they call the corresponding check function with the specific user/group and permission, wether it throws or not is irrelevant. I also added this logic in the tenant test (check `spy_check`)
The reason I use the side_effect here is because I want to stop the execution as soon as we receive a call to the check to stop it from calling any of the business logic code which will produce a lot of errors etc.

frontend/src/modules/Folders/views/FolderView.js
```diff
@@ -233,11 +233,11 @@ const FolderView = () => {
   const fetchItem = useCallback(async () => {
     setLoading(true);
     const response = await client.query(getDatasetStorageLocation(params.uri));
-    if (!response.errors && response.data.getDatasetStorageLocation !== null) {
+    if (response.data.getDatasetStorageLocation !== null) {
       setFolder(response.data.getDatasetStorageLocation);
       setIsAdmin(
         ['Creator', 'Admin', 'Owner'].indexOf(
-          response.data.getDatasetStorageLocation.dataset.userRoleForDataset
+          response.data.getDatasetStorageLocation.dataset?.userRoleForDataset
```
dlpzx, line 240: Should we dispatch all errors when the data is null?  or can we assume that the first error will correspond to the top-level query (I think so right? graphql will execute first the top resolver and then the field-resolvers)
- here in FolderView we are just dispatching the first error
- in DatasetListView we dispatch all errors with the forEach loop

petrkalos: Good catch, my intention was to dispatch all everywhere, will change it

backend/dataall/modules/s3_datasets/api/storage_location/resolvers.py
```diff
@@ -47,6 +50,13 @@ def resolve_dataset(context, source: DatasetStorageLocation, **kwargs):
     if not source:
         return None
     with context.engine.scoped_session() as session:
+        ResourcePolicyService.check_user_resource_permission(
+            session=session,
+            username=get_context().username,
+            groups=get_context().groups,
+            resource_uri=source.datasetUri,
+            permission_name=GET_DATASET,
+        )
```
dlpzx, line 59: As an alternative we could create a find_dataset method in DatasetService similar to `EnvironmentService.find_environment_by_uri(uri)` and that way avoid application logic and checks happening in the resolver

backend/dataall/core/organizations/services/organization_service.py
```diff
@@ -305,6 +305,13 @@ def resolve_organization_by_env(uri):
         context = get_context()
         with context.db_engine.scoped_session() as session:
             env = EnvironmentRepository.get_environment_by_uri(session, uri)
+            ResourcePolicyService.check_user_resource_permission(
```
dlpzx, line 308: We could just call `dataall.core.organizations.services.organization_service.OrganizationService.get_organization`

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+    ),
+    field_id('MetadataFormSearchResult', 'hasTenantPermissions'): TestData(
+        resource_ignore=IgnoreReason.USERROLEINRESOURCE, tenant_ignore=IgnoreReason.TENANT, tenant_admin_perm=True
+    ),
+    field_id('Mutation', 'DisableDataSubscriptions'): TestData(
+        tenant_perm=MANAGE_ENVIRONMENTS, resource_perm=ENABLE_ENVIRONMENT_SUBSCRIPTIONS
+    ),
+    field_id('Mutation', 'addConnectionGroupPermission'): TestData(
+        tenant_ignore=IgnoreReason.BACKPORT, resource_ignore=IgnoreReason.NOTREQUIRED
+    ),
```
dlpzx, line 372: In resource_ignore, can you change it to BACKPORT? Permissions are required but I do not want to pollute `main` with other PRs

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+        tenant_perm=MANAGE_METADATA_FORMS, resource_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('Mutation', 'createNetwork'): TestData(tenant_perm=MANAGE_ENVIRONMENTS, resource_perm=CREATE_NETWORK),
+    field_id('Mutation', 'createOmicsRun'): TestData(tenant_perm=MANAGE_OMICS_RUNS, resource_perm=CREATE_OMICS_RUN),
+    field_id('Mutation', 'createOrganization'): TestData(
+        tenant_perm=MANAGE_ORGANIZATIONS, resource_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('Mutation', 'createQuicksightDataSourceSet'): TestData(
+        tenant_ignore=IgnoreReason.TENANT, resource_ignore=IgnoreReason.NOTREQUIRED
+    ),
```
dlpzx, line 435: `'createQuicksightDataSourceSet'` is using `is_tenant_admin` in an indirect way. Can we add the `tenant_admin_perm` flag?

tests/test_permissions.py
dlpzx: Same for createRedshiftConnection

dlpzx: Same for deleteConnectionGroupPermission

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+    field_id('Mutation', 'deleteDataset'): TestData(tenant_perm=MANAGE_DATASETS, resource_perm=DELETE_DATASET),
+    field_id('Mutation', 'deleteDatasetStorageLocation'): TestData(
+        tenant_perm=MANAGE_DATASETS, resource_perm=DELETE_DATASET_FOLDER
+    ),
+    field_id('Mutation', 'deleteDatasetTable'): TestData(
+        tenant_perm=MANAGE_DATASETS, resource_perm=DELETE_DATASET_TABLE
+    ),
+    field_id('Mutation', 'deleteEnvironment'): TestData(
+        tenant_perm=MANAGE_ENVIRONMENTS, resource_perm=DELETE_ENVIRONMENT
+    ),
```
dlpzx, line 478: All glossary mutations have CUSTOM resource permissions

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+    field_id('Mutation', 'deleteMetadataFormVersion'): TestData(
+        tenant_perm=MANAGE_METADATA_FORMS, resource_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('Mutation', 'deleteNetwork'): TestData(tenant_perm=MANAGE_ENVIRONMENTS, resource_perm=DELETE_NETWORK),
+    field_id('Mutation', 'deleteNotification'): TestData(
+        tenant_ignore=IgnoreReason.APPSUPPORT, resource_ignore=IgnoreReason.APPSUPPORT
+    ),
+    field_id('Mutation', 'deleteOmicsRun'): TestData(
+        tenant_perm=MANAGE_OMICS_RUNS, resource_ignore=IgnoreReason.NOTREQUIRED
```
dlpzx, line 496: Delete omics runs is looping over delete_omics_run which checks DELETE_OMICS_RUN

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+        tenant_perm=MANAGE_ENVIRONMENTS, resource_perm=INVITE_ENVIRONMENT_GROUP
+    ),
+    field_id('Mutation', 'inviteGroupToOrganization'): TestData(
+        tenant_perm=MANAGE_ORGANIZATIONS, resource_perm=INVITE_ORGANIZATION_GROUP
+    ),
+    field_id('Mutation', 'markNotificationAsRead'): TestData(
+        tenant_ignore=IgnoreReason.APPSUPPORT, resource_ignore=IgnoreReason.APPSUPPORT
+    ),
+    field_id('Mutation', 'postFeedMessage'): TestData(
+        tenant_ignore=IgnoreReason.APPSUPPORT, resource_ignore=IgnoreReason.NOTREQUIRED
```
dlpzx, line 545: postFeedMessage has permissions checks with the latest changes in `main` they are TARGET_TYPE_PERM

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+        tenant_perm=MANAGE_SHARES, resource_perm=SUBMIT_SHARE_OBJECT
+    ),
+    field_id('Mutation', 'submitShareObject'): TestData(tenant_perm=MANAGE_SHARES, resource_perm=SUBMIT_SHARE_OBJECT),
+    field_id('Mutation', 'syncDatasetTableColumns'): TestData(
+        tenant_perm=MANAGE_DATASETS, resource_perm=UPDATE_DATASET_TABLE
+    ),
+    field_id('Mutation', 'syncTables'): TestData(tenant_perm=MANAGE_DATASETS, resource_perm=SYNC_DATASET),
+    field_id('Mutation', 'upVote'): TestData(
+        tenant_ignore=IgnoreReason.APPSUPPORT, resource_ignore=IgnoreReason.NOTREQUIRED
+    ),
```
dlpzx, line 603: Still to merge, but upVote will be limited to resource owners

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+        resource_perm=GET_ENVIRONMENT, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('Query', 'getDashboard'): TestData(resource_perm=GET_DASHBOARD, tenant_ignore=IgnoreReason.NOTREQUIRED),
+    field_id('Query', 'getDataPipeline'): TestData(resource_perm=GET_PIPELINE, tenant_ignore=IgnoreReason.NOTREQUIRED),
+    field_id('Query', 'getDataPipelineCredsLinux'): TestData(
+        resource_perm=CREDENTIALS_PIPELINE, tenant_perm=MANAGE_PIPELINES
+    ),
+    field_id('Query', 'getDataset'): TestData(
+        resource_ignore=IgnoreReason.NOTREQUIRED, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),  # TODO Review
```
dlpzx, line 733: Working on PRs that will change this behavior, but we can merge this first and then I'll update

tests/test_permissions.py
petrkalos: added is_owner check

backend/dataall/modules/datapipelines/services/datapipelines_service.py
```diff
@@ -255,6 +255,7 @@ def _delete_repository(target_uri, accountid, cdk_role_arn, region, repo_name):
 
     @staticmethod
     @TenantPolicyService.has_tenant_permission(MANAGE_PIPELINES)
```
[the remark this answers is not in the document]
dlpzx: @petrkalos here! hahahaha

tests/test_permissions.py
noah-paige: ++ Term and Category Mutations too I think

tests/test_permissions.py
```diff
@@ -0,0 +1,1268 @@
...
+        resource_perm=LIST_ENVIRONMENT_CONSUMPTION_ROLES, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('Query', 'listAllEnvironmentGroups'): TestData(
+        resource_perm=LIST_ENVIRONMENT_GROUPS, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('Query', 'listAllGroups'): TestData(
+        resource_ignore=IgnoreReason.NOTREQUIRED, tenant_ignore=IgnoreReason.USERLIMITED
+    ),
+    field_id('Query', 'listAttachedMetadataForms'): TestData(
+        resource_ignore=IgnoreReason.NOTREQUIRED, tenant_ignore=IgnoreReason.TENANT, tenant_admin_perm=True
```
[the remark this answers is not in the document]
noah-paige: same for `listEntityMetadataForms`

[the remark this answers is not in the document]
noah-paige: same for `listUserMetadataForms`

noah-paige (commented):
Mostly minor comments on updates to ignore cases

Couple of high level Qs:
- (1) Is there any way to assess that the FE client properly handles all partial data returns appropriately?

- (2 - Not necessarily in this PR but …) should we change everywhere on our FE client side to check if there is no data rather than if no errors (i.e. `if (!response.errors)`) ++ dispatch all error if exist rather than just the first (i.e. `dispatch({ type: SET_ERROR, error: response.errors[0].message });`)

petrkalos:
> Mostly minor comments on updates to ignore cases
> 
> Couple of high level Qs:
> 
> * (1) Is there any way to assess that the FE client properly handles all partial data returns appropriately?
> * (2 - Not necessarily in this PR but …) should we change everywhere on our FE client side to check if there is no data rather than if no errors (i.e. `if (!response.errors)`) ++ dispatch all error if exist rather than just the first (i.e. `dispatch({ type: SET_ERROR, error: response.errors[0].message });`)

1. I don't think so unless we visit all pages with all combinations of permissions. I think with the changes in this PR the impact is limited and easy to test manually. In #1729 and in #1727 therear a lot of those changes and @dlpzx is doing a manual validation. I think the risk isn't huge if something doesn't render properly it's easy to track and fix
3. I agree we should change this

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+    return f'{type_name}_{field_name}'
+
+
+@dataclass
+class TestData:
+    resource_ignore: IgnoreReason = None
+    resource_perm: str = None
+    tenant_ignore: IgnoreReason = None
+    tenant_perm: str = None
+    tenant_admin_ignore: IgnoreReason = IgnoreReason.NOTREQUIRED
```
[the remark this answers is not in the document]
dlpzx: at this moment we have 2 CUSTOM permissions besides the resources and tenant permissions:
- Tenant_admin_perm
- Glossary_admin_perm
Maybe at some point we could group CUSTOM permissions and have custom_perm = 'GLOSSARIES' | 'TENANT_ADMIN'

tests/test_permissions.py
```diff
@@ -0,0 +1,1268 @@
...
+        resource_ignore=IgnoreReason.PUBLIC, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('DatasetTable', 'GlueTableProperties'): TestData(
+        resource_perm=GET_DATASET_TABLE, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('DatasetTable', 'columns'): TestData(
+        resource_ignore=IgnoreReason.INTRAMODULE, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('DatasetTable', 'dataset'): TestData(
+        resource_ignore=IgnoreReason.INTRAMODULE, tenant_ignore=IgnoreReason.NOTREQUIRED
```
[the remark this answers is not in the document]
petrkalos: the incosistency between the 2 is addressed by #1727

tests/test_permissions.py
```diff
@@ -0,0 +1,1268 @@
...
+    ),
+    field_id('Mutation', 'approveShareObject'): TestData(tenant_perm=MANAGE_SHARES, resource_perm=APPROVE_SHARE_OBJECT),
+    field_id('Mutation', 'approveTermAssociation'): TestData(
+        tenant_perm=MANAGE_GLOSSARIES, resource_ignore=IgnoreReason.USERLIMITED
+    ),
+    field_id('Mutation', 'archiveOrganization'): TestData(
+        tenant_perm=MANAGE_ORGANIZATIONS, resource_perm=DELETE_ORGANIZATION
+    ),
+    field_id('Mutation', 'batchMetadataFormFieldUpdates'): TestData(
+        tenant_perm=MANAGE_METADATA_FORMS, resource_ignore=IgnoreReason.NOTREQUIRED
```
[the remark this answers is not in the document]
petrkalos: I will add tests that will cover `MetadataFormAccessService`

tests/test_permissions.py
petrkalos: I have `glossary_owner_perm=True` in all Term/Category/Glossary mutations

tests/test_permissions.py
```diff
@@ -0,0 +1,1268 @@
...
+    ),
+    field_id('Query', 'countUnreadNotifications'): TestData(
+        resource_ignore=IgnoreReason.USERLIMITED, tenant_ignore=IgnoreReason.USERLIMITED
+    ),
+    field_id('Query', 'countUpVotes'): TestData(resource_ignore=IgnoreReason.PUBLIC, tenant_ignore=IgnoreReason.PUBLIC),
+    field_id('Query', 'generateEnvironmentAccessToken'): TestData(
+        tenant_perm=MANAGE_ENVIRONMENTS, resource_perm=CREDENTIALS_ENVIRONMENT
+    ),
+    field_id('Query', 'getAttachedMetadataForm'): TestData(
+        resource_ignore=IgnoreReason.NOTREQUIRED, tenant_ignore=IgnoreReason.NOTREQUIRED
```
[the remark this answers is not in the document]
petrkalos: Currently there is no check in the backend, @SofiaSazonova do you think we should add something?

tests/test_permissions.py
```diff
@@ -0,0 +1,1268 @@
...
+        tenant_perm=MANAGE_REDSHIFT_DATASETS, resource_perm=IMPORT_REDSHIFT_DATASET
+    ),
+    field_id('Mutation', 'inviteGroupOnEnvironment'): TestData(
+        tenant_perm=MANAGE_ENVIRONMENTS, resource_perm=INVITE_ENVIRONMENT_GROUP
+    ),
+    field_id('Mutation', 'inviteGroupToOrganization'): TestData(
+        tenant_perm=MANAGE_ORGANIZATIONS, resource_perm=INVITE_ORGANIZATION_GROUP
+    ),
+    field_id('Mutation', 'markNotificationAsRead'): TestData(
+        tenant_ignore=IgnoreReason.APPSUPPORT, resource_ignore=IgnoreReason.APPSUPPORT
```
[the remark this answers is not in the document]
petrkalos: I think after 1597 we will have to implement another custom check (same way we did for tenant, category and mf)

tests/test_permissions.py
```diff
@@ -0,0 +1,1268 @@
...
+    ),
+    field_id('Group', 'tenantPermissions'): TestData(
+        resource_ignore=IgnoreReason.USERROLEINRESOURCE, tenant_ignore=IgnoreReason.TENANT, tenant_admin_perm=True
+    ),
+    field_id('MetadataForm', 'fields'): TestData(
+        resource_ignore=IgnoreReason.INTRAMODULE, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('MetadataForm', 'homeEntityName'): TestData(
+        resource_ignore=IgnoreReason.INTRAMODULE, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
```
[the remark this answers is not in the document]
petrkalos: I think I will mark this "SIMPLIFIED" as the data that it returns are just names thus it is not exposing anything critical. @noah-paige wdyt?

tests/test_permissions.py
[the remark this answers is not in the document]
petrkalos: @noah-paige I am marking this as PUBLIC

backend/dataall/core/vpc/services/vpc_service.py
```diff
@@ -96,7 +96,11 @@ def delete_network(uri):
     def get_environment_networks(environment_uri):
         with _session() as session:
             nets = []
-            all_nets = VpcRepository.get_environment_networks(session=session, environment_uri=environment_uri)
+            all_nets = VpcRepository.get_environment_networks(
+                session=session,
+                environment_uri=environment_uri,
+                groups=get_context().groups,
+            )
```
noah-paige, line 103: calling out that this change only filters by user's `context.groups` meaning the env admin will no longer be able to see all VPCs in the environment (which was the previous behavior). Would have to add filter only if the group is not the env admin (or similar logic) to keep same behavior as earlier

tests/test_permissions.py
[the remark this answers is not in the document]
noah-paige: I think Simplified is proper class for this

tests/test_permissions.py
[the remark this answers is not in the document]
noah-paige: explanation makes sense to me - sounds good

tests/test_tenant_unauthorized.py
```diff
@@ -1,49 +1,1156 @@
...
+        tenant_perm=MANAGE_METADATA_FORMS, resource_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('Mutation', 'cancelShareExtension'): TestData(
+        tenant_perm=MANAGE_SHARES, resource_perm=SUBMIT_SHARE_OBJECT
+    ),
+    field_id('Mutation', 'createAttachedMetadataForm'): TestData(
+        tenant_ignore=IgnoreReason.BACKPORT, resource_perm=ATTACH_METADATA_FORM
+    ),
+    field_id('Mutation', 'createCategory'): TestData(
+        tenant_perm=MANAGE_GLOSSARIES, resource_ignore=IgnoreReason.NOTREQUIRED
```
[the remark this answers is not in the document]
dlpzx: Outdated comment

tests/test_permissions.py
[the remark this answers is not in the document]
dlpzx: Resolving, the logic is needed for the way the fetching of testdata in the fixtures is implemented

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+    field_id('Dashboard', 'terms'): TestData(
+        resource_ignore=IgnoreReason.PUBLIC, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('Dashboard', 'upvotes'): TestData(
+        resource_ignore=IgnoreReason.APPSUPPORT, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('Dashboard', 'userRoleForDashboard'): TestData(
+        resource_ignore=IgnoreReason.INTRAMODULE, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
```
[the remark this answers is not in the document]
dlpzx: outdated

tests/permissions.py
```diff
@@ -0,0 +1,1144 @@
...
+        resource_perm=GET_DATASET, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('DatasetStorageLocation', 'terms'): TestData(
+        resource_ignore=IgnoreReason.PUBLIC, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('DatasetTable', 'GlueTableProperties'): TestData(
+        resource_perm=GET_DATASET_TABLE, tenant_ignore=IgnoreReason.NOTREQUIRED
+    ),
+    field_id('DatasetTable', 'columns'): TestData(
+        resource_ignore=IgnoreReason.CUSTOM, tenant_ignore=IgnoreReason.USERLIMITED
```
dlpzx, line 294: `CUSTOM` as I understand it is that permissions are checked, but they are checked through their own mechanisms. Are there any checks happening in the columns resolver?

dlpzx (approved):
Left some nit changes but overall I think it is good to go. I was not strict with Redshift or Metadata Forms

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+    field,
+    perm_type,
+    request,
+    mock_input_validation,
+):
+    fid = request.node.callspec.id.split('-')[-1]
+    perm, reason = EXPECTED_RESOLVERS[fid].get(perm_type)
+    msg = f'{fid} -> {field.resolver.__code__.co_filename}:{field.resolver.__code__.co_firstlineno}'
+    logging.info(msg)
+    assert_that(field.resolver).is_not_none()
```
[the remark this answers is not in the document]
petrkalos: you are right, if resolver is `None` the msg definition will throw

tests/test_permissions.py
```diff
@@ -0,0 +1,1254 @@
...
+    mock_check_resource,
+    mock_storage,
+    field,
+    perm_type,
+    request,
+    mock_input_validation,
+):
+    fid = request.node.callspec.id.split('-')[-1]
+    perm, reason = EXPECTED_RESOLVERS[fid].get(perm_type)
+    msg = f'{fid} -> {field.resolver.__code__.co_filename}:{field.resolver.__code__.co_firstlineno}'
```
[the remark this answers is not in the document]
petrkalos: good catch, I used `inspect.unwrap` to resolve the issue

tests/permissions.py
```diff
@@ -0,0 +1,1144 @@
...
+    field_id('Mutation', 'submitShareExtension'): TestData(
+        tenant_perm=MANAGE_SHARES, resource_perm=SUBMIT_SHARE_OBJECT
+    ),
+    field_id('Mutation', 'submitShareObject'): TestData(tenant_perm=MANAGE_SHARES, resource_perm=SUBMIT_SHARE_OBJECT),
+    field_id('Mutation', 'syncDatasetTableColumns'): TestData(
+        tenant_perm=MANAGE_DATASETS, resource_perm=UPDATE_DATASET_TABLE
+    ),
+    field_id('Mutation', 'syncTables'): TestData(tenant_perm=MANAGE_DATASETS, resource_perm=SYNC_DATASET),
+    field_id('Mutation', 'upVote'): TestData(
+        tenant_ignore=IgnoreReason.APPSUPPORT, resource_ignore=IgnoreReason.NOTREQUIRED
```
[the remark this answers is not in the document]
petrkalos: Yes I changed this after @noah-paige suggestion because it is checking the Confidentiality Classification...
```
    @staticmethod
    def paginate_active_columns_for_table(uri: str, filter=None):
        context = get_context()
            table: DatasetTable = DatasetTableRepository.get_dataset_table_by_uri(session, uri)
            dataset = DatasetRepository.get_dataset_by_uri(session, table.datasetUri)
            if (
                ConfidentialityClassification.get_confidentiality_level(dataset.confidentiality)
                != ConfidentialityClassification.Unclassified.value
            ) and (dataset.SamlAdminGroupName not in context.groups and dataset.stewards not in context.groups):
                raise exceptions.UnauthorizedOperation(
                    action='LIST_DATASET_TABLE_COLUMNS',
                    message='User is not authorized to view Columns for Confidential datasets',
                )
            return DatasetColumnRepository.paginate_active_columns_for_table(session, uri, filter)
```

tests/permissions.py
petrkalos: Yes I changed this after @noah-paige suggestion because it is checking the Confidentiality Classification...
```
    @staticmethod
    def paginate_active_columns_for_table(uri: str, filter=None):
        context = get_context()
            table: DatasetTable = DatasetTableRepository.get_dataset_table_by_uri(session, uri)
            dataset = DatasetRepository.get_dataset_by_uri(session, table.datasetUri)
            if (
                ConfidentialityClassification.get_confidentiality_level(dataset.confidentiality)
                != ConfidentialityClassification.Unclassified.value
            ) and (dataset.SamlAdminGroupName not in context.groups and dataset.stewards not in context.groups):
                raise exceptions.UnauthorizedOperation(
                    action='LIST_DATASET_TABLE_COLUMNS',
                    message='User is not authorized to view Columns for Confidential datasets',
                )
            return DatasetColumnRepository.paginate_active_columns_for_table(session, uri, filter)
```

noah-paige (approved):

PR #1727, opened 2024-12-03
Consistent get_<DATA_ASSET> permissions - S3_Datasets

dlpzx:
### Feature or Bugfix
- Feature/Bugfix

### Detail

This PR is the first part of a series and only handles S3_Datasets and its child components Tables and Folders

#### Current implementation
Most API calls on a particular resource are restricted by GET_RESOURCE permissions. But for resources that are indexed in the Catalog, some metadata is considered public as it is useful for data consumers to discover and understand the data assets.  Users will click on these resources from the Catalog view and call one of the following API calls:
- getDataset
- getDatasetStorageLocation
- getDatasetTable
- getDashboard
- getRedshiftDataset
- getRedshiftTable

From the above list, initially all are decorated with resource_permission checks except for getDataset and getDatasetTable. 

#### Target implementation
- Public information should be available for data consumers to explore, that means that we first need to remove the resource_permission checks from the list of APIs.
- Not all information is public, to get AWS information and other restricted metadata we still need to verify GET_X permissions on the resource. 
- For restricted metadata, we should provide default messages that do not break the frontend

In addition in this PR some unused fields are removed and `syncTables` is simplified to return an integer with the count of synced tables

### Testing
For each of the following I tested with a user with GET permissions and without GET permissions. FE views render and show information or unauthorized to view info placeholder
- [X] Dataset View, Dataset Edit Form and list Datasets
- [x] Dataset Data tab with list of tables and folders
- [X] Table view, Table Edit
- [X] Folder view and Folder Edit

Other checks
- [x] Complete share request
- [X] With requester check table and folder and view the details of the account...
- [X] Worksheets query with owned table
- [X] Worksheets query with shared table
- [X] Crawl dataset - correct S3 path
- [X] Sync tables - correct params

### Security

  - Is the input sanitized?


backend/dataall/modules/s3_datasets/services/dataset_location_service.py
```diff
@@ -135,3 +134,42 @@ def _delete_dataset_folder_read_permission(session, dataset: S3Dataset, location
...
+                    'S3BucketName': dataset.S3BucketName,
+                    'GlueDatabaseName': dataset.GlueDatabaseName,
+                    'IAMDatasetAdminRoleArn': dataset.IAMDatasetAdminRoleArn,
+                    'KmsAlias': dataset.KmsAlias,
+                    'importedS3Bucket': dataset.importedS3Bucket,
+                    'importedGlueDatabase': dataset.importedGlueDatabase,
+                    'importedKmsKey': dataset.importedKmsKey,
+                    'importedAdminRole': dataset.importedAdminRole,
+                }
+        except ResourceUnauthorized:
```
petrkalos, line 163: I'd let this exception bubble up and appear in the error section of the gql response

backend/dataall/modules/s3_datasets/services/dataset_location_service.py
```diff
@@ -135,3 +134,42 @@ def _delete_dataset_folder_read_permission(session, dataset: S3Dataset, location
...
+            with context.db_engine.scoped_session() as session:
+                ResourcePolicyService.check_user_resource_permission(
+                    session=session,
+                    username=context.username,
+                    groups=context.groups,
+                    resource_uri=folder.locationUri,
+                    permission_name=GET_DATASET_FOLDER,
+                )
+                dataset = DatasetRepository.get_dataset_by_uri(session, folder.datasetUri)
+                return {
```
petrkalos, line 151: What happens if we return the entire object? Would GQL fail because of the extra fields or will only return those that have mappings?
If that's not possible then maybe something like 
`return {field.name: getattr(dataset, field.name, None) for field in DatasetRestrictedInformation.fields}`
 would work better as we won't have to maintain those mappings manually. You can also export it to global method and reuse it an all places

petrkalos (changes requested):

backend/dataall/modules/s3_datasets/services/dataset_location_service.py
dlpzx: Why? The end user is authorized to get the object, but some fields are restricted

backend/dataall/modules/s3_datasets/services/dataset_location_service.py
dlpzx: You are right. Let me change it a bit

backend/dataall/modules/s3_datasets/services/dataset_location_service.py
petrkalos: They will get the top level object but an exception for the "restricted" subsection, that way it is more consistent to what the rest of the sub-resolvers are doing

dlpzx: I'll work on it, it is true that for consistency it would be better to not retrieve anything

dlpzx: I have submitted a new version and tested the following with a user with GET permissions and without GET permissions:
- Dataset view and all tabs
- Table view and tabs
- Folder view
- Worksheets dropdowns

backend/dataall/modules/s3_datasets/services/dataset_service.py
```diff
@@ -1,6 +1,7 @@
 import os
 import json
 import logging
+from sqlalchemy import inspect
```
petrkalos, line 4: unused

I usually do a `Code -> Optimize Imports` before commit to automatically remove unused imports

backend/dataall/modules/s3_datasets/services/dataset_location_service.py
```diff
@@ -1,9 +1,10 @@
+from sqlalchemy import inspect
```
petrkalos, line 1: unused

backend/dataall/modules/s3_datasets/services/dataset_location_service.py
```diff
@@ -1,9 +1,10 @@
...
 from dataall.modules.s3_datasets.indexers.dataset_indexer import DatasetIndexer
 from dataall.base.context import get_context
 from dataall.core.permissions.services.resource_policy_service import ResourcePolicyService
 from dataall.core.permissions.services.tenant_policy_service import TenantPolicyService
 from dataall.modules.catalog.db.glossary_repositories import GlossaryRepository
-from dataall.base.db.exceptions import ResourceShared, ResourceAlreadyExists
+from dataall.base.db.exceptions import ResourceUnauthorized, ResourceAlreadyExists
```
petrkalos, line 7: unused

backend/dataall/modules/s3_datasets/services/dataset_table_service.py
```diff
@@ -1,5 +1,5 @@
 import logging
-
+from sqlalchemy import inspect
```
petrkalos, line 2: unused

frontend/src/modules/Folders/components/FolderOverview.js
```diff
@@ -28,7 +28,7 @@ export const FolderOverview = (props) => {
       <Grid item lg={4} xl={3} xs={12}>
         <ObjectMetadata
           environment={folder.dataset.environment}
-          region={folder.dataset.region}
+          region={folder.restricted?.region || 'UNAUTHORIZED_INFO'}
```
petrkalos, line 31: * This is already handled in [ObjectMetadata](https://github.com/data-dot-all/dataall/blob/eca354e8df8189c2fcf31732579def69bad90bfb/frontend/src/design/components/ObjectMetadata.js#L127).
* I think it's better not to print region at all than printing "Region -> UNAUTHORIZED"

ditto for the rest of your special handling

petrkalos (changes requested):

frontend/src/modules/Folders/components/FolderOverview.js
dlpzx: In the latest commit 4b675f3 most views have been simplified except for the tables and folder view where table names are presented in the rows.

petrkalos (approved):

petrkalos (approved):

backend/dataall/modules/s3_datasets/api/dataset/types.py
```diff
@@ -23,6 +24,22 @@
     ],
 )
 
+DatasetRestrictedInformation = gql.ObjectType(
+    name='DatasetRestrictedInformation',
+    fields=[
+        gql.Field(name='AwsAccountId', type=gql.String),
+        gql.Field(name='region', type=gql.String),
+        gql.Field(name='S3BucketName', type=gql.String),
+        gql.Field(name='GlueDatabaseName', type=gql.String),
```
noah-paige, line 33: nit: A call out that these 2 fields (`S3BucketName` and `GlueDatabaseName`) are a part of index in opensearch (`backend/dataall/modules/s3_datasets/indexers/dataset_indexer.py`)  - not sure really how "restricted" they are treated if that information can be found elsewhere

backend/dataall/modules/s3_datasets/api/dataset/types.py
```diff
@@ -35,29 +52,13 @@
...
-        gql.Field(name='S3BucketName', type=gql.String),
-        gql.Field(name='GlueDatabaseName', type=gql.String),
-        gql.Field(name='GlueCrawlerName', type=gql.String),
-        gql.Field(name='GlueCrawlerSchedule', type=gql.String),
-        gql.Field(name='GlueProfilingJobName', type=gql.String),
-        gql.Field(name='GlueProfilingTriggerSchedule', type=gql.String),
-        gql.Field(name='IAMDatasetAdminRoleArn', type=gql.String),
-        gql.Field(name='KmsAlias', type=gql.String),
-        gql.Field(name='bucketCreated', type=gql.Boolean),
-        gql.Field(name='glueDatabaseCreated', type=gql.Boolean),
```
noah-paige, line 49: nit: making note we can remove the following from data model if we wanted to clean up sometime in the future

- bucketCreated
- glueDatabaseCreated
- iamAdminRoleCreated
- lakeformationLocationCreated
- bucketPolicyCreated

backend/dataall/modules/s3_datasets/api/table/types.py
```diff
@@ -21,6 +22,15 @@
         gql.Field(name='nodes', type=gql.ArrayType(TablePermission)),
     ],
 )
+DatasetTableRestrictedInformation = gql.ObjectType(
+    name='DatasetTableRestrictedInformation',
+    fields=[
+        gql.Field(name='AwsAccountId', type=gql.String),
+        gql.Field(name='GlueDatabaseName', type=gql.String),
+        gql.Field(name='GlueTableName', type=gql.String),
```
noah-paige, line 30: similar to call out above that `GlueDatabaseName` is a part of index in opensearch (`backend/dataall/modules/s3_datasets/indexers/table_indexer.py`) - not sure really how "restricted" they are treated if that information can be found elsewhere

++ to note I would think best if anything would be to restrict on open search and keep the other server side restricted logic the same

frontend/src/modules/Folders/views/FolderView.js
```diff
@@ -233,7 +233,7 @@ const FolderView = () => {
   const fetchItem = useCallback(async () => {
     setLoading(true);
     const response = await client.query(getDatasetStorageLocation(params.uri));
-    if (!response.errors && response.data.getDatasetStorageLocation !== null) {
+    if (response.data.getDatasetStorageLocation !== null) {
```
noah-paige, line 236: nit: update the error dispatch to dispatch all errors (1 or more)

frontend/src/modules/S3_Datasets/components/DatasetFolders.js
```diff
@@ -69,7 +70,7 @@ export const DatasetFolders = (props) => {
     const response = await client.query(
       listDatasetStorageLocations(dataset.datasetUri, filter)
     );
-    if (!response.errors) {
+    if (response.data.getDataset != null) {
```
noah-paige, line 73: nit: dispatch all errors

frontend/src/modules/S3_Datasets/components/DatasetTables.js
```diff
@@ -80,7 +80,7 @@ export const DatasetTables = (props) => {
         filter: { ...filter }
       })
     );
-    if (!response.errors) {
+    if (response.data.getDataset != null) {
```
noah-paige, line 83: nit: dispatch all errors

frontend/src/modules/S3_Datasets/components/DatasetUpload.js
```diff
@@ -153,7 +153,7 @@ export const DatasetUpload = (props) => {
               <TextField
                 fullWidth
                 helperText="Prefix without trailing slash at the end (e.g raw not raw/)"
-                label={`s3://${dataset.S3BucketName}/`}
+                label={`s3://${dataset.restricted.S3BucketName}/`}
```
noah-paige, line 156: nit: i know tab is limited to admin but any need to also protect this with `emptyPrintUnauthorized` or similar?

frontend/src/modules/S3_Datasets/components/DatasetStartCrawlerModal.js
```diff
@@ -132,7 +132,7 @@ export const DatasetStartCrawlerModal = (props) => {
                         error={Boolean(touched.prefix && errors.prefix)}
                         fullWidth
                         helperText={touched.prefix && errors.prefix}
-                        label={`s3://${dataset.S3BucketName}/`}
+                        label={`s3://${dataset.restricted.S3BucketName}/`}
```
noah-paige, line 135: nit: i know tab is limited to admin but any need to also protect this with `emptyPrintUnauthorized` or similar?

backend/dataall/modules/s3_datasets/api/dataset/types.py
```diff
@@ -23,6 +24,22 @@
...
+        gql.Field(name='AwsAccountId', type=gql.String),
+        gql.Field(name='region', type=gql.String),
+        gql.Field(name='S3BucketName', type=gql.String),
+        gql.Field(name='GlueDatabaseName', type=gql.String),
+        gql.Field(name='IAMDatasetAdminRoleArn', type=gql.String),
+        gql.Field(name='KmsAlias', type=gql.String),
+        gql.Field(name='importedS3Bucket', type=gql.Boolean),
+        gql.Field(name='importedGlueDatabase', type=gql.Boolean),
+        gql.Field(name='importedKmsKey', type=gql.Boolean),
+        gql.Field(name='importedAdminRole', type=gql.Boolean),
```
noah-paige, line 39: did we also want to include the following share expiration / config related information as part of restricted?
```
        gql.Field(name='autoApprovalEnabled', type=gql.Boolean),
        gql.Field(name='enableExpiration', type=gql.Boolean),
        gql.Field(name='expirySetting', type=gql.String),
        gql.Field(name='expiryMinDuration', type=gql.Integer),
        gql.Field(name='expiryMaxDuration', type=gql.Integer),
```

frontend/src/modules/S3_Datasets/views/DatasetView.js
```diff
@@ -161,7 +161,7 @@ const DatasetView = () => {
   const fetchItem = useCallback(async () => {
     setLoading(true);
     const response = await client.query(getDataset(params.uri));
-    if (!response.errors && response.data.getDataset !== null) {
+    if (response.data.getDataset !== null) {
```
noah-paige, line 164: nit: expose all errors

frontend/src/modules/Tables/views/TableView.js
```diff
@@ -222,7 +222,7 @@ const TableView = () => {
   const fetchItem = useCallback(async () => {
     setLoading(true);
     const response = await client.query(getDatasetTable(params.uri));
-    if (!response.errors && response.data.getDatasetTable !== null) {
+    if (response.data.getDatasetTable !== null) {
```
noah-paige, line 225: nit: expose all errors

frontend/src/utils/helpers/emptyPrintUnauthorized.js
```diff
@@ -0,0 +1,5 @@
+function emptyPrintUnauthorized(param) {
+  return param ? param : 'UNAUTHORIZED_INFO';
```
noah-paige, line 2: very nit: would it be a better UX to have the Unauth string be more like `******` to represent redacted/restricted info rather than a replacement string of `UNAUTHORIZED_INFO`

tests/modules/s3_datasets_shares/conftest.py
```diff
@@ -253,7 +254,12 @@ def dataset_confidential_fixture(env_fixture, org_fixture, dataset, group) -> S3
 
 @pytest.fixture(scope='module')
 def table_fixture(db, dataset_fixture, table, group, user):
-    table1 = table(dataset=dataset_fixture, name='table1', username=user.username)
+    dataset = dataset_fixture
+    dataset.GlueDatabaseName = dataset_fixture.restricted.GlueDatabaseName
+    dataset.region = dataset_fixture.restricted.region
+    dataset.S3BucketName = dataset_fixture.restricted.S3BucketName
+    dataset.AwsAccountId = dataset_fixture.restricted.AwsAccountId
+    table1 = table(dataset=dataset, name='table1', username=user.username)
```
noah-paige, line 262: nit: any reason handled this way here and updated `def table` factory for a similar/same required change at `tests/modules/s3_datasets/conftest.py`

noah-paige:
All minor enhancmenets / nits - the only pending comment I have before approval is regarding restrictions to additional dataset config info (re: https://github.com/data-dot-all/dataall/pull/1727#discussion_r1871981295) cc: @dlpzx

backend/dataall/modules/s3_datasets/api/dataset/types.py
dlpzx: I did not want to add a migration script because backfilling it in 2.6.2 was going to be a pain; but totally agree on the cleanup

backend/dataall/modules/s3_datasets/api/dataset/types.py
dlpzx: Initially yes, but this information is fetched when opening a share request. At the end the user needs it for the share request creation

backend/dataall/modules/s3_datasets/api/table/types.py
dlpzx: I think this might be the solution, restricting it also in OpenSearch

frontend/src/modules/S3_Datasets/components/DatasetUpload.js
dlpzx: This PR has gone back and forth. At the beginning I added checks everywhere, but finally I opted to use it only when it will be used. The same as if in the backend an if condition is not needed then we don't add it

frontend/src/utils/helpers/emptyPrintUnauthorized.js
dlpzx: Done!

frontend/src/modules/S3_Datasets/views/DatasetView.js
dlpzx: Same as with Dashboards. I don't know if adding more error banners is going to give users more useful info. If the top level query fails, then the errors that we might find in the nested resolvers are very unpredictable. For example if get_Dataset fails because the dataset is not found, the restricted is going to fail with something as 'URI is null'; which might be misleading.

tests/modules/s3_datasets_shares/conftest.py
dlpzx: This is not very elegant, agreed. The issue is that the `dataset_fixture` parameter takes in this case the value of the gql Dataset type. While the table factory accepts db S3Dataset model

frontend/src/modules/S3_Datasets/views/DatasetView.js
noah-paige: is it guaranteed the first error in the dispatch always the top level?

noah-paige:
@dlpzx - I think will have to merge latest from OS, resolve conflicts, and make changes to `test_permissions` for the changes in auth checks / nested resolvers 

Then good to go

tests/modules/s3_datasets_shares/conftest.py
noah-paige: I am not sure I follow - is it not the exact same dataset and table factories used in `tests/modules/s3_datasets/conftest.py` but in that conftest we define Table factor as

```
@pytest.fixture(scope='module', autouse=True)
def table(db):
    cache = {}

    def factory(dataset: S3Dataset, name, username) -> DatasetTable:
        key = f'{dataset.datasetUri}-{name}'
        if cache.get(key):
            return cache.get(key)
        with db.scoped_session() as session:
            table = DatasetTable(
                name=name,
                label=name,
                owner=username,
                datasetUri=dataset.datasetUri,
                GlueDatabaseName=dataset.restricted.GlueDatabaseName,
                GlueTableName=name,
                region=dataset.restricted.region,
                AWSAccountId=dataset.restricted.AwsAccountId,
                S3BucketName=dataset.restricted.S3BucketName,
                S3Prefix=f'{name}',
            )
...
```

(note the way we set GlueDatabaseName, region, S3BucketName, and S3BucketName)

noah-paige (approved):

Issue #1668, opened 2024-10-28
Metadata Form Enforcement


SofiaSazonova:
# Enforcement

Metadata forms can be obligatory to fill in on different levels. User can select the metadata form and entity types, that should have this form attached.  Enforcement affects selected entity types on all lower levels hierarchically.

## Table E.1. Metadata form enforcement rule

| Field | Description | Type | Possible Values |  
|-|-|-|-|
| Metadata Form | Identifier of metadata form, that is enforced | String | URI of existing form |
| Metadata Form Version| number of version, that is enforced | Int | |
| Level | Level of affected entities. All lower levels will be affected. | String | From table E.2 |
| Home Entity Uri | If level is not global, the Uri of the entity under which the MF is enforced | String |  |
| Entity Types | Entity type affected by this enforcement | Array [String] | Identifiers of data.all entities types (table E.2) |
| Severity |  | String | Obligatory/Recommended |

## Table E.2. Metadata form levels’ hierarchy
![meta levels](https://github.com/user-attachments/assets/4495ad4a-61e7-4f28-b28f-3860b33bc8fc)

Who can enforce: 

* Data.all admins can enforce any form on any level across the platform. They have full control over metadata form enforcement.
* Owners/admins of the data can enforce forms for this levels and levels below in the hierarchy. For example, an org admin can enforce a form for the org, all teams in that org, all environments in the org, all datasets in those environments, etc.
* Share approvers and requestors can enforce forms for a specific share they are involved with. However, they can only delete enforcement rules they created themselves - they cannot delete rules created by others

So in summary, enforcement capabilities cascade along with administrative privileges in the hierarchy. Global admins have full control, org/env admins can enforce for their sphere and below, dataset admins for the datasets and items in it, and share requesters and approvers for a specific share.

## How we enforce?
1. When MF is enforced, but entity misses it, on entity page there is a HUGE reminder about it.
2. If MF is mandatory, than NO actions are allowed until required MF is attached |
2.1. Orgs:

-  :no_entry: invite/remove groups
- :no_entry: link envs
- :no_entry: edit info

2.2. Envs
 - :no_entry: invite/remove
 - :no_entry: add consumption roles
 - :no_entry: add datasets
 - :no_entry: edit info
 
2.3. Datasets
- :no_entry: data actions (lauch crawler, add folders, delete tables, syncing data)
- :no_entry: create new shares. Existing shares are intact
- :no_entry: edit info

SofiaSazonova:
[...]

PR #1730, opened 2024-12-04
Metadata form enforcement

SofiaSazonova:
### Feature or Bugfix

- Feature

### Detail
**MF view**
1. Enforcement rule tab is added
2. MF owner can add Enforcement rules, others can view
3. For each Enforcement rule user can see entities affected by it and if the form is attached or not
4. Rule is automatically updated for latest MF versions when version is created/deleted

**Entity view**
1. If any enforced forms are missing from entity 'Metadata' tab on it's page has "alarm" icon
2. Missing MFs are highlighted  in the list of attached entities

**Backend**
1. New core component MetadataFormEntityManager. 
2. All eligible entities register in this manager, when their module is enabled
3. This manager is unused if metadata form module is not created
4. Metadata form module depends on DatasetBase module, because we have 'Dataset' enforcement level and logic binded to it.

**IMPORTANT**
Before this PR is merged, I will deregister modules, which don't have frontend for MF yet.

### Relates
- #1668 

### Security

  - Is the input sanitized?


SofiaSazonova:
https://github.com/user-attachments/assets/c989d438-0c6c-4bba-add6-e79c46eabcc1

backend/dataall/modules/metadata_forms/services/metadata_form_enforcement_service.py
```diff
@@ -0,0 +1,341 @@
...
+                return DatasetBaseRepository.get_dataset_by_uri(session, rule.homeEntity).label
+
+    @staticmethod
+    @TenantPolicyService.has_tenant_permission(MANAGE_METADATA_FORMS)
+    @MetadataFormAccessService.can_perform(ENFORCE_METADATA_FORM)
+    def delete_mf_enforcement_rule(uri, rule_uri):
+        with get_context().db_engine.scoped_session() as session:
+            rule = MetadataFormRepository.get_mf_enforcement_rule_by_uri(session, rule_uri)
+            session.delete(rule)
+            session.commit()
```
petrkalos, line 259: this logic belongs to the repo, also first `get` then `delete` suboptimal, perhaps there is a native way to do it something like `session.query(MetadataFormEnforcementRule).filter_by(id=rule_uri).delete()`

petrkalos (changes requested):
* there are some major architectural issues that we agreed to rectify at a later PR, I tried to write a detailed comment to convert directly to an issue in our backlog
* there are some that are bit more minor but I think they must be addressed (easy wins here and tehre)
* you will have to rebase and fix the persmission tests

frontend/src/modules/Organizations/views/OrganizationView.js
```diff
@@ -40,24 +42,7 @@ import {
 } from '../components';
 import { MetadataAttachment } from '../../Metadata_Forms/components';
 import { isModuleEnabled, ModuleNames } from '../../../utils';
```
dlpzx, line 44: nit: utils is a global package, we should use global imports; I think in general we should review UI imports in the metadata forms module

backend/dataall/modules/metadata_forms/api/resolvers.py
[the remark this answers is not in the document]
SofiaSazonova: I follow the notation in other resolvers 
e.g. `s3_dataset/dataset/resolvers`,  `share_base/api/resolvers` ...

backend/dataall/modules/metadata_forms/db/metadata_form_repository.py
```diff
@@ -331,3 +343,50 @@ def get_all_attached_metadata_forms(session, mf_uri, version=None):
...
+                    all_rules = all_rules.filter(MetadataFormEnforcementRule.entityTypes.any(etype))
+            if filter.get('level'):
+                all_rules = all_rules.filter(MetadataFormEnforcementRule.level == filter.get('level'))
+            if filter.get('home_entity'):
+                all_rules = all_rules.filter(MetadataFormEnforcementRule.homeEntity == filter.get('home_entity'))
+
+        return all_rules.all()
+
+    @staticmethod
+    def update_version_in_rules(session, uri, version):
```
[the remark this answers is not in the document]
SofiaSazonova: It's not rule version. It version of MF that is used in rule

backend/dataall/core/environment/db/environment_models.py
```diff
@@ -40,8 +42,18 @@ class Environment(Resource, Base):
...
+
+    def get_owner(self):
+        return self.SamlGroupName
+
+    def get_entity_name(self):
+        return self.label
+
 
 class EnvironmentGroup(Base):
+    __metaclass__ = MetadataFormEntity
```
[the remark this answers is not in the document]
dlpzx: +1 on using inheritance on the Resource ABC class

backend/dataall/modules/metadata_forms/services/metadata_form_enforcement_service.py
```diff
@@ -106,7 +106,7 @@ def get_affected_environments(uri, rule=None):
             if rule.level == MetadataFormEnforcementScope.Global.value:
                 return EnvironmentRepository.query_all_active_environments(session)
             if rule.level == MetadataFormEnforcementScope.Organization.value:
-                return EnvironmentRepository.get_all_envs_by_organization(session, rule.homeEntity)
+                return OrganizationRepository.query_organization_environments(session, uri=rule.homeEntity, filter=None)
```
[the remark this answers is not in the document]
SofiaSazonova: done

frontend/src/modules/Organizations/views/OrganizationView.js
SofiaSazonova: done

backend/dataall/core/organizations/db/organization_models.py
```diff
@@ -26,3 +39,16 @@ class OrganizationGroup(Base):
...
+
+    def get_owner(self):
+        return self.invitedBy
+
+    def get_entity_name(self):
+        return f'{self.groupUri}-{self.organizationUri}'
+
+
+MetadataFormEntityManager.register(Organization, MetadataFormEntityTypes.Organizations.value)
+MetadataFormEntityManager.register(OrganizationGroup, MetadataFormEntityTypes.OrganizationTeams.value)
```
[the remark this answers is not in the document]
SofiaSazonova: It's related to what we show to user.
If you insist on using model.name, we will anyway have to map it to user-readable names.

backend/dataall/modules/dashboards/db/dashboard_models.py
```diff
@@ -37,3 +39,9 @@ class Dashboard(Resource, Base):
     @classmethod
     def uri(cls):
         return cls.dashboardUri
```
dlpzx, line 41: the `get_uri` function is missing, is it because `uri` is used instead?

backend/dataall/core/organizations/db/organization_repositories.py
```diff
@@ -131,3 +131,7 @@ def find_group_membership(session, groups, organization):
             .first()
         )
         return membership
+
```
dlpzx, line 134: very nit: instead of returning the Query object, which is a very sqlalchemy-specific class, we could return directly the list of organizations and not call `.all()` in the metadata form service

backend/dataall/core/metadata_manager/metadata_form_entity_manager.py
```diff
@@ -0,0 +1,60 @@
...
+    OrganizationTeams = 'Organization Team'
+    Environments = 'Environment'
+    EnvironmentTeams = 'Environment Team'
+    S3Datasets = 'S3-Dataset'
+    RDDatasets = 'Redshift-Dataset'
+    Worksheets = 'Worksheets'
+    Dashboards = 'Dashboard'
+    ConsumptionRoles = 'Consumption Role'
+    Notebooks = 'Notebook'
+    MLStudioEntities = 'ML Studio Entity'
```
dlpzx, line 18: nit: rename `MLStudioEntities` to `MLStudioUser`
nit2: some names are in plural, other in singular

backend/dataall/core/metadata_manager/metadata_form_entity_manager.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def get_uri(self):
+        pass
+
+
+class MetadataFormEntityManager:
+    """
+    API for managing entities, to which MF can be attached.
+    """
+
```
dlpzx, line 41: instead of registering the resources with a [str, MetadataFormEntity] you could use the enum `MetadataFormEntityTypes` to ensure only those are registered

frontend/src/modules/Organizations/views/OrganizationView.js
```diff
@@ -40,24 +42,7 @@ import {
...
-  },
-  {
-    label: 'Teams',
-    value: 'teams',
-    icon: <SupervisedUserCircleRounded fontSize="small" />
-  }
-];
-
-const activeTabs = tabs.filter((tab) => tab.active !== false);
+import { listRulesThatAffectEntity } from '../../Metadata_Forms/services';
```
dlpzx, line 45: Alongside the whole PR and the implementation of the UI components of metadata forms there are violations of the modularization principles established in data.all v2.

The idea is that any of the FE modules can be replaced or removed without affecting other modules. Anything that is shared across modules is placed in the global packages. That means that imports such as `'../../Metadata_Forms/services';` that go 2 levels up and get an API from another module are "forbidden" because if we were now to remove or alter the Organizations module other modules would be affected. 

The whole MF module needs to be reviewed, but I think this task can be left out of this PR.

frontend/src/modules/Metadata_Forms/components/MetadataFormEnforcement.js
```diff
@@ -1,21 +1,781 @@
...
+    } catch (e) {
+      dispatch({ type: SET_ERROR, error: e.message });
+    }
+  };
+
+  useEffect(async () => {
+    if (client) {
+      await fetchEnvironments();
+      await fetchOrganizations();
+      await fetchDatasets();
```
dlpzx, line 137: I am not sure of this, but can we restrict the fetchX call to call only the one of the enforcementScope selected? If we are using the Organizations, we do not need to fetch Environments and Datasets and we can save 2 costly API calls

dlpzx (approved):
Left some comments, mostly minor

backend/dataall/modules/metadata_forms/services/metadata_form_enforcement_service.py
```diff
@@ -99,20 +104,22 @@ def get_affected_organizations(uri, rule=None):
             return []
 
     @staticmethod
-    def get_affected_environments(uri, rule=None):
+    def get_affected_environments(uri, rule=None) -> List[Environment]:
         with get_context().db_engine.scoped_session() as session:
             if not rule:
                 rule = MetadataFormRepository.get_mf_enforcement_rule_by_uri(session, uri)
```
petrkalos, line 110: is this change of indentation intentional?
previously if rule!=None you will run the logic of your function
now if rule!=None you will skip it

petrkalos (changes requested):
making sure that the change in `get_affected_environments` logic is intented

backend/dataall/modules/metadata_forms/services/metadata_form_enforcement_service.py
SofiaSazonova: You are right!

backend/dataall/core/environment/db/environment_models.py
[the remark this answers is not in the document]
SofiaSazonova: separarte issue opened

backend/dataall/modules/dashboards/db/dashboard_models.py
SofiaSazonova: done

backend/dataall/core/metadata_manager/metadata_form_entity_manager.py
SofiaSazonova: Let's do in in backend refactoring, when rewriting EntityManager

backend/dataall/modules/metadata_forms/api/resolvers.py
```diff
@@ -60,7 +64,11 @@ def get_metadata_form(context: Context, source, uri):
     return MetadataFormService.get_metadata_form_by_uri(uri=uri)
```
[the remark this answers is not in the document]
SofiaSazonova: It's main resolver
getMetadataForm = gql.QueryField(
    name='getMetadataForm',
    args=[gql.Argument('uri', gql.NonNullableType(gql.String))],
    type=gql.Ref('MetadataForm'),
    resolver=get_metadata_form,
    test_scope='MetadataForm',
)

subquery has different source

petrkalos (approved):

Issue #1420, opened 2024-07-18
notification improvements


zsaltys:
There are various actions in data.all that happen asynchronously such as a share being processed, dataset and environment stacks being updates, health of shares being verified etc. All of these actions can fail for whatever reason. Currently when things fail users are not able to find out until they check the logs or inspect data.all UI.

I'd like to propose that all asynchronous actions in data.all that can fail would send notifications and also be integrated with the recent reminder service that was implemented for shares to remind users that approvals for shares are pending.

A list of actions that should send notifications and to whom they should be sent:

a) An approved share fails (or revoked) to become successfully processed - both dataset owner and requester team should be notified and reminder. Optionally DA admins should also be able to subscribe to these notifications.

b) Dataset or Environment stack fails to update -environment admin team should be notified as well as the dataset owner team. Reminders should be sent. DA admins should be able to optionally subscribe as well.

c) Any background task such as catalog reindexing, share health verified - if it fails (throws an exception) then DA admins should be notified ONCE.

d) If a share becomes unhealthy both the requester and dataset owner should be notified and reminded. DA admins should be able to optionally subscribe also.

e) If attempting to re-apply a share fails both the requester and dataset owner should be notified ONCE. DA admins should be able to optionally subscribe also.

There are already a few tickets filed for this effort that we should combine into this major story:

https://github.com/data-dot-all/dataall/issues/1251
https://github.com/data-dot-all/dataall/issues/1299

There might be more that I was not able to find.

noah-paige:
Hi @zsaltys - also want to link these comments from an earlier discussion on persistent reminders feature that begin to start thinking through the above issue described

https://github.com/data-dot-all/dataall/issues/1248#issuecomment-2108546716
https://github.com/data-dot-all/dataall/issues/1248#issuecomment-2112705676


Ultimately I envision we would have a list of different "NotificationEvents" (i.e. `StackFailedNotification`, `ShareUnHealthyNotification`, `CatalogIndexerFailedNotification`, etc.) and so on that each have their own generic notification groups and reminder schedules depending on the event specified

Also this list of events should be easily extendable if tomorrow we wanted to add a notification for `NewTableMetricsNotification` OR `NewTeamInvitedNotification` for example.

I think it may be best to take some time and iron our a good generic and flexible design on how this notification system would look like in data.all. I think once we define it once it will be seamless to extend to more and more type of events or more and more type of delivery channels to send outbound notifications (on top of already in the UI and (optionally) via email)   

[...]

PR #1748, opened 2024-12-26
[GH-1420] Notification Improvements Stage 1

TejasRGitHub:
### Feature or Bugfix

- Feature
- Bugfix
- Refactoring

### Detail
- <feature1 or bug1>
- <feature2 or bug2>

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?


backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -268,12 +280,19 @@ def verify_share(
...
-                        processor.Processor(session, share_data, shareable_items).verify_shares()
+                        health_status = processor.Processor(session, share_data, shareable_items).verify_shares_health_status()
+                        share_object_item_health_status.append(health_status)
                     else:
                         log.info(f'There are no items to verify of type {type.value}')
                 except Exception as e:
                     log.error(f'Error occurred during share verifying of {type.value}: {e}')
+                    AdminNotificationService().notify_admins_with_error_log(
+                        process_error=f'Error occurred during verification of share with uri: {share_data.share.shareUri} for processor type: {type.value}  due to an unknown exception',
+                        error_logs=[str(e)], process_name='Sharing Service')
```
noah-paige, line 291: why not follow the same pattern as above in apply and revoke...
- (1) Why call `AdminNotificationService().notify_admins_with_error_log` here when also calling 
- (2) Why calling the `NotificationService` for each share item type

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -334,26 +354,45 @@ def reapply_share(cls, engine: Engine, share_uri: str) -> bool:
                                 log.info(f'Reapplying {type.value} succeeded = {success}')
                                 if not success:
                                     reapply_successful = False
+                                if success:
```
noah-paige, line 357: nit: `else`

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -334,26 +354,45 @@ def reapply_share(cls, engine: Engine, share_uri: str) -> bool:
...
                             else:
                                 log.info(f'There are no items to reapply of type {type.value}')
                         except Exception as e:
                             log.error(f'Error occurred during share reapplying of {type.value}: {e}')
+                            code_exception_list.append(f'Error occurred during reapplying of share with uri: {share_data.share.shareUri} for processor type: {type.value}  due to an unknown exception: {e}')
 
+                if not reapply_successful:
+                    log.info(
+                        f'Sending notifications since share object item(s) for share: {share_data.share.shareUri} are in unhealthy state after reapplying shares')
+                    ShareNotificationService(session=session, dataset=share_data.dataset,
```
noah-paige, line 370: could you have a case where some item types reapply successful so get a is healthy email but then also some fail so get a is unhealthy email? Would that not be confusing for end user? 

Why are we also sending notifications by share item type here but only on whole process for apply/revoke

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -334,26 +354,45 @@ def reapply_share(cls, engine: Engine, share_uri: str) -> bool:
...
+                log.exception(f'Error occurred during share reapply: {e}')
+                code_exception_list.append(str(e))
                 return False
 
+            finally:
+                if len(code_exception_list) > 0:
+                    AdminNotificationService().notify_admins_with_error_log(
+                        process_error=f'Error occurred during reapplying of share with uri: {share_data.share.shareUri}',
+                        error_logs=code_exception_list, process_name='Sharing Service')
+
```
noah-paige, line 395: why this here for re-apply? and why not everywhere else?

backend/dataall/modules/shares_base/services/share_notification_service.py
```diff
@@ -253,11 +257,96 @@ def notify_share_expiration_to_requesters(self):
...
+        share_link_text = ''
+        if os.environ.get('frontend_domain_url'):
+            share_link_text = (
+                f'<br><br> Please visit data.all <a href="{os.environ.get("frontend_domain_url")}'
+                f'/console/shares/{self.share.shareUri}">share link </a> '
+                f'to take action or view more details'
+            )
+        msg = (
+            f'Share request made for dataset: <b>{self.dataset.label}</b> with requestor principal: <b>{self.share.principalRoleName}</b> failed. <br><br>'
+            f'You can delete and resubmit the failed items in the share. If your share item still remains in the Share_Failed state then please get in touch with data.all admins.'
```
noah-paige, line 275: don't think this makes sense for revoke failing

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -268,12 +280,19 @@ def verify_share(
...
+                        share_object_item_health_status.append(health_status)
                     else:
                         log.info(f'There are no items to verify of type {type.value}')
                 except Exception as e:
                     log.error(f'Error occurred during share verifying of {type.value}: {e}')
+                    AdminNotificationService().notify_admins_with_error_log(
+                        process_error=f'Error occurred during verification of share with uri: {share_data.share.shareUri} for processor type: {type.value}  due to an unknown exception',
+                        error_logs=[str(e)], process_name='Sharing Service')
 
+            if False in share_object_item_health_status:
```
noah-paige, line 293: handle `share_object_item_health_status` as a list and check False in list here but everywhere else you handle it as a single boolean value starting as True and changing to False if one or more fails

backend/dataall/modules/shares_base/services/share_notification_service.py
```diff
@@ -253,11 +257,96 @@ def notify_share_expiration_to_requesters(self):
...
+            share_link_text = (
+                f'<br><br> Please visit data.all <a href="{os.environ.get("frontend_domain_url")}'
+                f'/console/shares/{self.share.shareUri}">share link </a> '
+                f'to take action or view more details'
+            )
+        msg = (
+            f'Hello Team, <br>'
+            f'Your share with share uri: <b>{self.share.shareUri}</b> has one or more unhealthy share items. <br><br>'
+            f'Once you visit your share link you can click on the Reapply button and this should correct your share and get it into an healthy state.<b> If this doesn\'t get your share in healthy state then please get in touch with data.all admins for your share.'
+            f'<br><br><b>Please note</b>: If you are using any terraform / cloudformation or any other IaC to also manage your bucket policy, kms policy and requestor IAM role, please make them aware of the data.all changes so that they don\'t wipe off data.all related policies'
```
noah-paige, line 305: nit: may want to make this messaging more generic... also should we just be highlighting S3 Dataset resources?

backend/dataall/modules/shares_base/db/share_object_repositories.py
```diff
@@ -42,11 +42,18 @@ def find_share(session, dataset: DatasetBase, env, principal_id, principal_role_
...
+        if share_statues:
+            query = query.filter(ShareObject.status.in_(share_statues))
+
+        return query.all()
 
     @staticmethod
-    def find_share_by_dataset_attributes(session, dataset_uri, dataset_owner, groups=[]):
+    def find_share_by_dataset_attributes(session, dataset_uri, dataset_owner, groups = None):
+        if groups is None:
+            groups = []
```
noah-paige, line 56: I am guessing there is an edge case where groups could be passed in explicitly as `None` before and not being set to `[]` so the query failed?

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
```diff
@@ -0,0 +1,261 @@
...
+        for share in share_dataset_map
+    ]
+
+
+def _get_unhealthy_share_notification(session):
+    unhealthy_share_objects: List[ShareObject] = ShareObjectRepository.get_share_object_with_health_status(
+        session=session, health_status_list=[ShareItemHealthStatus.Unhealthy.value])
+    log.info(f"Found {len(unhealthy_share_objects)} unhealthy share objects")
+    return [
+        NotificationResource(resource=share, resource_type='Share_object', resource_status='Unhealthy',
```
noah-paige, line 66: nit `Share Object` resource type above an `Share_object` here

noah-paige (commented):
part 1 Review - still have a couple more files on Dataset Table notifications and weekly digest

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
```diff
@@ -0,0 +1,261 @@
...
+                resource_bundle.__getattribute__(resource_bundle_type).append(notification)
+            else:
+                resource_bundle = NotificationResourceBundle()
+                resource_bundle.__getattribute__(resource_bundle_type).append(notification)
+                group_name_to_resource_bundle_map[receiver_group_name] = resource_bundle
+
+
+def send_reminder_email(engine):
+    task_exceptions = []
+    resources_type_tuple = ()
```
noah-paige, line 124: this should be a List to represent a list of tuples right?

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
```diff
@@ -0,0 +1,261 @@
...
+                (unhealthy_environment_notification_resources, "environment_object_notifications"))
+
+            # For each notification resource ( i.e. share notification, dataset notification, etc ),
+            # function _map_groups_to_resource_bundles maps each team name : resource bundle
+            # Equivalent to calling
+            # _map_groups_to_resource_bundles(list_of_notifications=pending_share_notification_resources,
+            #                                 resource_bundle_type="share_object_notifications")
+            # _map_groups_to_resource_bundles(list_of_notifications=unhealthy_share_objects_notification_resources,
+            #                                 resource_bundle_type="share_object_notifications") ....
+
```
noah-paige, line 153: nit: remove

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
```diff
@@ -0,0 +1,261 @@
...
+                error_logs=task_exceptions,
+                process_name="Weekly reminder task"
+            )
+
+
+def _construct_email_body(resource_bundle: NotificationResourceBundle):
+    msg_heading = """
+    Dear Team, <br><br>
+    
+    This email contains data.al resources where you need to take some actions. For resources which are in unhealthy state we request you to take actions ASAP so as to minimize any disruptions.<br><br>
```
noah-paige, line 184: nit: typo `data.all`

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
```diff
@@ -0,0 +1,261 @@
...
+    
+    <b>Helpful Tips:</b><br><br>
+    For shares which are in unhealthy state, you can re-apply share by clicking on the "Reapply share" button <br>
+    For environments and datasets which are in unhealthy state, you can go to the AWS account and check the stack associated with that environment/dataset and check the root cause of the stack. Once you address the root cause issue, you can click on "Update Stack" on the Stack Page. <br><br><br>
```
noah-paige, line 188: nit: specify `"Update Stack" on the Stack Page of the data.all resource in the data.all UI`

backend/dataall/modules/s3_datasets/tasks/tables_syncer.py
```diff
@@ -20,60 +22,84 @@
...
-        return processed_tables
+                        table_status_map = DatasetTableService.sync_existing_tables(session, uri=dataset.datasetUri, glue_tables=tables)
+
+                        if table_status_map:
+                            log.info('Sending email notification after dataset table updates were found')
+                            try:
+                                DatasetTableNotifications(dataset=dataset).notify_dataset_table_updates(session=session, table_status_map=table_status_map)
+                            except Exception as e:
+                                error_log = f"Error occurred while sending email to notify about changes to the glue tables for dataset with uri: {dataset.datasetUri} due to: {e}"
+                                task_exceptions.append(error_log)
```
noah-paige, line 63: `table_status_map` will always be defined so this will always execute right?

noah-paige: also should there be a try catch here or only on the loop over datasets?

noah-paige (commented):
some additional comments from pending file review

backend/dataall/modules/catalog/tasks/catalog_indexer_task.py
```diff
@@ -31,7 +32,14 @@ def index_objects(cls, engine, with_deletes='False'):
                     CatalogIndexerTask._delete_old_objects(indexed_object_uris)
                 return len(indexed_object_uris)
         except Exception as e:
+            error_log = f'Error occurred while indexing objects during the cataloging task. Exception: {e}'
+            log.error(error_log)
             AlarmService().trigger_catalog_indexing_failure_alarm(error=str(e))
+            AdminNotificationService().notify_admins_with_error_log(
```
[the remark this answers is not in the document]
TejasRGitHub: Alarm service is hooked up with a SNS topic where there are a lot of notifications ( from grapql lambda, esproxy lambda, etc ) sent from these services for health alerts ( like throttling , etc ) and they make up a lot of notitication. These notifications could be filtered via SNS subscription filters but haven't checked how they are used. 

I will add this another task in stage 2 to check if we can have only one ( combine Admin notifications into Alarm service , etc ) alerting notification system

backend/dataall/modules/notifications/services/admin_notifications.py
```diff
@@ -0,0 +1,32 @@
...
+        subject = f'Data.all alert | Attention Required | Failure in : {process_name}'
+        email_message = f"""
+            Following error occurred - <br><br> {process_error} <br><br>
+        """
+        for error_log in error_logs:
+            email_message += error_log + "<br><br>"
+
+        email_message += "Please check the logs in cloudwatch for more details"
+
+        SESEmailNotificationService.create_and_send_email_notifications(
```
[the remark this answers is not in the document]
TejasRGitHub: I could add UI notifications here but UI notifications are cumbersome to view and understand what error actually happened. It feels a bloated notification on the UI if there are more than 1 error logs ( coming from ECS tasks ). 

You are right in this case in which if the email notification is turned off then there won't be any notifications for failure. I think this might be a good reason to consider using the Alarm service notifications instead of the admin notifications.

backend/dataall/modules/notifications/services/ses_email_notification_service.py
```diff
@@ -60,3 +61,31 @@ def send_email_to_users(email_list, email_provider, message, subject):
...
+        This approach is used while sending email notifications in an ECS task ( e.g. persistent email reminder task, share expiration task, etc )
+        Emails send to groups mentioned in recipient_groups_list and / or emails mentioned in recipient_email_ids
+        """
+        if recipient_groups_list is None:
+            recipient_groups_list = []
+        if recipient_email_ids is None:
+            recipient_email_ids = []
+
+        share_notification_config = config.get_property(
+            'modules.datasets_base.features.share_notifications', default=None
```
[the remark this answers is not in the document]
TejasRGitHub: Yes, all the config related changes will happen in stage 2 of refactor and design changes

backend/dataall/modules/shares_base/services/sharing_service.py
TejasRGitHub: Thanks for pointing this out. I have missed adding this for apply and revoke. Let me add Admin notifications. Won't be calling AdminService notification service for each share item , instead will get all the exception logs and just send one admin notification email

backend/dataall/modules/shares_base/services/sharing_service.py
TejasRGitHub: Changing the variable names. Updated in next PR

backend/dataall/modules/shares_base/services/sharing_service.py
TejasRGitHub: Great catch. I don't intend on sending email for each share object item and only send an email for the share object. I think I might have mixed up some different version of branch her. I will correct it. Thanks for spotting. 

the logic for sending reapply notifications would be 
if reappply_successfull == false : 
    send unhealthy shares notification to the share owner 
else: 
   send healthy share notification

backend/dataall/modules/shares_base/services/sharing_service.py
TejasRGitHub: I want to make this pattern of sending email with the finally block everywhere in the sharing_service.py file . Thanks for spotting this. Let me update the PR with a common pattern.

backend/dataall/modules/shares_base/services/share_notification_service.py
TejasRGitHub: Oh yeah . You are right. I will make changes in this sentence . Thanks

backend/dataall/modules/shares_base/services/share_notification_service.py
TejasRGitHub: Hmm.. You have a good point there. I added this sentence only keeping in mind S3 bucket and glue share items. but I know in the future there could be more and I know there is redshift as a share item. 

I think it would be difficult to cover all the things in a statement when there are lot of share item types ( S3, glue tables, redshift, ... in the future we may have lets say GCP datasets, etc ) 

Maybe I should just remove it from here . Wydt ?

backend/dataall/modules/shares_base/db/share_object_repositories.py
TejasRGitHub: tbh, I added it because my editor was suggesting an improvement. I think this article does a fine job at explaining why - https://nikos7am.com/posts/mutable-default-arguments/

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
TejasRGitHub: Corrected to just have one. For these types of issues in which types are string where there can be such typing error / developer error in stage 2 I will try and make stricter types. Thanks for pointing out

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
TejasRGitHub: Correct. Added the correction and typing hints as well

backend/dataall/modules/s3_datasets/tasks/tables_syncer.py
TejasRGitHub: > table_status_map will always be defined so this will always execute right?

The `table_status_map` will be empty when there are no newly added tables, no table has been deleted , no table which was deleted previously has been added. And thus when it is empty the if part won't execute.

> also should there be a try catch here or only on the loop over datasets?

I have added this try catch explicitly so that that any email sending failures are caught. Also with this , the rest of the code deosn't get affected and I don't want it to get affected

backend/dataall/modules/shares_base/services/share_notification_service.py
TejasRGitHub: Removing this note about terraform and IaC as it would get difficult to generalize as we think about more and more share items ( redshift tbls, glue tables, folders, etc )

TejasRGitHub:
Hi @noah-paige , I have updated the PR. Please let me know if you have any additional code comments. Let me know if you need any clarification / discussion on few of my replies on your comments . 

Few additional things which I noticed and added as a part of new updated PR, 

1. Admin group config ( admin_notifications ) in config.json to enable/disable admin email notifications. I am only keeping email admin notifications and not creating UI notifications - as UI notifications will look bloated with a lot of content. 
2. Added try..catch in sharing_services file where missing and reformatted to have the same pattern of try except finally block across the file

backend/dataall/core/groups/db/constants.py
noah-paige, line 1: nit: should we also explicitly import `db` in the` __init__` file? (at `dataall/backend/dataall/core/groups/__init__.py`) -- think right now it is implicitly imported via `api`

backend/dataall/modules/notifications/services/admin_notifications.py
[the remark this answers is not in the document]
noah-paige: ack - maybe SNS should just be a different transport mechanism in Stage 2 of the refactor (so we would have UI, SNS, and email currently)

backend/dataall/core/groups/db/constants.py
TejasRGitHub: I wouldn't mind it getting imported implicitly. With a blank init all the group_models and constants file will be imported.

backend/dataall/modules/notifications/services/admin_notifications.py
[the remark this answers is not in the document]
TejasRGitHub: Yeah . Let me add that into the stage 2

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -598,6 +599,11 @@ def handle_share_failure(self, error: Exception) -> bool:
         S3ShareAlarmService().trigger_s3_bucket_sharing_failure_alarm(
             self.target_bucket, self.share, self.target_environment
         )
+        AdminNotificationService().notify_admins_with_error_log(
+            process_error=f'Error occurred while processing s3 bucket share request for share with uri: {self.share.shareUri}',
+            process_name='s3 bucket share processor',
```
petrkalos, line 604: nit: I'd probably use the classname `self.__class__.__name__` to make errors easier to trace in the codebase

backend/dataall/modules/shares_base/db/share_object_repositories.py
```diff
@@ -42,11 +42,18 @@ def find_share(session, dataset: DatasetBase, env, principal_id, principal_role_
         )
 
     @staticmethod
-    def find_dataset_shares(session, dataset_uri):
-        return session.query(ShareObject).filter(ShareObject.datasetUri == dataset_uri).all()
+    def find_dataset_shares(session, dataset_uri: str, share_statues: List[str] = None):
```
petrkalos, line 45: `share_statuses` 🗽🗽🗽

backend/dataall/modules/shares_base/db/share_object_repositories.py
```diff
@@ -42,11 +42,18 @@ def find_share(session, dataset: DatasetBase, env, principal_id, principal_role_
...
+        if share_statues:
+            query = query.filter(ShareObject.status.in_(share_statues))
+
+        return query.all()
 
     @staticmethod
-    def find_share_by_dataset_attributes(session, dataset_uri, dataset_owner, groups=[]):
+    def find_share_by_dataset_attributes(session, dataset_uri, dataset_owner, groups=None):
+        if groups is None:
+            groups = []
```
petrkalos, line 56: Is that to protect the case that someone will pass `None` and override the empty list?

backend/dataall/modules/shares_base/db/share_object_repositories.py
```diff
@@ -194,6 +201,16 @@ def get_share_data_items_by_type(session, share, share_type_model, share_type_ur
             query = query.filter(ShareObjectItem.healthStatus == healthStatus)
         return query.all()
 
+    @staticmethod
+    def get_share_object_with_health_status(session, health_status_list: List[str] = None):
+        query = (
+            session.query(ShareObject)
+            .join(ShareObjectItem, ShareObjectItem.shareUri == ShareObject.shareUri)
+            .filter(ShareObjectItem.healthStatus.in_(health_status_list))
```
petrkalos, line 209: won't that break if we call it with `health_status_list=None`? isn't it better to pass an empty list?

backend/dataall/modules/shares_base/db/share_object_repositories.py
```diff
@@ -194,6 +201,16 @@ def get_share_data_items_by_type(session, share, share_type_model, share_type_ur
             query = query.filter(ShareObjectItem.healthStatus == healthStatus)
         return query.all()
 
+    @staticmethod
+    def get_share_object_with_health_status(session, health_status_list: List[str] = None):
```
petrkalos, line 205: This and  find_dataset_shares are named very similarly and it's hard to tell how they are different. Would a name like `find_share_object_with_item_health_status()` describes it better?

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -56,82 +59,100 @@ def approve_share(cls, engine: Engine, share_uri: str) -> bool:
...
-
-            resources = [(share_data.dataset.datasetUri, share_data.dataset.__tablename__)]
-            resources.append(
-                (share_data.share.principalId, ConsumptionRole.__tablename__)
-                if share_data.share.principalType == PrincipalType.ConsumptionRole.value
-                else (
-                    f'{share_data.share.principalId}-{share_data.share.environmentUri}',
-                    EnvironmentGroup.__tablename__,
+        service_exceptions = []
+        share_successful = True
```
petrkalos, line 63: Do we need both? Can we assume that if there are service_errors the share was not successful?

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -152,87 +173,106 @@ def revoke_share(cls, engine: Engine, share_uri: str) -> bool:
...
+            service_exceptions.append(str(e))
+        finally:
+            if not revoke_successful:
+                ShareNotificationService(
+                    session=session, dataset=share_data.dataset, share=share_data.share
+                ).notify_share_object_failed()
+            if len(service_exceptions) > 0:
+                AdminNotificationService().notify_admins_with_error_log(
+                    process_error=f'Error occurred while revoking share with uri: {share_uri}',
+                    process_name='Sharing Service',
```
petrkalos, line 271: nit: maybe class name

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -152,87 +173,106 @@ def revoke_share(cls, engine: Engine, share_uri: str) -> bool:
...
+                    service_exceptions.append(str(e))
+                finally:
+                    existing_pending_items = ShareStatusRepository.check_pending_share_items(session, share_uri)
+                    if existing_pending_items:
+                        new_share_state = share_sm.run_transition(ShareObjectActions.FinishPending.value)
+                    else:
+                        new_share_state = share_sm.run_transition(ShareObjectActions.Finish.value)
+                    share_sm.update_state(session, share_data.share, new_share_state)
+
+        except Exception as e:
```
petrkalos, line 259: should you rethrow the exception as in the original code?

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -254,27 +294,50 @@ def verify_share(
...
+                        shareable_items = ShareObjectRepository.get_share_data_items_by_type(
+                            session,
+                            share_data.share,
+                            processor.shareable_type,
+                            processor.shareable_uri,
+                            status=status,
+                            healthStatus=healthStatus,
+                        )
+                        if shareable_items:
+                            health_status = processor.Processor(
```
petrkalos, line 314: you don't use the list so it might as well be a boolean `shares_health_status &= processor....`

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -295,64 +358,98 @@ def reapply_share(cls, engine: Engine, share_uri: str) -> bool:
...
+                except ResourceLockTimeout as timeout_exception:
+                    ShareStatusRepository.update_share_item_health_status_batch(
+                        session,
+                        share_uri,
+                        old_status=ShareItemHealthStatus.PendingReApply.value,
+                        new_status=ShareItemHealthStatus.Unhealthy.value,
+                        message=str(timeout_exception),
+                    )
+                    code_exception_list.append(str(timeout_exception))
+        except Exception as e:
```
petrkalos, line 435: should you raise the exception?

backend/dataall/modules/notifications/services/admin_notifications.py
```diff
@@ -0,0 +1,42 @@
...
+class AdminNotificationService:
+    """
+    Send email notifications to Admin Group i.e. DAAdministrators in data.all
+    Args -
+        1. process_error - string describing in short the error / exception details
+        2. error_logs - List of all the exception error logs
+        3. process_name - Code where the exception occurred. Example, inside an ECS task like cataloging task, etc or inside a graphql service
+    """
+
+    @classmethod
```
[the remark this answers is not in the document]
TejasRGitHub: Yes. Will do. Now it makes sense to have it as a static method. Earlier I have few class variables and it made sense to make it classmethod at that time, but now it should be staticmethod. Thanks for pointing that out @petrkalos

backend/dataall/modules/notifications/services/ses_email_notification_service.py
```diff
@@ -60,3 +69,30 @@ def send_email_to_users(email_list, email_provider, message, subject):
...
+    def create_and_send_email_notifications(subject, msg, recipient_groups_list=None, recipient_email_ids=None):
+        """
+        Method to directly send email notification instead of creating an SQS Task
+        This approach is used while sending email notifications in an ECS task ( e.g. persistent email reminder task, share expiration task, etc )
+        Emails send to groups mentioned in recipient_groups_list and / or emails mentioned in recipient_email_ids
+        """
+        if recipient_groups_list is None:
+            recipient_groups_list = []
+        if recipient_email_ids is None:
+            recipient_email_ids = []
```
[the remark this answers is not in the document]
TejasRGitHub: @petrkalos , if I put `[ ]` in the method definition my IDE complains and suggets me to the way it is right now. Here's an article I found for why setting lists to None as default vs [] for a list - https://nikos7am.com/posts/mutable-default-arguments/

backend/dataall/modules/notifications/services/ses_email_notification_service.py
```diff
@@ -60,3 +69,30 @@ def send_email_to_users(email_list, email_provider, message, subject):
...
+
+        if share_notification_config := config.get_property(
+            'modules.datasets_base.features.share_notifications', default=None
+        ):
+            for share_notification_config_type in share_notification_config.keys():
+                n_config = share_notification_config[share_notification_config_type]
```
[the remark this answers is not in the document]
TejasRGitHub: Thanks for the suggestions. Made the code changes in the next update. I will further make it more clean when I refactor for the stage 2 design once I get to it.

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
```diff
@@ -0,0 +1,278 @@
...
+from dataall.modules.shares_base.services.shares_enums import ShareItemHealthStatus
+
+log = logging.getLogger(__name__)
+
+"""
+A container for holding the resource ( e.g. dataset, share object, environment, etc ), receivers and health status ( resource_status ) for sending notifications
+"""
+
+
+class NotificationResource:
```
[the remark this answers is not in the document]
TejasRGitHub: Sure. This is a much better way of representation. Thanks for the suggestion @petrkalos

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
```diff
@@ -0,0 +1,278 @@
...
+    def __init__(self, resource, resource_type: str, resource_status: str, receivers: List[str] = None):
+        self.resource = resource
+        self.resource_type = resource_type
+        self.resource_status = resource_status
```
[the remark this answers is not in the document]
TejasRGitHub: Sorry, I didn't get this comment. Could you please elaborate on this comment

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
```diff
@@ -0,0 +1,278 @@
...
+
+    return table
+
+
+if __name__ == '__main__':
+    log.info('Starting weekly reminders task')
+    load_modules(modes={ImportMode.SHARES_TASK})
+    ENVNAME = os.environ.get('envname', 'dkrcompose')
+    ENGINE = get_engine(envname=ENVNAME)
+    group_name_to_resource_bundle_map: Dict[str, NotificationResourceBundle] = {}
```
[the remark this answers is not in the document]
TejasRGitHub: Made global. Thanks!

backend/dataall/modules/omics/tasks/omics_workflows_fetcher.py
```diff
@@ -16,48 +17,56 @@
...
-            log.info(f"Processing workflow name={workflow['name']}, id={workflow['id']}...")
-            existing_workflow = OmicsRepository(session).get_workflow_by_id(workflow['id'])
-            if existing_workflow is not None:
-                log.info(
-                    f"Workflow name={workflow['name']}, id={workflow['id']} has already been registered in database. Updating information..."
-                )
-                existing_workflow.name = workflow['name']
-                existing_workflow.label = workflow['name']
-                session.commit()
+    try:
```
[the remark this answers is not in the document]
TejasRGitHub: @petrkalos , The method `fetch_omics_workflows` is called inside the ECS task. For any exception which will occure ( without the try except ) block , yes it will bubble to the called of thie method ( the main function ) and log the exception and crash, 
Since this is the scenario, in which it is called in this ECS task, I am just wrapping it with a try..catch which won't change the execution logic / cause bugs but will just catch the exception and notify the admins. 

You are right that in case of exception , None will be returned. Function - `fetch_omics_workflows`'s return were previously not used though so I thought it would be okay. Do you think I should also raise the same exception to keep the signature same for now ? 

I will definitely double check all the tasks where I have added a try catch surrounding the function to check if there is change

backend/dataall/modules/omics/tasks/omics_workflows_fetcher.py
```diff
@@ -16,48 +17,56 @@
 
 def fetch_omics_workflows(engine):
```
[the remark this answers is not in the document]
TejasRGitHub: I really like this and had thought of creating a decorator to catch exception and such a decorator on each task which would have considerably reduced the use  of try.. catch. One of the things which I wanted to achieve was to get all the task exceptions and then send out an email with all of them combined. For this , I wanted to collect each and every exception occurring on that task function execution. If using a decorator, the level 1 exception would have been caught along with exceptions happening in the deeper level.  Please do correct me if I am wrong and let me know if I can still achieve this using decorators 

```python 
def task (): .... # Level1 
     get some value 
    try:  .... Level 2
         do something 
    except Exception e
````

File: backend/dataall/modules/s3_datasets/tasks/tables_syncer.py [hunk 36]
```diff
@@ -20,60 +22,85 @@
 
 
 def sync_tables(engine):
```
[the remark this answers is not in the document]
TejasRGitHub: Sure. Will do

backend/dataall/modules/s3_datasets_shares/aws/s3_client.py
```diff
@@ -173,6 +173,8 @@ def create_bucket_policy(self, bucket_name: str, policy: str, fix_malformed_prin
                 else:
                     log.error(f'Failed to create bucket policy. MalformedPolicy: {policy}')
                     raise e
+            else:
+                raise e
```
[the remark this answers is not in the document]
TejasRGitHub: Yes, I added this raise e , as it was missing here. When a bucket's access is removed from pivot role, then the share of S3 bucket should fail but it wasn't because this was missing. 

Suggestion noted and I will make an update for the same.

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -260,6 +261,11 @@ def process_approved_shares(self) -> bool:
                         ShareStatusRepository.update_share_item_health_status(
                             self.session, share_item, ShareItemHealthStatus.Unhealthy.value, str(e), datetime.now()
                         )
+                AdminNotificationService().notify_admins_with_error_log(
+                    process_error='Error occurred while processing redshift table share request',
+                    process_name='redshift table share processor',
+                    error_logs=[str(e)],
```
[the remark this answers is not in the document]
TejasRGitHub: Sure. Will do . I think reusing the same will be more helpful.

backend/dataall/modules/s3_datasets_shares/services/share_managers/lf_share_manager.py
```diff
@@ -654,6 +660,34 @@ def handle_revoke_failure(
...
+        self,
+        error: Exception,
+    ) -> True:
+        """
+        Handles share failure by raising an alarm to alarmsTopic
+        :param table: DatasetTable
+        :param error: share error
+        :return: True if alarm published successfully
+        """
+        logger.error(
```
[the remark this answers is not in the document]
TejasRGitHub: Are you suggesting it for all the logger.errors or just this one ?

backend/dataall/modules/shares_base/db/share_object_repositories.py
TejasRGitHub: Corrected in the next update

backend/dataall/modules/shares_base/db/share_object_repositories.py
TejasRGitHub: Its for this reason - https://nikos7am.com/posts/mutable-default-arguments/
My IDE was suggesting to make this change

backend/dataall/modules/shares_base/db/share_object_repositories.py
TejasRGitHub: You are right. Will make the correction. It will certainly return incorrect results if not break

backend/dataall/modules/shares_base/services/sharing_service.py
TejasRGitHub: Yes. the share_successful is tracking the status of process returned by the share processors. It is possible that the share processors might just return False but there were not exceptions . In that case, the share is still unsuccessful but with no exception. 

I could certainly combine service_exceptions and share_successful in some unusual nasty way but I think it will make it difficult to read code and track it. 

With that said, I think the nesting which is occurring due to all the try catches can be reduced but this would required considerable refactoring and testing. I am open to any suggestions regarding this, Let me know

backend/dataall/modules/shares_base/services/sharing_service.py
TejasRGitHub: I like the idea of having the class name as the process name . Let me check and replace it whereever possible

backend/dataall/modules/shares_base/services/sharing_service.py
TejasRGitHub: In the original code as well, the exception was not thrown. Can you please let me know if I am missing something

backend/dataall/modules/shares_base/services/sharing_service.py
TejasRGitHub: Not raising an exception and handling it here in the service layer itself. The caller of this method will get to know if this failed or not failed based on if this function returns True/False.

backend/dataall/modules/notifications/services/admin_notifications.py
```diff
@@ -0,0 +1,42 @@
...
+    def notify_admins_with_error_log(cls, process_error: str, error_logs: List[str], process_name: str = ''):
+        if (
+            config.get_property(
+                'modules.datasets_base.features.share_notifications.email.parameters.admin_notifications', default=False
+            )
+            is False
+        ):
+            log.info('Admin notifications are switched off')
+            return
```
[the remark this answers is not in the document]
TejasRGitHub: Created a decorator which can also be used later  at other places .

backend/dataall/modules/shares_base/db/share_object_repositories.py
TejasRGitHub: Corrected this. Thanks for pointing this out

backend/dataall/modules/s3_datasets_shares/services/share_managers/lf_share_manager.py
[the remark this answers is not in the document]
petrkalos: We need to evaluate it case by case. Sometimes having stacktraces everywhere might be annoying but sometimes (like this) when the error is printed by a utility function that is being called from multiple places then stack trace might be necessary in order to figure out what caused it.

backend/dataall/modules/shares_base/services/sharing_service.py
petrkalos: in the original code if there was an exception between lines `[156,177]` if would have bubble up. With your change it will be handled in the block [259,262]

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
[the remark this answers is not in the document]
TejasRGitHub: Created enums for both - resource_type and resource_status. Updated code in next push

backend/dataall/modules/omics/tasks/omics_workflows_fetcher.py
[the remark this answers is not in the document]
TejasRGitHub: @petrkalos , I have raised exception at places where the exceptions were naturally raise ( i.e. before the try..catch block ) . For places where the function returned either none Or True/False, I have kept the functionality the same.

backend/dataall/modules/omics/tasks/omics_workflows_fetcher.py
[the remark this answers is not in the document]
TejasRGitHub: Keeping it as how it is right now as this would required changes at many places and also required some research into if nested task_exceptions can be collected and pushed to a top level decorator.

backend/dataall/modules/s3_datasets_shares/services/share_managers/lf_share_manager.py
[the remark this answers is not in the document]
TejasRGitHub: Yes. I agree. I have added logger.exception for any internal ( for -loop , or internal if -else block ) try catch block where understanding the stack trace would help debugging. For any top level try catch that I introduced,I am keeping log.error() 

Please note, I am trying to cover most of the places where log.exception would make sense. There will still be places where log.error is used and log.exception would be more prudent there but I will keep this for some other refactoring PR and not combine it here

backend/dataall/modules/shares_base/db/share_object_repositories.py
TejasRGitHub: using the name `list_share_objects_with_item_health_status()`

backend/dataall/modules/shares_base/db/share_object_repositories.py
```diff
@@ -42,11 +42,15 @@ def find_share(session, dataset: DatasetBase, env, principal_id, principal_role_
         )
 
     @staticmethod
-    def find_dataset_shares(session, dataset_uri):
-        return session.query(ShareObject).filter(ShareObject.datasetUri == dataset_uri).all()
+    def find_dataset_shares(session, dataset_uri: str):
```
TejasRGitHub, line 45: Nothing has changed here. Only returning via query.allI() instead of on the same line

backend/dataall/modules/shares_base/db/share_object_repositories.py
TejasRGitHub: Removed the default value on the input argument argument. Will let it fail or return empty list is None is provided

backend/dataall/modules/shares_base/services/share_notification_service.py
```diff
@@ -332,30 +437,3 @@ def _create_notification_task(self, subject, msg):
                     log.info(f'Notification type : {share_notification_config_type} is not active')
         else:
             log.info('Notifications are not active')
-
```
TejasRGitHub, line 335: Removing this function from here and moving it to `ses_email_notification_service.py`

config.json
```diff
@@ -32,7 +32,8 @@
                         "active": false,
                         "persistent_reminders": false,
                         "parameters": {
-                            "group_notifications": true
+                            "group_notifications": true,
+                            "admin_notifications": true
```
TejasRGitHub, line 36: Adding this new parameter to switch on/off admin notifications. This can be later removed and admin notifications can be enabled by default when email notifications are enabled. Keeping this for now, to test and see how many notifications an admin might get with the current code change. There is a chance in which admins can be bombarded with a lot of notifications in which this will serve as a good way to stop creating admin notifications

TejasRGitHub:
Hi @petrkalos , @noah-paige , I have made code changes after taking a look at the review comments. Can you please take a look and let me if this PR looks good now 

I have re-tested after code changes 

1. Notifications after share is failed ( email + UI ) ✅ 
2. Notifications after share is Unhealthy ( email + UI ) ✅ 
3. Notifications after share is healthy ( email + UI ) ✅ 
4. Weekly digest reminder notifications  ( email ) ✅ 
5. Notifications to owner ( when table are deleted, added, synced ) and share owners ( if there exists a share on the table ) - ✅ 
6. Notifications when ECS or any service fails to Admins ✅

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -254,27 +294,49 @@ def verify_share(
...
-                except Exception as e:
-                    log.error(f'Error occurred during share verifying of {type.value}: {e}')
-
-        return True
+                    ShareNotificationService(
+                        session=session, dataset=share_data.dataset, share=share_data.share
+                    ).notify_share_object_items_unhealthy()
+        except Exception as e:
+            log.error(f'Unexpected error occurred while verifying share with uri: {share_uri} due to: {e}')
+            service_exceptions.append(str(e))
```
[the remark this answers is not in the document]
noah-paige: thinking again on this I guess it makes sense only to capture unhealthy responses from Processor since the notification is on health status  (unhealthy/healthy) rather than if the process was successful or failed - is this how you were thinking?

backend/dataall/modules/s3_datasets_shares/services/share_processors/glue_table_share_processor.py
```diff
@@ -3,6 +3,7 @@
 from warnings import warn
 from datetime import datetime
 from dataall.core.environment.services.environment_service import EnvironmentService
+from dataall.modules.notifications.services.admin_notifications import AdminNotificationService
```
noah-paige, line 6: nit: unused import

backend/dataall/modules/s3_datasets/services/dataset_table_notifications.py
```diff
@@ -0,0 +1,90 @@
...
+        <tr>    
+            <th align='center'>Glue Table Name</th>
+            <th align='center'>Status</th>
+        </tr>
+        """
+        table_body = """"""
+        for dataset_table, dataset_table_details in dataset_table_status_map.items():
+            table_body += f"""
+                <tr>
+                    <td align='center'>{dataset_table.GlueTableName}</td>
```
noah-paige, line 80: can we make sure we use `table.label` - the `GlueTableName` should not be used anywhere for end users who may not have access to the table already

noah-paige (commented):
thanks for all the work here Tejas - mostly nits from my end. 

Mainly can we make sure to use table label vs GlueTableName comment and remove the unused imports - then I can approve

backend/dataall/modules/shares_base/services/sharing_service.py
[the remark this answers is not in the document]
TejasRGitHub: `thinking again on this I guess it makes sense only to capture unhealthy responses from Processor since the notification is on health status (unhealthy/healthy) rather than if the process was successful or failed - is this how you were thinking?`
Yes you are right. For unexpected error in the share process, informing Admins makes more sense to me

backend/dataall/modules/shares_base/tasks/share_reapplier_task.py
```diff
@@ -17,37 +19,41 @@
...
+        share_object: ShareObject
+        processed_share_objects = []
+        task_exceptions = []
+        for share_object in share_objects:
+            try:
                 log.info(
-                    f'Re-applying Share Items for Share Object with Requestor: {share_object.principalId} on Target Dataset: {share_object.datasetUri}'
+                    f'Re-applying Share Items for Share Object, Share URI: {share_object.shareUri} ) with Requestor: {share_object.principalId} on Target Dataset: {share_object.datasetUri}'
                 )
                 processed_share_objects.append(share_object.shareUri)
```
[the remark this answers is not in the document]
TejasRGitHub: Yeah sure. Will do

backend/dataall/modules/omics/tasks/omics_workflows_fetcher.py
```diff
@@ -16,48 +17,57 @@
...
+                        owner=env.environmentUri,
+                    )
+                    OmicsRepository(session).save_omics_workflow(omicsWorkflow)
+        return True
+    except Exception as e:
+        log.error(f'Error occurred while processing omics workflow task due to: {e}')
+        AdminNotificationService().notify_admins_with_error_log(
+            process_error='Error occurred while processing omics workflow task',
+            error_logs=[str(e)],
+            process_name='Omics Workflow task',
```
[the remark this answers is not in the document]
TejasRGitHub: Since the method `fetch_omics_workflows` isn't wrapped under a class, I am using this label instead of class name

backend/dataall/base/feature_toggle_checker.py
```diff
@@ -56,3 +59,20 @@ def decorated(*args, **kwargs):
...
+
+        @functools.wraps(fn)
+        def decorated(*args, **kwargs):
+            value = config.get_property(config_property, default_value)
+            if not value:
+                log.info(f'Config - {config_property} is inactive')
+                return
+            return fn(*args, **kwargs)
+
+        return fn_decorator(decorated)
```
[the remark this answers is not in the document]
TejasRGitHub: Yes you are right . The method `is_feature_enabled `throws exception when the config is disabled which I didn't want to happen. Also the `is_feature_enabled` doesn't use default values. Instead of modifiying the existing method and make more changes I preferred to make this method.

backend/dataall/modules/omics/tasks/omics_workflows_fetcher.py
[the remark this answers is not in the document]
noah-paige: that is fine - see other occurrences you do the same if not wrapped in class

backend/dataall/base/feature_toggle_checker.py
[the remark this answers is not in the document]
noah-paige: think it is fine for now - maybe can think on way to consolidate LOC

backend/dataall/modules/s3_datasets/services/dataset_table_notifications.py
TejasRGitHub: @noah-paige , I didn't get what you mean by not having access to the table already. Could you please clarify more on that

noah-paige: we did a good amount of refactoring to ensure that we are not exposing any AWS Account related technical details on the shareable resources (including Glue Tables) in an earlier PR - see https://github.com/data-dot-all/dataall/blob/f5fbcfd3876d46ea477df2284746ab24d33b41f7/backend/dataall/modules/s3_datasets/api/table/types.py#L25

For the above let's use the table name (i.e. `label`) - this is typically the same as the technical Glue Table name but in the case they diverge we will still be protecting the appropriate parameter

TejasRGitHub: Oh I see. Got it.

TejasRGitHub: Made changes in the next PR update

TejasRGitHub:
@noah-paige, Updated PR and made minor changes as per the review comments

TejasRGitHub:
> @TejasRGitHub - looks like a test is failing and need to format with ruff

Updated PR . Now the ruff tests and the unit test should pass

noah-paige (commented):
Thanks for resolving ruff issues - looks like latest version of ruff had updated the way it formats f-strings causing a number of files to be updated for formatting outside the scope of this PR

backend/dataall/base/feature_toggle_checker.py
```diff
@@ -3,11 +3,14 @@
 """
 
 import functools
+import logging
```
TejasRGitHub, line 6: This file is a part of PR

backend/dataall/core/groups/db/constants.py
```diff
@@ -0,0 +1,2 @@
+class DataallGroups:
```
TejasRGitHub, line 1: This file is a part of PR

backend/dataall/modules/catalog/tasks/catalog_indexer_task.py
```diff
@@ -8,6 +8,7 @@
 from dataall.base.db import get_engine
 from dataall.base.loader import load_modules, ImportMode
 from dataall.base.utils.alarm_service import AlarmService
+from dataall.modules.notifications.services.admin_notifications import AdminNotificationService
```
TejasRGitHub, line 11: This file is a part of PR

backend/dataall/modules/notifications/services/admin_notifications.py
```diff
@@ -0,0 +1,35 @@
+import logging
```
TejasRGitHub, line 1: This file is a part of PR

backend/dataall/modules/notifications/services/ses_email_notification_service.py
```diff
@@ -37,19 +38,29 @@ def send_email_task(subject, message, recipient_groups_list, recipient_email_lis
         email_provider = SESEmailNotificationService.get_email_provider_instance(
             recipient_groups_list, recipient_email_list
         )
+        identityProvider = ServiceProviderFactory.get_service_provider_instance()
```
TejasRGitHub, line 41: This file is a part of PR

backend/dataall/modules/notifications/tasks/weekly_digest_reminder.py
```diff
@@ -0,0 +1,330 @@
+import logging
+import os
+from dataclasses import dataclass, field
+from typing import List, Dict, Any, Tuple, Set
+
```
TejasRGitHub, line 5: This file is a part of PR

backend/dataall/modules/notifications/tasks/weekly_notification_enums.py
```diff
@@ -0,0 +1,20 @@
+from enum import Enum
+
```
TejasRGitHub, line 2: This file is a part of PR

backend/dataall/modules/omics/tasks/omics_workflows_fetcher.py
```diff
@@ -5,6 +5,7 @@
 
 from dataall.core.environment.db.environment_models import Environment
```
TejasRGitHub, line 6: This file is a part of PR

backend/dataall/modules/redshift_datasets_shares/services/redshift_table_share_processor.py
```diff
@@ -2,6 +2,7 @@
 from datetime import datetime
 from typing import List
 from dataall.base.utils.naming_convention import NamingConventionService, NamingConventionPattern
+from dataall.modules.notifications.services.admin_notifications import AdminNotificationService
 from dataall.modules.shares_base.services.sharing_service import ShareData
```
TejasRGitHub, line 6: This file is a part of PR

backend/dataall/modules/s3_datasets/db/dataset_table_repositories.py
```diff
@@ -1,5 +1,7 @@
 import logging
+from dataclasses import dataclass, field
 from datetime import datetime
+from typing import Dict, List
```
TejasRGitHub, line 4: This file is a part of PR

backend/dataall/modules/s3_datasets/services/dataset_table_notifications.py
```diff
@@ -0,0 +1,90 @@
+from typing import Dict, List
+import logging
+from dataall.modules.notifications.services.ses_email_notification_service import SESEmailNotificationService
+from dataall.modules.s3_datasets.db.dataset_models import S3Dataset, DatasetTable
+from dataall.modules.s3_datasets.db.dataset_table_repositories import DatasetTableShareDetails
+from dataall.modules.shares_base.db.share_object_repositories import ShareObjectRepository
+
+log = logging.getLogger(__name__)
```
TejasRGitHub, line 8: This file is a part of PR

backend/dataall/modules/s3_datasets/services/dataset_table_service.py
```diff
@@ -1,12 +1,14 @@
 import logging
+from typing import Dict
+
```
TejasRGitHub, line 3: This file is a part of PR

backend/dataall/modules/s3_datasets/tasks/tables_syncer.py
```diff
@@ -2,13 +2,17 @@
...
+from typing import Dict, List
 
 from dataall.base.aws.sts import SessionHelper
 from dataall.core.environment.db.environment_models import Environment, EnvironmentGroup
 from dataall.core.environment.services.environment_service import EnvironmentService
 from dataall.base.db import get_engine
+from dataall.modules.notifications.services.admin_notifications import AdminNotificationService
 from dataall.modules.s3_datasets.aws.glue_dataset_client import DatasetCrawler
 from dataall.modules.s3_datasets.aws.lf_table_client import LakeFormationTableClient
+from dataall.modules.s3_datasets.db.dataset_table_repositories import DatasetTableShareDetails
```
TejasRGitHub, line 14: This file is a part of PR

backend/dataall/modules/s3_datasets_shares/aws/s3_client.py
```diff
@@ -163,16 +163,15 @@ def create_bucket_policy(self, bucket_name: str, policy: str, fix_malformed_prin
...
-                if fix_malformed_principals:
-                    log.info('MalformedPolicy. Lets try again')
-                    fixed_policy = SharePolicyVerifier.remove_malformed_principal(
-                        policy, DATAALL_BUCKET_SIDS, self._account_id, self.region
-                    )
-                    self.create_bucket_policy(bucket_name, fixed_policy, False)
-                else:
-                    log.error(f'Failed to create bucket policy. MalformedPolicy: {policy}')
-                    raise e
+            if e.response['Error']['Code'] == 'MalformedPolicy' and fix_malformed_principals:
```
TejasRGitHub, line 166: This file is a part of PR. This was changed since there was a bug in the code in which it didn't raise exception

backend/dataall/modules/s3_datasets_shares/services/share_managers/lf_share_manager.py
```diff
@@ -9,6 +9,7 @@
 from dataall.base.aws.sts import SessionHelper
 from dataall.base.db import exceptions
 from dataall.core.environment.services.environment_service import EnvironmentService
+from dataall.modules.notifications.services.admin_notifications import AdminNotificationService
```
TejasRGitHub, line 12: This file is a part of PR

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_access_point_share_manager.py
```diff
@@ -6,6 +6,7 @@
 from dataall.core.environment.services.environment_service import EnvironmentService
 from dataall.base.db import utils
 from dataall.base.aws.sts import SessionHelper
+from dataall.modules.notifications.services.admin_notifications import AdminNotificationService
```
TejasRGitHub, line 9: This file is a part of PR

backend/dataall/modules/s3_datasets_shares/services/share_managers/s3_bucket_share_manager.py
```diff
@@ -6,6 +6,7 @@
 from dataall.base.aws.sts import SessionHelper
 from dataall.core.environment.db.environment_models import Environment
 from dataall.core.environment.services.environment_service import EnvironmentService
+from dataall.modules.notifications.services.admin_notifications import AdminNotificationService
```
TejasRGitHub, line 9: This file is a part of PR

backend/dataall/modules/s3_datasets_shares/services/share_processors/glue_table_share_processor.py
```diff
@@ -346,15 +346,13 @@ def process_revoked_shares(self) -> bool:
                         log.info('Deleting target shared database...')
                         manager.delete_shared_database_in_target()
             except Exception as e:
-                log.error(
-                    f'Failed to clean-up database permissions or delete shared database {manager.shared_db_name} '
-                    f'due to: {e}'
-                )
+                manager.handle_revoke_clean_up_failure(error=e)
```
TejasRGitHub, line 349: This file is a part of PR

backend/dataall/modules/s3_datasets_shares/services/share_processors/s3_access_point_share_processor.py
```diff
@@ -172,7 +172,8 @@ def process_revoked_shares(self) -> bool:
 
         return success
 
-    def verify_shares(self) -> bool:
+    def verify_shares_health_status(self) -> bool:
```
TejasRGitHub, line 175: This file is a part of PR

backend/dataall/modules/s3_datasets_shares/services/share_processors/s3_bucket_share_processor.py
```diff
@@ -158,7 +156,8 @@ def process_revoked_shares(self) -> bool:
 
         return success
 
-    def verify_shares(self) -> bool:
+    def verify_shares_health_status(self) -> bool:
```
TejasRGitHub, line 159: This file is a part of PR

backend/dataall/modules/shares_base/services/share_notification_service.py
```diff
@@ -24,6 +25,9 @@ class DataSharingNotificationType(enum.Enum):
     SHARE_OBJECT_EXTENSION_REJECTED = 'SHARE_OBJECT_EXTENSION_REJECTED'
     SHARE_OBJECT_REJECTED = 'SHARE_OBJECT_REJECTED'
     SHARE_OBJECT_PENDING_APPROVAL = 'SHARE_OBJECT_PENDING_APPROVAL'
+    SHARE_OBJECT_FAILED = 'SHARE_OBJECT_FAILED'
```
TejasRGitHub, line 28: This file is a part of PR

backend/dataall/modules/shares_base/services/share_processor_manager.py
```diff
@@ -19,7 +19,7 @@ def process_revoked_shares(self) -> bool:
         ...
 
     @abstractmethod
-    def verify_shares(self) -> bool:
+    def verify_shares_health_status(self) -> bool:
```
TejasRGitHub, line 22: This file is a part of PR

backend/dataall/modules/shares_base/services/sharing_service.py
```diff
@@ -1,14 +1,17 @@
 import logging
 from dataclasses import dataclass
 
-from typing import Any
+from typing import Any, List
```
TejasRGitHub, line 4: This file is a part of PR

backend/dataall/modules/shares_base/tasks/persistent_email_reminders_task.py
```diff
@@ -17,20 +18,34 @@ def persistent_email_reminders(engine):
...
-        pending_shares = ShareObjectRepository.fetch_submitted_shares_with_notifications(session=session)
-        log.info(f'Found {len(pending_shares)} pending shares')
-        pending_share: ShareObject
-        for pending_share in pending_shares:
-            log.info(f'Sending Email Reminder for Share: {pending_share.shareUri}')
-            share = ShareObjectRepository.get_share_by_uri(session, pending_share.shareUri)
-            dataset = DatasetBaseRepository.get_dataset_by_uri(session, share.datasetUri)
-            ShareNotificationService(session=session, dataset=dataset, share=share).notify_persistent_email_reminder(
-                email_id=share.owner
+    task_exceptions = []
```
TejasRGitHub, line 21: This file is a part of PR

backend/dataall/modules/shares_base/tasks/share_expiration_task.py
```diff
@@ -19,56 +20,70 @@ def share_expiration_checker(engine):
...
-                    log.info(f'Revoking share with uri: {share.shareUri} as it is expired')
-                    # Put all share items in revoke state and then revoke
-                    share_items_to_revoke = ShareObjectRepository.get_all_share_items_in_share(
-                        session, share.shareUri, ['Share_Succeeded']
-                    )
-                    item_uris = [share_item.shareItemUri for share_item in share_items_to_revoke]
-                    revoked_items_states = ShareStatusRepository.get_share_items_states(
-                        session, share.shareUri, item_uris
-                    )
+    task_exceptions = []
```
TejasRGitHub, line 23: This file is a part of PR

backend/dataall/modules/shares_base/tasks/share_manager_task.py
```diff
@@ -2,6 +2,7 @@
 import os
 import sys
 
+from dataall.modules.notifications.services.admin_notifications import AdminNotificationService
```
TejasRGitHub, line 5: This file is a part of PR

backend/dataall/modules/shares_base/tasks/share_reapplier_task.py
```diff
@@ -1,7 +1,9 @@
 import logging
 import os
 import sys
+from typing import List
 
+from dataall.modules.notifications.services.admin_notifications import AdminNotificationService
```
TejasRGitHub, line 6: This file is a part of PR

backend/dataall/modules/shares_base/tasks/share_verifier_task.py
```diff
@@ -19,20 +21,37 @@ def verify_shares(engine):
...
-        log.info(f'Found {len(all_share_objects)} share objects  verify ')
-        share_object: ShareObject
-        for share_object in all_share_objects:
-            log.info(
-                f'Verifying Share Items for Share Object with Requestor: {share_object.principalId} on Target Dataset: {share_object.datasetUri}'
-            )
-            processed_share_objects.append(share_object.shareUri)
-            SharingService.verify_share(
-                engine, share_uri=share_object.shareUri, status=ShareItemStatus.Share_Succeeded.value, healthStatus=None
+    task_exceptions = []
```
TejasRGitHub, line 24: This file is a part of PR

noah-paige (approved):

TejasRGitHub:
> Thanks for resolving ruff issues - looks like latest version of ruff had updated the way it formats f-strings causing a number of files to be updated for formatting outside the scope of this PR

Yes there are abut 60 - 70 files which got formatted 😢 . @petrkalos I have commented on the files which are a part of this PR . You can also view the changes made in the PR update prior to the formatting changes

Issue #1513, opened 2024-09-03
Support for Aurora serverless 2


zsaltys:
Aurora serverless 1 is getting deprecated and replaced by version 2. Data.all should switch to v2 which supports better autoscaling. We also need to figure out an upgrade procedure for those who are already deployed with version 1.

 https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-serverless-v2.upgrade.html#sv1-to-sv2-apg

I couldn't find official AWS EOL notice but I was told so on an AWS ticket and was able to find this: https://www.infoq.com/news/2024/01/aurora-serverless-v1-retirement/#:~:text=We%20are%20reaching%20out%20to,Serverless%20version%201%20(v1).

dlpzx:
Hi @zsaltys , Thanks for opening the ticket, there is actually an issue already opened for this #548

PR #1759, opened 2025-02-02
Aurora upgrade cdk build

SofiaSazonova:
### Feature or Bugfix

- Feature

### Detail
- Depending on CDK parameter CodeBuild project is created to copy data from one AuroraCluster to Another
- Can be used for Aurora upgrade or emergency restore
- To deploy, add to `cdk.json` to `DeploymentEnvironments` section

```
"aurora_migration_enabled": true,
"old_aurora_connection_secret_arn": ...arn...
```
Copy source DB connection secret before deploing! That's going to be `old_aurora_connection_secret_arn`

### Upgrade Recommendations
Once deployed it will create a new v2 cluster with an empty database and a CodeBuild job that when it runs (manual trigger) will copy the data from the old cluster to the new one. Using the CodeBuild job is provided as a helper and is not mandatory to be used. Users can opt to perform the copy using their own scripts.

Suggested steps

1. Communicate with stakeholders a downtime of data.all platform
2. If you want to use the provided CodeBuild migration job you must set the following in your cdk.json
    1. Upon upgrade the db secret will be ovewritten hence you need to make a backup and provide this arn to the config
3. Put data.all into maintenance mode, login as Admin, go to Admin Settings → Maintenance → Pick “NO-ACCESS” → Click “Start Maintenance”
4. If previously you used the potgres version != 13, change the code of the codebuild job: change the line 38 of [deploy/stacks/aurora_migration_task.py](https://github.com/data-dot-all/dataall/pull/1759/files#diff-8998dd662f43e4d3b967edab21060e7fc6e30965f550ddd509868f2790e4c56a)
5. Push the changes in your code repo
6. Wait for the backend deployment to complete
7. Copy from the old Aurora cluster to the new
    1. If using the CodeBuild job from step 2 locate it and start the build. All the requiredinputs will be prepopulated
8. Validate that data.all have all the data, is operating properly and the pipeline is green
9. Stop “maintenance mode”
10. If CodeBuild job was used you can now remove the relevant configs from your cdk.json to delete it.

### IMPORTANT!
1. Do not run the pipeline right away. First, try to create CodeBuild job and v2 cluster manually to test. Once you eliminate v1 cluster from your pipeline, you won't be able to create another v1 cluster again.
2. Do not delete old cluster until you are completely sure.

### Relates
-  #1513 

### Security

  - Is the input sanitized?

### DIFF
--- a/.checkov.baseline
+++ b/.checkov.baseline
@@ -376,19 +376,7 @@
                     ]
                 },
                 {
-                    "resource": "AWS::S3::Bucket.dataalldevfrontend64065639",
-                    "check_ids": [
-                        "CKV_AWS_18"
-                    ]
-                },
-                {
-                    "resource": "AWS::S3::Bucket.dataalldevlogging0F6723EE",
-                    "check_ids": [
-                        "CKV_AWS_18"
-                    ]
-                },
-                {
-                    "resource": "AWS::S3::Bucket.dataalldevuserguide5964DC13",
+                    "resource": "AWS::S3::Bucket.dataalldevcloudfrontaccesslogsCAF85B96",
                     "check_ids": [
                         "CKV_AWS_18"
                     ]
@@ -452,13 +440,7 @@
                     ]
                 },
                 {
-                    "resource": "AWS::S3::Bucket.pipelineartifactsbucketE44F7DE9",
-                    "check_ids": [
-                        "CKV_AWS_18"
-                    ]
-                },
-                {
-                    "resource": "AWS::S3::Bucket.sourcecodebucket464EEFA3",
+                    "resource": "AWS::S3::Bucket.dataallaccesslogsEDA9BC4F",
                     "check_ids": [
                         "CKV_AWS_18"
                     ]
@@ -551,6 +533,12 @@
                         "CKV_AWS_115",
                         "CKV_AWS_116",
                         "CKV_AWS_117"
+                    ]
+                },
+                {
+                    "resource": "AWS::CloudTrail::Trail.S3CloudTrail9B4C955D",
+                    "check_ids": [
+                        "CKV_AWS_35"
                     ]
                 },
                 {
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.pdf filter=lfs diff=lfs merge=lfs -text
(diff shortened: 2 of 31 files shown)


SofiaSazonova:
[Enhanced monitoring billing details](https://aws.amazon.com/rds/faqs/#:~:text=What%20impact%20does%20Enhanced%20Monitoring%20have%20on%20my%20monthly%20bills%3F)

deploy/stacks/aurora_migration_task.py
```diff
@@ -0,0 +1,119 @@
...
+                            ]
+                        },
+                        'build': {
+                            'commands': [
+                                'export PGHOST=$SRC_HOST PGPORT=$SRC_PORT PGUSER=$SRC_USER PGPASSWORD=$SRC_PWD',
+                                '/usr/lib/postgresql/$SRC_PGVER/bin/pg_isready',
+                                '/usr/lib/postgresql/$SRC_PGVER/bin/pg_dump -x -Fc -v > db.dump',
+                                'export PGHOST=$TGT_HOST PGPORT=$TGT_PORT PGUSER=$TGT_USER PGPASSWORD=$TGT_PWD',
+                                '/usr/lib/postgresql/$TGT_PGVER/bin/pg_isready',
+                                '/usr/lib/postgresql/$TGT_PGVER/bin/pg_restore -v -x -O -C -c -d postgres db.dump',
```
petrkalos, line 103: [according to pg_restore docs](https://www.postgresql.org/docs/current/app-pgrestore.html)
* `-c` clean will DROP objects that will be restored
* `-C` will create the database (not sure what happens if it already exists)
* The combination of two flags will DROP and RECREATE the entire db

I am a bit concerned that customers might rerun this job after they have migrated and the two dbs have diverged something that will effectively restore the target db to the point in time of the old db.

Perhaps I am overthinking it but is there anything we can do to prevent this?

I think we need the `-c` because db_migrations trigger func will run after the deployment of the new cluster and will "pollute" the database. But maybe we don't need the `-C` because you already create the default database in CDK. wdyt?

SofiaSazonova: The new cluster is created with DB in it and also, smth (e.g. permissions) is already there. So, it's easier and safer to drop everything and restore. It works the second time as well. 
This code will give us a copy of old DB in new cluster. Thats' exactly what we need.

It's not a 'normal' thing to do, so I think that's the best option.

petrkalos (approved):

petrkalos (approved):

dlpzx (approved):

PR #1792, opened 2025-03-26
Fix/custom auth 500

petrkalos:
### Feature or Bugfix
- Bugfix

### Detail
Currently when access is denied we raise exceptions which have 2 side effects
* APIGW returns 500 Internal Server Error to the user which is confusing
* Deny cache is not populated hence increasing the traffic that IdP is receiving

In this change we return an explicit Deny policy for all known scenarios and raise an exception otherwise.
The known scenarios include
* We receive 401 from Cognito which means invalid access token (unauthorized access)
* We receive 403 from Cognito which means throttling
* We fail to validate the jwt token returned by Cognito

### Relates
- <URL or Ticket>

### Security

  - Is the input sanitized?

### DIFF
--- a/deploy/custom_resources/custom_authorizer/auth_services.py
+++ b/deploy/custom_resources/custom_authorizer/auth_services.py
@@ -24,7 +24,9 @@ class AuthServices:
 
     @staticmethod
     def generate_deny_policy(incoming_resource_str: str):
-        return AuthServices.generate_policy({'sub': None, EMAIL_CLAIM: None}, 'Deny', incoming_resource_str)
+        return AuthServices.generate_policy(
+            {'sub': None, EMAIL_CLAIM: None, USER_ID_CLAIM: None}, 'Deny', incoming_resource_str
+        )
 
     # Generates Policy document containing policy to allow the API invocation for allowed API Endpoints
     # Also attaches the claims which are present in the token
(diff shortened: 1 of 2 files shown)


deploy/custom_resources/custom_authorizer/auth_services.py
```diff
@@ -15,6 +22,10 @@ def generate_resource_str(resource, api_resource_name):
         resource_list[5] = '/'.join(api_gateway_arn)
         return ':'.join(resource_list)
 
+    @staticmethod
+    def generate_deny_policy(incoming_resource_str: str):
+        return AuthServices.generate_policy({'sub': None, EMAIL_CLAIM: None}, 'Deny', incoming_resource_str)
```
noah-paige, line 27: do we need to add USER_ID: None to this dict too? - i.e. `{'sub': None, EMAIL_CLAIM: None, USER_ID: None}`

Otherwise do we run the risk of getting KeyError when updating context as `'user_id': verified_claims[USER_ID_CLAIM],`?

petrkalos: I could yes but currently it's working (I tested it) because both `email` and `user_id` env variables are equal to `email`

noah-paige (approved):

PR #1830, opened 2025-07-02
update node v22

petrkalos:
### Feature or Bugfix
- Feature

### Detail
* drop deprecated node v18 and upgrade to latest LTS v22
* define default codebuild options once and remove per codebuildstep redefinitions


deploy/stacks/pipeline.py
```diff
@@ -201,12 +199,17 @@ def __init__(
...
             code_build_defaults=pipelines.CodeBuildOptions(
                 build_environment=codebuild.BuildEnvironment(
                     environment_variables={
                         'DATAALL_REPO_BRANCH': codebuild.BuildEnvironmentVariable(value=git_branch),
-                    }
-                )
+                    },
+                    build_image=codebuild.LinuxBuildImage.AMAZON_LINUX_2_5,
+                ),
+                partial_build_spec=BuildSpec.from_object(
```
anushka-singh, line 210: Why do we need a partial build spec?

petrkalos: `AMAZON_LINUX_2_5` comes with 3 node versions installed 18,20 and 22 and by default it's using 18 that's why we see all the warnings. With the partial buildspec we force all CodeBuild jobs associated with this pipeline to use 22. See the docs [here](https://docs.aws.amazon.com/codebuild/latest/userguide/runtime-versions.html#:~:text=Custom%20runtime%20buildspec%20example)

TejasRGitHub: In the future, if we move away from Amazon linux 2_5 Or if Amazon Linux starts using a more better version maybe node 23 or something then we'd have to update this. Should we deprecate using explicitly mentioning node js version after maybe 2.10 or 2.11 ?

PR #1850, opened 2025-07-31
upgrade python to 3.12

### LINKED ISSUE #1842: Python 3.9 EOL Oct 30th 2025
Communication from AWS

`Hello,

[AWS Health may periodically trigger reminder notifications about this communication if resources remain unresolved.]

We are contacting you as we have identified that your AWS Account currently has one or more AWS Lambda functions using the Python 3.9 runtime.

We are ending support for Python 3.9 in Lambda on December 15, 2025. This follows Python 3.9 End-Of-Life (EOL) reached on October 30, 2025 [1]. End of support does not impact function execution. Your functions will continue to run. However, they will be running on an unsupported runtime which is no longer maintained or patched by the AWS Lambda team.

As described in the Lambda runtime support policy [2], end of support for language runtimes in Lambda happens in several stages.

- Starting on December 15, 2025, Lambda will no longer apply security patches and other updates to the Python 3.9 runtime used by Lambda functions, and functions using Python 3.9 will no longer be eligible for technical support. Also, Python 3.9 will no longer be available in the AWS Console, although you can still create and update functions using Python 3.9 via AWS CloudFormation, the AWS CLI, AWS Serverless Application Model (SAM), or other tools.

- Starting January 15, 2026, you will no longer be able to create new Lambda functions using the Python 3.9 runtime.

- Starting February 15, 2026, you will no longer be able to update existing functions using the Python 3.9 runtime.

We recommend that you upgrade your existing Python 3.9 functions to the latest available Python runtime in Lambda before December 15, 2025.

Your impacted Lambda functions using the Python 3.9 runtime are listed on the 'Affected resources' tab of your AWS Health Dashboard.

This notification is generated for functions using the Python 3.9 runtime for the $LATEST function version. The following command shows how to use the AWS CLI [3] to list all functions in a specific region that is using Python 3.9, including published function versions. To find all such functions in your account, repeat the following command for each region:

aws lambda list-functions --region us-east-1 --output text --query "Functions[?Runtime=='python3.9'].FunctionArn"

Starting 180 days before deprecation, you can also use Trusted Advisor to identify functions using the Python 3.9 runtime [4].

If you have any concerns or require further assistance, please contact AWS Support [5].`

petrkalos:
## Overview
This commit upgrades the data.all framework from Python 3.9 to Python 3.12, modernizing the runtime environment and improving performance, security, and compatibility.

### 🐍 **Python Runtime Upgrade**
- **Core upgrade**: Python 3.9 → Python 3.12 across all environments
- **New runtime configuration**: Added `deploy/stacks/runtime_options.py` to centrally manage Python version settings
- **Lambda runtime**: Updated to use Python 3.12 runtime for AWS Lambda functions

### 📦 **Dependency Updates** (either because they were incompatible with 3.12 or because they raised known vulnerabilities)
- **ariadne**: 0.17.0 → 0.26.2 (GraphQL library upgrade)
- **fastapi**: 0.115.5 → 0.116.1 (minor version bump)
- **PyYAML**: 6.0 → 6.0.2 (patch update)
- **aws-cdk-lib**: 2.177.0 → 2.208.0 (significant CDK upgrade)

### 🐳 **Docker Infrastructure Improvements**
- **Replace dev/Dockerfile**: Refactored prod Dockerfiles and compose to use those in our local setup which reduces duplication but also the local environment is close to the prod.
- **Simplified base images**: Removed custom Elasticsearch and PostgreSQL Dockerfiles, now using official images
- **Health checks**: Added comprehensive health checks for all services in docker-compose
- **Service dependencies**: Improved service dependency management with condition-based startup
- **Removed wait-for-it.sh**: Eliminated custom wait script in favor of Docker health checks
- **Use Configs, Secrets and Watch**:  Replace volume mounts with the more modern Configs (bind configs in containers), Secrets (bind secrets in containers) and Watch (monitor for changes and sync them in the container)

### 🏗️ **Container Orchestration Enhancements**
- **PostgreSQL**: Switched to official `postgres:16.4` image (match the version used in Aurora)
- **Elasticsearch**: Switched to official `docker.elastic.co/elasticsearch/elasticsearch:7.9.3` image
- **Service isolation**: Better container naming and networking configuration
- **Startup reliability**: Enhanced service startup sequence with proper health monitoring

### 🧹 **Code Cleanup**
- **Removed files**: 
  - `backend/docker/dev/wait-for-it.sh` (replaced by health checks)
  - `compose/elasticsearch/Dockerfile` and config (using official image)
  - `compose/postgres/Dockerfile` (using official image)

Resolve #1842

### DIFF
--- a/backend/docker/prod/ecs/Dockerfile
+++ b/backend/docker/prod/ecs/Dockerfile
@@ -8,6 +8,7 @@ ARG PYTHON_VERSION=python3.12
 ARG CONTAINER_USER=cuser
 ARG CONTAINER_USER_GROUP=cuser
 ARG CONTAINER_UID=999
+ARG PORT=8080
 
 # Clean cache
 RUN dnf upgrade -y;\
@@ -85,7 +86,7 @@ COPY --chmod=755 <<EOT /uvicorn.sh
 #!/usr/bin/env bash
 set -e
 . ~/.nvm/nvm.sh
-${PYTHON_VERSION} -m uvicorn \$@
+${PYTHON_VERSION} -m uvicorn --port ${PORT} \$@
 EOT
 
-CMD ["/uvicorn.sh", "cdkproxymain:app", "--host", "0.0.0.0", "--port", "8080"]
+CMD ["/uvicorn.sh", "cdkproxymain:app", "--host", "0.0.0.0"]
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -14,8 +14,9 @@ services:
       dockerfile: backend/docker/prod/ecs/Dockerfile
       args:
         CONTAINER_UID: ${UID}
+        PORT: '2805'
     healthcheck:
-      test: [ "CMD", "curl", "-s", "http://localhost:8080" ]
+      test: [ "CMD", "curl", "-s", "http://localhost:2805" ]
       interval: 60s
       timeout: 10s
       retries: 10
@@ -23,7 +24,7 @@ services:
     expose:
       - 2805
     ports:
-      - 2805:8080
+      - 2805:2805
     depends_on:
       db:
         condition: service_healthy
@@ -61,8 +62,10 @@ services:
   graphql:
     build:
       context: ./
+      additional_contexts:
+        dataall-graphql-base: "service:graphql-base"
       dockerfile_inline: |
-        FROM dataall-graphql-base
+        FROM dataall-graphql-base 
         RUN python3.12 -m pip install uvicorn --prefix=/usr/local
     entrypoint:
       - "bash"


backend/docker/dev/Dockerfile
```diff
@@ -69,15 +65,24 @@ COPY ./requirements.txt requirements.txt
 COPY ./dataall/base/cdkproxy/requirements.txt cdk.requirements.txt
 
 # Install App requirements
-RUN /bin/bash -c "${PYTHON_VERSION} -m pip install setuptools --prefix /usr/local"
```
TejasRGitHub, line 72: Is removing `/bin/bash -c` a simplification ? Or its not required anymore ?

backend/docker/dev/Dockerfile
```diff
@@ -69,15 +65,24 @@ COPY ./requirements.txt requirements.txt
...
 COPY ./dataall /dataall
-ADD ./cdkproxymain.py /cdkproxymain.py
-ADD ./local_graphql_server.py /local_graphql_server.py
 
 WORKDIR /
 
-ENTRYPOINT [ "/bin/bash", "-c", ". ~/.nvm/nvm.sh && uvicorn cdkproxymain:app --host 0.0.0.0 --port 8080" ]
+COPY --chmod=755 <<EOT /uvicorn.sh
+#!/usr/bin/env bash
+set -e
```
TejasRGitHub, line 79: Don't we need to source node js env variables by using `. ~/.nvm/nvm.sh` required by cdk ?

backend/dataall/runtime_options.py
```diff
@@ -0,0 +1,12 @@
+from aws_cdk.aws_lambda import Runtime
+
+"""
+Changing this will change the python runtime in:
+* PROD Dockerfiles (lambda and ecs)
+* Lambda runtimes
+Will not change (so must manually be changed) in:
+* dev/Dockerfile
```
TejasRGitHub, line 8: Why can't we do the same for the dev Docker files. Can we pass build arguments while building docker files ?

backend/dataall/runtime_options.py
```diff
@@ -0,0 +1,12 @@
+from aws_cdk.aws_lambda import Runtime
```
TejasRGitHub, line 1: Should we put this file in the root folder instead of backend?

deploy/stacks/container.py
```diff
@@ -99,7 +100,7 @@ def __init__(
                 ecs.CfnTaskDefinition.ContainerDefinitionProperty(
                     image=cdkproxy_image.image_name,
```
TejasRGitHub, line 101: Is it possible to change the default version of python in the image ? I am thinking by default meaning, any command executed with `python -m ....` will execute with python3.12 ? Is that possible ?

TejasRGitHub: AL2023 can use python3.9 wherever it wants but if possible can we change the default version of python in the image?

backend/docker/dev/Dockerfile
petrkalos: simplification and not required, python executable is in the `PATH`

backend/docker/prod/ecs/Dockerfile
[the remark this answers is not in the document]
petrkalos: that's a good catch, it was throwing errors even from before this change. I think it's not needed, will try to remove it.

frontend/docker/dev/Dockerfile
```diff
@@ -29,6 +29,6 @@ COPY --chown=${CONTAINER_USER}:root ./version.json /
 ENV DISABLE_ESLINT_PLUGIN=true
 
 # Set the port to serve the application
-ENV PORT=80
+ENV PORT=8080
```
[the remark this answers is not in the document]
petrkalos: No and in fact ports below 1023 shouldn't really be used for such things. I just switched it to 8080 so that the message at the end is matching what the hosts sees and one can click the url directly.

deploy/stacks/runtime_options.py
```diff
@@ -0,0 +1 @@
+../../backend/dataall/runtime_options.py
```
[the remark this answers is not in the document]
petrkalos: No we can't because we won't be able to import it, `dataall` is the root module

backend/dataall/base/cdkproxy/cdk_cli_wrapper.py
```diff
@@ -216,7 +216,7 @@ def describe_stack(stack, engine: Engine = None, stackid: str = None):
 
 
 def cdk_installed():
-    cmd1 = ['.', '~/.nvm/nvm.sh']
+    cmd1 = ['bash', '~/.nvm/nvm.sh']
```
TejasRGitHub, line 219: Would this also source the environment variables needed by cdk ?

petrkalos: This is just a health check, I don't think it's being used anywhere but I saw the error and fixed it

TejasRGitHub (approved):
+1

